PR #18, opened 2026-05-15
feat(inference): QuaRot step 2 — rotation absorption math (ADR-044)

ohdearquant:
> _PR description authored by Claude (Anthropic agent) on behalf of @ohdearquant. Attribution per AGENTS.md (PR #17)._

## Summary

**Step 2 of 4** in QuaRot 4-bit rotated quantization (ADR-044). Step 1 (Hadamard primitives) merged in #15.

This PR adds the **rotation-absorption math**:

```text
absorb_input_rotation:  W ← W · R^T   so y = W · x = W' · (R · x)
absorb_output_rotation: W ← R · W     so R · y = W' · x
```

Both row-major, in-place, with **f32** and **f64** variants. The f64 path keeps the rotation math itself out of single-precision rounding; whether that f64 precision survives downstream depends on the bridge to Q4 chosen in step 3 (see ADR-044 step 3 row — two candidate paths, both pending measurement on step 4's perplexity benchmark).

**Scope honest:** this PR is a small algebraic staging patch — math primitives + correctness validation, nothing more. It is useful only when step 3 lands with the real `RotationPlan` + binary that consumes these primitives. If step 3 stalls, this leaves public API surface in `lattice-inference::quant::quarot::rotation` with no consumer; the trade-off was discussed in ADR-044 §Consequences and accepted given the multi-PR step-table commitment.

## Files changed

- `crates/inference/src/quant/quarot/rotation.rs` (new, ~470 lines including tests)
- `crates/inference/src/quant/quarot/mod.rs` — `pub mod rotation;`
- `docs/adr/ADR-044-quarot-rotated-quantization.md` — step table updated; stale "2× disk" removed from both consequence sites; step 3 row now documents the f64→BF16 vs `quantize_f32_to_q4` bridge options without claiming precision parity

## Scope change vs original ADR-044

Original ADR step 2 included "read SafeTensors → absorb → save rotated SafeTensors as intermediate." This PR drops the intermediate write because step 3's `quantize_quarot` binary will read → absorb → quantize → write Q4 in a single offline pass — saving and re-loading an intermediate SafeTensors is wasteful.

| Step | PR | Scope |
|---|---|---|
| 1 | PR #15 (merged) | Hadamard primitives (f32 + f64) |
| 2 | **this PR** | Rotation absorption math + synthetic equivalence tests, no I/O |
[...]

PR #15, opened 2026-05-15
feat(inference): QuaRot Hadamard primitives (ADR-044)

ohdearquant:
> _PR description authored by Claude (Anthropic agent) on behalf of @ohdearquant. Attribution per the policy in PR #17._

## Summary

**Step 1 of 4** in QuaRot 4-bit rotated quantization (Ashkboos et al., NeurIPS 2024, [arxiv:2404.00456](https://arxiv.org/abs/2404.00456)). This PR is **primitives only**: rotation absorption, q4 integration, and the perplexity benchmark land in PRs 2–4 per the ADR step table.

- `quant::quarot::hadamard::walsh_hadamard_in_place` — iterative butterfly Walsh-Hadamard, O(n log n), power-of-2 only (f32)
- `quant::quarot::hadamard::walsh_hadamard_orthonormal_in_place` — isometry variant (||y|| = ||x||) (f32)
- `quant::quarot::hadamard::walsh_hadamard_f64_in_place` + `walsh_hadamard_orthonormal_f64_in_place` — f64 variants for the absorption-pass precision budget per ADR-044 §Risks
- `quant::quarot::hadamard::RandomizedHadamard` — seeded R = H · D rotation, `apply` + `apply_inverse` (f32) and `apply_f64` + `apply_inverse_f64` for the absorption pass. Sign vector stored once and cast at apply time.
- ADR-044 documents v0 as 4 sequential PRs, model coverage table (decoder 0.6B + hybrid 0.8B in v0; encoder 4B + BERT deferred to v1)

## Why QuaRot

Existing Q4_0 RTN quantization clips outlier channels in Qwen3 weights. QuaRot applies a random Hadamard rotation before quantization — the rotation is orthogonal (model output is mathematically identical) but spreads outlier energy across all dimensions, so the resulting distribution is uniformly quantizable. Paper reports 0.47 perplexity loss at 4-bit weights+activations+KV on LLaMA-2-70B; we'd re-measure on our pipeline before claiming numbers externally.

## Test plan

- [x] `cargo test -p lattice-inference --lib quant::` — **19 unit tests pass**
- [x] Known small-case results (size 1, 2, 4 Walsh-Hadamard) — f32 and f64
- [x] Double-application scales by n (unnormalized form)
- [x] Orthonormal form is isometry (||Hx|| = ||x||) — f32 and f64
- [x] Orthonormal form is involution (H(Hx) = x) — f32 and f64
- [x] Power-of-two and length-mismatch guards
[...]

PR #19, opened 2026-05-15
feat(inference): QuaRot step 3a — RotationPlan + apply_tensor_rotation (ADR-044)

ohdearquant:
> _PR description authored by Claude (Anthropic agent) on behalf of @ohdearquant. Attribution per AGENTS.md (PR #17)._

## Summary

**Step 3a** of QuaRot (ADR-044). PR #15 (primitives) and PR #18 (absorption math) merged; this PR adds the **architecture-specific recipe** that says "in Qwen3.5, the attention `q_proj` takes the residual-stream rotation on its input side."

The math primitives in `quant::quarot::rotation` are architecture-agnostic. This module owns the Qwen3.5-specific tensor-name patterns.

### Coverage table (also in module doc)

| Tensor | Reads from residual | Writes to residual | Absorption |
|---|---|---|---|
| `attn.q_proj`, `attn.k_proj`, `attn.v_proj` | ✓ | — | input-side |
| `attn.o_proj` | — | ✓ | output-side |
| `mlp.gate_proj`, `mlp.up_proj` | ✓ | — | input-side |
| `mlp.down_proj` | — | ✓ | output-side |
| `lm_head` | ✓ | — | input-side |
| `embed_tokens` | — | ✓ | output-side |

## Files changed

- `crates/inference/src/quant/quarot/plan.rs` (new, ~400 lines including tests)
- `crates/inference/src/quant/quarot/mod.rs` — `pub mod plan;`

## Test plan

- [x] `cargo test -p lattice-inference --lib quant::` — **35 unit tests pass** (28 from prior PRs + 7 new plan tests)
- [x] `qwen35_plan_covers_dense_residual_stream_tensors` — all 9 dense tensor patterns resolve to the correct side
- [x] `qwen35_plan_misses_moe_expert_weights` — documents v1 deferral for MoE
- [x] `qwen35_plan_misses_gdn_linear_attention_weights` — documents v1 deferral for GatedDeltaNet
- [x] `apply_tensor_rotation_skips_unplanned` — unplanned tensor (GDN) not mutated
- [x] **`apply_tensor_rotation_mlp_layer_pair_rotates_output`** — the absorption identity at the MLP layer pair: output equals `R · y_original` (not `y_original` itself, because `down_proj` writes back to the rotated residual). Initially had this wrong and the test caught it during development — the corrected expected value reflects how absorption actually behaves at the residual-stream boundary
- [x] `f64_apply_matches_f32_apply_pattern` — 1e-5 parity between f32 and f64 paths
- [x] `cargo clippy --workspace -- -D warnings` — clean

## Out of scope for step 3a (deferred to 3b/3c/3d)

- **Step 3b**: SafeTensors I/O reader (reuse existing memmap parser from `bin/quantize_q4.rs`)
- **Step 3c**: Forward-equivalence assertion on a real model (`||rotated_forward − original_forward|| < 1e-5` on a small batch)
- **Step 3d**: Q4 bridge (f64→BF16 vs new `quantize_f32_to_q4` entry point — decided by measurement per ADR-044 step 3 row) + the `quantize_quarot` binary
- RMSNorm scale rotation: residual-stream rotation needs to commute through pre/post RMSNorm. For Hadamard rotations specifically this is the QuaRot identity; needs its own helper before a real Qwen3.5 model runs end-to-end. To be added in step 3c or 3d.

## Scope honest

Like PR #18, this is staging — useful only when the downstream slices (3b through 3d) land. The plan-as-data structure does buy something concrete: the deferred-to-v1 boundary (MoE experts, GDN layers) is now expressed as code that REJECTS those tensor names rather than as prose in an ADR. Future agents touching the QuaRot pipeline will trip the test if they accidentally widen scope.

## References

- ADR-044
- PR #15 (step 1: Hadamard primitives, merged)
- PR #18 (step 2: absorption math, merged)


ohdearquant:
> _Comment authored by Claude (Anthropic agent) on behalf of @ohdearquant. Attribution per AGENTS.md (PR #17)._

Round 1 review addressed in commit c8b58b1. Three of four findings were real bugs/scope errors — accepting all, here's the honest accounting:

**1. (Blocker) embed_tokens.weight OutputSide → InputSide** — real math bug. Verified against `loading.rs:260` (`[vocab_size, hidden]`) and the tied-embedding fallback at `weights.rs:51` (`logits_weight()` returns `&self.embed_tokens` when `lm_head` is None). My original OutputSide would have tried to match a hidden-dim rotation (1024) against vocab_size rows (248320) and failed at runtime. The correct absorption is **input-side**: each row IS the embedding vector, and `W ← W · R^T` rotates each row, producing `R · embedding` as the output. The same absorption covers both uses (untied embedding output AND tied-as-lm_head input): `W · R^T · (R · hidden) = W · hidden = logits_original`. Module doc now spells out both directions in the coverage table.

**2. (Blocker) Skips GDN → now includes GDN** — also accepting this. My "deferred to v1" framing was unjustified. GDN reads from and writes back to the residual stream the same way GQA does, just with different tensor names. Added 5 GDN tensor rules: `linear_attn.in_proj_qkv/z/b/a` input-side, `linear_attn.out_proj` output-side. Renamed `qwen35_residual_stream` → `qwen35_residual_stream_linear_layers` to be accurate about what it covers (linear-layer weights only — RMSNorm gamma is a separate class).

**3. (Major) RMSNorm absent** — real blocker for end-to-end correctness, accepting fully. Qwen3.5's shifted RMSNorm `(1 + gamma)` (norm.rs:16) does not commute with Hadamard. The QuaRot paper handles this by **fusing `(1 + gamma)` into the immediately-following linear layer's weights** before rotation absorption; after fusion, the normalize-only step is rotation-invariant. This is significant model surgery, NOT just rotation absorption. I'm pushing it to step 3c/3d explicitly:
   - Module doc has a "Known gaps (BLOCKERS for real-model conversion)" section listing this
   - The plan does NOT silently include RMSNorm gamma tensors; new test `qwen35_plan_misses_rmsnorm_and_scalar_weights` asserts that
   - Step 3c/3d MUST implement fusion before any real-model conversion claim. Documented as a refuse-on-fail.

**4. (Major) Suffix matching too loose** — also valid. Added `validate_coverage(tensor_names)` returning a `CoverageReport` with `unmatched_rules` and `matched_tensors` fields. Two new tests:
   - `validate_coverage_flags_missing_rules`: pure-GQA tensor list flags 5 unmatched GDN rules
   - `validate_coverage_complete_for_qwen35_hybrid_tensor_set`: Qwen3.5 hybrid tensor mock yields complete coverage
   
   Step 3b's binary will refuse-on-fail when `unmatched_rules.is_empty()` is false — that signals a plan-vs-architecture mismatch (e.g., trying to convert a Llama-3 with the Qwen3.5 plan).

**On the framing "tests bless the omissions"** — accepted as a real critique of round-1 of this PR. The new tests now bless the *positive* coverage (14 rules matching real Qwen3.5 hybrid tensor names) AND the *intentional* misses (MoE experts → v1, RMSNorm gamma → step 3c/3d fusion required). The previous "should not match v0 plan" assertions for GDN tensors are gone since GDN is now correctly covered.

**Verification:**
- `cargo test -p lattice-inference --lib quant::` — 33 passed (14 rules × ≥1 case + GDN+MoE+RMSNorm absence tests + coverage validation + MLP layer-pair identity + f32/f64 parity)
- `cargo clippy --workspace -- -D warnings` — clean
- `cargo fmt --all -- --check` — clean

Plan now has **14 rules** (was 9 with the embed_tokens bug + missing GDN). Two known gaps still need addressing in step 3c/3d before any real conversion can be claimed correct: RMSNorm gamma fusion + bias rotation. These are now documented as code-visible blockers (module doc + test assertions), not buried in prose.

ohdearquant:
> _Comment authored by Claude (Anthropic agent) on behalf of @ohdearquant. Attribution per AGENTS.md (PR #17)._

Round 2 review addressed in commit 7d24923. All 5 findings accepted as valid — the review went deep this round and 4 of them were architecturally important.

**1. (Blocker) `lm_head` as required rule fails for tied embeddings** — fair. Verified: Qwen3.5 default `tie_word_embeddings=true` (`qwen35_config.rs:177`), and the loader only loads `lm_head.weight` when `!tie_word_embeddings` (`loading.rs:16`). My `validate_coverage` would have rejected every default Qwen3.5-0.8B build. Added `RuleRequirement::{Required, Optional}` enum, marked `lm_head` as Optional. `is_complete()` now ignores Optional rules with zero matches; they surface in `unmatched_optional_rules` for informational use.

**2. (Blocker) Tied + final-RMSNorm fusion mutually incompatible** — this was the deepest finding. Walked the math: if the tied tensor must serve as embedding (rows = `R · E[i]`) AND as lm_head after `(1 + g_final)` fusion (rows = `R · diag(1 + g_final) · E[i]`), one shared tensor can't be both unless `g_final == 0`. Module doc now spells out the two paths step 3c/3d must choose between: (a) untie embeddings during conversion + materialize a separate `lm_head` with both rotation AND fused `(1 + g_final)`, or (b) keep final RMSNorm online and don't fuse. Plan covers the linear-tensor side of both options; the decision is left to the conversion binary with explicit guidance.

**3. (Major) `validate_coverage` didn't validate what its docs claimed** — yep. Extended `CoverageReport` from 2 fields to 5: `matched_tensors`, `unplanned_tensors`, `ambiguous_tensors`, `unmatched_required_rules`, `unmatched_optional_rules`. `is_complete()` requires zero unmatched-required AND zero ambiguous (pattern collisions). Optional rules being unmatched is fine; unplanned tensors are caller's call per ADR-043.

**4. (Major) Shape doc wrong in two places**:
   - `q_proj.weight` is `[2*q_dim, hidden]` for Qwen3.5 (Q + gate_z fused per `weights.rs:8`), not `[q_dim, hidden]`. Module doc fixed. The absorption still works because `hidden = cols` regardless of the row count — input-side absorption applies row-wise so the gate-z half rotates correctly alongside the Q half. Added that detail to the table.
   - RMSNorm fusion direction: I wrote "rows" but for row-major `[out, hidden]` weights an input-dim scale is a column multiply (`W[i,j] *= (1 + g[j])`, equivalent to `W ← W · diag(1 + g)`). Fixed.

**5. (Major) Test gap** — accepting, also painful because previous rounds *had* a positive-coverage test that got dropped during the edit churn. Added 6 new tests:
   - `qwen35_plan_covers_residual_stream_tensors`: all 14 rules with positive cases
   - `qwen35_plan_misses_rmsnorm_and_scalar_weights`: intentional misses
   - `validate_coverage_tied_embeddings_is_complete_without_lm_head`: the exact blocker case from finding #1
   - `validate_coverage_untied_embeddings_matches_lm_head`: positive untied case
   - `validate_coverage_flags_unmatched_required_rule`: drop o_proj, watch `is_complete()` go false
   - `validate_coverage_lists_unplanned_tensors`: stray MTP / input_layernorm tensors

The required-tensor suffix list (`qwen35_required_residual_tensor_suffixes`) is now a test helper that mirrors what the loader at `loading.rs:16, 33` actually expects — step 3b can promote this to a non-test helper when wiring the SafeTensors reader.

**Verification:**
- `cargo test -p lattice-inference --lib quant::` — **39 passed** (33 → 39)
- `cargo clippy --workspace -- -D warnings` — clean
- `cargo fmt --all -- --check` — clean

**Honest meta-take:** round 2 surfaced architectural decisions (tied + RMSNorm fusion, optional rules, coverage report semantics) that round 1 didn't reach. The plan abstraction is meaningfully tighter now. Still: this is staging — useful only when step 3b/c/d land. The conversion-binary decisions (which fusion path, what to do with tied embeddings post-fusion) are flagged but not made here.

ohdearquant:
> _Comment authored by Claude (Anthropic agent) on behalf of @ohdearquant. Attribution per AGENTS.md (PR #17)._

Round 3 review addressed in commit 4015d17. All 3 findings accepted — the first one was the most important, and the review is right that a wrong comment is worse than a missing one because future converters cargo-cult it.

**1. (Major) "Keep final-RMSNorm online" path was mathematically wrong** — walked the math:

Qwen3.5 runtime (norm.rs:16, forward.rs:81): `n = D · normalize(h)` then `logits = W_lm · n`, where `D = diag(1 + g_final)`.

After QuaRot, residual is rotated (`h_rot = R · h`). Using normalize's rotation-invariance:
```
n_rot       = D · normalize(h_rot) = D · R · normalize(h)
logits_run  = lm_head' · n_rot     = lm_head' · D · R · normalize(h)
```

For `logits_run = logits_original = W_lm · D · normalize(h)`:
```
lm_head' · D · R = W_lm · D
=> lm_head' = W_lm · D · R^T · D^{-1}
```

That's **not** a clean input-side rotation of `W_lm` — `D` and `R` don't commute (D diagonal, R dense Hadamard). So any "rotate lm_head input-side and leave the final norm alone" recipe is wrong. For both tied and untied.

Replaced the two-paths framing with a single correct recipe. Step 3c MUST:
1. **Untie** the embeddings (copy `embed_tokens → lm_head` pre-rotation when `tie_word_embeddings=true`)
2. **Fuse** final-norm scale: `lm_head_fused = W_lm · diag(1 + g_final)` (column multiply per the Known gaps section)
3. **Rotate**: `lm_head_final = lm_head_fused · R^T = W_lm · D · R^T`
4. **Zero out runtime final_norm.weight** (set `g_final = 0` so `D = I`)

Runtime then computes `lm_head_final · R · normalize(h) = W_lm · D · R^T · R · n = W_lm · D · n = logits_original`. ✓

This matches your prescription exactly. Module doc spells out the math step-by-step so a future converter implementer cannot misread it.

**2. (Medium) `validate_coverage` is suffix-presence, not per-layer** — fair. A model with 24 layers but only 1 layer's tensors present passes the current check. Docstring narrowed to say "**suffix-presence only**, NOT per-layer coverage." Step 3b's conversion binary will need a stricter check using the loader's config-derived expected tensor list — pointing at a TODO `expected_tensor_names(config)` helper that step 3b will add. The conversion binary will call both checks: this method for the architecture sanity (right names exist), per-layer-coverage for completeness (all layers covered).

**3. (Minor) Misplaced docstring** — accepted. The `RotationPlan` prose ended up orphaned on `RuleRequirement` after the round-2 edit (RuleRequirement got inserted between the prose and the struct it described). Moved the docstring back to `RotationPlan` and added a Construction/Use paragraph. RuleRequirement's own docstring stays on RuleRequirement.

**Verification:**
- `cargo test -p lattice-inference --lib quant::` — 39 passed
- `cargo clippy --workspace -- -D warnings` — clean
- `cargo fmt --all -- --check` — clean

**Honest meta:** review round 3 caught exactly the dangerous kind of error — a confidently-written comment about model math that was wrong. That comment would have shipped to step 3c as a cargo-cult target. Glad it got caught here. The fact that the unit tests in this PR could not catch this (it's a comment about runtime semantics, not the plan's local behavior) is why prose review on math-heavy modules matters.

ohdearquant:
> _Comment authored by Claude (Anthropic agent) on behalf of @ohdearquant. Attribution per AGENTS.md (PR #17)._

Round 4 review addressed in commit d5d2c25. All 5 findings accepted as valid. The harsh framing is fair — none of these are bugs in the rotation math itself, but the prose around it was misleading enough that step 3c could have cargo-culted into a broken converter. The fixes are all docs / scope discipline.

**1. (Blocker) Tied-embedding untying needs config flip too** — verified `loading.rs:266`: the loader skips `lm_head.weight` entirely when `cfg.tie_word_embeddings=true`. Without flipping the saved config to false, the runtime falls back to `embed_tokens` via `weights.rs:51`'s `logits_weight()` and the fused `lm_head` is never consulted. Step 3c's offline procedure now lists this as a separate REQUIRED step ("Flip the output config to `tie_word_embeddings=false`. ... **This is a config-mutation requirement, NOT just a weight transform.**"). The contradictory "tied fallback handles both roles" line is gone.

**2. (Major) `is_complete()` can report success for wrong conversion** — fair. The plan captures rotation rules; RMSNorm fusion + config flip are separate transforms. Made the boundary explicit:
   - Module §Known gaps opens with "**This plan is rotation-rule data only.** ... A converter that rotates every planned tensor and stops there will produce wrong logits."
   - `is_complete()` doc now says "**NOT a correctness gate**" with the full list of what the conversion binary's separate full-validity check must verify (per-layer coverage, RMSNorm fusion applied, config flipped, fused lm_head materialized, forward-equivalence delta below threshold).

**3. (Medium) Helper name wrong** — verified: the real helper is `qwen_required_tensor_names` at `loading.rs:12`, re-exported `#[cfg(test)]` only at `mod.rs:44`. Doc updated with the correct name + explicit note that step 3b must promote the re-export out of test cfg.

**4. (Medium) Misidentified GDN linear_attn.norm** — verified at `gdn.rs:425, 440` and `gdn_fused.rs:127`. GDN's gated-RMSNorm multiplies plain `gamma[i]`, NOT `(1 + gamma[i])`. More importantly: this norm runs INSIDE the GDN block between linear-attention compute and `out_proj`. The residual-stream rotation enters at `in_proj_*` and exits at `out_proj` — it does NOT cross this internal norm. So the plain-gamma GDN norm does NOT need QuaRot fusion. Module doc now separates the shifted-gamma input/post/final RMSNorms (need fusion, with file:line citations) from GDN's plain-gamma internal norm (must NOT be attempted, with citations).

**5. (Medium) LoRA runtime compatibility silently broken** — accepting. The forward path at `forward.rs:249, 467` and `gdn_fused.rs:385` adds LoRA delta to base matmul output using the same activation. Rotated base produces output in a different basis than un-rotated adapter delta. v0 now explicitly marks QuaRot-converted models as LoRA-runtime-incompatible. Two future-paths documented (rotate adapter on load using converter seed stored in .q4 metadata, OR refuse-on-compose at adapter load time), neither in v0.

**Verification:**
- `cargo test -p lattice-inference --lib quant::` — 39 passed (no behavior change; all changes are docs)
- `cargo clippy --workspace -- -D warnings` — clean
- `cargo fmt --all -- --check` — clean

**Honest meta:** four rounds of review on a small Rust module have been about prose quality, not local code correctness. That's the right shape for a math-heavy data-plan abstraction. The unit tests verify the plan's local invariants (suffix matching, side dispatch, ambiguity detection); the prose verifies the conversion-correctness contract that the downstream converter will encode. Adversarial doc review at this stage is cheap relative to shipping a converter that silently produces wrong logits — exactly the failure class the QuaRot literature has hit repeatedly.

If you want me to keep iterating, I'm happy to keep going. Alternatively, this round closes the obvious correctness gaps in the prose; the remaining work for a mergeable PR may now be limited to whether `is_complete()` should be renamed to something less authoritative-sounding (e.g., `is_suffix_coverage_complete`) to remove temptation, or whether the current doc-level discipline is sufficient.

ohdearquant:
> _Comment authored by Claude (Anthropic agent) on behalf of @ohdearquant. Attribution per AGENTS.md (PR #17)._

Round 5 addressed in commit 204915c. Both findings accepted — both are internal-doc consistency, no code-path change.

**1. (Medium) Stale tied-fallback comment at line 297** — yep, that comment in `qwen35_residual_stream_linear_layers()` still said `logits_weight()` "handles both roles" while the module §Tied embeddings section says exactly the opposite. Replaced with a pointer to the module doc + an explicit "tied-fallback path is NOT correctness-preserving" reminder.

**2. (Minor) "Two known blockers" undercount** — fair. The list at module §Known gaps is now five items (RMSNorm fusion, config flip, fused lm_head materialization, helper-export promotion, LoRA incompatibility). Reworded the constructor doc to point at the module doc without enumerating a count, so future additions can land in §Known gaps without this site going stale again.

**Verification:**
- `cargo test -p lattice-inference --lib quant::` — 39 passed (no behavior change)
- `cargo clippy --workspace -- -D warnings` — clean
- `cargo fmt --all -- --check` — clean

PR #48, opened 2026-05-19
feat(inference): add GDN self-speculative decoding

ohdearquant:

## Summary

Self-speculative decoding using the model's own GDN (linear attention) layers as draft and GQA (full attention) layers as verifier. Zero additional model weights or memory overhead — exploits Lattice's unique hybrid architecture (18 GDN + 6 GQA in Qwen3.5).

- `forward_step_gdn_only`: runs embedding + all GDN layers + final head in a single Metal command buffer, skipping GQA layers entirely. Does not advance KV cache seq_len.
- `generate_greedy_self_spec`: draft/verify/rollback loop. Drafts up to 4 tokens via GDN-only forward, verifies with full model via `verify_tokens_batched`, accepts longest agreeing prefix, rolls back GDN S-matrices + KV cache on rejection.
- Gated by `LATTICE_SELF_SPEC=1` + greedy sampling + at least one linear attention layer present.
- Checkpoint pool shared with MTP: sized to `max(MTP_VERIFY_MAX_TOKENS, SELF_SPEC_MAX_DRAFT)`.

## Why this is novel

No other inference engine has a hybrid linear/full attention architecture to exploit for free self-speculation. Every other self-speculative approach (Medusa, Eagle, early-exit) requires extra trained heads or a separate draft model. Here the GDN layers ARE the draft model — they're already in the weights.

## Test plan

- [x] `self_spec_gdn_only_does_not_advance_kv_cache` — KV cache seq_len unchanged during draft
- [x] `self_spec_gdn_only_logits_are_finite` — draft logits are valid
- [x] `self_spec_checkpoint_pool_allocated_on_env` — pool created when env var set
- [x] `self_spec_generate_respects_token_budget` — generation stays within max_new_tokens
- [x] 576/576 tests pass, clippy clean, fmt clean

### DIFF
--- a/crates/inference/src/attention/gdn.rs
+++ b/crates/inference/src/attention/gdn.rs
@@ -97,6 +97,22 @@ impl GatedDeltaNetState {
         self.conv_buffer.fill(0.0);
     }
 
+    /// Snapshot the live recurrent state for speculative rollback.
+    /// See ADR-052: snapshot-before-draft protocol.
+    pub fn snapshot(&self) -> GdnLayerSnapshot {
+        (self.s_matrices.clone(), self.conv_buffer.clone())
+    }
+
+    /// Restore recurrent state from a prior [`snapshot`]. The snapshot must have been taken
+    /// from a state with the same key/value dims; in debug builds this is enforced via
+    /// `debug_assert_eq!` on buffer lengths.
+    pub fn restore_from(&mut self, snap: &GdnLayerSnapshot) {
+        debug_assert_eq!(self.s_matrices.len(), snap.0.len());
+        debug_assert_eq!(self.conv_buffer.len(), snap.1.len());
+        self.s_matrices.copy_from_slice(&snap.0);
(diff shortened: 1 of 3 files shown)


ohdearquant:
**Review flagged 3 blockers and 1 major issue.**

**Blockers:**
1. Metal self-spec tests don't compile (BpeTokenizer::for_testing missing, enable_thinking missing, unsafe env mutation)
2. Checkpoint pool off-by-one: pool sized for 4 but verify_tokens_batched needs 5 slots with 4 drafts → panic
3. Rollback restores wrong GDN state: slot 0 overwritten during verification before rollback reads it. Partial acceptance + rejection leaves GDN state inconsistent with KV cache.

**Major:**
1. ADR-052's snapshot/restore API not implemented — Metal-only checkpoint pool doesn't satisfy the general trait contract.

The Metal checkpoint/blit approach is sound but the state machine is broken. Rework needed per ADR-052.

PR #71, opened 2026-05-24
perf(inference): SIMD elementwise ops (rms_norm, silu, elementwise_mul)

ohdearquant:
## Summary
- Adds NEON (AArch64) and AVX2 (x86_64) SIMD paths for `rms_norm`, `silu_inplace`, and `elementwise_mul` in `crates/inference/src/forward/cpu/elementwise.rs`
- Uses `simd_config()` dispatch pattern from `softmax.rs` — runtime fallback to scalar
- Schraudolph fast-exp for silu sigmoid (same bit trick as softmax fast_exp)
- Criterion bench at hidden_size=[896, 2048, 4096] with `Throughput::Bytes` reporting

## Test plan
- [ ] 6 new tests: known-value correctness + SIMD-vs-scalar comparison at multiple sizes
- [ ] `cargo test -p lattice-inference` passes
- [ ] `cargo bench --bench elementwise_bench --no-run` compiles

### DIFF
--- a/benches/BASELINE.md
+++ b/benches/BASELINE.md
@@ -1,0 +1,73 @@
+# Lattice Performance Baseline
+
+**Date**: 2026-05-23
+**Commit**: ac38f20 (main, post ADR-057 LoRA lifecycle merge)
+**Platform**: Apple Silicon (aarch64 / Darwin)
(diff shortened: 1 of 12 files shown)


ohdearquant:

PR #73, opened 2026-05-24
perf(inference): optimize matmul_bt scalar fallback + NEON prefetch

ohdearquant:
## Summary
- Adds 8x k-loop unrolled `matmul_bt_scalar_m1` fast path for single-token decode (m=1)
- Software prefetch of next k-chunk B-rows in NEON microkernel via `prfm pldl1keep` inline asm
- Criterion benchmark at 3 sizes: single-token decode, short-sequence prefill, FFN projection

## Test plan
- [ ] `cargo test -p lattice-inference` passes (771 tests)
- [ ] `cargo bench --bench matmul_opt_bench --no-run` compiles
- [ ] NEON prefetch compiles only on `target_arch = "aarch64"` + `not(target_os = "macos")`

### DIFF
--- a/benches/BASELINE.md
+++ b/benches/BASELINE.md
@@ -1,0 +1,73 @@
+# Lattice Performance Baseline
+
+**Date**: 2026-05-23
+**Commit**: ac38f20 (main, post ADR-057 LoRA lifecycle merge)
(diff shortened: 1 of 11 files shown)


ohdearquant:

PR #75, opened 2026-05-24
perf(inference): decode attention kernel + benchmark

ohdearquant:
## Summary
- Adds specialized `decode_attention` kernel for the seq_len=1 case (single-token decode against full KV cache)
- Scalar implementation with decode-specific score computation and weighted V sum
- Criterion benchmark at kv_seq_len=[128, 512, 2048] for standard and GQA configurations

## Test plan
- [ ] 4 unit tests for decode attention correctness
- [ ] `cargo test -p lattice-inference` passes (769 tests)
- [ ] `cargo bench --bench attn_opt_bench --no-run` compiles

### DIFF
--- a/benches/BASELINE.md
+++ b/benches/BASELINE.md
@@ -1,0 +1,73 @@
+# Lattice Performance Baseline
+
+**Date**: 2026-05-23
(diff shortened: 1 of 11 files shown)


ohdearquant:

PR #188, opened 2026-06-04
perf(inference): chunked batched prefill for long prompts on Metal

ohdearquant:
## chunked batched prefill for long prompts (Metal)

Replaces the per-token prefill loop with a chunked, batched Metal path for long
prompts. Long-context prompts no longer pay a GPU dispatch per token during prefill.

### Why
For prompts longer than the prefill window, the old path fell back to a per-token
loop — one GPU dispatch per token — which dominated time-to-first-token on long
contexts. This chunks the request and batches the per-chunk work.

### Result (Qwen3.5-0.8B, Apple Silicon Metal, 1000-token prompt)
| | Prefill tok/s | TTFT |
|---|---:|---:|
| before (per-token fallback) | 99 | 10.2 s |
| after (chunked batched) | **184** | **5.5 s** |

~1.86× prefill, bit-exact (parity max_abs_diff = 0.000000). Decode path unchanged.

### Notes
- First of a 3-PR stack (prefill perf). PR2 adds batched attention; PR3 adds the
  parallel GatedDeltaNet scan.
- No new crates; no library `unwrap()`.

---
**Stack:** #188 (this) → #189 (batched attention) → #190 (GDN parallel scan)

### DIFF
--- a/crates/inference/src/bin/bench_decode_ab.rs
+++ b/crates/inference/src/bin/bench_decode_ab.rs
@@ -36,7 +36,7 @@ fn run() -> Result<(), Box<dyn std::error::Error>> {
     use lattice_inference::forward::metal_qwen35::{ChatMessage, MetalQwen35State};
     use lattice_inference::model::qwen35::Qwen35Model;
     use lattice_inference::model::qwen35_config::{GenerateConfig, Qwen35Config};
-    use lattice_inference::tokenizer::BpeTokenizer;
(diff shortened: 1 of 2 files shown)


ohdearquant:
Superseded by #228 (merged) — same engine change, rebased clean onto main with critic fixes (MINOR-1/3) + the e2e long-prefill regression trip-wire. Verified 1.78× TTFT, bit-for-bit greedy parity, e2e HF parity green.

PR #189, opened 2026-06-04
perf(inference): batched causal prefill attention + elementwise batching

ohdearquant:
## batched causal prefill attention + elementwise batching

**Stacked on #188.** Replaces the per-token decode-attention loop in prefill with a
single batched causal-attention pass, and batches the elementwise prefill ops so they
no longer issue a GPU dispatch per token.

### Why
After the chunked-prefill fix, the remaining per-token cost in prefill was the
attention loop and the elementwise ops running once per position. This collapses both
into batched passes over the prompt.

### Result
Additional prefill speedup on top of PR1 (cumulative interleaved A/B tracked
internally; a fresh same-process A/B will be attached before merge rather than quoting
a number I can't reproduce in-session). Decode path unchanged; prefill argmax parity
preserved across runs.

### Notes
- Removes an unnecessary `unsafe` in the attention fallback.
- No new crates; no library `unwrap()`.

### DIFF
--- a/crates/inference/src/forward/metal_qwen35.rs
+++ b/crates/inference/src/forward/metal_qwen35.rs
@@ -2358,6 +2358,290 @@ kernel void moe_shared_gate_add(
     if (gid >= hidden) return;
     scratch_out[gid] = fma(gate_val, expert_out[gid], scratch_out[gid]);
 }
+
+// ===== Batch scatter Q and gate from interleaved q_proj output (Win 1) =====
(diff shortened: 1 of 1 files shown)


ohdearquant:
Superseded by #229 — same batched-causal-attention change, rebased clean onto post-#228 main as a single commit. Validated: 5.31× cumulative TTFT (1825ms), argmax-identical parity. Under critic + CI gate now.

PR #191, opened 2026-06-20
feat(tune): exact-gradient LoRA backward + multi-layer training pipeline

ohdearquant:
## Summary

- Exact CPU backward pass through GQA attention (with Qwen3.5 sigmoid-gated Q) and GDN recurrence layers, with numerical gradcheck verification (11/11 GQA + 3/3 GDN checks green)
- Multi-layer training pipeline (train_grad_full) that materialises layers [19..23] and trains q/v LoRA on the two GQA layers with Adam optimizer
- Per-key Adam timestep fix (shared counter over-advanced t) + MLX-matched LoRA-A init (1/sqrt(in))
- Held-out validation eval integrated into the training loop
- Differential test examples (diff_attn_layer23, diff_gdn_layer) that verify materialised forward matches the real model (max-diff 3.58e-6)

## Convergence verification (100-step matched comparison vs MLX-LM)

Properly matched config: q/v LoRA on layers [19,23] only, 106K params, rank=8, scale=20, lr=1e-4, completion-only loss, 16 train / 8 valid samples, seq_len=128.

| step | MLX true val | lattice held-out | MLX delta | lattice delta |
|------|-------------|-----------------|-----------|---------------|
| 0    | 4.9061      | 4.9056          | --        | --            |
| 20   | **4.5966**  | 4.7113          | **-0.310**| -0.194        |
| 40   | 4.7604      | **4.6141**      | -0.146    | **-0.292**    |
| 60   | 4.9951      | 4.7499          | +0.089    | -0.156        |
| 80   | 5.1378      | 4.9656          | +0.232    | +0.060        |
| 100  | 5.5380      | 5.0287          | +0.632    | +0.123        |

- Base alignment: 4.9061 vs 4.9056 (delta 0.0005), identical eval
- Best checkpoint: MLX 4.5966 (step 20) vs lattice 4.6141 (step 40), gap 0.018 NLL
- Lattice overfits 5x less severely at step 100 (+0.123 vs +0.632)
- **Verdict: on par** (0.018 NLL gap at best checkpoint, lattice more stable)

## Files (20 files, +6422 lines)

- backward/attention_gqa.rs: GQA backward with sigmoid-gate gradient
- backward/ops.rs: elementwise backward ops (RMSNorm, SiLU, etc.)
- backward/gradcheck.rs: finite-difference gradcheck harness
- backward/tape.rs: forward tape for backward replay
- attention/gdn_backward.rs: GDN recurrence backward
- model/qwen35/{cache,eval,model}.rs: capture-tap + materialised layer accessors
- bin/train_grad_full.rs: multi-layer trainer (layers 19-23)
- bin/train_grad_layer23.rs: single-layer trainer
- lora/optimizer.rs: per-key Adam fix

## Test plan

- [x] cargo clippy --workspace -- -D warnings clean
- [x] 11/11 GQA gradcheck + 3/3 GDN gradcheck (self-consistency)
- [x] Differential test: materialised forward matches real model (max-diff < 1e-3)
- [x] 100-step training run with held-out eval matches MLX-LM quality
- [ ] make ci full pass


ohdearquant:

## Review of PR #191 (round 1)

### Verdict: REQUEST CHANGES

### Prior review context
- No prior human review threads were visible through `gh pr view`; this is a first-round review.
- This round focused on the requested numerical backward paths, gradcheck coverage, optimizer state, feature gating, dependency policy, and trainer edge cases.

### What this round examined
- GQA backward gate handling against the Qwen3Next reference shape: q-projection is split into query and gate, attention output is multiplied by `sigmoid(gate)`, then `o_proj` runs.
- GDN reverse-time recurrence, including the adjoint state carried through `S_t = g*S_{t-1} + outer(k, delta)` and the `kv_mem = S_{t-1}^T k` retrieval path.
- Finite-difference settings in the unit gradchecks and the full trainer gradcheck mode.
- Differential examples for real-model materialised forward parity.
- Adam per-key timestep implementation and dependency/feature-gating policy.

### Findings (# total: 0 blockers, 1 major, 2 mediums, 1 minor)

#### Major
1. [crates/tune/src/bin/train_grad_full.rs:701](crates/tune/src/bin/train_grad_full.rs#L701) Long-sequence handling can panic or exhaust memory instead of failing cleanly.
   - **Current**: `--seq-len` is parsed without any upper bound at lines 701-703, then accepted samples are sent through `build_caches`, which calls `model.capture_attn_io(&s.tokens, first_layer)` at lines 157-160. `capture_attn_io` only rejects empty token lists before calling `forward_step` for every position at [crates/inference/src/model/qwen35/eval.rs:208](crates/inference/src/model/qwen35/eval.rs#L208)-224, while `Qwen35Model::max_context` documents that `forward_step` panics when the position exceeds the precomputed RoPE table at [crates/inference/src/model/qwen35/model.rs:87](crates/inference/src/model/qwen35/model.rs#L87)-92.
   - **Issue**: The trainer does handle an empty training set at lines 853-855, but it does not handle the large-sequence edge case called out in the review brief. A user can pass `--seq-len` above `model.max_context()` and admit long samples that later panic in the forward path. Separately, `FullFwd` stores `HeadPos { logits: Vec<f32> }` for every completion source position at lines 300-305 and pushes a full vocab-sized logits vector in the loop at lines 460-471. With Qwen3.5's vocab, this scales as `completion_positions * vocab_size * 4` just for cached logits, before backward buffers. Large sequence settings can abort the process via allocation failure rather than returning a bounded diagnostic.
   - **Fix**: Validate `seq_len_cap <= model.max_context()` before loading samples, reject or clamp larger values with an explicit error, and add a memory budget/streaming head path. At minimum do not store all per-position logits in `FullFwd`; store the final hidden/inv/target and compute CE/VJP one position at a time, or fail early with an estimated memory requirement.
   - **Test**: Add a trainer CLI/unit test for `--seq-len model.max_context()+1` returning `Err`, plus a small-fixture test proving `forward_full` does not retain `seq * vocab` logits if the streaming fix is implemented.
   - **Confidence**: derived from the trainer and `capture_attn_io` code paths; not observed by intentionally triggering an OOM.

#### Mediums
1. [crates/tune/src/lora/optimizer.rs:350](crates/tune/src/lora/optimizer.rs#L350) The prior Adam timestep bug is fixed in code but not regression-tested against multiple keys.
   - **Current**: `AdamState::step` now advances a per-key counter at lines 93-106, which is the right implementation for the prior shared-counter bug. The only optimizer state test calls `step` once for `"test"` and asserts that key's count is `1` at lines 350-374.
   - **Issue**: That test does not exercise the failure mode described in the PR: one optimizer step updates several tensors (`a_q`, `b_q`, `a_v`, `b_v`) and a shared/global counter over-advances bias correction for later keys. A regression could reintroduce cross-key advancement while still passing the current one-key test shape.
   - **Fix**: Add a test that steps `a`, `b`, `c`, `d` once each and asserts each key has `t == 1`, then steps only `a` again and asserts `a == 2` while the other keys remain `1`. Optionally assert the first update magnitude for identical gradients is identical across first-time keys.
   - **Test**: `cargo test -p lattice-tune --features train-backward adam_per_key_timestep_does_not_cross_advance`.
   - **Confidence**: derived from the test coverage; observed `cargo test -p lattice-tune test_adam --features train-backward` passed the existing single-key test.

2. [crates/tune/src/bin/train_grad_full.rs:992](crates/tune/src/bin/train_grad_full.rs#L992) The full-depth finite-difference check can miss sparse missing-gradient bugs because it probes top analytic entries only.
   - **Current**: Gradcheck mode selects `top_k_indices(agrad, probe.min(alen))` at lines 968-969, where `top_k_indices` sorts by `grad[b].abs().partial_cmp(...).unwrap()` at lines 664-669. It then computes FD only for those analytic-top entries.
   - **Issue**: This is good for non-vacuous high-signal entries, but it is not sufficient as the only assembled full-depth real-model check. If a structural bug zeros or under-populates an entire gradient group, top-by-analytic can choose arbitrary zero entries; it may not sample the entries where FD would be non-zero. The direct GQA unit test finite-differences all q/v LoRA arrays, but the real full-depth test is the one that is supposed to catch integration mistakes through frozen GDN layers.
   - **Fix**: Keep the top-k probes, but add deterministic random probes per array and at least one scan/assertion that the analytic nonzero pattern is compatible with FD on sampled entries. For the prior failure class, include a fixed probe set covering gate-half `b_q` rows in lower and top GQA layers.
   - **Test**: A `--gradcheck --probe` run should fail if the GDN backward is bypassed or if the gate half of `b_q` is zeroed.
   - **Confidence**: derived from the gradcheck sampling strategy; observed the unit `gradcheck` filter passed.

#### Minor
1. [crates/tune/src/bin/train_grad_full.rs:710](crates/tune/src/bin/train_grad_full.rs#L710) `--log-every 0` panics.
   - **Current**: `log_every` is parsed directly at lines 710-712 and used as `step % log_every` at line 1158.
   - **Issue**: This is a small CLI hardening gap, but it is still an avoidable runtime panic in the training binary.
   - **Fix**: Reject `--log-every 0` during argument validation or treat `0` as "log only final".
   - **Test**: CLI parse test or a small unit around argument validation.
   - **Confidence**: derived from the modulo operation.

- GQA sigmoid gate backward: [crates/inference/src/backward/attention_gqa.rs:127](crates/inference/src/backward/attention_gqa.rs#L127)-139 correctly splits `d_context_gated` into `d_context_ungated = d_context_gated * sigmoid(gate_z)` and `d_gate_z = d_context_gated * context_ungated * sigmoid'(gate_z)`, and lines 216-227 re-interleave the raw gate-half gradient into the full `2*q_dim` q-projection gradient.
- GQA forward cache stores the ungated context before applying the gate at [crates/inference/src/backward/attention_gqa.rs:558](crates/inference/src/backward/attention_gqa.rs#L558)-589, so the gate derivative has the right activation.
- GDN backward carries reverse-time `d_s` state and differentiates both the recurrent update and retrieval paths at [crates/inference/src/attention/gdn_backward.rs:438](crates/inference/src/attention/gdn_backward.rs#L438)-650. Its forward-save formulation matches the fused forward's pre-decay retrieval plus `g * kv_mem` delta convention at [crates/inference/src/attention/gdn_fused.rs:481](crates/inference/src/attention/gdn_fused.rs#L481)-499.
- Differential examples do load the on-disk Qwen3.5 model and compare materialised forward against real captured forward. `diff_attn_layer23` uses layer 23 and fails at `max_diff >= 1e-3` ([crates/inference/examples/diff_attn_layer23.rs:35](crates/inference/examples/diff_attn_layer23.rs#L35)-58, 75-98, 123-128). `diff_gdn_layer` loads a configurable real GDN layer, defaults to layer 2, and uses the same `< 1e-3` max-diff threshold ([crates/inference/examples/diff_gdn_layer.rs:42](crates/inference/examples/diff_gdn_layer.rs#L42)-65, 88-102, 131-136).
- No `unwrap()`/`expect()` or `unsafe` hits were found in `crates/inference/src/backward/*.rs` or `crates/inference/src/attention/gdn_backward.rs` outside tests; `rg` returned no matches for those target files.
- The new backward modules are gated behind `train-backward` at [crates/inference/src/lib.rs:53](crates/inference/src/lib.rs#L53)-54 and [crates/inference/src/attention/mod.rs:7](crates/inference/src/attention/mod.rs#L7)-8.
- Dependency direction looks acceptable: `lattice-inference` did not gain internal dependencies, and `lattice-tune` adds an optional renamed dependency on the leaf `lattice-inference` with `version` plus `path` at [crates/tune/Cargo.toml:52](crates/tune/Cargo.toml#L52).

### Observed test results
- `cargo test -p lattice-inference --features train-backward gradcheck`: passed. 11 tests passed, including GQA LoRA gradcheck, GDN single-head/multi-head gradchecks, and end-to-end LoRA gradcheck.
- `cargo test -p lattice-tune test_adam --features train-backward`: passed. 1 optimizer test passed.

### What I did NOT check
- I did not run `make ci`, full workspace clippy, benches, or the real-model differential examples; the model-backed examples require local model files and were reviewed by code path only.
- I did not reproduce the PR body's MLX-LM convergence table or the claimed `3.58e-6` differential max-diff; I only verified that the examples enforce `< 1e-3`.
- I did not manually audit the SIMD kernels touched in `metal_qwen35.rs`; the requested backward paths are CPU-only and no new unsafe was added in them.

### Sources consulted
- Hugging Face Transformers Qwen3Next source: `Qwen3NextAttention` defines `q_proj` with `2 * num_attention_heads * head_dim`, splits into query and gate, multiplies attention output by `sigmoid(gate)`, then applies `o_proj`: https://github.com/huggingface/transformers/blob/v5.12.0/src/transformers/models/qwen3_next/modeling_qwen3_next.py

- After the trainer edge-case and regression-test fixes, the next useful round should run the two real-model differential examples plus `train_grad_full --gradcheck` on the actual model artifacts, then inspect any remaining bench-regression rationale.

Domain utility: LOW — khive returned no matching review domains, so the repo guidelines, code, tests, and authoritative Qwen3Next source were the useful references.

ohdearquant:
## Bench regression override rationale

The gate failures (3 aarch64, 25 x86_64) are all in `lattice-embed` SIMD benchmarks (simd_batch_cosine, simd_query_batch_dot_product, etc.). This PR does not touch any embed SIMD code — all changes are:

1. `backward/*.rs` — new modules, feature-gated behind `train-backward`
2. `attention/gdn_backward.rs` — new module, feature-gated
3. `model/qwen35/{cache,eval,model}.rs` — capture-tap accessors, feature-gated
4. `forward/metal_qwen35.rs` — batch fused Metal dispatch helpers (GPU-only, not benchmarked in CI)
5. `tune/` — training binaries and optimizer

None of these affect the CPU SIMD kernels measured by the gate. The regressions are CI runner variability (note identical base/new ns values in many rows). Adding `bench-allow-regression` per ADR-058 override policy.

PR #226, opened 2026-06-23
fix(inference): guard max_new_tokens == 0 in both generate paths

ohdearquant:
## What

Both non-streaming generate paths sampled and returned **one token even when the caller passed `max_new_tokens == 0`**, violating the contract. The streaming variant (`generate_streaming`) already guarded this; the two siblings did not:

- `Qwen35Model::generate` (`crates/inference/src/model/qwen35/generation.rs`)
- free-fn `generate` (`crates/inference/src/generate.rs`, `QwenModel` path)

## Fix

Mirror the existing `generate_streaming` guard verbatim — early-return an empty `GenerateOutput` before prefill/sampling. **Pure early return**; the decode hot path and the greedy e2e-parity path are byte-identical.

## Severity: LOW

Found by a correctness sweep, which rated it `high`. Honest reassessment: this is a **degenerate-config off-by-one** (`max_new_tokens == 0` is a config real callers almost never pass). Symptom is one unexpected token — no crash, no memory unsafety, no corruption. Worth fixing for contract-correctness and to make the three generate paths consistent, but not high-severity.

## Test

`generate_max_new_tokens_zero_returns_empty` — synthetic-model regression for the `Qwen35Model` path (fails pre-fix: forward+sample yields 1 token; passes post-fix: guard returns empty). The free-fn path is disk-gated (`from_directory` only), so it has no cheap unit fixture; its guard mirrors the now-tested pattern verbatim.

## Verification

- `cargo test -p lattice-inference --lib` → 915 pass, 7 ignored
- `cargo clippy -p lattice-inference --lib -- -D warnings` → clean
- `bench-compare`: correctness-only change, no hot-path delta (added branch is never taken when `max_new_tokens > 0`, which holds for all generation + benchmarks). Numbers added below.
- `e2e-parity.yml` runs on this PR (touches `inference/src`) — the relevant correctness gate.

### DIFF
--- a/crates/inference/src/generate.rs
+++ b/crates/inference/src/generate.rs
@@ -197,6 +197,18 @@ pub fn generate(
         return Err(InferenceError::InvalidInput("Empty prompt".into()));
     }
 
+    // max_new_tokens == 0 means "generate nothing": return before prefill/sampling
+    // so we never emit a token the caller did not ask for.
+    if config.max_new_tokens == 0 {
+        return Ok(GenerateOutput {
+            text: String::new(),
+            token_ids: Vec::new(),
+            prompt_tokens: prompt_len,
+            generated_tokens: 0,
+            stopped_by_eos: false,
+        });
+    }
+
     // 2. Initialize KV cache and scratch (allocate once per request)
     let max_seq = prompt_len + config.max_new_tokens;
     let cache_cfg = FlatKVCacheConfig::for_qwen3(
--- a/crates/inference/src/model/qwen35/generation.rs
+++ b/crates/inference/src/model/qwen35/generation.rs
@@ -26,6 +26,18 @@ impl Qwen35Model {
             return Err(InferenceError::Inference("empty prompt".into()));
         }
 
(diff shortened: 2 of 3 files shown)


ohdearquant:
### bench-compare (origin/main `74ea1e75d` vs HEAD `863b74683`, --quick)

**No significant regression — all inference micro-bench change deltas `p > 0.05` (within noise).**

This is a correctness-only guard: the added `if max_new_tokens == 0` branch sits *before* prefill/sampling and is never taken during generation or benchmarks (`max_new_tokens > 0` always holds), so the decode/prefill hot loops are byte-identical.

A couple of quick-mode outliers (`+69%`, `+10%`) appeared but at `p = 0.10`/`0.08` (not significant) — these are the known two-worktree harness artifacts (10-sample quick mode, wide variance), not real deltas. The only borderline-`p=0.05` lines were small *negative* (faster) deltas, which the change cannot have caused. Aggregate A/B table failed to render (`embed` `tier_prepared_*` groups missing base estimates — unrelated to this inference-only change).

Verdict: bench-compare shows no change attributable to this PR.

PR #238, opened 2026-06-23
feat(inference): opt-in f16 KV cache for Metal GQA path (#154)

### LINKED ISSUE #154: feat(inference): wire f16 KV cache into default Metal decode path
## Context

ADR-062 Phase 1 (#130) landed the f16 KV cache infrastructure in v0.3.0 — the storage format, conversion kernels, and cache management are implemented. But the default decode path (`encode_gdn_layer`, `encode_gqa_layer`) still reads/writes f32 KV buffers.

## Goal

Make f16 KV cache the default for Metal decode, halving KV memory from `seq_len × kv_dim × 4` to `seq_len × kv_dim × 2` bytes. For Qwen3.5-0.8B with 4096 context: 24 layers × 4096 × 256 × 4 = 96 MB → 48 MB.

## Tasks

- [ ] Wire f16 KV write path into `dispatch_copy_offset` (K/V cache stores)
- [ ] Wire f16 KV read path into `dispatch_decode_attention` 
- [ ] Update attention kernels to read `device const half*` from KV cache
- [ ] Benchmark: confirm no throughput regression (f16 reads should be faster due to halved bandwidth)
- [ ] PPL gate: confirm ≤0.05 PPL delta vs f32 KV (expected: negligible for attention scores)
- [ ] Make configurable via env var (`LATTICE_KV_F16=0` to disable) as escape hatch

## Acceptance

- [ ] `bench_decode_ab` shows ≥160 tok/s (no regression)
- [ ] `ppl_metal` shows ≤0.05 PPL delta vs current f32 path
- [ ] Memory reduction confirmed via Metal GPU frame capture or allocation tracking

ohdearquant:
## Summary

Closes #154. Adds an **opt-in** f16 KV cache to the Metal Qwen3.5 attention path, gated by `LATTICE_KV_F16` (default **OFF**). When OFF, the path is byte-identical to `main` (f32 KV). When ON (`LATTICE_KV_F16=1`), the device-side KV buffers are f16 (2 bytes/elem instead of 4), halving KV memory for the full-attention (GQA) layers.

Qwen3.5-0.8B is GatedDeltaNet-dominated (18 GDN / 6 GQA layers); only the 6 GQA layers use a KV cache, so this touches the GQA attention path only. GDN is untouched.

## Design: additive, byte-identical f32 path

The three KV-reading kernels (`decode_attention`, `decode_attention_flash_partial`, `prefill_attention_batched_causal`) are generated from a **single MSL `#define` macro** instantiated twice — for `float` and for `half`. KV reads inside the macro are wrapped in `float(...)` — a no-op cast for `float`, the widening cast for `half`. This guarantees the f32 expansion is the same logic as the prior kernel (no hand-duplication, no drift). f16 adds sibling write kernels (`copy_buf_offset_f16`, `copy_kv_cache_batch_f16`), dtype-aware allocation, a dtype-aware `append_kv`, a dtype-aware ADR-064 assertion, and pipeline/dispatch selection. **No f32-kernel body was edited.**

## Verification (Qwen3.5-0.8B, Metal, local M-series)

**Greedy parity — `LATTICE_KV_F16=1` vs `=0`:** exact full-sequence match (not just first-5-token) at short, ~2.8k-char, and ~7.4k-char prompts. Because the f32 path is byte-identical to `main` (which passes the HF e2e-parity gate), an exact f16-vs-f32 match means the f16 path inherits HF parity.

**Perplexity (Metal `ppl_metal`):**

| KV dtype | NLL | PPL |
|----------|-----|-----|
| f32 (OFF, == main) | 1.467063 | 4.3365 |
| f16 (ON) | 1.467067 | 4.3365 |

Δ ≈ 4e-6 NLL — well under the 0.05 gate. The tiny nonzero delta confirms the f16 path is genuinely active (not silently falling back to f32).

**Decode throughput at 1k ctx (`bench_decode_ab` slope, runs=5):** 71.6 (OFF) vs 72.8 tok/s (ON) — **neutral within noise**.

## Honest finding: f16 KV is decode-NEUTRAL here, and that's expected

This PR does **not** speed up decode, and the bench shows it. First-principles: decode reads ~800 MB of Q8 weights per token; the KV cache at 1k ctx for 6 GQA layers is ~6 MB read per token (<1% of bandwidth). Halving KV bandwidth therefore can't move decode throughput on this model. **The real decode lever is weight quantization (Q4 vs Q8), not KV dtype.** This re-frames the #185 prioritization that promoted #154 as *the* decode lever — it isn't.

What f16 KV **does** buy is **memory**: it halves the GQA KV buffers (~50 MB at 4096 ctx, ~400 MB at 32K). That matters for the Pi-5 edge direction and for long-context headroom, not for decode tok/s on an 0.8B model.

## Why default OFF

Correctness is proven and memory is halved, but no speed benefit was measured, and long-context decode could not be characterized cleanly (the quick harness is thermal-noise-bound past ~2k ctx on this machine). f32 stays the zero-risk default. Set `LATTICE_KV_F16=1` to opt in.

## bench-compare (ADR-058)

This touches `crates/inference/`. The relevant A/B is the f16-vs-f32 decode-slope table above (71.6 → 72.8 tok/s at 1k ctx, neutral). Standard `make bench-compare` (origin/main vs HEAD with the default f32 path) shows no change because the default path is byte-identical to main — the f16 path is exercised only with `LATTICE_KV_F16=1`, which the A/B harness above toggles directly.

## Test / build status

- `cargo build --release -p lattice-inference --features "f16,metal-gpu"` — clean
- `cargo clippy --workspace -- -D warnings` (CI's no-features invocation) — clean
- 86 metal-gpu tests pass (incl. `test_decode_attention_edge_cache_len_{1,257,513,1025}`, `kv_cache_determinism`, `from_q4_dir`)
- f32 path (`LATTICE_KV_F16=0`) confirmed byte-identical to f16-off / origin/main

## Known CI gap (separate follow-ups, not blocking)

The Metal GQA path that this PR touches is **not** exercised by CI's `e2e-parity.yml` (it builds `qwen35_generate` with `--features f16` only, which uses the CPU path), nor by CI clippy (no-features, never lints metal-gpu-gated code). The verification above was run manually on the Metal path. Filing these as follow-up issues.

---

### DIFF
--- a/crates/inference/src/forward/metal_qwen35.rs
+++ b/crates/inference/src/forward/metal_qwen35.rs
@@ -479,144 +479,122 @@ kernel void per_head_rms_norm(
 // GQA group simultaneously. QK computed once per (query_head, cache_tile), scores stored in
 // threadgroup memory and reused for V accumulation. Q loaded once into threadgroup memory.
 // Grid: [num_kv_heads, 1, 1].  Threads: [256, 1, 1] — one thread per output dimension.
-kernel void decode_attention(
-    device const float* q        [[buffer(0)]],
-    device const float* k_cache  [[buffer(1)]],
-    device const float* v_cache  [[buffer(2)]],
-    device float* out            [[buffer(3)]],
-    constant uint& cache_len     [[buffer(4)]],
-    constant uint& head_dim      [[buffer(5)]],
-    constant uint& num_q_heads   [[buffer(6)]],
-    constant uint& num_kv_heads  [[buffer(7)]],
-    constant uint& q_dim         [[buffer(8)]],
-    constant uint& kv_dim        [[buffer(9)]],
-    constant float& scale        [[buffer(10)]],
-    uint gid  [[threadgroup_position_in_grid]],
-    uint lid  [[thread_position_in_threadgroup]],
-    uint tgs  [[threads_per_threadgroup]])
-{
-    if (cache_len == 0) return;
-    constexpr uint HEAD_DIM    = 256;
-    constexpr uint MAX_GRP     = 8;   // max GQA group size supported
-    constexpr uint TILE_TOKENS = 256; // cache tokens per tile; one thread per token during QK
-
-    if (head_dim != HEAD_DIM || num_kv_heads == 0) return;
-    const uint kvh = gid;
-    if (kvh >= num_kv_heads) return;
-    if ((num_q_heads % num_kv_heads) != 0) return;
-    const uint group_size = num_q_heads / num_kv_heads;
-    if (group_size == 0 || group_size > MAX_GRP) return;
-    const uint qh_base = kvh * group_size;
-
-    // Threadgroup memory (~17 KB total for MAX_GRP=8, HEAD_DIM=256, TILE_TOKENS=256)
-    threadgroup float q_s    [MAX_GRP * HEAD_DIM];      // Q for all heads in this KV group
-    threadgroup float score_s[MAX_GRP * TILE_TOKENS];   // raw QK scores, then unnorm weights
-    threadgroup float reduce_s[TILE_TOKENS];             // reduction scratch
-    threadgroup float m_s    [MAX_GRP];                  // running max per Q head
-    threadgroup float l_s    [MAX_GRP];                  // running denominator per Q head
-    threadgroup float alpha_s[MAX_GRP];                  // exp(m_old - m_new) per Q head
-
-    // Initialize running softmax state
-    if (lid < group_size) {
-        m_s[lid] = -INFINITY;
-        l_s[lid] = 0.0f;
-    }
-
-    // Load Q for all query heads in this GQA group into threadgroup memory (H4)
-    for (uint idx = lid; idx < group_size * HEAD_DIM; idx += tgs) {
-        q_s[idx] = q[qh_base * HEAD_DIM + idx];
-    }
-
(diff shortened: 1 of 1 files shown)


ohdearquant:
## Adversarial review: approved

Ran an adversarial review (high effort) against `origin/main`, specifically to verify the byte-identical-f32 invariant. No functional findings. It independently confirmed:

- The `float`-instantiated macro bodies of all three read kernels (`decode_attention`, `decode_attention_flash_partial`, `prefill_attention_batched_causal`) are byte-equivalent in behavior to `origin/main` — only macro wrapping + no-op `float(...)` casts added, control flow and math preserved.
- `append_kv` f32 branch unchanged; f32 allocation stays 4 bytes/element.
- `MetalKvCache::new` gates buffer sizes correctly (×2 f16, ×4 f32), incl. the ADR-064 layout check using `element_size = use_kv_f16 ? 2 : 4`.
- All changed read/write KV dispatch sites are format-aware (`_f16` vs non-`_f16`); `use_kv_f16` is threaded consistently from `LATTICE_KV_F16` → `new_session` → `MetalQwen35State` → `MetalKvCache::new`.
- No new `unwrap!`/`expect!`/`panic!` in non-test library paths.

Combined with the exact greedy parity, PPL Δ≈4e-6, and decode-neutral A/B in the PR body, the f16 path is correct and the default-OFF f32 path is provably unchanged.

Issue #243, opened 2026-06-23
fix(inference): n-gram speculative decoding is lossy — verify_draft commits draft[0] unverified (diverges from greedy)


ohdearquant:
## Bug

`verify_draft` (crates/inference/src/speculative.rs:1835) commits the **first** draft token without verifying it against the target model's prediction at the current position. This makes n-gram speculative decoding **lossy** — its output can diverge from plain greedy decoding, when speculative decoding is supposed to be exact (token-identical to greedy).

```rust
// speculative.rs ~1846
for (i, &draft_token) in draft_tokens.iter().enumerate() {
    let logits = forward_fn(draft_token, position_start + i);
    let model_choice = argmax(&logits);
    all_logits.push(logits);
    accepted += 1;                                   // draft[i] counted accepted unconditionally
    if i + 1 < draft_tokens.len() && model_choice != draft_tokens[i + 1] as usize {
        break;                                       // only ever rejects draft[i+1], never draft[0]
    }
}
```

`forward_fn(draft[0], pos)` conditions on `draft[0]` already being the token at `pos` and returns the logits for `pos+1`. The target's actual greedy choice **at `pos`** (i.e. `argmax(forward_fn(all_tokens.last(), pos))`) is never computed, so `draft[0]` is never checked. A length-1 draft is therefore always accepted verbatim.

## Trigger

`prompt_tokens = [1,1,1]` (n-gram speculator drafts `[1]`), `max_new_tokens = 1`, `forward_fn` whose greedy choice at the current position is `0` (not `1`): the function returns `[1]`, but greedy decoding would return `[0]`.

## Impact

Speculative decoding silently changes generation output vs the non-speculative path. Any parity/correctness expectation that "speculation is a pure speedup" is violated. Confidence: medium-high (requires the n-gram path to draft a token greedy wouldn't pick — realistic on repetitive context).

## Fix (needs design)

Restructure verification so the current-position target distribution gates `draft[0]`:
- Either run a forward pass for the current position first and verify `draft[0] == argmax(current_logits)` before consuming it, or
- Reframe the speculator so `draft[0]` is the already-verified bonus token from the prior step (standard speculative-decoding bookkeeping).

This is an algorithmic change to the verification loop, not a one-line clamp — filing separately from the budget-overrun bug. Discovered via a sweep, TBV-confirmed against the source.

PR #281, opened 2026-06-24
fix(inference): speculative argmax first-wins to match greedy decode (#280)

### LINKED ISSUE #280: speculative::argmax tie-break is last-wins; engine/sampling/HF greedy are first-wins (speculation not token-identical to greedy)
## Summary

`speculative::argmax` breaks ties toward the **last** max index, but every other greedy path in the engine breaks ties toward the **first** index. Since speculative decoding is documented as a *pure speedup* (output must be token-identical to plain greedy), a tied top logit makes speculative output diverge from non-speculative greedy. This is the same class of bug as #243/#258, found by an adversarial audit of the (CI-blind) speculative path.

## Evidence

- `crates/inference/src/speculative.rs:103-110` — `argmax` uses `iter().enumerate().max_by(|a,b| a.partial_cmp(b).unwrap_or(Equal))`. Rust's `max_by` returns the **last** element among equals (the doc comment even states "returns the last occurrence"). → **last-wins**.
- `crates/inference/src/forward/metal_qwen35.rs:7548-7553` — production greedy decode: `if v > best_val { ... }` (strict `>`). → **first-wins**.
- `crates/inference/src/sampling.rs:372` (`argmax_f32_scalar`) and `:415` (`argmax_f32_neon`, which *explicitly* tie-breaks to the smaller index: `lane_val == best_val && lane_idx < best_idx`) → **first-wins**.
- `torch.argmax` (the HF reference the `e2e-parity` gate compares against) → **first-wins**.

`speculative::argmax` is the **only** last-wins implementation in the crate. The existing test `argmax_tie_last_wins` (`speculative.rs:2084`) codifies the wrong behavior — it looks like an accidental `max_by` footgun rather than an intentional choice.

## Why it bites in practice

The n-gram speculator fires on **repetitive context** — which is exactly where top-logit ties are most likely (and quantization coarsens logits into more collisions). So the bug's trigger conditions correlate with the speculator's activation conditions; this is not a purely theoretical tie.

Concrete divergence: with a `forward_fn` returning tied top logits (e.g. `[0.0, 0.0, -1.0, ...]`), speculative emits the *later* tied index while plain greedy (engine / HF) emits the *earlier* one.

## Fix

Make `speculative::argmax` first-wins (strict `>` loop), matching `sampling`, the engine, and `torch.argmax`. NaN handling becomes consistent with the canonical paths as a side effect (NaN is skipped, matching `sampling.rs:376`). Update `argmax_tie_last_wins` → first-wins and add a regression test that cross-checks against the engine's first-wins loop.

## Severity

High for the correctness contract ("pure speedup, token-identical to greedy"), though only observable on tied top logits. CI-blind: no workflow exercises the speculative path, so this would not be caught by `e2e-parity`.

## Follow-ups (separate)

- Add a generation-level greedy-equivalence test for the speculative path to CI (would have caught both this and #243).
- Audit `mtp_verify_draft` error paths for cache-rollback symmetry (draft generation mutates caches before a possible `Err`; rollback only runs on the success path).

_Found via an adversarial audit of `speculative.rs`; claims independently verified against source._

ohdearquant:
Closes #280.

## Problem

`speculative::argmax` broke ties toward the **last** max index (`Rust max_by`), but every other greedy path breaks ties toward the **first**:
- engine greedy decode `forward::metal_qwen35.rs:7548-7553` (`if v > best_val`)
- `sampling::argmax_f32_scalar` (`sampling.rs:372`) and `argmax_f32_neon` (`:415`, explicit `lane_idx < best_idx`)
- `torch.argmax` (the HF reference the `e2e-parity` gate compares against)

Speculative decoding is documented as a **pure speedup — token-identical to plain greedy**. A tied top logit therefore made speculative output diverge from non-speculative greedy. Same class as #243/#258. The trigger correlates with activation: the n-gram speculator fires on repetitive context, which is exactly where top-logit ties (and quantization-induced logit collisions) are most likely.

Found by an adversarial audit of the speculative path, which has **zero CI coverage** (no workflow exercises speculation).

## Fix

`argmax` → a first-wins loop matching `sampling::argmax_f32_scalar`. NaN entries are now skipped consistently with the canonical paths (was: `partial_cmp().unwrap_or(Equal)`).

## Verification (TDD)

Tests written first, run against the **unfixed** code to prove they catch the bug:
```
argmax_tie_first_wins              left: 1  right: 0   FAILED
argmax_tie_matches_engine_first_wins  left: 3  right: 0   FAILED   (argmax returned 3 = last of three tied 0.0; engine returns 0)
```
After the fix:
```
test result: ok. 56 passed; 0 failed   (full speculative:: module)
cargo clippy -p lattice-inference --lib -- -D warnings   clean
```
`argmax_tie_matches_engine_first_wins` cross-checks against an inline copy of the engine's first-wins loop, so it guards the *contract* (agreement with greedy), not just a hardcoded index. `e2e-parity` triggers on this PR (touches `crates/inference/src`).

## Scope / follow-ups (not in this PR)

- Generation-level greedy-equivalence test for the speculative path in CI (would have caught both this and #243).
- Audit `mtp_verify_draft` error-path cache-rollback symmetry (review flagged: draft generation mutates caches before a possible `Err`; rollback only runs on the success path). Low severity.

## Note on merge

This is a behavior change (greedy output on tied logits). It is regression-gated and proven, but I am **not** auto-merging — flagging for human review since it changes tie-break semantics on the hot path.

### DIFF
--- a/crates/inference/src/speculative.rs
+++ b/crates/inference/src/speculative.rs
@@ -2117,6 +2117,17 @@ mod tests {
     }
 
     #[test]
+    fn argmax_skips_nan_and_never_panics() {
+        // NaN loses to finite values; among finite ties the first index wins.
+        // Matches sampling::argmax_f32_scalar and the engine greedy decode.
+        assert_eq!(argmax(&[f32::NAN, 1.0, 1.0]), 1);
+        assert_eq!(argmax(&[1.0, f32::NAN, 1.0]), 0);
+        // All-NaN and all-(-inf) fall back to index 0 — must never panic.
+        assert_eq!(argmax(&[f32::NAN, f32::NAN]), 0);
+        assert_eq!(argmax(&[f32::NEG_INFINITY, f32::NEG_INFINITY]), 0);
+    }
+
+    #[test]
     fn argmax_negative_values() {
         assert_eq!(argmax(&[-3.0, -1.0, -2.0]), 1);
     }
--- a/crates/inference/src/speculative.rs
+++ b/crates/inference/src/speculative.rs
@@ -98,15 +98,22 @@ impl NgramSpeculator {
 ///
 /// Return the index of the maximum element in `logits`.
 ///
-/// When multiple elements share the maximum value, returns the last
-/// occurrence (Rust `max_by` semantics). Returns 0 on empty input.
+/// When multiple elements share the maximum value, returns the **first**
+/// occurrence. Speculative decoding must be token-identical to plain greedy, so
+/// this tie-break must match the engine greedy decode (`forward::metal_qwen35`),
+/// the `sampling` module, and `torch.argmax` — all of which are first-wins (#280).
+/// Non-finite (`NaN`) entries are skipped, matching `sampling::argmax_f32_scalar`.
+/// Returns 0 on empty input.
 pub fn argmax(logits: &[f32]) -> usize {
-    logits
-        .iter()
-        .enumerate()
-        .max_by(|(_, a), (_, b)| a.partial_cmp(b).unwrap_or(std::cmp::Ordering::Equal))
-        .map(|(i, _)| i)
-        .unwrap_or(0)
+    let mut best_idx = 0;
+    let mut best_val = f32::NEG_INFINITY;
+    for (i, &v) in logits.iter().enumerate() {
+        if v > best_val {
+            best_val = v;
+            best_idx = i;
+        }
+    }
+    best_idx
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


ohdearquant:
## Critic gate: review flagged one soft issue

Adversarial review of the fix commit. Core fix confirmed correct on every substantive dimension:

- **A — tie-break direction (PASS):** engine greedy (`metal_qwen35.rs:7548-7553`), `sampling::argmax_f32_scalar`, `sampling::argmax_f32_neon` (explicit `lane_idx < best_idx`), and `torch.argmax` are all first-wins. Speculative was the lone last-wins outlier; the fix aligns it.
- **B — NaN/non-finite (PASS):** new loop skips NaN, treats ±inf coherently, all-NaN/all-(-inf)/empty → index 0, never panics — same contract as `sampling::argmax_f32_scalar`, avoiding the old `partial_cmp(..).unwrap_or(Equal)` pitfalls.
- **C — no hidden last-wins dependence (PASS):** all ~12 non-test callers are greedy/verification paths that want greedy-equivalent tie-breaking; none rely on last-wins.
- **E — scope/style (PASS):** minimal single-file diff, no `unwrap`.

**One TBV correction to the review:** the review's Dimension-D note claimed the input `[0.0, 0.0, -1.0, 0.0]` in `argmax_tie_matches_engine_first_wins` "would return 0 in both old and new," i.e. wouldn't distinguish the bug. That's incorrect — the fail-first run proved the **old** `argmax` returned index **3** for that input (`left: 3, right: 0`), because Rust `max_by` is last-wins (the very bug). The test does distinguish old/new. That slip is based on the same last-wins-vs-first-wins assumption this PR fixes.

I did act on the review's valid coverage suggestion: added `argmax_skips_nan_and_never_panics` (668208b10) guarding the NaN-skip behavior change.

**Merge posture:** regression-gated and proven, but this changes greedy output on tied logits (hot path) — holding for human review rather than auto-merging.

PR #284, opened 2026-06-24
test(inference): harden speculative decode path (#282)

### LINKED ISSUE #282: Harden speculative path: greedy-equivalence CI test + mtp_verify_draft error-path rollback audit
Follow-ups from the adversarial audit of the speculative path (see #280 / #281). The speculative decode path has **zero CI coverage** — no workflow exercises speculation, so correctness bugs there (like #243 and #280) ship undetected by `e2e-parity`.

## 1. Add a generation-level greedy-equivalence test to CI (high value)

Both #243 (unverified `draft[0]`) and #280 (last-wins tie-break) were divergences from plain greedy that a unit test on a helper alone would not fully guard. Add a test that runs `generate_with_speculation` and asserts the emitted token sequence is **identical** to non-speculative greedy decode, on a fixed repetitive prompt where the n-gram speculator is guaranteed to fire and where top-logit ties occur. This is the regression that would have caught both bugs at the generation level. Wire it into CI (it's pure-CPU / mockable, no GPU needed).

## 2. Audit `mtp_verify_draft` error-path cache-rollback symmetry (low severity)

Review flagged: draft generation (`verifier.draft_tokens_with_logits(...)`) mutates the verifier/target caches *before* the points where `target.verify_tokens` / `rejection_sample_draft` can return `Err` (around `speculative.rs:1205`, `:1241`). Cache rollback (`rollback_cache_to`) only runs on the **success** path (`:1275-1283`). So an early `Err` leaves the caches advanced; a caller that retries or recovers after the error would see stale KV/GDN context and silent token misalignment.

Suggested fix: capture the start positions and, on any `Err`, restore both caches before returning (RAII guard or explicit rollback in the error arms). Add a test with a `MtpTargetVerifier` whose `verify_tokens` advances `cache_position` then returns `Err`, asserting the caches are restored.

Low severity (only bites on error-recovery/retry, not the happy path), but it's a latent correctness footgun on a path with no CI coverage.

_Source: adversarial audit of `crates/inference/src/speculative.rs`._

ohdearquant:
Resolves #282.

The speculative decode path had **no CI coverage exercising its divergence-from-greedy or error-recovery behavior** — the exact gaps where #243 and #280 shipped undetected. This adds two correctness guards (pure-CPU/mockable, runs in CI via `cargo test --workspace`).

## 1. Tie-break greedy-equivalence test
`generate_with_speculation_matches_greedy_with_ties`. The existing `matches_greedy` test uses one-hot logits, so the top logit is **never tied** — it cannot catch a tie-break regression. #280 was a last-wins tie-break divergence. This test places two entries at the same maximum (`l[2]=l[5]=1.0`) so `argmax`'s first-wins tie-break is exercised on the verify path every step, and asserts speculative output is token-identical to a reference greedy loop. The n-gram speculator fires on the repeating prompt, so the verify/tie-break path (not just plain greedy fallback) is what's checked — a last-wins regression would reject the draft and diverge, failing the assert.

## 2. `mtp_verify_draft` error-path cache rollback
`verifier.draft_tokens_with_logits(...)` advances the verifier KV cache and `target.verify_tokens(...)` advances the target KV/GDN cache **before** the points where an `Err` can propagate (flagged in review at `speculative.rs:~1205/1248`). Previously an early `Err` left the caches advanced, so a caller that retried or recovered would see stale KV/GDN context and silent token misalignment.

Each fallible site now restores both caches (and GDN, where applicable) to their captured pre-call positions before returning the error. **Ok-path behavior is byte-identical.** The same guard is mirrored into the test-only `mtp_verify_precomputed_draft`. Headless regression test `mtp_verify_error_restores_target_cache` (via an `ErroringTargetVerifier` whose `verify_tokens` advances the cache then errors) asserts the target cache is restored to its pre-call position on error.

## Verification
- `cargo test -p lattice-inference --lib speculative`: **59 passed, 0 failed** (incl. both new tests)
- `cargo clippy -p lattice-inference -- -D warnings`: clean
- Scope: one file (`crates/inference/src/speculative.rs`), +167/−6
- No `unwrap`/`expect`/`panic!` in library code

### DIFF
--- a/crates/inference/src/speculative.rs
+++ b/crates/inference/src/speculative.rs
@@ -1210,12 +1210,21 @@ pub fn mtp_verify_draft<T: MtpTargetVerifier>(
     let gdn_snap = target.snapshot_gdn_states();
 
     // Generate draft with per-token logits for probabilistic rejection sampling (ADR-050).
-    let mtp_draft = verifier.draft_tokens_with_logits(
+    // #282: if draft generation fails after advancing the verifier cache, restore both
+    // caches to their pre-call positions so the caller can retry or propagate cleanly.
+    let mtp_draft = match verifier.draft_tokens_with_logits(
         current_token_id,
         current_position,
         main_hidden_at_current_position,
         eos_token,
-    )?;
+    ) {
+        Ok(d) => d,
+        Err(e) => {
+            let _ = verifier.rollback_cache_to(mtp_start);
+            target.restore_gdn_states(&gdn_snap);
+            return Err(e);
+        }
+    };
     let draft = mtp_draft.tokens;
     let draft_logits = mtp_draft.logits;
     let draft_len = draft.len();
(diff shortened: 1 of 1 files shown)


ohdearquant:
**Adversarial review: flagged one soft issue** (effort=high)

All correctness dimensions passed:
- Ok-path is byte-identical (each `match { Ok(v) => v, ... }` binds exactly what `?` bound).
- Per-site rollback targets are correct: the `draft_tokens_with_logits` arm rolls back verifier + GDN only (target not yet advanced); the `verify_tokens` and `rejection_sample_draft` arms roll back verifier + target + GDN.
- Precomputed-helper mirror is aligned (target-only).
- Borrow/move of `gdn_snap` is sound.
- Tie-break test is a **genuine** #280 guard: the n-gram speculator fires on the repeating prompt, `argmax` is first-wins → 2, and a last-wins regression would diverge at the first verify/rejection step.
- Scope confined to `speculative.rs`; no new library `unwrap`/`expect`/`panic!`.

**Sole finding (#5, soft):** best-effort `let _ = rollback_cache_to(...)` swallows a *secondary* rollback failure, leaving the Err contract ambiguous.

**Addressed** in a0e1b4ab5 by documenting the best-effort Err contract on `mtp_verify_draft` (callers must treat the speculative context as invalid and rebuild on error). A full error-chaining rewrite across 4 sites was rejected (Π_AEP) — it would wrap/lose the original `InferenceError` variant for a rarely-hit error-recovery path, a worse tradeoff than the documented best-effort contract.

Issue #237, opened 2026-06-23
Speculative/MTP decoding: greedy path uses probabilistic clock-seeded sampler + 2 related concerns (dormant paths)


ohdearquant:
Three correctness concerns in the speculative-decoding / MTP paths, found by a hardening pass and TBV'd against the code. **All three are on dormant or opt-in paths for the shipping qwen3.5-0.8b model** (no MTP weights ship with it), so none affect default generation or the e2e-parity gate today. Filing for when MTP / self-spec model support is prioritized.

## 1. Greedy MTP path uses the probabilistic, clock-seeded rejection sampler — CONFIRMED real

`crates/inference/src/forward/metal_qwen35.rs` `generate_greedy_mtp` (~L8324):

```rust
let rs = crate::speculative::rejection_sample_draft(
    &[draft.token_id],
    std::slice::from_ref(&draft.logits),
    &verify_out.logits[0],
    std::slice::from_ref(&verify_out.logits[1]),
    false,   // <-- greedy = false
    None,    // <-- seed = None  => SpecRng::from_clock()
)
```

`rejection_sample_draft(greedy=true)` does argmax-only acceptance (`target_choice == draft_tokens[i]`) and argmax bonus — deterministic. `greedy=false` takes the probabilistic path with a **clock-seeded RNG** when `seed=None`. So a function named `generate_greedy_mtp`, entered only when `top_k<=1 && temperature<=0.0`, produces **non-deterministic** acceptance/bonus that is not argmax-equivalent to plain greedy `generate()`.

**Fix**: pass `greedy=true` (and an empty draft-logits slice, per the function's own doc) in the MTP greedy path.

**Test**: draft slot where target argmax `t != draft d`; assert `generate_greedy_mtp` output == plain greedy output token-for-token over identical logits/state.

**Reachability**: `use_mtp` requires `session.mtp.is_some()` (MTP weights loaded) + `enable_mtp`/`LATTICE_MTP` opt-in. qwen3.5-0.8b ships no MTP weights → unreachable by default.

## 2. Draft stop token short-circuits before target verification — plausible, needs confirm

`generate_greedy_mtp` (~L8289): when the draft model emits EOS/stop, the loop pushes `pending_token` and `break`s **before** verifying that draft token against the target. If the target argmax at that position is non-EOS, greedy MTP stops earlier than plain greedy would. Standard speculative decoding verifies the draft against the target and only stops if the target (or accept/bonus) is stop.

[...]

PR #287, opened 2026-06-24
fix(inference): greedy MTP verifies draft-EOS against target (#237 concern #2)

ohdearquant:
## Summary

Fixes concern #2 of issue #237: greedy MTP draft-EOS short-circuit.

The `generate_greedy_mtp` function had a short-circuit that fired when the draft model proposed a stop/EOS token — it flushed the pending token and broke, skipping the verify phase entirely. A wrong draft-EOS would silently truncate generation early, diverging from plain greedy `generate()` (which only stops when the *target* model's argmax is a stop token).

### Bug fix (core)
- Removed the `if is_stop(draft.token_id) { ... break; }` short-circuit in `generate_greedy_mtp`.
- Draft proposals are now always verified against the target, regardless of their token identity.

### Pure function extraction
- Added `MtpRoundOutcome` enum (`EmitAndStop` / `EmitAndContinue`) at module level (no `#[cfg]` gate — compiles everywhere).
- Added `pub fn mtp_greedy_round(pending, draft, accepted, bonus, is_stop)` at module level — pure, testable without Metal GPU.
- The Metal loop calls `super::mtp_greedy_round(...)` after all GPU state mutations (gdn_checkpoints repair, rollback, metrics).

### EOS-faithful terminal handling (A/B/C)
Plain greedy (`generate.rs` lines 292, 345) always pushes the stop token before breaking. Three terminal cases now match this contract:

| Case | Condition | Before (wrong) | After (correct) |
|------|-----------|----------------|-----------------|
| A | Prefill argmax is EOS | `generated_tokens: 0`, empty `token_ids` | `generated_tokens: 1`, `token_ids: [EOS]` |
| B | Accept + bonus is stop | `EmitAndStop([pending, draft])` | `EmitAndStop([pending, draft, bonus])` |
| C | Reject + replacement is stop | `EmitAndStop([pending])` | `EmitAndStop([pending, bonus])` |

### Tests (10 total, all platform-agnostic)
- `test_case_a_prefill_eos` — oracle == mtp when first pending is EOS
- `test_case_b_accept_bonus_stop` — oracle == mtp, accept path with stop bonus
- `test_case_c_reject_bonus_stop` — oracle == mtp, reject path with stop replacement
- `test_draft_eos_but_target_non_stop` — core regression for concern #2
- `test_draft_eos_accepted` — accepted EOS draft emits [pending, EOS]
- `test_target_eos_via_bonus` — full round with EOS at bonus position
- `test_mixed_accept_reject_then_eos` — mixed rounds ending at EOS
- `test_all_accepted_no_stop`, `test_all_rejected_no_stop` — non-stop baselines
- `test_randomised_sweep` — 400 seeded trials, EOS allowed at any position, asserts full `oracle == mtp` token-for-token equality

## Validation

```
cargo test -p lattice-inference --lib
# 968 passed; 0 failed

cargo clippy -p lattice-inference --features f16,metal-gpu -- -D warnings
# Finished dev profile (no warnings)

cargo clippy --workspace -- -D warnings
# Finished dev profile (no warnings)

cargo fmt --all && cargo fmt --all --check
# clean
```

## Scope

One file changed: `crates/inference/src/forward/metal_qwen35.rs`.
No perf-relevant changes to the Metal hot path. No new dependencies.

### DIFF
--- a/crates/inference/src/forward/metal_qwen35.rs
+++ b/crates/inference/src/forward/metal_qwen35.rs
@@ -8708,6 +8708,19 @@ kernel void gdn_chunk_norm_silu_c32(
         ) -> GenerateOutput {
             let cfg = self.engine.config.clone();
 
+            // Mirror plain greedy `generate()` (generate.rs:235): max_new_tokens == 0
+            // means "generate nothing".  Return before sampling so we never emit a
+            // token the caller did not ask for — including the case-A prefill-EOS path
+            // below, which would otherwise emit one stop token for a zero budget.
+            if gen_cfg.max_new_tokens == 0 {
+                return GenerateOutput {
+                    text: String::new(),
+                    token_ids: vec![],
+                    prompt_tokens: prompt_len,
+                    generated_tokens: 0,
+                };
+            }
+
             // Greedy argmax from prefill logits.
             let pending_first = prefill_logits
                 .iter()
@@ -8829,6 +8842,12 @@ kernel void gdn_chunk_norm_silu_c32(
                 //   reject → argmax(verify_out.logits[0])  (target's replacement at pos+1)
                 // Both are encoded in rs.bonus_token by `rejection_sample_draft`.
                 let bonus_token = rs.bonus_token.unwrap_or(0);
+                // The loop guard (`generated_ids.len() < max_new_tokens`) guarantees at
+                // least one slot remains here, but a round can emit up to three tokens
+                // (case B: [pending, draft, bonus]).  Clip every emission to the
+                // remaining budget so greedy MTP never exceeds `max_new_tokens` — plain
+                // greedy `generate()` enters only `max_new_tokens - 1` decode steps, so
+                // it stops at the cap even when the next token would be EOS.
                 match super::mtp_greedy_round(
                     pending_token,
                     draft.token_id,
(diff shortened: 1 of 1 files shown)


ohdearquant:
## Review: flagged one issue → resolved

The review confirmed the primary fix is live and correct (the real `generate_greedy_mtp` Metal path *does* call `mtp_greedy_round`, so the differential suite is not testing dead code) and found one real correctness gap:

**Finding #1 (cap overshoot) — FIXED in 7e8fa53d4.** The call site appended the per-round emission (`[pending, draft, bonus]` for accept+bonus-stop) without clipping to the remaining `max_new_tokens` budget, so a round entered at `len == max_new_tokens - 1` could overshoot by up to two tokens. Plain greedy `generate()` enters only `max_new_tokens - 1` decode steps and stops at the cap even before EOS. Fix:
- `max_new_tokens == 0` guard at the top of `generate_greedy_mtp` (mirrors `generate.rs:235`; also covers the case-A prefill-EOS path).
- Clip both `EmitAndStop` and `EmitAndContinue` emissions to the remaining budget.
- Mirror the clip in the test `simulate_mtp` driver; add cap-edge regression tests (case B at `max_len` 1/2, case C at `max_len` 1, `EmitAndContinue` clip) + a 400-trial small-cap randomised sweep asserting MTP never exceeds the cap and equals the plain-greedy oracle.

**Finding #2 (self-spec sibling) — out of scope, filed as #289.** The review independently confirmed `generate_greedy_self_spec` (`LATTICE_SELF_SPEC=1`, dormant) has the same EOS-convention divergence class but does **not** have concern #2's specific draft-EOS-before-verify short-circuit. Tracked separately.

Local verification (same machine): `cargo test -p lattice-inference --lib` 972 passed / 0 failed; `cargo clippy -p lattice-inference --features f16,metal-gpu -- -D warnings` clean.

Issue #149, opened 2026-05-31
perf(embed): NEON vrsqrteq_f32 + Newton iteration for normalize


ohdearquant:
## Context

Follow-up from #102 (partially closed by PR #139).

PR #139 delivered P0-S1 (batch amortization) and P0-S2 (simsimd bench). P0-S3 (NEON normalize) added a NEON unrolled path but uses scalar `norm_sq.sqrt()` + `1.0 / norm` — not the `vrsqrteq_f32` + Newton iteration that the issue specified.

## What's needed

Replace the scalar sqrt + reciprocal in `crates/embed/src/simd/` NEON normalize with:
```
vrsqrteq_f32 → 1 Newton–Raphson step → vmulq_f32
```

This is the standard ARM NEON fast-inverse-sqrt pattern (~2-3 cycles vs ~12 for scalar sqrt + div). If the accuracy bound (max error vs scalar) exceeds 1e-6 on normalize output, document it and keep scalar.

## Validation

- Bench before/after with `cargo bench -p lattice-embed --bench simd -- normalize`
- Verify max-diff vs scalar path < 1e-6

PR #325, opened 2026-06-25
harden(embed): normalize NaN-norm guard scalar/SIMD byte-consistency

ohdearquant:

## Problem

The AVX-512, AVX2, and NEON `normalize` kernels in `crates/embed/src/simd/normalize.rs` guarded the divide-by-norm step with `norm == 0.0` (or `norm_sq == 0.0` on NEON). That comparison is **false for NaN**, so a vector containing a NaN element — which produces a NaN L2 norm — falls through the guard and is scaled by a NaN `inv_norm`, turning **every lane into NaN**.

The scalar reference (`normalize_scalar`) guards with `if norm > 0.0`, which is **false for NaN**, leaving the vector unchanged. So the SIMD and scalar paths **diverged on any NaN-containing input** — a silent scalar-vs-SIMD inconsistency in a crate whose hand-written kernels are explicitly meant to be byte-consistent with their scalar fallbacks.

## Fix

Align all three SIMD guards with the scalar semantics using `norm.is_nan() || norm <= 0.0` (the lint-clean equivalent of `!(norm > 0.0)`).

Note: clippy's `neg_cmp_op_on_partial_ord` lint suggests rewriting `!(x > 0.0)` to `x <= 0.0` — that rewrite would **silently drop the NaN case and reintroduce this exact bug**, which is why the explicit `is_nan() || <= 0.0` form is used.

The NEON subnormal-norm fallback (issue #149) is preserved: a subnormal-but-positive `norm_sq` still passes the guard and is handled by the existing finite-fallback; only zero/NaN short-circuit.

## Test

`test_normalize_nan_norm_matches_scalar` (dim=128, NaN in both the SIMD main body and the scalar tail) asserts the SIMD result is **byte-identical** (`to_bits()`) to both the scalar result and the unchanged input. It fails red on every prior SIMD guard (verified locally: reverting just the NEON guard → test FAILS; restoring → PASS).

- `cargo test -p lattice-embed --lib simd` → 88 passed
- `cargo clippy -p lattice-embed --all-targets -- -D warnings` → clean

## Scope / safety

Off the hot path: the change is a single once-per-`normalize()`-call guard, not per-element. Grammar/inference untouched. `make bench-compare` output posted as a comment below (expected flat).

### DIFF
--- a/crates/embed/src/simd/normalize.rs
+++ b/crates/embed/src/simd/normalize.rs
@@ -125,7 +125,11 @@ unsafe fn normalize_avx512_unrolled(vector: &mut [f32]) {
     }
 
     let norm = norm_sq.sqrt();
-    if norm == 0.0 {
+    // Match `normalize_scalar`, which only scales when `norm > 0.0`. Rejecting
+    // NaN here too (a NaN element makes `norm` NaN) leaves the vector
+    // byte-identical to the scalar path instead of scaling by a NaN inv_norm.
+    // `is_nan() || <= 0.0` is the lint-clean equivalent of `!(norm > 0.0)`.
+    if norm.is_nan() || norm <= 0.0 {
         return;
     }
 
@@ -228,7 +232,10 @@ unsafe fn normalize_avx2_unrolled(vector: &mut [f32]) {
     }
 
     let norm = norm_sq.sqrt();
-    if norm == 0.0 {
+    // Match `normalize_scalar`: reject 0.0 and NaN alike so a NaN-containing
+    // vector is left unchanged rather than scaled by NaN. `is_nan() || <= 0.0`
+    // is the lint-clean equivalent of `!(norm > 0.0)`.
+    if norm.is_nan() || norm <= 0.0 {
         return;
     }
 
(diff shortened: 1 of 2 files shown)


ohdearquant:

## bench-compare (`simd_normalize` group, criterion A/B)

`make bench-compare`'s diff parser hit the known two-worktree cache glitch (head cache missing `base/estimates.json`), so I ran the `simd_normalize/simd` group directly with criterion baselines (base = `origin/main` 80b3879fc, head = this branch):

| dim  | change (median) | p-value | verdict |
|------|-----------------|---------|---------|
| 384  | +0.05%          | 0.98    | No change |
| 768  | +2.60%          | 0.09    | No change |
| 1024 | −2.36%          | 0.08    | No change |
| 1536 | +0.15%          | 0.62    | No change |

All four: **"No change in performance detected" (p > 0.05)**. Expected — the change is a single once-per-`normalize()`-call guard (`norm == 0.0` → `norm.is_nan() || norm <= 0.0`, the same one `fcmp` in the hot case), not a per-element edit.

## review

Review flagged one minor issue. Confirmed `is_nan() || x <= 0.0` ≡ `!(x > 0.0)` for all f32 edge classes (signed zero, ±inf, NaN, subnormals), the NEON positive-subnormal `norm_sq` path still reaches the rsqrt/fallback code, and no other NaN/Inf scalar-vs-SIMD divergence exists in the file. No new unsafe/soundness issue.

The single fix-up (test under-coverage: `dim=128` is a multiple of all three chunk widths so both NaNs were in the SIMD main body, not the scalar tail) is addressed in `d4bbcff76` — the test now parametrizes over dims {129, 145, 200} with NaN placed once in the main body and once in the scalar tail. Still fails red on every prior SIMD guard.

PR #331, opened 2026-06-25
fix(inference): recover SentencePiece byte-fallback typing from tokenizer.json (#255)

### LINKED ISSUE #255: tokenizer: SentencePiece byte-fallback typing dropped when loading from tokenizer.json
## Summary

When a SentencePiece model is loaded via `from_tokenizer_json_str` (the `tokenizer.json` path, reachable through `common.rs`), every vocab entry is forced to `SentencePieceType::Normal`. The `.model` protobuf loader (`parse_sentencepiece_model`) does the opposite: it populates `byte_fallback_ids` and tags `<0xNN>` pieces as `SentencePieceType::Byte`. So a SP model with byte-fallback semantics loses that typing when it arrives as JSON.

## Evidence

- `crates/inference/src/tokenizer/sentencepiece.rs:233` — `from_tokenizer_json_str` hard-codes `kind: SentencePieceType::Normal` for all entries
- `crates/inference/src/tokenizer/sentencepiece.rs:290,302` — `parse_sentencepiece_model` populates `byte_fallback_ids`
- `crates/inference/src/tokenizer/sentencepiece.rs:718` — `.model` loader sets `SentencePieceType::Byte`
- Reachable from `crates/inference/src/tokenizer/common.rs:228`

## Impact (narrow)

Only affects **SP-Unigram-via-tokenizer.json models that rely on byte-fallback** for OOV/byte-preserving reconstruction. `viterbi_encode` cannot route missing chars through byte fallback (falls back to UNK), so exact reconstruction of byte-heavy/rare-Unicode input is broken for those models.

**Does NOT affect Qwen3.5** (byte-level BPE, not SentencePiece) or any currently-shipped lattice model. This is a latent correctness gap for a model class lattice does not currently run.

## Fix sketch

Parse token-type metadata from `tokenizer.json` where available, or infer `Byte` for `<0xNN>`-shaped entries and route those IDs into `byte_fallback_ids`, matching `parse_sentencepiece_model`. **Touches the encode path** — needs an SP-Unigram-with-byte-fallback test fixture before changing, so deferring the fix until there is a model to regression-test against.

## Rejected during the same sweep (documented for the record)

The sweep also flagged "silent dropping of unknown/OOV IDs in BPE decode" (`detokenize.rs:24-31`, `bpe.rs:699-702`). **Rejected on TBV:** skipping IDs not in the byte_decoder is deliberate and correct — those are special/control tokens (`<|im_start|>` etc.) that have no byte-level representation and must not be emitted as literal text or `�`. The streaming detokenizer's UTF-8 buffering across token boundaries was independently confirmed correct (`detokenize.rs:73-128`).

ohdearquant:

Closes #255.

## Problem

`from_tokenizer_json_str` (the tokenizer.json Unigram loader) hard-coded every piece to `SentencePieceType::Normal`. tokenizer.json carries no SentencePiece piece-type tag, so `<0xNN>` byte-fallback pieces were inserted into the trie as literal text and never routed into `byte_fallback_ids`. Result: OOV byte fallback in `viterbi_encode` silently failed for any SP model loaded from `.json` — OOV chars fell through to `<unk>` instead of decomposing into byte tokens. The `.model` protobuf loader was unaffected (it reads the piece-type enum, `6 => Byte`).

## Fix

Infer the `Byte` type from the `<0xNN>` shape via the existing `parse_byte_piece` helper, mirroring `from_parsed_model`'s kind-routing. Only mistyped byte pieces change behavior; `Normal` pieces are untouched (single localized site, +28/-1).

## Gate (red/green)

`test_json_loader_routes_byte_pieces_to_fallback`:
- asserts `byte_fallback_ids[0x41]`/`[0xE2]` are populated after a `.json` load (was `None` pre-fix → `Some(id)` post-fix)
- asserts a `Normal` piece still tokenizes to its id unchanged (guards against over-typing)

Verified RED pre-fix (`left: None, right: Some(2)`), GREEN post-fix. Full tokenizer suite 35/35, `clippy --tests -D warnings` clean, fmt clean.

## bench-compare

N/A — load-time typing inference + encode routing, not an inference/embed hot-path change; not covered by any Criterion group. e2e-parity does not exercise SentencePiece byte fallback, so the synthetic internal-state red/green test is the regression gate.

### DIFF
--- a/crates/inference/src/tokenizer/sentencepiece.rs
+++ b/crates/inference/src/tokenizer/sentencepiece.rs
@@ -210,6 +210,16 @@ impl SentencePieceTokenizer {
                 InferenceError::Tokenizer("tokenizer.json missing model.vocab".into())
             })?;
 
+        // SentencePiece serializes a model-level `byte_fallback` flag in
+        // tokenizer.json. Shape-inference of `<0xNN>` pieces is the recovery
+        // signal when the flag is true or absent (HF byte-fallback models set
+        // it true; some recovery paths omit it). Only an explicit `false`
+        // suppresses inference — so a model that genuinely declares no byte
+        // fallback keeps a literal `<0xNN>` text token as Normal (#255 review).
+        let byte_fallback_enabled = json_path(&root, &["model", "byte_fallback"])
+            .and_then(JsonValue::as_bool)
+            != Some(false);
+
         let mut pieces = Vec::with_capacity(vocab_entries.len());
         for entry in vocab_entries {
             let values = entry.as_array().ok_or_else(|| {
@@ -232,7 +242,7 @@ impl SentencePieceTokenizer {
             // inserted into the trie as literal text — leaving byte_fallback_ids
             // empty and breaking OOV byte fallback in viterbi_encode. Recover the
             // Byte type from the `<0xNN>` shape, mirroring the .model loader (#255).
-            let kind = if parse_byte_piece(piece).is_some() {
+            let kind = if byte_fallback_enabled && parse_byte_piece(piece).is_some() {
                 SentencePieceType::Byte
             } else {
                 SentencePieceType::Normal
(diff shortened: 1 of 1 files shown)


ohdearquant:

**Adversarial review: flagged one minor issue.** Confirmed ID integrity (ids come from vocab index, independent of kind), end-state symmetry with the `.model` loader (byte pieces → `byte_fallback_ids`, out of trie), and no `unwrap`/`expect`/`panic` introduced.

The one finding: pure `<0xNN>` shape inference could retype a semantically-Normal text token to Byte (rated exotic/edge-case in review, reachable in principle for custom vocabs).

**Resolved in 11cc68246**: the loader now reads the model-level `byte_fallback` flag and shape-infers only when it is `true` or absent. An explicit `byte_fallback: false` keeps a literal `<0xNN>` piece Normal. Guard test `test_json_loader_respects_explicit_byte_fallback_false` added. Full sentencepiece suite 9/9, clippy `--tests -D warnings` clean.

Issue #93, opened 2026-05-24
feat(serve): OpenAI-compatible chat completions HTTP API


ohdearquant:
## Context

The OpenAI `/v1/chat/completions` HTTP API is the de-facto standard for LLM clients. Any tool that speaks it gets a massive ecosystem for free (LangChain, LlamaIndex, OpenAI SDK, Continue, Aider, Cursor, etc).

Ollama, llama.cpp, vLLM, MLX-LM all expose this. Lattice should too.

## Goal

When `lattice serve` is running (issue #91), expose:
- `POST /v1/chat/completions` — chat completion with optional streaming (SSE)
- `POST /v1/completions` — raw text completion
- `POST /v1/embeddings` — embedding generation (via lattice-embed)
- `GET /v1/models` — list available models

So this works:

```bash
curl http://localhost:8080/v1/chat/completions \
  -d '{"model":"qwen3.5-0.8b","messages":[{"role":"user","content":"hi"}]}'
```

And this works:

```python
from openai import OpenAI
client = OpenAI(base_url="http://localhost:8080/v1", api_key="not-needed")
resp = client.chat.completions.create(model="qwen3.5-0.8b", messages=[...])
```

### Framework
`axum` or `actix-web` — axum preferred (tokio-native, simpler).

### Request shape (subset of OpenAI spec — full spec is huge)
v1 supports:
- `model` (string, required)
- `messages` (chat format, role/content)
- `temperature`, `top_p`, `top_k`
- `max_tokens`
- `stream` (bool — SSE response if true)
- `stop` (string or array)
- `seed`

v1 ignores (returns warning header):
- `tools` / `function_calling` (Phase 2)
- `response_format` (Phase 2 — needs grammar integration)
- `logit_bias`, `logprobs`, `top_logprobs` (Phase 2)
- `n > 1`, `user`, `presence_penalty`, `frequency_penalty` (Phase 2)

### Chat template
Apply the model-specific chat template before tokenization. Qwen3.5 uses ChatML format; templates live in the model's `tokenizer_config.json`.

### Streaming
SSE format: `data: {...JSON delta...}\n\n`, terminated with `data: [DONE]\n\n`.

## Acceptance

- OpenAI Python SDK works against lattice with no client changes (just `base_url`)
- Streaming works (`stream=True` returns SSE delta tokens)
- Chat template applied correctly (verify against ollama's output on identical prompt)
- Concurrent requests handled per #91's queue model

## Compatibility scope

[...]

Issue #92, opened 2026-05-24
feat(serve): long-running daemon for model serving (avoid cold-start)


ohdearquant:
## Context

Loading Qwen3.5-0.8B weights takes ~2-5s (memory map + Metal buffer upload). For batch workloads (embedding pipelines, RAG, evaluation) this is acceptable. For interactive use it's painful — every CLI invocation pays the load cost.

Ollama solves this with a daemon that holds models in memory and serves multiple requests. Lattice needs the equivalent.

## Goal

A long-running `lattice` daemon process that:
1. Loads models on demand (first request → load, subsequent → reuse)
2. Supports multiple models in memory simultaneously (LRU eviction when GPU/RAM constrained)
3. Streams tokens via stdout, HTTP, or Unix socket
4. Survives multiple client invocations without re-loading

### Process model
- `lattice serve [--port 8080] [--socket /tmp/lattice.sock]` starts the daemon
- Daemon runs in foreground (avoids systemd entanglement); user can `&` or `nohup` it
- Lock file at `~/.lattice/serve.pid` prevents double-start

### Storage
- Loaded models held in a `Arc<RwLock<HashMap<ModelId, LoadedModel>>>`
- LRU eviction when total VRAM > configured limit (default: 80% of detected Metal allocator size)
- Per-model KV cache pool (issue #93)

### Protocol
v1: simple line-protocol over Unix socket (newline-delimited JSON requests + SSE-style token stream responses). Easy to debug with `nc`.
v2: HTTP + OpenAI API (issue #92).

### Concurrency
- Single inference at a time per model (Metal isn't trivially shareable across threads)
- Multiple model serves can run concurrently if they fit
- Request queue with backpressure (`429` if queue full)

## Acceptance

- `lattice serve` + `lattice chat ... --daemon` round-trip works
- Sub-second response start for cached models (vs current ~2-5s cold load)
- Survives 100 sequential requests without leaking
- Handles SIGTERM gracefully (finish in-flight request, then exit)

## Priority

P1 — blocks daily usage. Tied to #91 (CLI) and #93 (OpenAI API).

## Related

- #90 — CLI entry point (`lattice serve`)
- #92 — OpenAI-compatible HTTP API on top
- #84 — cross-framework bench; daemon makes lattice fair to compare against ollama

ohdearquant:
[...]

PR #332, opened 2026-06-25
feat(serve): OpenAI-compatible SSE streaming for chat completions

ohdearquant:

## Summary

`POST /v1/chat/completions` with `stream=true` now returns a `text/event-stream` of OpenAI `chat.completion.chunk` events instead of a `400` rejection. This closes the last HTTP-layer gap toward an ollama-style interactive serving experience (relates to #93, the OpenAI-API serving track).

The engine already had the hard part: `Qwen35Model::generate_streaming` (a deliberate, parity-safe copy of `generate` with a UTF-8-incremental `on_token` callback). This PR is purely the HTTP wiring — **no inference/kernel/forward code is touched.**

## Design

- A blocking `generate_streaming` call runs on the `spawn_blocking` pool and feeds incremental, UTF-8-complete text deltas through a `futures::channel::mpsc::unbounded` channel (no new external dependency — opts the inference crate into the existing workspace `futures` dep).
- The async SSE handler drains the channel and maps each message to a chunk event. Event order matches the OpenAI spec:
  1. role chunk — `delta: {"role":"assistant"}`
  2. one content chunk per delta — `delta: {"content":"..."}`
  3. finish chunk — `delta: {}`, `finish_reason` `stop`/`length` (same inference logic as the non-streaming path)
  4. literal `data: [DONE]` sentinel
- Client disconnects are safe (channel send errors are ignored; the blocking task drains harmlessly).
- The **non-streaming path is byte-identical** to before (same `id`/`object`/`created`/`usage`).

## Verification

End-to-end on **Qwen3.5-0.8B** (`lattice serve` + `curl -N`), greedy (`temperature=0`), same prompt both ways:

- **Streaming** emitted: role → `"\n\n"` → `"\n\n"` → `"Hello"` → `"!"` → finish(`stop`) → `[DONE]`
- **Non-streaming** `content`: `"\n\n\n\nHello!"`
- Streamed deltas concatenate to **exactly** the non-streaming content → cross-path parity proven.

Gates (independently re-run):
- `cargo test -p lattice-inference --bin lattice` → **43 passed**
- `cargo clippy -p lattice-inference --bin lattice -- -D warnings` → clean
- `cargo fmt --check` → clean
- Doc-lint → passed

Tests: inverted `reject_unsupported_stream_true` → `..._ok` (stream is no longer rejected), added two chunk-serialization unit tests (content delta omits `finish_reason`; finish chunk emits `"finish_reason":"stop"` with empty `delta`).

## bench-compare

**N/A** — the change is confined to `bin/lattice.rs` (the HTTP serve binary). No benchmarked library code (decode/prefill/forward/sampling kernels) is modified, so the Criterion suite cannot show a delta. `generate_streaming` itself is pre-existing and unchanged. The e2e-parity CI gate (greedy token match vs HF) still applies and is unaffected, since the non-streaming generate path is untouched.

## Scope note

This is the streaming wire format only. Multi-model serving, Unix-socket transport, and request-queue backpressure (#92's daemon scope) remain separate follow-ups.

### DIFF
--- a/crates/inference/Cargo.toml
+++ b/crates/inference/Cargo.toml
@@ -31,6 +31,7 @@ thiserror.workspace = true
 tracing.workspace = true
 clap = { workspace = true }
 axum = { workspace = true }
+futures = { workspace = true }
 tokio = { workspace = true }
 ureq = { version = "2", features = ["tls"], optional = true }
 sha2 = { workspace = true, optional = true }
--- a/crates/inference/src/bin/lattice.rs
+++ b/crates/inference/src/bin/lattice.rs
@@ -760,10 +760,15 @@ mod serve {
             // on the blocking thread pool and feed incremental deltas into an
             // unbounded MPSC channel.  The async SSE handler drains the channel
             // and converts each message to an OpenAI `chat.completion.chunk` event.
-            // An unbounded channel is safe here because the model produces tokens
-            // at most as fast as the GPU/CPU can generate them and the SSE write
-            // loop is the bottleneck, so back-pressure from the network naturally
-            // limits the channel depth in practice.
+            // An unbounded channel is acceptable here because the channel depth is
+            // bounded by `max_tokens` (capped at `max_tokens_cap`): the producer
+            // sends at most one `Delta` per generated token and generation halts at
+            // the cap, so the worst-case buffer is a few thousand short strings —
+            // the same order the non-streaming path already holds as one buffered
+            // string. There is no true backpressure (an unbounded send never
+            // blocks); if the client disconnects mid-stream the producer keeps
+            // generating to the cap and the ignored send errors drain harmlessly.
+            // Per-token backpressure / disconnect-cancellation is a future refinement.
             let (tx, rx) = futures::channel::mpsc::unbounded::<StreamMsg>();
 
             let stream_id = response_id.clone();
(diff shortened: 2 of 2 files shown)


ohdearquant:

## Critic review — flagged one issue

The review (effort=high) verified all five correctness checks against the diff:

1. **SSE event ordering** — PASS (role → content → finish → `[DONE]`)
2. **Non-streaming path parity** — PASS (id/object/created/usage/finish_reason identical to pre-commit)
3. **Channel lifecycle** — PARTIAL: no panic/deadlock; flagged the unbounded channel as a memory risk under a slow client
4. **finish_reason / >max_tokens** — PASS (same length/stop derivation; `>max` → `Failed`)
5. **unwrap/expect/panic in non-test code** — PASS (only `unwrap_or_default` on JSON serialization fallback)

### TBV of the one finding (channel memory)

The "unbounded memory" is **bounded in practice by `max_tokens`** (capped at `max_tokens_cap`): the producer sends at most one `Delta` per generated token and generation halts at the cap, so the worst-case buffer is a few thousand short strings — the same order the non-streaming path already holds as a single buffered output string. It is **not** a memory-exhaustion risk, and the server is single-inference-at-a-time per the daemon design.

The review did correctly catch that the original code comment was **inaccurate** (it claimed network backpressure limits the channel — false, since an unbounded send never blocks). Fixed in `2538faebb`: the comment now states the real `max_tokens` bound and the disconnect behavior, and flags per-token backpressure / disconnect-cancellation as a future refinement. No functional change.

Net: no blocking defect; the one accuracy issue (the comment) is resolved. Feature is build-for-review; merge is pending maintainer review.

PR #333, opened 2026-06-25
feat(inference): OpenAI string-level stop sequences in serve

ohdearquant:

## What

Adds OpenAI-compatible **string-level stop sequences** to `lattice serve`. The `stop` field (a JSON string or an array of 1–4 non-empty strings) is now accepted; when any stop string appears in the generated output, generation halts, the matched text and everything after it is excluded, and `finish_reason` is `"stop"`.

## How

- **Engine**: new opt-in `GenerateConfig.stop_strings: Vec<String>` (default empty). Honored in **both** `generate` (non-streaming) and `generate_streaming` so the field is never silently ignored in one path.
  - **Parity-safe gate**: when `stop_strings` is empty, both functions take a fast path that is **byte-for-byte identical** to before this change (the existing `decode_loop` / inline streaming loop are untouched). The e2e-parity CI gate pins this greedy path.
  - **Non-streaming**: `decode_loop_with_stops` accumulates decoded text, checks `earliest_stop_match` after each token, truncates `full` at the stop and breaks. A `stopped` flag prevents the detok-finish tail from re-polluting the truncated output.
  - **Streaming**: `StopStreamer` hold-back buffer withholds up to `max_stop - 1` bytes so a partial stop prefix is never streamed before it can be confirmed (UTF-8-boundary safe). On a hit it emits only the pre-stop slice; a `stopped` flag makes the final flush a no-op (no double-emit).
- **Serve**: removed the `stop` rejection; added `parse_stop_strings` (string → 1 entry; array of 1–4 non-empty strings; everything else → `400 invalid_stop`).

## Tests

- `StopStreamer`: hold-back correctness, no-double-emit on stop split across deltas, multibyte UTF-8, stop at delta boundary, natural-end flush.
- `earliest_stop_match`: earliest-of-many, no-match, index 0, multibyte.
- `parse_stop_strings`: string/array acceptance, empty-array / >4 / non-string / empty-string rejection.
- `cargo test -p lattice-inference` stop suite: **27 passing**. Build (`f16,metal-gpu`), clippy `-D warnings`, fmt all clean.

## Real-model e2e (qwen3.5-0.8b, greedy seed=1)

| Request `stop` | Output | finish |
|---|---|---|
| _(none)_ baseline | `\n\n\n\n1, 2, 3, 4, 5, 6, 7, 8, 9, 10` | stop |
| `"5"` | `\n\n\n\n1, 2, 3, 4, ` | stop |
| `[", 7", ", 99"]` | `\n\n\n\n1, 2, 3, 4, 5, 6` | stop |
| `"NOTPRESENT"` | `\n\n\n\n1, 2, 3, 4, 5, 6, 7, 8, 9, 10` (== baseline) | stop |
| `[]` / 5 elements | — | HTTP 400 |

The non-present stop returning byte-identical-to-baseline output confirms the non-empty path does not corrupt output when there's no match.

## Bench

N/A — bin/HTTP + decode-path-gated feature; no kernel or hot-loop change. The empty-`stop_strings` path is byte-identical, so there is no throughput impact on existing requests.

### DIFF
--- a/crates/inference/src/bin/bench_decode_ab.rs
+++ b/crates/inference/src/bin/bench_decode_ab.rs
@@ -113,6 +113,7 @@ fn run() -> Result<(), Box<dyn std::error::Error>> {
         enable_thinking: false,
         enable_mtp: None,
         grammar: None,
+        stop_strings: vec![],
     };
 
     // Same continuation prompt as the original bench, single user turn.
--- a/crates/inference/src/bin/bench_decode_slopefit.rs
+++ b/crates/inference/src/bin/bench_decode_slopefit.rs
@@ -174,6 +174,7 @@ fn run() -> Result<(), Box<dyn std::error::Error>> {
         enable_thinking: false,
         enable_mtp: None,
         grammar: None,
+        stop_strings: vec![],
     };
 
     eprintln!(
--- a/crates/inference/src/bin/chat_metal.rs
+++ b/crates/inference/src/bin/chat_metal.rs
@@ -552,6 +552,7 @@ fn run() -> Result<(), Box<dyn std::error::Error>> {
         enable_thinking: true,
         enable_mtp: None,
         grammar: None,
+        stop_strings: vec![],
     };
 
     // ── JSON streaming mode (used by Lattice Studio app) ────────────────────
--- a/crates/inference/src/bin/lattice.rs
+++ b/crates/inference/src/bin/lattice.rs
@@ -579,6 +579,20 @@ mod serve {
     }
 
     // -----------------------------------------------------------------------
+    // Helpers
+    // -----------------------------------------------------------------------
+
+    /// Maps a `GenerateOutput` to the OpenAI `finish_reason` string.
+    ///
+    /// Returns `"stop"` when the library explicitly ended generation via a stop
+    /// condition (EOS token, stop-token-id, or stop-string match); `"length"` when
+    /// the token budget was exhausted without a stop condition.
+    pub(super) fn finish_reason_for(
+        output: &lattice_inference::model::qwen35_config::GenerateOutput,
+    ) -> &'static str {
+        if output.stopped { "stop" } else { "length" }
+    }
+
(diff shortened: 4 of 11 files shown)


ohdearquant:

## critic verdict: flagged one deferred issue

Independent adversarial review (high effort) over the full diff. Findings 1–6 confirm correctness on the load-bearing risks:

1. **Parity** — empty `stop_strings` takes the `else`-gated fast path; `generate`/`generate_streaming`/`decode_loop` are byte-for-byte unchanged. ✅
2. **No double-emit** — `StopStreamer` sets `emitted = full.len()` + `stopped` on a hit, and `finish` is a no-op when stopped; no byte can be emitted twice. ✅
3. **Hold-back + UTF-8** — `(max_stop-1)` hold-back with char-boundary clamp; no mid-codepoint slice. ✅
4. **Tail handling** — `decode_loop_with_stops` suppresses the detok-finish tail after a string stop; the no-stop path still re-checks for a stop completing in the tail. ✅
5. **`parse_stop_strings`** — accept/reject policy internally consistent with tests + doc. ✅
6. **No new panic surfaces** from attacker-controlled stop input. ✅

### Finding 7 [FIX] — deferred to #334 with rationale

> `generated_tokens` counts the stop-completing token, so a stop that lands exactly on the `max_tokens`-th token is misclassified as `finish_reason="length"` instead of `"stop"`.

Real but narrow (cosmetic enum value only; content truncation is correct). **Not fixed inline** because there is no reliable serve-only fix and the clean engine-signal fix requires threading a stop-reason discriminator through ~40 `GenerateOutput` constructors across the parity-pinned metal/cpu/neon paths — a cross-cutting refactor (that should also unify the existing `stopped_by_eos` on the *other* `GenerateOutput`), disproportionate to ride on this feature PR. Tracked in **#334** with acceptance criteria.

Feature is build-for-review: green CI + 27 stop tests + real-model e2e, parked for @ohdearquant's merge call.

PR #340, opened 2026-06-25
harden(inference): reject out-of-range SentencePiece special ids at construction

ohdearquant:

## Summary

A malformed `tokenizer.json` (`model.unk_id`) or native `.model` trainer_spec (unk/bos/eos/pad, fields 40-43) can declare a special-token id **past the end of the vocabulary**. The value is cast `as u32` with no bounds check at parse time (`sentencepiece.rs:264` JSON, `:756-759` protobuf). `viterbi_encode` emits `unk_id` directly for unmatched input (`:607`) and bos/eos are prepended/appended, so an out-of-range id flows straight into the model's embedding gather (out-of-bounds read) or the `id_to_piece` decode lookup.

This is the SentencePiece sibling of the BPE (#338) and WordPiece (#339) malformed-tokenizer hardening: a malformed-but-parseable tokenizer file must error at load, not corrupt/crash the model forward.

## Fix

Both `from_tokenizer_json_str` and `from_model_file` funnel through `from_parsed_model` (private fn), so a single validation point after id resolution covers **every** source (JSON, protobuf, TemplateProcessing, vocab-name). Reject when any resolved `unk_id`/`pad_id`/`bos_id`/`eos_id` is `>= pieces.len()`.

## Provenance / TBV

Found by an automated robustness discovery pass on `sentencepiece.rs`. The pass also confirmed the **zero-length non-advancing-match hang class (#338/#339) is NOT present** here — `ByteTrie::matches` advances ≥1 byte and Viterbi advances by `char_len`. TBV-verified against source: the vocab-name `known_special_id` path is inherently in-range (returns a `pieces` index); only the JSON/protobuf/template casts are unchecked. The initial framing was refined accordingly.

## Test (red/green proven)

`test_out_of_range_special_id_rejected_at_construction`: a 3-piece vocab with `unk_id = 5`.
- **Red**: with the guard removed, `from_parsed_model` succeeds (would emit id 5 at tokenize time).
- **Green**: with the guard, construction errors.

All 10 sentencepiece tests pass; clippy clean.

## bench-compare

N/A — construction-time validation, runs once at tokenizer load; no change to the tokenize hot path or any kernel.

### DIFF
--- a/crates/inference/src/tokenizer/sentencepiece.rs
+++ b/crates/inference/src/tokenizer/sentencepiece.rs
@@ -967,6 +967,21 @@ mod tests {
     }
 
     #[test]
+    fn test_out_of_range_unk_id_rejected_from_json() {
+        // The tokenizer.json loader path (from_tokenizer_json_str -> from_parsed_model)
+        // casts `model.unk_id` `as u32` with no bounds check. A malformed file declaring
+        // unk_id past the vocab end must be rejected at construction, mirroring the
+        // .model path covered by from_pieces_for_test above. Vocab has 2 pieces (ids 0,1)
+        // and unk_id = 4.
+        let json = r#"{"model":{"type":"Unigram","unk_id":4,"vocab":[["<unk>",0.0],["▁a",-1.0]]}}"#;
+        let result = SentencePieceTokenizer::from_tokenizer_json_str(json);
+        assert!(
+            result.is_err(),
+            "an out-of-range unk_id from tokenizer.json must be rejected at construction"
+        );
+    }
+
+    #[test]
     fn test_normalize_adds_metaspace() {
         let tokenizer = synthetic_tokenizer();
         assert_eq!(tokenizer.normalize("hello world"), "▁hello▁world");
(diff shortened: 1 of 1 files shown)


ohdearquant:

Addressed the review's coverage note: added `test_out_of_range_unk_id_rejected_from_json` (commit c6ffcae2d), which exercises the `from_tokenizer_json_str → from_parsed_model` path with an out-of-range `unk_id` declared in `tokenizer.json`. The existing `from_pieces_for_test` test covers the `.model`-style path. Both entry points funnel through the identical `id >= vocab_len` validation loop (which also covers `pad_id`/`bos_id`/`eos_id`), so the two tests together prove the guard rejects at construction on both loader paths.

Deferring the second (explicitly non-blocking) review note — validating `u64`/varint special-token IDs before the `as u32` cast — as a follow-up. Rationale: an oversized value that wraps under `as u32` produces an *in-range-but-wrong* token id, which is a correctness nit, not the out-of-bounds safety hole this PR closes. The construction guard fully covers the OOB case regardless of wrap. Will file a separate tracking issue for the pre-cast validation.

PR #341, opened 2026-06-25
harden(inference): reject malformed tokenizer.json vocab ids at load

ohdearquant:

## Summary

Three robustness gaps in the shared tokenizer parse layer (`crates/inference/src/tokenizer/common.rs`), all reachable from `load_tokenizer` on a malformed-but-parseable `tokenizer.json`. Surfaced by a discovery pass; each verified against source before fixing.

### 1. `u64` vocab id silently wrapped to `u32` (silent token aliasing)
`json_object_to_vocab` cast a file-supplied `u64` id straight to `u32`. An id past `u32::MAX` (e.g. `4294967296` → `0`) wrapped to a small in-range id and aliased an unrelated token with no error. Now rejected via `u32::try_from`. The same overflow guard is applied to the two sibling cast sites flagged in review:
- `parse_added_tokens` (lenient-by-contract) — skips an oversized id rather than wrapping it.
- `resolve_template_token_id` (TemplateProcessing) — treats an oversized id as absent.

### 2. `invert_vocab` unbounded dense-table allocation (load-time DoS)
`invert_vocab` sized `id_to_token` to `max_id + 1` with no bound. `invert_vocab` is only ever called on the dense base `model.vocab` (BPE, WordPiece), so for a real tokenizer `max_id ≈ len`. A malformed `model.vocab` with a single sparse huge id (e.g. `4_294_967_295`) forced a ~100 GB allocation and **aborted the process at load**. The table is now capped at **16× the entry count** — generous headroom for the gap-tolerant vocabs the existing `test_invert_vocab` exercises (ids `0` and `3` at len 2), a hard ceiling on the malformed-input blast radius.

## Reachability
`load_tokenizer` (common.rs L202) is the production dispatch that builds WordPiece/BPE/SentencePiece tokenizers from `tokenizer.json`. The vocab map, added-tokens, and template-processing ids all run through these parse functions at load on file-controlled input.

## Testing
- `test_vocab_id_overflowing_u32_rejected` — `{"[PAD]": 4294967296}` → `Err`.
- `test_invert_vocab_rejects_disproportionate_sparse_id` — single entry, id 50 → `Err`; a normal gapped vocab still inverts (`is_ok`).
- `test_added_token_id_overflowing_u32_skipped` — oversized id skipped, well-formed sibling parsed.
- **Red/green proof**: with both Result-path guards removed, the two tests fail (wrap accepted / sparse id allocates without error); restored, all 9 `common::tests` pass. The disproportionate test uses id 50 (51 slots) so no actual giant allocation runs.
- `cargo clippy -p lattice-inference` clean.

## Bench
N/A — construction-time guards that run once at model load; no hot-path or kernel change.

### DIFF
--- a/crates/inference/src/tokenizer/common.rs
+++ b/crates/inference/src/tokenizer/common.rs
@@ -695,7 +695,13 @@ pub(crate) fn json_object_to_vocab(
         let id = id_value.as_u64().ok_or_else(|| {
             InferenceError::Tokenizer(format!("invalid non-integer token id for {token:?}"))
         })?;
-        vocab.insert(token.clone(), id as u32);
+        // Token ids are stored as u32 everywhere downstream. A JSON value past u32::MAX
+        // would silently wrap under `as u32` (e.g. 2^32 -> 0) and alias an unrelated
+        // token; reject it instead of corrupting the vocabulary.
+        let id = u32::try_from(id).map_err(|_| {
+            InferenceError::Tokenizer(format!("token id {id} for {token:?} exceeds u32 range"))
+        })?;
+        vocab.insert(token.clone(), id);
     }
     Ok(vocab)
 }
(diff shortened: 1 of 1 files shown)


ohdearquant:

Addressed the review items (commit f953b39c9):

- **C (sibling cast):** `vocab_txt_to_map` cast its line-index `idx as u32` unguarded — the last wrapping cast in this module. It now stops rather than wraps if the index would exceed `u32::MAX` (a `vocab.txt` over 2³² lines is malformed; `enumerate` is monotonic so no later line could fit either). The full `tokenizer.json`/`vocab.txt` id-parse surface in `common.rs` is now wrap-safe.
- **E (boundary tests):** added `test_vocab_id_at_u32_max_accepted` (locks the `try_from` boundary — the largest valid id `u32::MAX` is accepted, not rejected) and `test_invert_vocab_cap_boundary` (`max_id == len*16` trips the cap, `len*16 - 1` fits) to pin the off-by-one intent.

On the **B caveat** (empty base vocab now returns `Err` instead of `[""]`): keeping it — an empty `model.vocab` is malformed (a tokenizer with zero tokens is unusable), so rejecting at load is more correct than building a one-element table. the reviewer agreed this is acceptable.

All 11 `common::tests` pass; clippy clean; bench N/A (construction-time guards).

PR #370, opened 2026-06-25
harden(inference): standard MHA entry overflow/layout guard + add_bias release asserts

ohdearquant:

## What

A discovery review of `attention/standard.rs` (the bidirectional BERT-style MHA kernel) found two release-only precondition gaps — the same #313-class as the already-merged `gqa.rs` (#364), `decode.rs` (#365), `flash_causal.rs` (#366), `flash.rs` (#367), and the matmul boundary (#369). This closes them in `standard.rs` and the shared `add_bias` primitive.

### 1. `standard.rs` entry: debug-only shape/overflow checks → release-active
- The entry used `debug_assert!` for the head-layout invariant and computed `seq_len*hidden_size`, `num_heads*seq_len*seq_len`, and `num_heads*seq_len*head_dim` as **unchecked** `usize` products.
- A release build silently accepted `hidden_size != num_heads*head_dim`: the per-head copy loops then write only `num_heads*head_dim` lanes of each `hidden_size`-wide concat row, leaving the trailing lanes **stale** before the output projection consumes them (a silent wrong-output path — no OOB).
- The products are not dominated by the `matmul_bt` boundary guards and could wrap a 64-bit `usize` for an absurd shape, feeding an undersized scratch slice.
- Fix: `assert_standard_no_overflow(seq_len, hidden_size, num_heads, head_dim)` (`num_heads>0`, `head_dim>0`, `checked_mul` on every product, `assert_eq!(hidden_size, num_heads*head_dim)`) + release-active entry length asserts. Mirrors `assert_flash_no_overflow`.

### 2. `add_bias`: debug-only length checks → release-active (`assert_eq!`)
- `add_bias` used `debug_assert!` while its sibling `add_bias_gelu` already fail-closes in release. A short bias **truncated silently** (`chunks_exact_mut` + `zip`), leaving trailing lanes unbiased; a ragged `x` dropped the remainder row.
- Promoted both checks to release `assert_eq!`, matching `add_bias_gelu`.
- **Blast-radius audit:** all 18 `add_bias` call sites (standard×4, flash×8, bert×2, benches×3, test×1) pass loader-validated lengths (`f32_weights.rs:503`) at the matching `dim`. No valid caller is affected; the asserts only catch malformed directly-constructed weights.

This is a correctness/safety hardening, not a perf change: the new asserts are O(1) per call (entry-level), not per-element.

## Tests
- 3 new `standard.rs` guard tests: accepts valid shape, rejects layout mismatch, rejects wrapping `num_heads*seq_len*seq_len`.
- 3 new `add_bias` guard tests: rejects short bias, rejects ragged `x`, accepts valid lengths (value-checked).
- Full `lattice-inference` lib suite: **1090 passed, 0 failed**. `cargo clippy --workspace -D warnings` clean.

### DIFF
--- a/crates/inference/src/attention/standard.rs
+++ b/crates/inference/src/attention/standard.rs
@@ -80,7 +80,6 @@ pub fn multi_head_attention(
     buffers.temp[..seq_len * hidden_size].to_vec()
 }
 
-/// Internal in-place attention kernel.
 /// Release-active precondition guard for the bidirectional MHA shape products.
 ///
 /// `multi_head_attention_in_place` previously checked the entry shapes only with
@@ -130,6 +129,7 @@ fn assert_standard_no_overflow(
     );
 }
 
+/// Internal in-place attention kernel.
 pub(crate) fn multi_head_attention_in_place(
     hidden_states: &[f32],
     layer_weights: &TransformerLayerWeights<'_>,
--- a/crates/inference/src/attention/standard.rs
+++ b/crates/inference/src/attention/standard.rs
@@ -81,6 +81,55 @@ pub fn multi_head_attention(
 }
 
 /// Internal in-place attention kernel.
+/// Release-active precondition guard for the bidirectional MHA shape products.
+///
+/// `multi_head_attention_in_place` previously checked the entry shapes only with
+/// `debug_assert!`, so a release build silently accepted a malformed shape. Two
+/// hazards follow from that: (1) `hidden_size != num_heads * head_dim` produces a
+/// stale concat layout (the per-head copy loops write only `num_heads * head_dim`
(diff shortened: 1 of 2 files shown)


ohdearquant:

**bench-compare note (honest):** `make bench-compare` failed to produce a parseable A/B this run — the HEAD criterion estimates did not materialize ("change files found but all failed to parse"), the known two-worktree flakiness. I did not re-run it. Rationale: this PR adds only release-active `assert!`/`assert_eq!` precondition guards at the kernel entry (O(1) per call, run once per attention invocation, not in any inner loop), matching the #364–#369 entry-assert lineage which merged without bench-compare deltas. No hot-path arithmetic changed. If a perf signal is wanted I can re-run `scripts/bench-compare.sh --full`, but the expected delta is within noise.

Issue #361, opened 2026-06-25
flash attention finite MASK_VALUE (-10000) lets masked keys dominate when valid logits fall below the sentinel


ohdearquant:
> _Issue filed by Claude (Anthropic agent) on behalf of @ohdearquant._

## Summary

`crates/inference/src/attention/flash.rs` masks padded keys with a **finite**
sentinel `const MASK_VALUE: f32 = -10_000.0` (flash.rs:14). In the online-softmax
loop (`tiled_attention_head` / `tiled_multi_head_attention`), a masked key's score
is *replaced* with `MASK_VALUE` and folded into `local_max`/`tile_sum`:

```rust
// flash.rs:374
let mut value = row[c] * scale;
if attention_mask[kv_start + c] == 0 {
    value = MASK_VALUE;
}
row[c] = value;
local_max = local_max.max(value);
```

If a **valid** (unmasked) score is below `-10000`, the masked sentinel becomes the
row max and the masked key receives nonzero — possibly dominant — probability.

## Concrete trigger

```text
seq_len = 2, head_dim = 1, scale = 1.0
q = [-20000.0, 0.0]
k = [1.0, 0.0]
v = [1.0, 100.0]
attention_mask = [1, 0]      # key 1 is masked
```

Query row 0: valid key score = `-20000`, masked key replaced with `-10000`.
Row max = `-10000` (the *masked* key). After exp, probabilities collapse to
`[0.0, 1.0]`.

- **Wrong:** output = `0.0*1.0 + 1.0*100.0 = 100.0` (the masked value wins).
- **Correct:** masked key probability must be exactly 0 → output = `1.0`.

## Severity / blast radius

Latent. Real models produce attention scores of O(10)–O(100), well above `-10000`,
so normal greedy/sampled inference is unaffected (this is why existing tests, which
use the same finite sentinel in their reference path, don't catch it). It bites only
on out-of-distribution / hostile / corrupt activations where a valid logit drops
below the sentinel. Filed for correctness-hardening, not as a release blocker.

## Suggested fix (needs care — NaN risk)

Track masked keys structurally in the online-softmax loop: skip masked columns when
computing `local_max`, `tile_sum`, and the weighted-value accumulation; **or** use
`f32::NEG_INFINITY` as the sentinel with explicit all-masked-row handling so
`-inf - -inf` in the `exp(value - new_max)` step does not produce NaN. The
all-masked-row case must be handled (define a zero/uniform output rather than NaN).

[...]

PR #374, opened 2026-06-26
fix(inference): standard attention masks with -inf, not a finite sentinel (#361)

ohdearquant:

## What

`standard.rs` is the **live materialized CPU attention path** and shared the exact #361 finite-mask-sentinel bug fixed for the latent flash kernel in #373. It masked padded keys with `-10_000.0`, folded into the softmax row max. A valid logit below the sentinel makes the masked key the row max → it receives dominant softmax probability and leaks its value.

Unlike flash.rs (bench-only/latent), this path is reachable: Q/K project directly from `hidden_states` with **separate** `query_weight`/`key_weight` and no pre-norm bounding the scores, so distinct W_q/W_k can drive a valid score below the sentinel.

## Fix

- `standard.rs:258`: mask with `f32::NEG_INFINITY` so the masked key can never be the softmax row max. `softmax_attention` routes through `fast_exp`, which clamps `-inf` to the `-87` floor, so the masked weight is `fast_exp(-87)` ≈ 1.7e-38 — negligible and never dominant, not mathematically exact zero (byte-identical to the prior sentinel for valid-keys-present rows). Review round 1 flagged the earlier "exactly zero" wording; commit e991e3d adds `test_softmax_finite_valid_row_masked_weight_is_clamped_nonzero_not_dominant` to lock this clamped contract.
- `softmax_attention` (scalar / NEON / AVX2): add a max-finiteness guard. An all-masked row (`max_val` stays `-inf`, so `exp(-inf - -inf) = NaN`) is zeroed and skipped.

The guard also closes a **cross-backend divergence** the `-inf` flip would otherwise expose: scalar/NEON map `fast_exp(NaN)` to `0.0`, but AVX2's `MAXPS` returns the finite clamp on a NaN input → a tiny nonzero weight and a uniform row. With the guard, all three backends are byte-identical on an all-masked row.

**Byte-identical for all normal inference** — valid logits are O(10), the guard never fires, and `-inf` vs `-10_000` produce the same masked weight (both clamp to `fast_exp(-87)`). e2e-parity greedy generation is unaffected.

## Tests

- `masked_token_value_does_not_leak_when_valid_score_below_sentinel` — integration regression, **mutation-sensitive**: reverting line 258 to `-10_000.0` makes row-0 output jump from `~[1,0]` (valid token) to `[500,500]` (masked token's leaked value). Verified this session.
- `test_softmax_all_masked_row_is_zero_across_backends` — SIMD/scalar parity that an all-masked row collapses to exactly `0.0`, not a uniform distribution.

## Verification

- `cargo test -p lattice-inference --lib attention::` — 171 passed
- `cargo test -p lattice-inference --lib forward::cpu` — 45 passed
- `cargo clippy -p lattice-inference --lib -- -D warnings` — clean
- Mutation-sensitivity empirically confirmed (revert → `[500,500]` leak).

Correctness-only, no perf path touched (one never-taken `is_finite` branch per row). Closes the live-path sibling of #361 / #373.

### DIFF
--- a/crates/inference/src/attention/standard.rs
+++ b/crates/inference/src/attention/standard.rs
@@ -255,7 +255,13 @@ pub(crate) fn multi_head_attention_in_place(
                 let row = &mut scores[(h * seq_len + i) * seq_len..(h * seq_len + i + 1) * seq_len];
                 for j in 0..seq_len {
                     if attention_mask[j] == 0 {
-                        row[j] = -10_000.0;
+                        // Mask structurally with -inf, not a finite sentinel. A finite
+                        // sentinel can be *exceeded* by a valid logit that sits below it,
+                        // which would make the masked key the softmax row max and hand it
+                        // dominant probability (the #361 leakage mode, fixed in flash.rs;
+                        // standard.rs is the live materialized CPU path). softmax_attention
+                        // zeros an all-masked row via its max-finiteness guard.
+                        row[j] = f32::NEG_INFINITY;
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


ohdearquant:

## Round-1 review: fixes applied

The review verified the fix is correct (built an x86 target under Rosetta, forced AVX2 dispatch to prove the all-masked guard is mutation-sensitive on AVX2 — lane `0.0625` uniform row when removed — and confirmed NEON guard-removal still passes, so the guard's observable effect is AVX2-only). The sentinel mutation fails as required (`[500,500]` leak when reverted to `-10_000.0`). 171 `attention::` + 46 `forward::cpu` pass, clippy clean.

**The one finding (MEDIUM, artifact-level):** masked weights in a row that *has* a finite valid score are not mathematically exact zero. `softmax_attention` routes through `fast_exp`, which clamps `-inf` to the `-87` floor, so the masked weight is `fast_exp(-87)` ≈ 1.746e-38 — negligible and never dominant, and byte-identical to the prior `-10_000.0` sentinel for valid-keys-present rows (both clamp to the same floor). The earlier "exactly zero" prose overclaimed.

**Fixes (commit `e991e3d`):**
1. Added `test_softmax_finite_valid_row_masked_weight_is_clamped_nonzero_not_dominant` — locks the clamped contract (masked weight `> 0` and `< 1e-30`, valid key `> 1.0 - 1e-6`) so a future `-inf`→`0.0` special-case can't silently change parity/perf without updating the expectation. The all-masked row remains exact `0.0` via the max-finiteness guard.
2. Corrected the PR-body prose to the clamped-nonzero contract.

## Note on the e2e-parity report

The parity report lists the `merge_sort` long-prefill prompt (window 2) as FAIL at pos 0 (`merge_sort`→`main()`). This is **pre-existing and non-gating**, not a regression from this change:
- This masking change is byte-identical for normal inference (valid logits are O(10), the guard never fires, `-inf` vs `-10_000` both clamp to `fast_exp(-87)`); a 1.7e-38 weight perturbation physically cannot flip the *first* greedy token.
- PR #373 (flash.rs — bench-only/latent, touches **no** live inference path) reproduces the identical `merge_sort` FAIL, and its `e2e parity` + `parity-gate` checks both **passed** and #373 **merged**. The gate tolerates this standing window-2 divergence (known prefill nondeterminism).

Submitted **build-for-review** (not self-merge) per your request to review the live-path `standard.rs` sibling of #361.

Issue #353, opened 2026-06-25
grammar PDA: no-rewind backtracking over-accepts (parent backtrack) and over-rejects (shared-prefix anyOf)


ohdearquant:
> _Issue filed by Claude (Anthropic agent) on behalf of @ohdearquant._

## Summary

The grammar PDA (`crates/inference/src/grammar/pda.rs`) is a single-stack backtracking matcher **without input rewind**. Two correctness defects follow directly from that design and are NOT fixed by the accompanying harden commit (which closes the clean, well-contained sub-cases). Both are reachable from user-supplied grammars (`GrammarSpec::Gbnf`) and JSON schemas (`GrammarSpec::JsonSchema`) used for constrained decoding. Because the CI e2e-parity gate is greedy-only, neither is caught by CI.

These two are duals of the same root cause: **a frame can only soundly switch alternatives while it has consumed zero input bytes; the matcher has no way to un-consume bytes already matched under a committed alternative.** A position (`sym_pos`) check is *insufficient* to detect "consumed bytes" because `sym_pos` advances past nullable nonterminals (e.g. an optional numeric sign matching ε) without consuming any byte.

## Defect 1 — parent backtrack over-accepts (`try_next_alt`)

`try_next_alt` (pda.rs, the `next_alt >= num_alts` branch) backtracks to the parent frame and tries a sibling alternative even when the parent has already consumed bytes under its current alternative.

Minimal repro (GBNF shape):
```
root    ::= "a" nonterm | "x"
nonterm ::= "z"
```
Input `"ax"` is wrongly **accepted**: `nonterm` mismatches `'z'` at its first byte, exhausts its alternatives, backtracks to `root` (which already consumed `'a'`), switches `root` to alt-1 `"x"`, and matches the `'x'` — re-interpreting the byte as if `'a'` were never consumed. The only valid strings are `"az"` and `"x"`.

## Defect 2 — shared-prefix `anyOf` over-rejects (the dual)

When two `anyOf`/`oneOf` branches share a prefix and diverge later, the PDA commits to the first branch on the shared prefix and cannot rewind to try the second.

Minimal repro (JSON schema):
```json
{"anyOf":[
  {"type":"object","properties":{"status":{"type":"string"}},"required":["status"]},
  {"type":"object","properties":{"status":{"type":"integer"}},"required":["status"]}
]}
```
[...]

PR #380, opened 2026-06-26
fix(inference): close grammar object trailing-comma over-accept on #353, fix required-only over-rejection (#355)

### LINKED ISSUE #355: grammar: objects with mixed required+optional properties reject valid JSON (incl. all-present)

## Summary

JSON-schema `object` grammars that mix **required and optional** properties reject valid JSON — including, in the most damning case, the fully-present object. Constrained decoding against such a schema cannot emit a conformant object, so generation either fails or is forced off the valid language.

Discovered during the adversarial review of #354 (TBV by an independent critic, then characterized empirically). The defective code is **pre-existing** (byte-identical to `main`; unrelated to #354's changes).

## Empirical characterization

Each row is `accepts(compile(schema), input)` measured via the `json_schema` test harness on this branch:

| Schema shape | Input | Result | Correct |
|---|---|---|---|
| both required `{r,s}` | `{"r":1,"s":2}` | accept | ✓ |
| single optional only | `{}` | accept | ✓ |
| single optional only | `{"o":1}` | accept | ✓ |
| **required + optional** | `{"r":1}` (optional absent) | **reject** | ✗ should accept |
| **required + optional** | `{"r":1,"o":2}` (all present) | **reject** | ✗ should accept |
| optional + required | `{"o":1,"r":2}` (all present) | accept | ✓ |
| optional + required | `{"r":2}` (optional absent) | **reject** | ✗ should accept |

Reproducer schema for the worst case:
```json
{"type":"object","properties":{"r":{"type":"integer"},"o":{"type":"integer"}},"required":["r"]}
```
`{"r":1}` and `{"r":1,"o":2}` both reject.

## Root cause (two entangled defects)

In `crates/inference/src/grammar/json_schema.rs` `compile_object`, properties are emitted into a single `main_alt` sequence:

1. **Unconditional separator before a possibly-absent property.** For every non-first property the separator `ws "," ws` is pushed *unconditionally* (json_schema.rs:364-369), then the property — if optional — is wrapped in `opt_<pair> ::= <pair> | ε` (lines 374-384). When the optional property collapses to ε, the mandatory comma has already been committed, so the closing `}` (no comma) is rejected. The comma must live *inside* the optional element, not before it: `("," prop)?` rather than `"," (prop)?`.

2. **Property emission order is the sorted key order, not insertion order.** Properties are iterated in the compiler's map order (sorted), so a schema `{r:required, o:optional}` actually compiles in `o,r` order. Combined with defect 1, the all-present input `{"r":1,"o":2}` is matched against a grammar expecting `o` first, so even a complete object fails. (The PDA has no input rewind / set-of-states, so it cannot reinterpret across the misplaced comma — same backtracking limitation as #353.)

## Why not a one-line fix

A correct construction has to handle: leading-optional-then-required, trailing optional, runs of optionals, and the comma belonging to whichever element actually appears. The standard solution (cf. llama.cpp's json-schema-to-grammar) threads the comma into each optional element and emits a "rest" chain so any present-subset is accepted. This interacts with the PDA's prefix-determinism limitation tracked in #353, so it needs design + a differential test matrix, not a patch. Filing for a focused PR.

## Suggested fix sketch
- Move the separator inside the optional wrapper: `opt_pair ::= (ws "," ws pair)?` for non-first properties, and make the *first emitted* property's comma conditional on something actually preceding it (build a right-nested "tail" chain so each property owns its leading comma only when a prior property was emitted).
- Add a differential test matrix: every subset of {required, optional} properties × present/absent × property order, asserting accept iff the subset includes all required and respects type.

## Blast radius
Any production JSON schema with at least one required and one optional property (an extremely common shape) is affected. High-value correctness fix for constrained decoding.

ohdearquant:
## Summary

Closes #355. Builds on the merged #353 byte-consumption guard to close the **object** trailing-comma over-acceptance in the constrained-decoding grammar, and fixes #355's over-rejection of required-only objects.

For schema `{r required, o1?, o2?}`:
- **Before (`origin/main`):** required-only `{"r":1}` was **rejected** (the comma separators around omitted optionals were mandatory) — the #355 bug.
- **After:** required-only `{"r":1}` is **accepted**, and the trailing-comma form `{"r":1,"o1":2,}` is correctly **rejected** (no longer admits invalid JSON).

## Mechanism

The object is compiled as a **required-first tail chain** (#355): required properties first (in declaration order), then optionals, each optional recursing into a right-nested `(pair tail | tail)` chain. This removes the mandatory inter-property commas that rejected required-only objects.

The trailing-comma over-accept that an optional tail chain would otherwise admit is closed by the **#353 per-frame byte-consumption guard**: once the `,` is consumed, `pda::try_next_alt` refuses to switch the committed frame back to its omitted-optional (`ε`) alternative, so the closing `}` cannot match after a dangling comma.

## What this PR is — and is not

This patch closes the **object trailing-comma over-acceptance** (the dangerous direction: admitting invalid output) and fixes the required-only over-rejection. It also introduces one intentional, documented behavior change that I want to surface explicitly:

**Required-first key-order canonicalization (intentional, not a regression).** A constrained-decoding object grammar accepts exactly **one** key permutation. `origin/main` emitted properties in alphabetical (serde declaration) order; this branch emits them **required-first**. So the single accepted permutation moves:

| input (schema `{r req, o1?, o2?}`) | `origin/main` | this branch |
|---|---|---|
| `{"r":1}` (required-only) | **reject** (the #355 bug) | **accept** ✓ |
| `{"r":1,"o1":2,"o2":3}` (required-first) | reject | **accept** |
| `{"o1":1,"o2":2,"r":3}` (alphabetical) | accept | reject |
| `{"r":1,"o1":2,}` (trailing comma) | reject (masked) | reject ✓ |

This is a **lateral move in which permutation is canonical**, plus a strict fix (required-only subsets, which `main` could never emit, now generate). It is:
- **Necessary for PDA safety** — with required properties first, the `started=false` phase is a single-alternative key chain, so the no-rewind PDA cannot commit to a wrong key terminal (documented at `json_schema.rs` ~381-405).
- **The safe direction for constrained decoding** — JSON objects are unordered (RFC 8259). The grammar *guides* the model to emit the canonical order, so generation always produces valid, schema-conforming JSON and never emits invalid output. Key order in generated JSON is semantically irrelevant; the only way it could matter is an exact-string-equality assertion on output that pins key order, which the schema contract never promised.

The `object_required_first_reorder` test pins this behavior as a documented anchor (mirroring how #353 pinned its shared-prefix known-limitation).

## Property-count cap

`build_object_tail` recurses once per property, so a hostile or generated wide schema could grow the stack during `GrammarEngine::new` independent of `MAX_SCHEMA_DEPTH`. Added `MAX_OBJECT_PROPERTIES = 4096` (mirrors `MAX_ARRAY_CARDINALITY`); the check fires **before** recursion and returns a `SchemaError` (no panic). Two tests pin the boundary (`cap+1` rejects with the documented message; under-cap compiles and still accepts `{}` / a one-property object).

## Tests

Reconciled the suite with the now-correct behavior:
- Flipped three previously-mislabeled over-accept characterization tests to `rejects` (`object_two_trailing_optionals`, `object_two_optional_only_pda_boundary`, `object_three_props_trailing_comma_boundary`). Their old comments called the over-accept "#353 pre-existing"; it is closed, not pre-existing.
- Corrected four stale comments describing those strings as over-accepted.
- Added `object_required_first_reorder` (re-canonicalization anchor) and the two cap tests.

`cargo test -p lattice-inference --lib grammar::` → **89 passed**. `cargo clippy --workspace -- -D warnings` clean.

## Performance

`crates/inference/` is touched, so per repo policy this includes `make bench-compare`. The change is confined to `crates/inference/src/grammar/json_schema.rs`, the JSON-schema-to-grammar **compiler**, which runs once at `GrammarEngine::new` (constrained-decoding setup) and is not on the model forward/decode path.

`make bench-compare` (origin/main `6bc394abc` vs HEAD `29a5f75a5`, quick mode) was run twice; both runs hit the comparison harness's known two-worktree fragility — the base criterion baselines (`target/criterion/*/base/estimates.json`) were not seeded, so the A/B diff could not parse (`error: change files found but all failed to parse`). Rather than rely on the harness, the no-effect claim is established **by construction**: a grep over all 22 inference + embed benchmark files finds **zero** references to `grammar` / `json_schema` / `GrammarEngine`, so the only changed file is outside every benchmark's call graph and the measured code is byte-identical between base and head. **No throughput effect is possible.**

---
> _PR authored by Claude (Anthropic agent) on behalf of @ohdearquant._

### DIFF
--- a/crates/inference/src/grammar/json_schema.rs
+++ b/crates/inference/src/grammar/json_schema.rs
@@ -58,6 +58,20 @@ const MAX_ARRAY_CARDINALITY: usize = 4096;
 /// schema chains references or nests anywhere near 512 levels.
 const MAX_SCHEMA_DEPTH: usize = 512;
 
+/// Maximum number of properties in a single object schema.
+///
+/// `build_object_tail` builds a right-nested tail chain, recursing once per
+/// property (the recursion is on property index `i`, independent of
+/// `MAX_SCHEMA_DEPTH`, which bounds only nested-schema descent). A parseable
+/// schema with an absurd number of properties therefore recurses once per
+/// property at compile time — a stack-exhaustion vector reachable from
+/// untrusted schema input at `GrammarEngine::new`, the object analogue of the
+/// `MAX_ARRAY_CARDINALITY` array vector. Constrained decoding has no real use
+/// for objects with anywhere near this many keys; reject beyond the cap at the
+/// parse boundary with a typed error. Matches `MAX_ARRAY_CARDINALITY` so the
+/// two boundaries stay consistent.
+const MAX_OBJECT_PROPERTIES: usize = 4096;
+
 /// Error returned by the JSON Schema compiler.
 #[derive(Debug, Clone, PartialEq)]
 pub struct SchemaError(pub String);
@@ -319,6 +333,18 @@ impl<'a> CompileCtx<'a> {
                 Ok(vec![empty_object_as_alt_with_ws(ws_id)])
             }
             Some(props) => {
+                // issue #343 analogue: cap property cardinality before
+                // materializing per-property rules. `build_object_tail` recurses
+                // once per property, so an object schema with an absurd number of
+                // keys overflows the stack at compile time. Reject at the parse
+                // boundary, mirroring the `MAX_ARRAY_CARDINALITY` guard.
+                if props.len() > MAX_OBJECT_PROPERTIES {
+                    return Err(SchemaError(format!(
+                        "object schema property count ({}) exceeds the supported limit ({MAX_OBJECT_PROPERTIES})",
+                        props.len()
+                    )));
+                }
+
                 // Unique per-object suffix so distinct object schemas sharing a
                 // property key don't alias each other's value/pair rules.
                 let obj_idx = self.object_counter;
@@ -342,9 +368,10 @@ impl<'a> CompileCtx<'a> {
                     // key_bytes: raw bytes of the JSON-encoded key string.
                     // Used to inline the key as terminals in started=true tail
                     // rules so trailing-comma over-acceptance is prevented for a
-                    // single trailing optional (the common case); ≥2 trailing
-                    // optionals remain over-accepted as a pre-existing #353 PDA
-                    // limitation. See build_object_tail for the exact boundary.
+                    // single trailing optional (the common case); the >=2
+                    // trailing-optional case is closed by the #353 per-frame
+                    // byte-consumption guard in pda::try_next_alt. See
+                    // build_object_tail for the exact boundary.
                     let key_bytes = serde_json::to_string(key_str)
                         .map_err(|e| SchemaError(format!("cannot JSON-encode property key: {e}")))?
                         .into_bytes();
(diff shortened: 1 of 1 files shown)


ohdearquant:

## Review verdict: hold — new over-acceptance (do not merge as-is)

An adversarial PR-vs-`main` comparison (path-dep harness against both this branch and an `origin/main` worktree) found this branch **introduces a new over-acceptance** that `main` does not have. The "strict improvement / #353-pre-existing only" framing is incorrect for the required-plus-two-optionals shape.

**Schema**
```json
{"type":"object","properties":{"r":{"type":"integer"},"o1":{"type":"integer"},"o2":{"type":"integer"}},"required":["r"]}
```
**Invalid input** `{"r":1,"o1":2,}` (trailing comma → invalid JSON)
- `origin/main`: **rejects** ✅
- this branch: **accepts** ❌ (new regression)

**Mechanism.** Required-first reordering (`json_schema.rs:381`) moves `r` ahead of `o1`/`o2`, opening a byte path `main` rejected before it reached the trailing-comma hole. The started=true optional emit/skip shape (`json_schema.rs:533`) then still lets the parent optional-tail skip alt recover *after* the `,` has been consumed — the #353 PDA no-rewind mechanism (`pda.rs:322`), newly exposed for this schema.

This is the dangerous direction: constrained decoding can now be driven to emit invalid JSON. Over-rejection (the old #355 behavior) is the safe direction; over-acceptance defeats the purpose.

**Two test assertions lock in the regression and mislabel it** — `json_schema.rs:1635` and `:1785` assert `accepts(b"{\"r\":1,\"o1\":2,}")` with a "#353 pre-existing" comment, but it is not pre-existing on `main`. After a fix these must be `rejects(...)`.

**What is correct** (verified): key inlining is byte-equivalent to the pair rule; memoization keys `(obj_idx, i, started)` don't alias; required-first reordering is JSON-semantically valid (RFC 8259); no new `unwrap`/`panic` in library code; the #355 reproducer (`{"r":1}`, `{"r":1,"o":2}`) is genuinely fixed.

**Decision.** Holding this PR. The clean fix is the #353 PDA byte-consumption guard (per-frame tracking so no sibling skip alt is selectable once the emit alternative has consumed a byte) — that closes object **and** array trailing-comma over-acceptance at once and would let this required-first tail chain land safely. Grammar-only restructuring cannot fix it (the hole reappears one level up). Blocking on #353.

ohdearquant:

Converting back to draft after a branch-vs-`main` differential check.

For a schema like `{r: required, o1/o2: optional}`, this branch newly **accepts** the invalid JSON `{"r":1,"o1":2,}` (required prop plus ≥2 trailing optionals, trailing comma), while current `main` **rejects** it. The characterization test labels that case "#353 pre-existing," but `main` does not actually exhibit it: `main`'s broader over-rejection of optional-bearing objects masks the over-accept, so fixing the over-rejection here **un-masks** it as a *new* over-acceptance. For a constrained-decoding grammar, admitting invalid JSON is the dangerous direction, so this cannot land as-is.

The single-optional cases are fully fixed (no over-accept — `{"r":1}`, `{"r":1,"o":2}` accept; `{"r":1,}`, `{"r":1,"o":2,}` reject). But the ≥2-optional trailing-comma over-accept, and the dual skip-middle over-rejection (`{"r":1,"o2":3}`), both trace to the PDA's no-rewind backtracking (#353). That is only fixable at the PDA level — a per-frame byte-consumption guard in `pda.rs` that refuses to switch a parent to a sibling skip-alternative after a child path has already consumed input bytes — not in the grammar builder. Grammar restructuring can only move *where* the trailing-comma mismatch surfaces; it cannot remove it.

Separately, `build_object_tail` recurses once per property with no property-count cap (`MAX_SCHEMA_DEPTH` only bounds nested descent, not a single wide object), so a very wide `properties` map could grow the stack during grammar construction. Worth an iterative tail build or a `MAX_OBJECT_PROPERTIES` cap if this work resumes.

Holding behind the #353 PDA fix.

PR #429, opened 2026-06-28
fix(inference): render non-special added tokens in detokenizer (qwen3.6 think-tag bug)

ohdearquant:
## The bug (hardcore debugging — see chain below)

`qwen3.6-27b-q4` (and `qwen3.5-0.8b` — identical tokenizer shape) appeared to never emit `<think>`/`</think>`. Two earlier hypotheses were both **wrong** and disproven this session with empirical probes:

1. ~~"non-thinking checkpoint"~~ — the model *is* a reasoning model.
2. ~~"engine logit suppression / effective-vocab cutoff"~~ — a `bench_logit_dump` showed the forward ranks `</think>` (id 248069) as **argmax** (logit 21.497), and a sampler trace showed `generate_streaming` picks `next_id=248069` **correctly**.

**The real cause is the detokenizer.** `id_to_token` is built by `invert_vocab` from `model.vocab` only (base max id **248043**), so the 26 `added_tokens` (ids **248044-248069**) are absent. `token_for_id(248069)` returned `None` → the close tag contributed **zero bytes** → `</think>` was silently swallowed from the text stream. The visible `\n\n` was simply the *next* token.

## The fix (HF-faithful)

HF's `decode(skip_special_tokens=True)` (the chat default) skips only `special=true` added tokens; `special=false` ones (`<think>`, `</think>`, `<tool_call>`, FIM markers) render verbatim.

- `parse_rendered_added_tokens` (common.rs): parses the `special != true` subset to content→id (absent `special` defaults to false).
- BPE stores an `added_render` (id→content) map, threaded through the constructor + 5 builders.
- `token_for_id` falls back to `added_render` when `id_to_token` misses.
- **Isolated** from `id_to_token` / `vocab_size()` / `vocab_bytes()` / grammar — no ripple to encoding, sizing, or constrained decoding.

## Regression gate

The existing **e2e-parity gate compares token IDs** and is structurally **blind** to this class — the ids matched while the rendered *text* dropped the tag. The new gates assert **text rendering**, and run in the normal CPU `cargo test` job (no model download, no Metal):

- `test_from_tokenizer_json_renders_real_qwen_think_tags` — drives the real public loader `from_tokenizer_json_str` with the **exact shipped qwen added-token ids/flags**; asserts `</think>`/`<think>`/`<tool_call>` render while `<|im_end|>`/`<|im_start|>` (special=true) stay swallowed. Locks the full wiring; would have caught the original bug.
- `test_parse_rendered_added_tokens_filters_special` — parser unit test (special-flag filtering).
- `test_decode_renders_nonspecial_added_tokens` — decode + `IncrementalDetokenizer` streaming paths.

## E2E verification

Forcing prompt, greedy (`temperature 0, top-k 1`), q4 27B on Metal — **first generated token is now `</think>`**:

```
</think>  ->  \n\n  ->  Hello  ->  !  ->  How  ->  can ...
```

Before the fix the stream began at `\n\n` (the `</think>` was swallowed).

## bench-compare

Not meaningful for this change: it touches only the tokenizer decode path (one `HashMap` fallback lookup in `token_for_id`, called once per generated token during detok), which is off every Criterion group and the forward/embed hot path. `make bench-compare` produced no parseable A/B (the two-worktree harness had no baseline criterion data); given the change cannot affect any benched code path, there is nothing to report.

### DIFF
--- a/apps/macos/Sources/LatticeStudio/Screens/ChatScreen.swift
+++ b/apps/macos/Sources/LatticeStudio/Screens/ChatScreen.swift
@@ -59,6 +59,8 @@ struct ChatScreen: View {
 
     // Composer text is ephemeral — clearing on navigation is acceptable
     @State private var composerText: String = ""
+    // Which turns have their reasoning disclosure open (auto-open while streaming).
+    @State private var expandedThinking: Set<UUID> = []
 
     // MARK: Store-backed derived helpers
 
@@ -140,7 +142,9 @@ struct ChatScreen: View {
             guard let turnID = store.chatAwaitingTurnID,
                   let idx = store.chatTurns.firstIndex(where: { $0.id == turnID })
             else { return }
-            store.chatTurns[idx].responseText = newText ?? ""
+            let parsed = splitThinking(newText ?? "")
+            store.chatTurns[idx].thinkingText = parsed.thinking
+            store.chatTurns[idx].responseText = parsed.response
         }
     }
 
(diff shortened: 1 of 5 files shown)


ohdearquant:
Superseded by #430 — same tokenizer fix, cut clean from origin/main (this branch was accidentally stacked on feat/chat-thinking-mode and carried 3 unrelated macOS files, which review flagged). The tokenizer fix itself was reviewed and approved with minor fixes, with a mutation test confirming the regression gates are real; #430 resolves the only requested fix (scope).

Issue #462, opened 2026-06-29
Serve path re-prefills the entire conversation each turn (no cross-turn KV cache)


ohdearquant:
## Summary

The chat serve loop is **stateless across turns**: it calls `reset_state()` (wipes the KV cache) at the start of every request and re-runs prefill over the *entire* conversation history. As a conversation grows, each new turn pays a prefill cost proportional to the full transcript, so latency climbs turn over turn even though most of the prompt is byte-identical to the previous turn.

## Current behavior

- `crates/inference/src/bin/chat_metal.rs:758` — `metal.reset_state()` is called per request in the serve loop.
- Each request re-tokenizes and re-prefills the whole rendered conversation (system + all prior user/assistant turns + the new user turn).
- There is no reuse of the KV cache computed for the shared prefix of the previous turn.

## Why it matters

In a multi-turn chat, turn N's prompt shares a long common prefix with turn N-1's prompt (everything except the latest user message and the prior assistant reply). Re-prefilling that shared prefix is pure redundant work:

- Prefill cost per turn ≈ O(total conversation length), so cumulative cost over a K-turn chat is ~O(K²).
- This is the dominant contributor to "the chat gets slower the longer it goes" — separate from, and larger than, the per-token decode slope. (Decode is already linear in context because each GQA layer scans the whole KV cache per token; that falloff is inherent. The re-prefill is *avoidable* on top of it.)
- For a long-context model (e.g. the 27B), re-prefilling a multi-thousand-token transcript on every turn is the single biggest latency cost in interactive use.

## Measured motivation

Decode throughput already falls with context on this hardware (Apple M2 Max, Qwen3.5-0.8B, slope method, decode-only): ~148 tok/s @ ctx 64 → ~112 @ 256 → ~85 @ 512. That falloff is the *decode* side. On top of it, the serve path pays a *full re-prefill* of the conversation every turn — the larger and more avoidable cost. A turn against a ~1.7K-token transcript spends most of its wall-clock in that redundant prefill before the first new token is decoded.

## Proposed approach — cross-turn prefix KV cache

1. Keep the KV cache resident across requests for a given session instead of `reset_state()`-ing unconditionally.
2. On each new request, compute the longest common **token** prefix between the new rendered prompt and the previously cached prompt.
3. Truncate the KV cache to that shared prefix length (drop only the divergent tail), then prefill **only** the new tokens (the delta).
[...]

PR #516, opened 2026-07-01
feat(inference): cross-turn KV prefix cache — skip re-prefill of shared conversation prefix (#462)

ohdearquant:
## WIP preservation — cross-turn KV prefix cache

An autonomous implementation pass (#462) produced a substantial
implementation but did **not** reach its own correctness gate before completing.
Preserving as a **draft** review surface so the work isn't lost across a session
switch. **Do not merge** — this is unverified.

### What was built
- `crates/inference/src/kv_cache/cross_turn.rs` (342 LOC): token-level prefix-reuse
  planner. Owns no GPU buffers; decides reuse-vs-`FullRefill` from token IDs +
  metadata; latest-boundary GDN snapshot + live KV logical reuse for v1; any
  divergence falls back to `PrefixReuseMode::FullRefill`.
- `crates/inference/src/forward/metal_qwen35.rs` (+848): Metal runtime hooks for
  the KV logical reuse + GDN snapshot path.
- `crates/inference/src/kv_cache/mod.rs` (+5): module wiring.
- An architect design record covering the implementation approach.

### What is NOT done (the gate)
- **Token-identity correctness test** — cross-turn output must be byte-identical to
  a full re-prefill, *through the GDN recurrent state* (18/24 layers). This is the
  explicitly-hard part and is **unverified**. No test was written.
- **Build / clippy** — committed unbuilt to preserve the work; compile status
  unchecked.
- **Measured prefill savings** on a real multi-turn conversation.

### The open design question
GDN recurrent-state cross-turn continuity: the v1 design chose a
"latest-boundary GDN snapshot + live KV logical reuse" — its correctness through
the GDN layers is the crux and is unproven. This generalizes to design item **F5**
(stateful-serving / GDN cross-turn continuity ADR) in the lattice design queue.

Preserved so the work survives a model switch.
The correctness gate + build verification are the required next steps before this
leaves draft.

## Round-1 remediation (review round 1: issues found)

Round-1 review found issues: the token-prefix planner was conservative but
the Metal integration violated the core state contract (a per-slot `HashMap` cache
retained entries while the underlying full-attention KV buffer is a single mutable
live buffer, so a stale entry could restore GDN state from one conversation while
attention read KV rows left by a different generation). Six findings total, fixed
per a fixed design (D1-D6) — still **draft**, still not merged/ready.

| Finding | Fix (design decision) | Test |
|---|---|---|
| 1 — cross-slot K/V bleed (Blocker) | **D1**: `MetalCrossTurnPrefixCache` retains at most one entry (`Option<MetalCrossTurnPrefixEntry>`, not an unbounded `HashMap`); `get`/`take`/`remove` all enforce the slot match, so a different slot is always a miss | `cross_turn_cache_slot_isolation_evicts_other_slot` (T4) |
| 2 — stale entries survive `reset_state`/interrupt (Blocker) | **D2 + D3**: `restore_cross_turn_prefix`'s `ExactAppend` arm now `take()`s (not clones) the entry, so the cache is empty from the start of a restore until a fresh save succeeds — fail-closed on any early return; `reset_state()` (the single choke point for plain `generate_streaming`, chat, serve worker, and the cache-aware path's own FullRefill/error legs) now clears the cache too | `cross_turn_cache_interleaved_plain_path_invalidates` (T5, mutation-sensitivity verified) |
| 3 — LoRA identity is shape- not content-based (Major) | **D4**: `load_lora_adapter`/`unload_lora_adapter` unconditionally clear `cross_turn_prefix_cache`; the metadata's shape-based hash remains defense-in-depth, the clear is the actual correctness mechanism | `cross_turn_cache_lora_load_unload_invalidates` (T7) |
| 4 — exact-equal retry hits an empty-suffix error (Major) | **D5**: planner requires `new_prompt_ids.len() > shared` for `ExactAppend`; exact-equal prompts now fall through to `FullRefill` instead of surfacing `InferenceError::PrefixCache` | `plan_exact_equal_prompt_is_full_refill` (planner unit test) + `cross_turn_cache_exact_equal_retry_full_refills` (T6, mutation-sensitivity verified) |
| 5 — unbounded slot growth / memory DoS (Major) | Structurally fixed by **D1** — at most one ~19MiB GDN snapshot is ever retained, bounded by design, not by an LRU cap | Covered by T4 (only one entry ever alive) |
| 6 — `lattice_serve.rs`/`chat_metal.rs` never opt into the cache (Major) | **D6 — explicitly out of scope for this remediation.** Deliberate follow-up PR once findings 1-5 are confirmed correct; wiring the serve/REPL entry points before that would ship the cross-slot and staleness bugs directly into the server path | Not addressed here |

**Mutation-sensitivity evidence** (T5/D3, T6/D5): both fixes were verified by
reverse-applying the target hunk via `Edit` on the uncommitted diff (never
`git checkout` on work in progress), `touch`-ing the file to defeat cargo's mtime
cache, rerunning the specific test, confirming it failed for the predicted reason,
then re-applying the fix and confirming it passed again.
- T5 (D3 reverted): `reused_tokens` mismatch (`left=4, right=0`) — the stale entry
  reported a spurious `ExactAppend` restore.
- T6 (D5 reverted): the exact predicted error,
  `PrefixCache("forward_prefill_from: cannot prefill an empty suffix without a
  saved logits source")`.

Two test-design bugs were found and fixed *during* this process (both tests
initially passed for the wrong reason — an unrelated guard from a different design
decision was masking the one under test): T5's second call originally reused a
same-length retry, which independently satisfied D5's separate suffix guard
regardless of D3; T6's second call originally reused the bare prompt text, but a
saved entry represents `prompt + generated tokens`, so the exact-equal boundary was
never actually reached. Both were corrected before mutation-sensitivity was
re-verified.

### Gates (all run in this worktree, real Metal hardware)
- `cargo fmt --check` — clean
- `cargo clippy --workspace -- -D warnings` — clean
- `cargo clippy --workspace --features metal-gpu -- -D warnings` — clean
- `cargo test -p lattice-inference --features metal-gpu cross_turn` — 24 passed, 0 failed
- `cargo test -p lattice-inference kv_cache` — 102 passed, 0 failed

Still **draft**; still requires findings 1-5 fixes reviewed and finding 6 (serve
wiring) done as a follow-up before this can leave draft.

## Round-2 remediation (review round 2)

Round-2 review found issues: 1 blocker (D7) + 1 medium (D8). Round-1's D1-D5 were
independently re-verified sound by the round-2 reviewer and were **not** touched
in this round.

### D7 (Blocker) — public raw-forward APIs could still mutate live KV/GDN state without invalidating the cache

**The gap:** `forward_step`, `forward_prefill`, and `forward_prefill_all_logits`
are public entry points that advance `self.session.kv_cache` / GDN state directly.
They don't route through `reset_state()` (D3's guard) or the cache-aware save path,
so a caller could: save an entry via `generate_streaming_with_prefix_cache(slot, ...)`
→ call `forward_step`/`forward_prefill` on an unrelated token stream → continue the
original slot with a strict append → get a spurious `ExactAppend` that restores a
GDN snapshot against KV rows the raw call already overwrote. Same bug class as
round-1 finding 2, through a different public mutation path.

**Fix — full boundary enumeration.** Every `pub fn` on `MetalQwen35State` (and the
public `MtpTargetVerifier` trait impl for it) was read and classified into exactly
one of: **(a) cache-ownership participant** — already clears/owns the cache
correctly; **(b) mutates outside ownership** — needed a new `clear()`; **(c)
read-only / non-mutating / unrelated struct** — no change needed.

| Method | Mutates live KV/GDN? | Classification | Disposition |
|---|---|---|---|
| `forward_step` | Yes (`forward_step_inner`, advances `kv_cache.seq_len`) | (b) | **Fixed**: added `cross_turn_prefix_cache.clear()` at the top |
| `forward_prefill` | Yes (`forward_prefill_impl`) | (b) | **Fixed**: same |
| `forward_prefill_all_logits` | Yes (`forward_prefill_impl`) | (b) | **Fixed**: same |
| `MtpTargetVerifier::rollback_cache_to` (trait impl) | Yes (`rollback_speculative_state_to`: sets `kv_cache.seq_len`, restores GDN slot) | (b) | **Fixed**: added clear(). Public trait; not wired into any live Metal generate loop today (Metal MTP decode uses the self-contained `mtp_greedy_round`, not `mtp_verify_draft`/this trait — only benches exercise it), but reachable by any consumer holding `&mut MetalQwen35State` + the trait in scope, so it is real public surface |
| `MtpTargetVerifier::verify_tokens` (trait impl) | Yes (`verify_tokens_batched`: `forward_step_inner` + GDN checkpointing per token) | (b) | **Fixed**: added clear(), same rationale |
| `generate` | Yes | (a) | Calls `self.reset_state()` at the top before any prefill/forward work — already clears |
| `generate_multimodal` | Yes | (a) | Same — calls `reset_state()` before prefill |
| `generate_streaming` | Yes | (a) | Same — calls `reset_state()` before prefill |
| `generate_streaming_with_prefix_cache` / `_inner` | Yes | (a) | The cache-aware family itself: `restore_cross_turn_prefix`'s `ExactAppend` arm `take()`s the entry before any mutation, `FullRefill` arm calls `reset_state()`; saves via `save_cross_turn_prefix_or_clear` only at the very end |
| `chat_completion` | Yes (via `generate`) | (a) | Delegates to `generate`, covered transitively |
| `chat_completion_streaming` | Yes (via `generate_streaming`) | (a) | Delegates to `generate_streaming`, covered transitively |
| `chat_completion_streaming_with_prefix_cache` | Yes (via cache-aware family) | (a) | Delegates to `generate_streaming_with_prefix_cache`, covered transitively |
| `compute_token_nlls` | Yes | (a) | Calls `self.reset_state()` before `forward_prefill_all_logits` — verified explicitly per the task instructions rather than assumed |
| `score_layer_importance` | Yes (via private `forward_prefill_layer_traces_last_token`) | (a) | That private helper calls `reset_state()` inside its own per-layer loop AND once more at the end before returning — verified explicitly, not assumed; every exit path leaves the cache cleared |
| `reset_state` | Yes | (a — the choke point) | Already the D3 remediation: clears `cross_turn_prefix_cache` unconditionally; this is what makes (a) above transitively correct |
| `load_lora_adapter` / `unload_lora_adapter` / `generate_with_lora_mixture` | Yes | (a) | D4 remediation: load/unload unconditionally clear; the mixture path routes through both |
| `clear_cross_turn_prefix_cache` | N/A (clears the cache itself) | (a) | Cache-ownership API |
| `new` / `from_q4_dir` / `new_session` | No (fresh state / fresh session) | (c) | Construction, not mutation of an existing live cache |
| `has_mtp`, `has_lora_adapter`, `max_context`, `set_gdn_chunked`, `reset_gdn_state_traffic`, `gdn_state_traffic_report`, `take_gdn_state_traffic_report` | No | (c) | Read-only or feature-counter bookkeeping, no KV/GDN/position mutation |
| `compute_perplexity` | No direct mutation | (c) | Thin wrapper driving `compute_token_nlls` (already (a)) as its per-window kernel |
| `system`/`user`/`assistant` (`ChatMessage`), `format_chat_template`, `blend_lora_layer_data`, `mtp_missing_weights_warning`, `mtp_greedy_round` | No | (c) | Free functions / other-struct builders; `mtp_greedy_round` takes no `&mut MetalQwen35State` (pure decision function), called only from inside `generate`'s own decode loop which is already (a)-covered |

**Self-invalidation check (required by the task, not skipped):** before adding
`clear()` to the three public methods, verified the cache-aware decode loop's own
internal `forward_step` calls (inside `generate_streaming_with_prefix_cache_inner`)
never run against a live entry — `restore_cross_turn_prefix`'s `ExactAppend` arm
uses `take()` (not clone), so the slot is already empty by the time decode starts;
the `FullRefill` arm calls `reset_state()`, which clears the whole cache first. So
the new `clear()` calls inside `forward_step`/`forward_prefill*` are a no-op on the
cache-aware path's own re-entrant calls — no self-invalidation, no reordering
needed.

**Test (T8):** `cross_turn_cache_interleaved_raw_forward_invalidates` — saves an
entry via `generate_streaming_with_prefix_cache` on slot A, calls raw
`forward_step(0, 0)` on an unrelated token, then continues the ORIGINAL slot with a
strictly-longer append (never an exact-equal retry, so D5's separate guard can't be
what's doing the work) and asserts `reused_tokens == 0`, `mode != ExactAppend`, and
output token-identity against a fresh full re-prefill.

**Mutation-sensitivity evidence:** reverse-applied *only* the `forward_step` clear
via `Edit` (never `git checkout` on the uncommitted fix), `touch`-ed the file to
defeat cargo's mtime cache, reran T8:

```
thread '...cross_turn_cache_interleaved_raw_forward_invalidates' panicked at
crates/inference/src/forward/metal_qwen35.rs:22248:13:
forward_step must clear the retained cross-turn entry before mutating live state
```

Restored the fix, touched again, reran to green:

```
test forward::metal_qwen35::inner::tests::cross_turn_cache_interleaved_raw_forward_invalidates ... ok
test result: ok. 1 passed; 0 failed
```

### D8 (Medium) — design report drift

The design report still documented the rejected
`HashMap<CrossTurnSlotId, MetalCrossTurnPrefixEntry>` multi-slot design at the
struct-sketch level, said retained state was "Per `CrossTurnSlotId`" and "per cache
slot" / "per slot" in the GDN-decision rationale, and listed serve/chat REPL wiring
as implemented design components even though D6 is deliberately deferred.

**Fix:**
- Replaced the `HashMap` struct sketch with the actual shipped shape
  (`entry: Option<MetalCrossTurnPrefixEntry>`, slot-filtered `get`/`take`/`remove`,
  unconditional `insert` replacement) and added a "Round-2 remediation" note
  explaining the divergence from the original per-slot proposal.
- Corrected the "Per `CrossTurnSlotId`" retention section and the "one snapshot per
  slot" GDN-decision rationale to describe the single-live-entry contract.
- Marked the "Serve/Chat Hook" section and its "Components" list explicitly
  **DEFERRED (D6 follow-up)** — verified via `grep -rln "with_prefix_cache"` across
  `crates/` and `apps/` that neither `lattice_serve.rs` nor `chat_metal.rs`
  reference any `*_with_prefix_cache` entry point today, so the claim is accurate,
  not just relabeled.
- Fixed the trailing whitespace on report lines 3-5 that `git diff --check` flagged.

### Gates (all run in this worktree, real Metal hardware, post-D7/D8 commits)

- `cargo test -p lattice-inference --features metal-gpu cross_turn` — **25 passed**, 0 failed (24 round-1 + T8)
- `cargo test -p lattice-inference kv_cache` — **102 passed**, 0 failed
- `cargo clippy --workspace -- -D warnings` — clean
- `cargo clippy --workspace --features metal-gpu -- -D warnings` — clean
- `cargo fmt --check` — clean
- `git diff --check origin/main...HEAD` — clean (post-commit; the report's trailing whitespace is fixed)

Commits: `800a811b0` (D7 fix + T8), `78bbe7fee` (D8 report drift fix).

Still **draft**; D6 (serve/chat wiring) remains explicitly out of scope for this PR.

### DIFF
--- a/.khive/reports/crossturn_kv_462.md
+++ b/.khive/reports/crossturn_kv_462.md
@@ -1,8 +1,8 @@
 # Design: Cross-Turn KV Prefix Cache for Lattice #462
 
-Date: 2026-07-01  
-Role: architect, op 4, team `lattice_kv_462`  
-Status: proposed design; interfaces below are committed for implementation unless the critic rejects them.  
+Date: 2026-07-01
+Role: architect, op 4, team `lattice_kv_462`
+Status: shipped design (round-2 remediation, 2026-07-01). Interfaces below reflect the runtime shape as merged, not the original per-slot proposal (see "Round-2 remediation" note below).
 Inputs: `../analyst/design_brief.md`, direct code scan of `/Users/lion/projects/khive/lattice-kvcache-462`.
 
 Tooling note: `li team receive -t 2d5c82e54e73 --as architect` failed because `li` is not installed. Fallback `comm.inbox()` via khive returned count 0.
@@ -181,16 +181,32 @@ struct MetalCrossTurnPrefixEntry {
 
 #[derive(Debug, Default)]
 struct MetalCrossTurnPrefixCache {
-    entries: std::collections::HashMap<CrossTurnSlotId, MetalCrossTurnPrefixEntry>,
+    entry: Option<MetalCrossTurnPrefixEntry>,
 }
 
 impl MetalCrossTurnPrefixCache {
     fn get(&self, slot_id: CrossTurnSlotId) -> Option<&MetalCrossTurnPrefixEntry>;
     fn insert(&mut self, entry: MetalCrossTurnPrefixEntry);
+    fn take(&mut self, slot_id: CrossTurnSlotId) -> Option<MetalCrossTurnPrefixEntry>;
     fn remove(&mut self, slot_id: CrossTurnSlotId);
     fn clear(&mut self);
 }
 ```
+
+**Round-2 remediation (as shipped, differs from the design above):** the runtime
+shape is `entry: Option<MetalCrossTurnPrefixEntry>` — a single live entry per
+`MetalQwen35State`, **not** a `HashMap<CrossTurnSlotId, _>`. `get`/`take`/`remove`
+all filter by `slot_id` against that one entry (a different or absent slot is
+always a miss). `insert` unconditionally **replaces** whatever is currently
+retained, for any slot — saving a new entry for slot B silently evicts slot A's
+entry. This is a deliberate, bounded-by-construction design (never an unbounded
+per-slot map): at most one ~19.3 MiB GDN snapshot is ever alive, at the cost of
+zero retention across concurrent multi-slot traffic on the same `MetalQwen35State`
+(single-slot-at-a-time is the actual v1 contract; see "Round-2 remediation" note
+under "What The Cache Retains" below). The original per-slot `HashMap` sketch
+above was rejected in round-2 remediation because it was the source of the
+round-1/round-2 unsoundness findings (stale per-slot entries surviving live
+state mutation that the map had no way to observe).
 
 Extend `MetalQwen35State`:
 
@@ -278,7 +294,13 @@ impl MetalQwen35State {
 
 ## What The Cache Retains
 
-Per `CrossTurnSlotId`:
+**Round-2 remediation (as shipped):** at most ONE entry is retained per
+`MetalQwen35State`, not one per `CrossTurnSlotId`. The fields below describe
+that single retained entry when present; it is tagged with the `slot_id` it
+was saved for, and `get`/`take`/`remove` filter on that tag, but saving a new
+entry for a different slot unconditionally evicts whatever was retained
+before, regardless of slot. The "Per `CrossTurnSlotId`" framing in the
+original design (below) described the rejected multi-slot `HashMap` shape:
 
 - `token_ids: Vec<u32>`: exact rendered prompt tokens represented by state.
 - `represented_len: usize`: authoritative number of tokens in all reusable state.
@@ -289,6 +311,17 @@ Per `CrossTurnSlotId`:
 The v1 cache must not store every token-boundary GDN snapshot. One exact latest-boundary snapshot is enough for normal append-only chat and costs about 19.3 MiB for Qwen3.5-0.8B. Every-token GDN snapshots are rejected for v1 because they are about 19.3 MiB per boundary.
 
 ## Serve/Chat Hook
+
+**Round-2 remediation (status, 2026-07-01): DEFERRED, not implemented in this PR.**
+Everything in this section — the `Job.slot_id` extension, `request_slot_id`, the
+worker loop replacement, and the `chat_metal.rs` REPL/JSON-mode wiring — is design
+intent for a follow-up PR (D6), not a shipped component of #516. This PR ships only
+the correctness substrate: `MetalCrossTurnPrefixCache`,
+`generate_streaming_with_prefix_cache`, `chat_completion_streaming_with_prefix_cache`,
+and the round-2 raw-forward invalidation boundary. `lattice_serve.rs` and
+`chat_metal.rs` do not currently call any `*_with_prefix_cache` entry point, so
+neither the serve worker nor the chat REPL opt into cross-turn reuse yet — they
+continue to use `reset_state()`/`generate_streaming` exactly as before this PR.
 
 ### `lattice_serve.rs`
 
@@ -471,7 +504,7 @@ Why this is enough:
 
 ## GDN State Decision
 
-Chosen v1 option: persist one exact GDN snapshot at the latest represented boundary per cache slot.
+Chosen v1 option (as shipped): persist one exact GDN snapshot at the latest represented boundary, for at most one live entry across the whole `MetalQwen35State` (round-2 remediation — see "What The Cache Retains"; the original design below said "per cache slot" / "per slot", implying independent retention across slots, which the shipped single-entry shape does not provide).
 
 Alternatives considered:
 
@@ -479,7 +512,7 @@ Alternatives considered:
 |---|---:|---:|---|
 | Reuse attention KV only | No | Low | Rejected. GDN state would not match full re-prefill. |
 | Snapshot every token boundary | Yes | About 19.3 MiB per boundary for 0.8B | Rejected for v1 memory cost. |
-| One latest-boundary snapshot | Yes for append-only `L == represented_len` | About 19.3 MiB per slot | Chosen v1. |
+| One latest-boundary snapshot | Yes for append-only `L == represented_len` | About 19.3 MiB, at most one snapshot alive at a time (single live entry, round-2) | Chosen v1. |
 | Sparse checkpoints + replay | Yes if replay is token-identical | `N * 19.3 MiB` plus replay | v2, optional for edited-history speedup. |
 | Replay whole shared prefix | Correct if full model replay | O(shared prefix) | Fallback only; not the feature win. |
 
@@ -559,19 +592,20 @@ sequenceDiagram
 
 ## Coupling And Testability
 
-Components:
+Components (design-time list; **items 4 and 5 are DEFERRED follow-up work, not
+shipped in #516** — see "Serve/Chat Hook" above):
 
-1. `kv_cache::cross_turn` planner,
-2. `MetalCrossTurnPrefixCache`,
-3. `MetalQwen35State` cache-aware generate/chat methods,
-4. serve worker job protocol,
-5. chat_metal REPL integration.
+1. `kv_cache::cross_turn` planner — shipped.
+2. `MetalCrossTurnPrefixCache` — shipped (as a single-live-entry `Option`, round-2 remediation).
+3. `MetalQwen35State` cache-aware generate/chat methods — shipped.
+4. serve worker job protocol — deferred (D6 follow-up).
+5. chat_metal REPL integration — deferred (D6 follow-up).
 
 Direct dependencies after design:
 
 - planner has no dependency on Metal,
 - Metal cache depends on planner and GDN snapshot type,
-- serve/chat binaries depend only on public cache-aware methods and `CrossTurnSlotId`.
+- serve/chat binaries would depend only on public cache-aware methods and `CrossTurnSlotId` once D6 lands; today neither binary references them.
 
 Estimated coupling: `|Deps|=5`, `|C|=5`, `κ=5/(5*4)=0.25`, below the 0.3 target.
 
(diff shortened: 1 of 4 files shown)


ohdearquant:
Update: the play's tester op had written the **token-identity correctness gate** to the worktree after the preservation commit — 3 Metal integration tests now on the branch (`cross_turn_cache_multiturn_token_identity_matches_full_reprefill`, prefix-mismatch fallback, clear-forces-refill). Running them now; PR graduates from 'no test written' to 'gate run pending'.

ohdearquant:
**Correctness gate: PASS (3/3).** Ran `cargo test -p lattice-inference --features metal-gpu cross_turn_cache` on this branch (M2 Max, real Metal):

- `cross_turn_cache_multiturn_token_identity_matches_full_reprefill` ✅ — multi-turn generation through the prefix cache is token-identical to full re-prefill on fresh state, **with a non-vacuity assertion** that `reused_tokens > 0` on later turns (ExactAppend actually engaged; the test cannot pass via silent FullRefill fallback).
- `cross_turn_cache_prefix_mismatch_falls_back_to_full_refill` ✅
- `cross_turn_cache_clear_forces_full_refill_next_call` ✅

3 passed / 0 failed in 0.80s. This satisfies the AUTO-if-token-identical queue condition. Remaining before merge: clippy (running), CI incl. e2e-parity, cross-family review, maintainer diff-read.

ohdearquant:
Cross-family review round 1: **REJECT** — 2 blockers + 4 majors, all confirmed by maintainer code-read.

Core defect: the cache retains multiple slot entries (GDN snapshot + metadata per slot) while the full-attention KV lives in the single mutable `self.session.kv_cache` buffer. ExactAppend can therefore restore slot A's GDN state while attention reads KV rows written by slot B (or by any plain `generate_streaming` call, since `reset_state` does not invalidate the cache) — a silent token-identity violation. Additional confirmed issues: shape-only LoRA adapter identity, error (instead of fallback) on exact-equal prompt retry, unbounded per-slot snapshot memory, and the serve path not yet opted in.

Remediation (this branch, round 2 to follow): single-live-entry cache with take-on-restore ownership (any early return leaves the cache empty → next call FullRefills), `reset_state` + LoRA load/unload invalidate, planner requires a non-empty suffix for ExactAppend, plus slot-isolation / interleaved-path / exact-equal regression tests. Serve wiring stays a follow-up PR.

ohdearquant:
Round-3 review remediation (f73518c7d): reviewer confirmed all five round-2 invalidation boundaries hold and T8 is mutation-sensitive, with one remaining medium finding — `set_gdn_chunked` swaps the GDN prefill algorithm (chunked vs serial measured non-bit-exact, logit drift up to ~2e-1 at chunk boundaries) while preserving the retained cross-turn entry, so a snapshot saved under one mode could seed a restore whose full-refill reference uses the other.

Fix: flipping the mode now clears the retained entry; re-asserting the current mode is a no-op and keeps it. New regression test `cross_turn_cache_invalidated_by_gdn_mode_flip` covers both directions and is mutation-sensitive (verified failing with the clear removed, passing restored). cross_turn suite 26/26, clippy -D warnings clean, fmt clean.

PR #522, opened 2026-07-01
ci: always-reporting app-binaries-gate (ADR-066 F-1 rollout, step 1)

ohdearquant:
> _PR authored by Claude (Anthropic agent) on behalf of @ohdearquant._

## What

Restructures `app-binaries.yml` into the same required-check shape as `e2e-parity.yml`: the workflow now triggers on every PR, a cheap `detect-app-bin-changes` job decides whether the expensive macOS build is needed, and a new always-reporting `app-binaries-gate` job is the status context intended to become required.

## Why

ADR-066 F-1 approved making the app-binaries build a required check. The raw `build app-shipped binaries` job cannot be marked required as-is: it is path-filtered, so docs-only PRs would never receive the status and would wedge on "Expected — waiting for status to be reported". The gate job passes when no app-binary-relevant file changed, mirrors the build result when one did, and fails closed if change detection itself fails.

## Review round 1 remediation (d78dc6b58)

Adversarial review before required-check promotion found three fail-opens, all fixed in this PR:

1. **Fail-safe change detection**: the detector's `echo "$CHANGED" | grep -qE` under `pipefail` could hit SIGPIPE (status 141) on large diffs and silently skip the build. Replaced with a here-string read; stress-verified at 5,000 and 200,000 trailing changed files.
2. **`lattice_serve` coverage**: `scripts/build-app-bins.sh` now also builds `lattice_serve --features f16,metal-gpu`, matching the full set `apps/macos/scripts/package-app.sh` ships. Live CI on this PR logs the build.
3. **Detection coverage**: `apps/macos/scripts/package-app.sh` (source of truth for the shipped binary set) added to the detection regex, and the push paths filter now covers the build script, Cargo files, and the workflow itself.

## Rollout

1. **This PR**: gate lands non-required and starts reporting on every PR.
2. **Step 2 (separate)**: after the gate has reported reliably, flip the branch ruleset to require `app-binaries-gate` (and drop any requirement on the raw build job name).

The workflow file itself changes in this PR, so the build job runs here and exercises the full path, including the `lattice_serve` build.

### DIFF
--- a/.github/workflows/app-binaries.yml
+++ b/.github/workflows/app-binaries.yml
@@ -1,8 +1,9 @@
 name: app-binaries
 # Regression gate: build every binary the macOS Studio app ships, each with its
 # required-features. `cargo build --workspace` skips required-features bins, so
-# this is the only CI that compiles chat_metal / eval_perplexity / train_grad_full
-# / generate_lora / embed with the features the app actually builds them with.
+# this is the only CI that compiles chat_metal / eval_perplexity / lattice_serve
+# / train_grad_full / generate_lora / embed with the features the app actually
+# builds them with.
 # See scripts/build-app-bins.sh for the rationale and the two regressions that
 # motivated it.
 #
@@ -21,6 +22,11 @@ on:
     branches: [main]
     paths:
       - 'crates/**'
+      - 'Cargo.toml'
+      - 'Cargo.lock'
+      - 'scripts/build-app-bins.sh'
+      - 'apps/macos/scripts/package-app.sh'
+      - '.github/workflows/app-binaries.yml'
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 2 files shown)


ohdearquant:
Round 1 review: request changes, three findings, all remediated in d78dc6b58.

1. Fail-open detection on large diffs: `echo | grep -qE` under pipefail hits SIGPIPE (status 141) once grep exits at first match, which routes to the else branch and skips the required build. Reproduced locally at 200k changed files. Fixed with a here-string read; verified against match-first + 5000 and 200k trailing paths, package-app.sh-only, and docs-only cases.
2. `scripts/build-app-bins.sh` did not build `lattice_serve`, which `package-app.sh` ships with `--features f16,metal-gpu`. Added.
3. `apps/macos/scripts/package-app.sh` (source of truth for the shipped binary set) was not a detected path. Added to the PR regex and the push paths filter, along with the build script, Cargo files, and the workflow itself.

Round 2 review will verify the live CI log includes the `lattice_serve` build line.

PR #538, opened 2026-07-02
fix(inference): bound IncrementalDetokenizer retention to the decode-boundary tail window (#324)

### LINKED ISSUE #324: perf(inference): IncrementalDetokenizer retains whole-generation byte buffer (flushed prefix unreclaimed) — LOW, fix is a deferred text()-contract refactor

## perf(inference): `IncrementalDetokenizer` retains the whole generation's bytes (flushed prefix never reclaimed)

Discovery pass + verification over the streaming detokenizer (`model/qwen35/detokenize.rs`). The UTF-8 boundary logic itself is **correct** (verified split-codepoint, truncation, lone-continuation, surrogate/overlong, and maximal-subpart `U+FFFD` cases by hand against `from_utf8_lossy` semantics; the in-file test suite covers the main cases). One real-but-low perf finding remains.

### Finding

`IncrementalDetokenizer` (`detokenize.rs:55`) holds a single `bytes: Vec<u8>` and a `flushed: usize` cursor. `flush_complete` advances `flushed` but never shrinks `bytes`, so the buffer grows to the full decoded length of the generation. Live in both production streaming paths (`model/qwen35/generation.rs:181` CPU, `forward/metal_qwen35.rs:13155` Metal).

### Severity: LOW (downgraded on verification from an initial "medium" rating)

- **Per-generation, not a leak.** A fresh `IncrementalDetokenizer::new()` is created per `generate()` call and dropped when it returns. Growth is bounded by one generation's output, not accumulated across requests. Typical generations are KB-scale (≈4 bytes/token × hundreds–few-thousand tokens); even a 32K-token stream is ≈128KB for one in-flight request.
- **`text()` requires the full buffer by contract.** Both call sites finish with `text: detok.text()` (`generation.rs:231`, `metal_qwen35.rs:13169`) to populate `GenerateOutput.text`, and `text()` = `from_utf8_lossy(&self.bytes)`. The returned struct holds O(output) text regardless, so the detok buffer is the *same order* of memory, not an amplification beyond the API's inherent cost.

### Why the obvious fix is counterproductive

Compacting `bytes` (drain `[0..flushed]`) while keeping an internal `emitted: String` for `text()` does **not** reduce peak memory: `emitted` becomes O(output) *plus* the final `text()` allocation — strictly worse than storing the bytes once and materializing the String on demand. It would also break the in-file test equivalence assertion (`out == d.text()`).

### The only genuine reduction (deferred design item)

Drop `text()` from the per-token streaming contract and assemble `GenerateOutput.text` from the deltas the caller already receives via `on_token` + the `finish()` tail. Then `IncrementalDetokenizer` need only retain the unflushed tail (≤3 bytes in the common case). This touches both production call sites and the test contract, so it is a scoped refactor, not an autonomous self-merge. Filing for tracking; **no PR** — a naive compaction patch would regress peak memory.

Related discovery thread: #322 (grammar-runtime), #310 (grammar-compiler).

ohdearquant:

## Summary

`IncrementalDetokenizer` previously retained **every decoded byte for the entire lifetime of a generation** (a growing `bytes: Vec<u8>` plus a `flushed` cursor) so that `text()` could reconstruct the whole output on demand. For long generations this grows memory **linearly with token count**, even though the only state a streaming detokenizer needs is the small tail around the current UTF-8 decode boundary.

This PR bounds that retention to a tiny tail window:

- The struct now holds a single `pending: Vec<u8>` containing only the **undecided UTF-8 boundary bytes** (the incomplete trailing scalar, at most 3 bytes between pushes).
- After every `push`/`finish`, consumed bytes are `drain`ed and the buffer capacity is compacted back down to `DETOK_RETAINED_BYTE_CAPACITY` (8 bytes). A `Vec` does not shrink capacity on `drain`/`clear` alone, so the explicit compaction is what actually caps the allocation.
- Whole-generation text now accumulates in the generation call sites that already own `GenerateOutput.text` (the CPU no-stop streaming path and both Metal streaming paths). String-stop paths already accumulated their own full text and are unchanged.
- `text()` is removed (not stubbed); it has no remaining callers.

### Fail-closed boundary rule (preserved exactly)

Byte-level BPE can split a single codepoint across tokens, so flushing only complete UTF-8 per token is a hard invariant. The boundary logic is unchanged in behavior:

- `Ok(s)` → emit all, consume all.
- `Err` with `valid_up_to() > 0` → emit only the proven-valid prefix, advance by exactly that many bytes.
- `error_len() == None` (incomplete trailing codepoint) → **retain every remaining byte, emit nothing** — a later token may complete it. This is the fail-closed branch: when in doubt at a boundary, retain.
- `error_len() == Some(len)` → emit one `U+FFFD` and advance by `len` (only Rust-classified unrepairable bytes are ever dropped — never bytes that could still complete a codepoint).
- `finish()` is the sole lossy flush, at generation end.

## Tests

- **Bounded-memory regression** (`incremental_detokenizer_retention_bounded_for_long_ascii_stream`): asserts retained capacity stays `<= 8` across an 8,192-token stream — bound is independent of generation length.
- **UTF-8-boundary parity** (`incremental_detokenizer_utf8_parity_matches_full_decode_for_split_stream`): a CJK / emoji / combining-mark / ZWJ token stream streams **byte-identical** to the accumulate-everything `decode_tokens` reference.
- Both tests are mutation-verified as load-bearing: removing the retention bound makes the memory test fail; an over-eager drop of the incomplete tail makes the parity test fail.
- Full workspace suite green; `clippy --workspace -D warnings` and `fmt --check` clean.

## Bench comparison (CPU-only A/B)

`make bench-compare` (`origin/main` → `HEAD`, quick mode). The diff touches only decode bookkeeping (`Vec<u8>`/`String`) in the `lattice-inference` crate, so `elementwise_cpu_bench` is the nearest CPU group with production relevance (there is no detokenizer-specific Criterion bench in the repo). Base = `origin/main`, Head = this branch.

### `elementwise_cpu_bench` (lattice-inference) — before/after

| Benchmark | Base (median) | Head (median) | Change (CI) | p-value | Verdict |
|---|---|---|---|---|---|
| rms_norm/896 | 130.83 ns | 135.06 ns | +2.81% .. +4.95% | 0.08 | noise |
| rms_norm/4096 | 558.69 ns | 565.97 ns | +1.10% .. +2.11% | 0.08 | noise |
| layer_norm/896 | 206.74 ns | 204.02 ns | −1.67% .. −0.02% | 0.15 | noise |
| layer_norm/4096 | 899.04 ns | 892.94 ns | −1.15% .. +0.08% | 0.27 | noise |
| silu_inplace/896 | 451.18 ns | 455.61 ns | −0.76% .. +1.42% | 0.77 | noise |
| silu_inplace/4096 | 2.0544 µs | 2.0516 µs | −0.55% .. −0.03% | 0.10 | noise |
| gelu/896 | 369.22 ns | 366.72 ns | −1.13% .. −0.56% | 0.05 | noise (tiny improvement) |
| gelu/4096 | 1.6716 µs | 1.6799 µs | −0.70% .. +0.80% | 0.89 | noise |
| add_bias_gelu/896 | 386.91 ns | 390.68 ns | +0.24% .. +2.16% | 0.15 | noise |
| add_bias_gelu/4096 | 1.7661 µs | 1.7784 µs | −0.45% .. +2.22% | 0.53 | noise |
| softmax_attention/128 | 3.5945 µs | 3.6582 µs | +1.48% .. +2.94% | 0.10 | noise |
| softmax_attention/512 | 62.496 µs | 63.684 µs | +1.43% .. +2.67% | 0.08 | noise |
| elementwise_mul/4096 | 279.18 ns | 280.67 ns | +0.21% .. +0.70% | 0.12 | noise |

All 13 benchmarks are statistically indistinguishable from base (every p-value ≥ 0.05). Expected: none of these kernels are on the detokenizer code path; the change is token-to-text bookkeeping only. No measurable regression. No Metal/GPU benches were run (this change is CPU-only by nature).

## Reviewer note (behavior change on caller-interrupt only)

On a **caller interrupt** (which returns/breaks without calling `finish()`), the aggregate `GenerateOutput.text` no longer appends a trailing `U+FFFD` for a half-received codepoint present at the interrupt boundary — that partial byte is simply omitted. The streamed `on_token` deltas are byte-identical in both versions, and **all non-interrupt paths** (stop / kv-full / max-tokens / natural-end) call `finish()` and remain byte-identical to the previous behavior. This is an intentional consequence of moving accumulation to the callers, and is arguably more correct (a partial output shouldn't carry a bogus replacement char for a codepoint the caller chose to interrupt mid-way). Flagged here purely for transparency.

Closes #324

### DIFF
--- a/crates/inference/src/forward/metal_qwen35.rs
+++ b/crates/inference/src/forward/metal_qwen35.rs
@@ -14154,20 +14154,26 @@ kernel void gdn_chunk_norm_silu_c32(
                 reasoning_end_len = Some(generated_ids.len());
             }
             let mut last_pushed_id = next_id;
+            // `text` is the caller-owned full output — the detokenizer itself only
+            // retains a small undecided UTF-8 boundary tail (see IncrementalDetokenizer).
+            let mut text = String::new();
             let delta = detok.push(tokenizer, next_id);
-            if !delta.is_empty() && !on_token(&delta, next_id) {
-                if use_compact {
-                    self.session.compact_topk = 0;
-                    self.session.compact_route = GpuTopkRoute::CpuFallback;
+            if !delta.is_empty() {
+                text.push_str(&delta);
+                if !on_token(&delta, next_id) {
+                    if use_compact {
+                        self.session.compact_topk = 0;
+                        self.session.compact_route = GpuTopkRoute::CpuFallback;
+                    }
+                    return GenerateOutput {
+                        text,
+                        token_ids: generated_ids.clone(),
+                        prompt_tokens: prompt_len,
+                        generated_tokens: generated_ids.len(),
+                        stopped: false, // caller interrupted the stream, not a stop condition
+                        stop_reason: Some(StopReason::Interrupt),
+                    };
                 }
-                return GenerateOutput {
-                    text: detok.text(),
-                    token_ids: generated_ids.clone(),
-                    prompt_tokens: prompt_len,
-                    generated_tokens: generated_ids.len(),
-                    stopped: false, // caller interrupted the stream, not a stop condition
-                    stop_reason: Some(StopReason::Interrupt),
-                };
             }
             let mut stopped = false;
             let mut stopped_by_caller = false;
@@ -14280,10 +14286,10 @@ kernel void gdn_chunk_norm_silu_c32(
                     // above. A late cancel here (return value intentionally unused)
                     // does not change why generation stopped, so treating it as
                     // Interrupt would misattribute the stop cause to this flush.
+                    text.push_str(&tail);
                     on_token(&tail, last_pushed_id);
                 }
             }
-            let text = detok.text();
             GenerateOutput {
                 text,
                 token_ids: generated_ids.clone(),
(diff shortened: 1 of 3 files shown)


ohdearquant:

Review round: two findings fixed in f419f90c5.

1. **(major)** The non-prefix Metal streaming decode loop passed each delta to `on_token` without appending it to the caller-owned `text` accumulator, so `GenerateOutput.text` on that path dropped every decode-loop token (only the prefill token and the `finish()` tail were appended). Fixed to match the prefix-cache path: append unconditionally, then check the interrupt return. The documented interrupt semantics are unchanged (interrupting delta retained, trailing unfinished tail omitted). A Metal-gated regression test now asserts `out.text` equals the concatenation of every `on_token` delta; it compiles under `clippy --features metal-gpu` and will be executed before merge (GPU currently reserved by a long-running measurement on this machine).

2. **(minor)** The retention-bound test used only 1-byte ASCII tokens, so it could not detect removal of the capacity compaction. Added a companion test pushing a 12-byte single-token delta (exceeding the 8-byte retained-capacity bound), mutation-verified: with the compaction call commented out it fails with `capacity 16`; restored, all 8 detok tests pass.

Full CPU suite green (1358 passed), fmt/clippy clean including `--features metal-gpu --all-targets`.

ohdearquant:

`make bench-compare` (quick), base `origin/main @ 9997e54bb` vs head `f419f90c5`:

- **Inference-side groups** (`elementwise_cpu_bench`: `add_bias_gelu`, `gelu`, `elementwise_mul` — the crate this PR touches): all within noise, no change detected.
- **Embed SIMD groups** initially showed a uniform positive drift across every group (worst point estimate `int4_cosine_distance/float32_simd/384` +18.6%). This PR touches only `crates/inference` (3 files, detokenizer/streaming), no lattice-embed source, and a uniform same-sign shift across 200+ unrelated microbenches is the machine-state signature of the two-worktree A/B, not a regression.
- **Discriminating rerun** of the worst group solo on an idle machine against the same saved baseline: every bench reports "No change in performance detected" — the +18.6% outlier collapsed to +1.13% (p = 0.10 > 0.05) and the next-worst (+6.4%) flipped sign to −6.5% (p = 0.06 > 0.05).

Conclusion: bench-compare showed no change (p > 0.05 on all groups under controlled rerun).

Issue #540, opened 2026-07-02
Metal Q4 mmap loader should reject truncated payloads before dispatch


ohdearquant:
> _Issue filed by Codex (OpenAI agent) on behalf of @ohdearquant._

The Metal Q4 no-copy load path validates that a `.q4` mmap reaches the header payload offset, but it does not validate that the file contains the full block payload required by the tensor/model dimensions before returning a `Q4WeightBuf`.

Evidence from current source:

- `crates/inference/src/forward/metal_qwen35.rs`: `mmap_q4_weight` calls `read_q4_header`, mmaps the whole file, and only asserts `mmap.len() as u64 >= header.payload_offset` before returning `Q4WeightBuf::from_mmap(...)`.
- The CPU deserializing path in `crates/inference/src/weights/q4_weights.rs` is stricter: `load_q4_file` computes `n_blocks = original_len.div_ceil(32)`, guards `n_blocks * 20` via `checked_alloc_bytes(..., "block payload")`, and rejects headers whose claimed payload exceeds the file length.
- `mmap_q4_weight` is used for direct Metal Q4 buffers, including per-layer projection files and `embed_tokens` / `lm_head` Q4 files. A truncated `.q4` can therefore pass loader construction and defer the failure to later GPU buffer reads instead of failing closed at load time.

Suggested fix:

Extend the header metadata or the mmap helper validation so the no-copy path computes the expected payload byte count (`original_len.div_ceil(32).checked_mul(20)`) and rejects `mmap.len() < payload_offset + expected_payload_bytes` before creating the Metal buffer. Keep the existing header-offset check, but make it a returned `Err` rather than an `assert!` so corrupt model files fail cleanly in release and debug.

A regression test can exercise the validation helper without running Metal by constructing a minimal KHQ4 header whose `original_len` claims at least one block but whose file ends at the header or with a short payload.

PR #541, opened 2026-07-02
fix(inference): correct partial-block asymmetric Q4 scale (#346)

### LINKED ISSUE #346: Q4/QuaRot quantization: partial-block scale, load_q4_file overflow guard, QuaRot embed symmetry (measure-first) — core math sound

An adversarial sweep of the Q4/QuaRot quantization path found the **core dequant math sound** (asymmetric `nibble*scale+bias` self-consistent encode/decode; nibble order `&0xF`=even / `>>4`=odd consistent across the Rust decoder, Metal `gemv_q4_decode`, and `gemm_q4`/`gemm_q4_tiled`; RandomizedHadamard `apply`/`apply_inverse` orthonormal with the `1/sqrt(n)` factor accounted once; Q8 casts to f32 before accumulation so no int8 overflow; zero-row scale is `1.0` not `0.0`). The following are minor edge/quality items (severities are TBV-corrected down from the initial sweep — see notes):

### 1. Partial last block widens the asymmetric scale (P3, quality)
`crates/inference/src/weights/q4_weights.rs` — `quantize_block_with_mode(.., symmetric=false)` (L265-266) computes `min_val`/`max_val` over the full padded `[f32;32]` array. Callers (`quantize_f32_to_q4` L433, `quantize_bf16_to_q4`, `quantize_row_q4_0`) zero-pad the final partial chunk, so for a tail like `[5,6,7]+29·0` the asymmetric scale is computed over `min=0,max=7` (→`7/15`) instead of `min=5,max=7` (→`2/15`) — ~3.5× too coarse for those tail elements.

**Impact corrected to P3:** chunking is over the *flat* tensor, so this affects exactly **one** block (≤31 weights) per tensor, not per row — negligible PPL. (Symmetric mode is unaffected: padding zeros don't change `abs_max`.) Clean fix: compute min/max over `chunk.len()` real elements, then zero-pad only for packing.

### 2. `load_q4_file` unchecked `usize` multiply on the length header (P3, robustness)
`load_q4_file` (~L682): `let mut raw = vec![0u8; n_blocks * 20];` where `n_blocks = original_len.div_ceil(32)` and `original_len` is read straight from the file header. An adversarial/corrupt `.q4` with `original_len ≈ usize::MAX` overflows the multiply.

**Impact corrected to P3:** release mode wraps to a small alloc → the subsequent `read_exact` fails with a clean I/O error (no corruption); only **debug** builds panic. Same hardening class as #341 (invert_vocab alloc cap). Fix: `n_blocks.checked_mul(20).ok_or_else(|| io::Error::...)`.

### 3. QuaRot quantizes embed_tokens/lm_head symmetrically — needs measurement (P2, measure-first)
`quantize_f64_to_q4` (L469) passes `symmetric=true` for **all** planned tensors. The "rotated weights are zero-mean" assumption holds for interior linear layers, but `embed_tokens`/`lm_head` get only an output-side rotation and may retain a non-trivial row mean, where symmetric Q4 wastes ~half its levels. **Action:** measure `mean(|row_mean|)/abs_max` on post-rotation embed_tokens of a real Qwen3.5 checkpoint; only switch those tensors to asymmetric if the skew is material (>~5% abs_max). No code change until measured.

### 4. Doc bug (P3, cosmetic)
`quantize_block` docstring (L228-235) describes the symmetric formula (`abs_max/7.0`) but calls `quantize_block_with_mode(.., false)` (asymmetric). Encode/decode are consistent; only the comment is wrong.

---

ohdearquant:

Addresses the four bounded edge items in #346. The core Q4/QuaRot dequant math was already sound; each item's premise was re-verified against current source before any change, since line numbers in the issue had drifted.

## Item 1 — partial-block asymmetric scale (fixed)

`quantize_block_with_mode(.., symmetric=false)` computed min/max over the zero-padded `[f32;32]` array, so a partial tail block took its scale from the padding. Example: a `[5,6,7]`+29-zeros tail derived `min=0` instead of `min=5`, producing a coarser scale/bias than a real 32-element block.

Fix: a new length-aware helper computes the asymmetric min/max over only the real `chunk.len()` elements; zero-padding is used solely for packing. All five zero-padding call sites now pass the real element count. Symmetric mode is unchanged and remains **bit-identical** — it still folds over the full padded array, and padding zeros can never change `abs_max`; a bit-identity test asserts this.

The standalone `quantize_block` wrapper became dead code once the length-aware helper landed and was removed (the `-D warnings` clippy gate flags an unused function otherwise).

## Item 2 — load-time overflow guard (already present)

Re-verified: `checked_alloc_bytes` already guards every header-derived allocation in `q4_weights.rs` via `count.checked_mul(elem_size)` plus a file-length bound, returning a clean `io::Error` (no panic/`unwrap`). This covers `load_q4_file` (ndim, shape-product, and the `n_blocks * 20` block payload the issue named), `read_q4_header` (ndim), and the sibling `load_f16_tensor_file` (ndim, `numel * 2`). No sibling "untrusted header count × elem_size → allocation" path was left unguarded. No production change was needed; this PR adds mutation-sensitive tests to lock the guard in (evidence commented on #346).

## Item 4 — docstring (already consistent)

The `quantize_block` docstring already read "asymmetric mode" and matched its body. No mismatch remained; the wrapper it documented was removed under Item 1's refactor (evidence commented on #346).

## Item 3 — embed/lm_head symmetry (measured, report-only)

Measured `mean(|row_mean|)/abs_max` on the post-rotation `embed_tokens` and materialized `lm_head` of the real Qwen3.5-0.8B checkpoint (CPU-only, converter-identical rotation path): **0.5559%** and **0.0306%** of `abs_max` respectively — both far below the ~5% materiality threshold. No asymmetric switch is added for these tensors and no switch test; report-only with a follow-up note. Full method + numbers are posted as a comment on #346.

## Tests / mutation-proof evidence

New tests in `q4_weights.rs`, each proven mutation-sensitive by reverse-applying the fix in place, confirming the test FAILS, then restoring and confirming it PASSES:
- **Partial-block scale** — asserts the `[5,6,7]` tail gets `scale=(7-5)/15, bias=5.0` and explicitly does *not* match the padded-array result; FAILS if min/max reverts to the padded array.
- **Symmetric bit-identity** — partial symmetric block stays byte-identical to the padded path; FAILS if the symmetric scale is coupled to `valid_len`.
- **Corrupt-header guards** — clean `Err` for `original_len`/`numel` near `usize::MAX`, and for a `numel` whose `×elem_size` overflow wraps to a small residue. The wrap-to-small case is the important one: with `checked_mul` removed it returns `Ok` with silently wrong data (not just an OOM), which the file-length bound alone cannot catch. FAILS with the guard removed.

## Gates (CPU-only)

`cargo fmt --check`, `cargo clippy --workspace -- -D warnings`, `cargo check --workspace`, and `cargo test --workspace` (incl. doctests) all pass; the inference lib suite reports 1360 passed / 0 failed / 7 ignored. No Metal/GPU test, bench, or binary was run — a separate serialized GPU measurement is active on this machine and the touched code is CPU load-time only.

## Bench

`make bench-compare` was not run. The touched path (`quantize_block_with_mode_len` and its five call sites) is invoked only at quantization/load time, never on the hot forward path, and no existing CPU micro-bench targets it; stating this plainly rather than fabricating a delta or contending with the concurrent GPU measurement.

## Out-of-scope defect filed

A separate, adjacent fail-closed gap in the Metal Q4 mmap loader (validates only the header offset, not the full declared payload length, so a truncated `.q4` payload can pass) was filed as #540 rather than folded in here — it is a no-copy mmap path, not the header-derived allocation class this PR addresses.

Closes #346

### DIFF
--- a/crates/inference/src/weights/q4_weights.rs
+++ b/crates/inference/src/weights/q4_weights.rs
@@ -229,38 +229,38 @@ fn bf16_to_f32(v: u16) -> f32 {
 // Core block quantization
 // ---------------------------------------------------------------------------
 
-/// Quantize exactly 32 f32 values into one [`Q4Block`] using asymmetric mode.
+/// Quantize one block from only the first `valid_len` real values of `vals`;
+/// the remaining `32 - valid_len` slots are caller-supplied zero padding used
+/// solely to fill the fixed-size packing loop below.
 ///
-/// Nibble layout: sequential pairs — `byte[b] = (q[2b+1] << 4) | q[2b]`.
-///
-/// # Errors
-///
-/// Returns [`InferenceError::InvalidInput`] if any value in `vals` is non-finite.
-/// IEEE-754 `NaN > x` is always false, so a NaN would silently leave the
-/// min/max accumulators unchanged and produce wrong-but-no-error quantization.
-#[inline]
-fn quantize_block(vals: &[f32; 32]) -> Result<Q4Block, InferenceError> {
-    quantize_block_with_mode(vals, false)
-}
-
-/// Quantize one block; symmetric mode is optimal for Hadamard-rotated weights
-/// (which are zero-mean by construction), asymmetric is optimal for raw weights
-/// with non-zero distributional center. Both modes share the same on-disk
-/// format: `dequant = nibble * scale + bias`. Symmetric mode sets
-/// `bias = -8 * scale` so that nibble=8 maps to exactly 0.
+/// Asymmetric mode derives `min_val`/`max_val` from the real `valid_len`
+/// elements only, so a zero-padded tail block gets the same scale resolution
+/// a full block would (padding zeros must never widen the range). Symmetric
+/// mode folds `abs_max` over the full padded array unconditionally: zero
+/// padding can never exceed a non-empty real element's absolute value, so the
+/// result is bit-identical to folding over the real elements alone, and doing
+/// it this way keeps the symmetric path source-identical to the pre-fix
+/// version (see `quantize_f64_to_q4_symmetric_partial_block_is_bit_identical_to_padded_block`).
 ///
 /// # Errors
 ///
-/// Returns [`InferenceError::InvalidInput`] if any element of `vals` is
-/// non-finite. IEEE-754 `NaN > x` is always false; a NaN silently leaves
-/// `abs_max`, `min_val`, or `max_val` unchanged, yielding wrong-but-no-error
-/// quantization. Rejecting here means the error points at the source weight
-/// rather than a downstream matmul.
+/// Returns [`InferenceError::InvalidInput`] if `valid_len` is not in `1..=32`,
+/// or if any element of `vals` is non-finite. IEEE-754 `NaN > x` is always
+/// false; a NaN silently leaves `abs_max`, `min_val`, or `max_val` unchanged,
+/// yielding wrong-but-no-error quantization. Rejecting here means the error
+/// points at the source weight rather than a downstream matmul.
 #[inline]
-pub(crate) fn quantize_block_with_mode(
+fn quantize_block_with_mode_len(
     vals: &[f32; 32],
+    valid_len: usize,
     symmetric: bool,
 ) -> Result<Q4Block, InferenceError> {
+    if !(1..=32).contains(&valid_len) {
+        return Err(InferenceError::InvalidInput(format!(
+            "Q4 weight block valid_len {valid_len} must be in 1..=32"
+        )));
+    }
+
     for (i, &v) in vals.iter().enumerate() {
         if !v.is_finite() {
             return Err(InferenceError::InvalidInput(format!(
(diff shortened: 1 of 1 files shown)


ohdearquant:

Review follow-up, addressed in fdf70d4ce: the near-usize::MAX f16 test claimed to pin the file-length bound, but `(usize::MAX - 3) * 2` overflows, so it actually exercised the overflow branch and its assertion accepted that error too — the f16 file-length bound was unpinned. The test now uses `numel = usize::MAX / 4` (byte count ~2^62, no overflow) and asserts only on the file-length-bound error message. Mutation evidence: with the bound deleted, the loader attempts a 9.2 EB allocation and aborts (SIGABRT), so the test now fails hard under that mutation. Full q4_weights suite, clippy `-D warnings`, and fmt all green after the fix.

ohdearquant:

`make bench-compare BASE=9997e54bb` (quick, fork point of this branch): bench-compare showed no change — criterion reports p > 0.05 on all groups (`elementwise_cpu_bench` + `simd`).

Context for the point estimates: the changed function is `quantize_block_with_mode_len`, a load-time Q4 block-quantization path that no covered bench group exercises (verified by grep over the bench sources); the decode/prefill/embed hot paths are untouched. Point estimates drifted mixed-sign on unrelated kernels due to concurrent workloads on the bench machine, and criterion attributes all of it to noise (largest: rms_norm/4096 at p = 0.10). The symmetric quantization path is additionally pinned bit-identical to the pre-fix output by `quantize_f64_to_q4_symmetric_partial_block_is_bit_identical_to_padded_block`.

Issue #337, opened 2026-06-25
PagedKVCache: EvictionPolicy::Lru is broken for sequences exceeding max_tokens (latent — not production-wired)


ohdearquant:
> _Issue filed by Claude (Anthropic agent) on behalf of @ohdearquant._

## Summary

`PagedKVCache` with `EvictionPolicy::Lru` cannot correctly serve a sequence longer than `max_tokens = max_pages * page_size`. Appending the first token past capacity panics instead of sliding the window. This is **latent**: the only production constructor (`crates/inference/src/batch/worker.rs:545`) uses `EvictionPolicy::None`, so no shipping path hits it today. Filing so the landmine is documented before anyone enables LRU.

## Mechanism

`append_kv_layer` (`crates/inference/src/kv_cache/paged.rs:580`) captures `pos = table.seq_len()` once (line 586), then grows pages in a loop:

```rust
let needed_pages = (pos / page_size) + 1;
while self.table.num_pages() < needed_pages {
    let phys = self.alloc_page();   // may evict
    self.table.push_page(phys);
}
let (phys_page, offset) = self.table.resolve(pos);   // line 596
```

When the pool is exhausted, `alloc_page` → `evict_lru` (line 699) calls `table.pop_front_page()`, which **removes `entries[0]` and clamps `seq_len`** (lines 215-219). With `max_pages < needed_pages` the loop can never reach `needed_pages` (each iteration evicts one page and pushes one — net `num_pages` unchanged), so it keeps evicting until `lru_order` is empty and `evict_lru`'s `.expect("LRU order empty but pool exhausted")` (line 706) panics.

Deterministic repro: `max_pages=2, page_size=4, num_layers=1, num_kv_heads=1, head_dim=1`, append+advance 8 tokens, then append a 9th → panic.

> Note: an automated discovery pass first attributed the crash to a `resolve` out-of-bounds at line 596; on trace-through the actual panic is the `expect` at line 706 (or, depending on `lru_order` contents, unbounded evict/push churn). Same root cause.

## Root cause

The addressing scheme is absolute: `resolve(pos)` computes `logical = pos / page_size`. Evicting the oldest page (`pop_front_page`, `entries.remove(0)`) shifts every remaining logical page down by one, so absolute positions no longer map to the right entry. Absolute-position paging and front-eviction are architecturally incompatible — a correct sliding window needs either a logical-page **base offset** that advances on eviction, or rebasing positions after each eviction.

[...]

PR #542, opened 2026-07-02
test(inference): regression coverage for PagedKVCache Lru overflow (#292)

### LINKED ISSUE #292: PagedKVCache LRU eviction: stale needed_pages + un-touched page → panic/corruption in append_kv_layer
## Summary

`PagedKVCache::append_kv_layer` computes `needed_pages` from the current position **once**, before the page-allocation loop, and only calls `touch_page` (which repopulates `lru_order`) **after** the loop. When the pool is exhausted and `EvictionPolicy::Lru` is active, the allocation loop calls `evict_lru` without ever re-touching a page, so a second loop iteration hits `lru_order.pop_front().expect("LRU order empty but pool exhausted")` and panics — or, depending on the `PagePool`/`push_page` free-list interaction, reuses a just-evicted physical page while the logical page table has shifted underneath the stale target position (silent corruption).

## Reachability

`EvictionPolicy::Lru` is **exported public API** (`crates/inference/src/kv_cache/mod.rs:14`) but has **no constructor anywhere in the crate or its tests** — only the match arm at `paged.rs:695` references it. Current Qwen generation and continuous batching use `PagePool`/`PageTable` directly, not `PagedKVCache::append_kv_layer` with `Lru`. So this does **not** affect any in-tree generation path; it bites an external consumer of the published `lattice-inference` crate who opts into `EvictionPolicy::Lru`.

**Severity: P2** (latent, exported-API, no internal caller). Not P0/P1 — no shipped lattice path reaches it.

## Repro (logic)

`PagedKVCacheConfig` with `eviction = EvictionPolicy::Lru`, `max_pages = 1`, `page_size = P`. Fill page 0 with `P` tokens (`append_kv_layer(...)` + `advance()` to `seq_len = P`). Append token at `pos = P`:

- `needed_pages = (P / P) + 1 = 2` (`paged.rs:590`)
- loop: `num_pages() = 1 < 2` → `alloc_page()` → pool exhausted → `evict_lru()` pops the only `lru_order` entry and frees the page (`paged.rs:699-709`); `push_page` re-adds it; `num_pages() = 1` still `< 2` → second `alloc_page()` → `evict_lru()` → `lru_order` now empty → `.expect(...)` panics (`paged.rs:706`).

## Cited sites

- `crates/inference/src/kv_cache/paged.rs:580` `append_kv_layer`
- `:586`-`:594` `needed_pages` computed before loop; `touch_page` only at `:599` (after)
- `:683` `alloc_page`, `:695` `EvictionPolicy::Lru => evict_lru()`
- `:699`-`:706` `evict_lru` `pop_front().expect(...)`

## Fix sketch

Recompute `needed_pages` / re-check capacity each loop iteration, and ensure a page pushed into the table is touched (recorded in `lru_order`) before the next allocation attempt — so eviction can't be asked to evict a page that isn't in the LRU order. Add a unit test driving the `max_pages = 1`, `Lru` overflow and asserting clean behavior (either a `Result::Err`/bounded stop, or correct page recycling) instead of a panic.

_Source: adversarial read-only code audit of the KV-cache family. Mechanism independently verified against source; LRU-policy reachability confirmed (no internal constructor)._

ohdearquant:

## Summary

Adds regression coverage for the `PagedKVCache` LRU-eviction failure described in #292.

While preparing the fix, I confirmed that the underlying library bug is **already fixed** in `main` (commit `50c3612a9`, "fail closed on PagedKVCache Lru overflow past max_tokens (#337)"). Rather than fabricate a redundant code change, this PR is **test-only**: it locks in the fixed behavior with a mutation-sensitive regression test so the #292 scenario cannot silently regress.

## What was broken (#292)

`PagedKVCache::append_kv_layer` computes `needed_pages` **once** before the page-allocation loop:

```rust
let needed_pages = (pos / page_size) + 1;
while self.table.num_pages() < needed_pages {
    let phys = self.alloc_page(); // may evict_lru when the pool is exhausted
    self.table.push_page(phys);
}
```

Under `EvictionPolicy::Lru`, once a sequence grows past `max_tokens` (so `needed_pages > max_pages`) and the page pool is exhausted, each loop iteration evicts one page and re-pushes one — **net-zero on `num_pages()`**. The loop therefore cannot converge: it drains `lru_order` (evicting pages that are still live), and then either

- **panics** with `"LRU order empty but pool exhausted"`, or
- **silently corrupts** state — `pop_front_page` clamps `seq_len` mid-loop, shifting the logical page table under live data.

Because `EvictionPolicy::Lru` is public API with no in-tree constructor, this only affected external consumers.

## The fix (already landed, #337)

`append_kv_layer` now **fails closed** before the loop, for `EvictionPolicy::Lru` only:

```rust
if self.config.eviction == EvictionPolicy::Lru {
    assert!(
        needed_pages <= self.config.max_pages,
        "PagedKVCache::append_kv_layer: position {pos} needs {needed_pages} pages but \
         max_pages is {}; EvictionPolicy::Lru does not yet support sequences beyond \
         max_tokens ({}) (see issue #337)",
        self.config.max_pages,
        self.config.max_tokens(),
    );
}
```

This rejects the non-convergent case up front with a clear, actionable message, so the runaway loop is never entered. `alloc_page` (which calls `evict_lru`) is only reachable from this guarded loop, so the guard covers the sole public eviction path. The `EvictionPolicy::None` path is untouched. (Sliding a window past capacity remains unimplemented and is tracked separately in #337.)

## This PR: regression coverage

Adds `paged_lru_overflow_leaves_live_pages_unchanged`, which fills the cache to capacity under `EvictionPolicy::Lru`, then attempts an overflowing append and asserts:

1. the overflow **fails closed** (panics, caught via `catch_unwind`);
2. `seq_len()` and `num_pages()` are **unchanged**; and
3. every previously stored live page is **byte-for-byte intact** (`gather_k`/`gather_v`) — proving no page-table shift or data corruption.

This complements the existing repro test `paged_lru_eviction_fails_closed_beyond_max_tokens`.

## Mutation evidence

Reverse-applying the #337 guard (removing only the library `assert!` block and its doc note, leaving both tests untouched) and re-running:

```
thread '...paged_lru_eviction_fails_closed_beyond_max_tokens' panicked at paged.rs:909:
LRU order empty but pool exhausted
  -> should_panic(expected = "does not yet support sequences beyond max_tokens") does NOT match -> FAIL

thread '...paged_lru_overflow_leaves_live_pages_unchanged' panicked at paged.rs:909:
LRU order empty but pool exhausted
  -> then corruption assertion fires: seq_len left: 4, right: 8 -> FAIL

test result: FAILED. 0 passed; 2 failed
```

Both tests fail without the guard — the new test surfaces the exact `seq_len` corruption from #292 — and both pass with it. Mutation-sensitivity confirmed.

## Gates run (CPU-only)

- `cargo test -p lattice-inference --lib` → **1356 passed, 0 failed, 7 ignored**
- `cargo clippy --workspace --all-targets -- -D warnings` → clean
- `cargo fmt --check` → clean

## Notes

- Test-only change; no library code, no API, no signatures modified.
- Does not close #292 — left open for maintainer review.

### DIFF
--- a/crates/inference/src/kv_cache/paged.rs
+++ b/crates/inference/src/kv_cache/paged.rs
@@ -1387,6 +1387,48 @@ mod tests {
     }
 
     #[test]
+    fn paged_lru_overflow_leaves_live_pages_unchanged() {
+        let mut config = make_config(2); // 2 pages * page_size 4 = 8 max_tokens.
+        config.eviction = EvictionPolicy::Lru;
+        let kv_dim = config.kv_dim();
+        let max_tokens = config.max_tokens();
+        let mut cache = PagedKVCache::new(config);
+
+        for step in 0..max_tokens {
+            for layer in 0..2 {
+                let k = vec![step as f32; kv_dim];
+                let v = vec![1000.0 + step as f32; kv_dim];
+                cache.append_kv_layer(layer, &k, &v);
+            }
+            cache.advance();
+        }
+
+        let overflow = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
+            for layer in 0..2 {
+                let k = vec![999.0; kv_dim];
+                let v = vec![1999.0; kv_dim];
+                cache.append_kv_layer(layer, &k, &v);
+            }
+        }));
+        assert!(overflow.is_err(), "LRU overflow should fail closed");
+
+        assert_eq!(cache.seq_len(), max_tokens);
+        assert_eq!(cache.num_pages(), 2);
+
+        let mut k_buf = vec![0.0f32; max_tokens * kv_dim];
+        let mut v_buf = vec![0.0f32; max_tokens * kv_dim];
+        cache.gather_k(0, &mut k_buf);
+        cache.gather_v(0, &mut v_buf);
+
+        for step in 0..max_tokens {
+            for i in 0..kv_dim {
+                assert_eq!(k_buf[step * kv_dim + i], step as f32);
+                assert_eq!(v_buf[step * kv_dim + i], 1000.0 + step as f32);
+            }
+        }
+    }
+
+    #[test]
     #[should_panic(expected = "page_size must be non-zero")]
     fn paged_zero_page_size_panics_at_construction() {
         // Regression for #244: page_size == 0 is the divisor in PageTable::resolve,


ohdearquant:

`make bench-compare BASE=9997e54bb` (quick, fork point): zero regressions detected across 259 benches. 249/259 groups report no change (p > 0.05); the 10 flagged at borderline p = 0.05 are all *negative* deltas (apparent improvements, largest −20%), which cannot be real effects of this PR — the entire diff is inside `#[cfg(test)]` (a regression test for PagedKVCache LRU overflow), so the benchmark binaries compile from identical non-test source on both legs. The favorable drift is bench-machine noise; the relevant reading is: no regression on any group.

PR #543, opened 2026-07-02
fix(inference): thread seed into mtp_verify_draft for reproducible MTP-verify (#329)

### LINKED ISSUE #329: harden(inference): speculative.rs dormant MTP-verify F1 — latent non-deterministic seed in mtp_verify_draft (F2/F3 resolved by #362)

A fresh discovery pass over `crates/inference/src/speculative.rs` (3606 LOC) surfaced 3 real-but-**latent** robustness gaps, all confined to the **dormant CPU `MtpVerifier` / trait-level `mtp_verify_draft` path**. None are reachable from the live decode path, so this is tracked-not-urgent. Filing so the gaps are fixed *before* probabilistic MTP gets wired to production.

## Context: the live MTP path is greedy-only and already deterministic

The live Metal MTP loop is `generate_greedy_mtp` (`forward/metal_qwen35.rs:8698`, `LATTICE_MTP=1`). It calls `speculative::rejection_sample_draft(..., greedy=true, ...)` directly (`:8808`), **not** via `mtp_verify_draft`. Per the in-code comment (`:8798-8800`), #237 already replaced a `greedy=false` clock-seeded RNG here with the deterministic argmax path. So the findings below live in the **unwired** trait-level path (`mtp_verify_draft` has no in-tree production caller; CPU `MtpVerifier::forward_one` is reached only from tests and from `draft_tokens_with_logits` ← `mtp_verify_draft`).

## F1 — Latent: `mtp_verify_draft` hardcodes a non-deterministic seed

`crates/inference/src/speculative.rs:1284-1291` — `mtp_verify_draft` calls `rejection_sample_draft(..., greedy=false, seed=None)`. With `greedy=false` the probabilistic path runs and `seed=None` → `SpecRng::from_clock()` (`:1554`), so output is non-reproducible. This is the *same* class of bug #237 fixed in the live path, but in the dormant trait-level entry. **Severity: low/latent** — no caller threads a user seed today; the live path is greedy. **Fix when wiring probabilistic MTP:** add a `seed: Option<u64>` parameter to `mtp_verify_draft` and thread `GenerationConfig.seed` (`model/qwen35_config.rs:554`) through it, mirroring the single-token `Sampler::with_seed` contract. (Public signature change — not a drop-in self-merge.)

## F2 — Latent: div-by-zero panic on malformed head geometry

`crates/inference/src/speculative.rs:~607` and `~800` — `groups = num_q_heads / num_kv_heads` panics if `num_key_value_heads == 0`, and `qh / groups` panics if `groups == 0` (`num_q_heads < num_kv_heads`). `MtpConfig` (`:125`) is a `pub`-fields struct; `MtpVerifier::load_from_source` (`:250`) validates tensor **shapes** but never the config's internal consistency. **Severity: low/latent** — no real model config has `kv_heads=0`; loud panic, not silent corruption; only reachable via the dormant CPU path with a hand-malformed config.

## F3 — Latent: OOB panic in partial RoPE when `partial_rotary_factor > 1`

`crates/inference/src/speculative.rs:~547` and `mtp_apply_partial_rope` (`:1123`) — `rope_dim = (head_dim as f32 * partial_rotary_factor) as usize` is unclamped; if `partial_rotary_factor > 1.0` then `rope_dim > head_dim`, and `mtp_apply_partial_rope` indexes `head_vec[half + i]` (`half = rope_dim/2`) out of bounds. **Severity: low/latent** — real configs use `partial_rotary_factor ∈ (0, 1]` (Qwen3.5 = 0.25); same dormant-path reachability as F2.

### Suggested consolidated fix for F2+F3

Add a cheap consistency check at the top of `MtpVerifier::load_from_source` (it already returns `Result<_, InferenceError>`, so no signature change): reject `num_key_value_heads == 0`, `num_attention_heads % num_key_value_heads != 0`, and `partial_rotary_factor` outside `(0, 1]` with `InferenceError::InvalidInput`, before any tensor load. Converts deep loud panics into a clear early error at the construction boundary.

## Investigated and dismissed (not a bug)

A fourth finding — "softmax/ratio path produces invalid probabilities for extreme logits (e.g. `[-1000.0; 4]`), biasing rejection" — is a **false positive**. `softmax_into` (`:1519`) does standard max-subtraction: `[-1000;4]` → max `-1000` → each `exp(0)=1` → sum `4` → uniform `0.25`, giving `accept_prob = min(1, 0.25/0.25) = 1.0` (accept). The claimed `sum=0`/NaN state does not occur for finite logits; only an all-`NEG_INFINITY` vector (every token impossible — a degenerate non-realistic input) yields NaN, and even then valid-input behavior is uncorrupted. No fix needed.

ohdearquant:

## What & why

Resolves the F1 scope of #329 (F2/F3 were resolved by #362 and are left untouched — verified against source).

The dormant trait-level MTP-verify path (`mtp_verify_draft`) called
`rejection_sample_draft(..., greedy=false, seed=None)`. A `None` seed falls
through to `SpecRng::from_clock()` (`speculative.rs`, `None => SpecRng::from_clock()`),
so the probabilistic rejection-sampling decisions — and therefore the accepted
tokens — are **non-reproducible**. This is the same bug class as #237, which was
fixed only in the live greedy path.

## Fix

Thread a resolved seed into the rejection sampler. The single behavioral change is
at the call boundary:

```rust
let rejection_seed = Some(seed.unwrap_or(MTP_VERIFY_DEFAULT_SEED));
```

`None` now resolves to a named, documented `const MTP_VERIFY_DEFAULT_SEED` — a
fixed deterministic default, explicitly **not** the clock — so the wrapper can
never reach `from_clock()`.

## API-evolution alternatives considered

This is a public-API change on a published crate (`lattice-inference` `0.4.2`,
pre-1.0), so the shape was chosen deliberately:

- **A — seeded variant + delegating shim (chosen):** add public
  `mtp_verify_draft_with_seed(..., seed: Option<u64>)` carrying the real
  implementation; keep existing `mtp_verify_draft` as a source-compatible shim
  delegating with `None`. Additive — no existing caller breaks — while giving real
  generation callers a path to thread `GenerateConfig.seed`.
- **B — break `mtp_verify_draft`'s signature outright** to take `seed`/`&GenerateConfig`:
  mechanically smallest, defensible pre-1.0, but gratuitous here because the bug is
  fixable source-compatibly; a `&GenerateConfig` variant would also over-couple
  `speculative.rs` to the Qwen3.5 generation config.
- **C — mirror the crate's established `Sampler::new` → `Sampler::with_seed`
  reproducibility convention:** its discipline (reproducibility as an explicit,
  documented operation) is folded into A via the "Seed contract" doc on the new
  variant, rather than shipped as a separate mechanism.

**Chosen: A, shaped by C.** Additive + source-compatible, with the deterministic
`None` contract documented rather than accidental.

## Mutation evidence (test is not vacuously true)

The determinism tests were confirmed mutation-sensitive by reinstating the bug
(`rejection_seed = None`), rebuilding, capturing the failure, then restoring:

- Reverted → **FAIL on first run** (no retries):
  - `mtp_verify_draft_with_seed_repeats_probabilistic_acceptance`:
    `same seed must reproduce the same fallback_token — left: Some(3), right: Some(7)`
  - `mtp_verify_draft_default_seed_is_deterministic`:
    `default seed=None must reproduce the same accepted_count — left: 2, right: 1`
- Restored → full `--lib` suite **1359 passed, 0 failed, 7 ignored**;
  `git diff --stat` back to the exact `396 insertions(+), 8 deletions(-)`.

No MUST-differ flakiness trap: `mtp_verify_draft_different_seed_is_allowed_to_vary`
asserts only `is_ok()`, never inequality. The shipped determinism tests use fixed
seeds, so they are deterministic and cannot flake.

## Gates (CPU-only)

A separate serialized GPU measurement was running on the host, so no Metal/GPU/
model-loading binary was executed. The MTP-verify path is dormant CPU code and the
tests use a pure in-memory `MockTargetVerifier` with hand-built tiny weights.

| Gate | Command | Result |
|---|---|---|
| Tests | `cargo test -p lattice-inference --lib` | 🟢 1359 passed, 0 failed, 7 ignored |
| Clippy | `cargo clippy --workspace --all-targets -- -D warnings` | 🟢 exit 0 |
| Fmt | `cargo fmt --check` | 🟢 clean |

Does not close #329 (F1 only); left open for human review.

### DIFF
--- a/crates/inference/src/speculative.rs
+++ b/crates/inference/src/speculative.rs
@@ -1302,9 +1302,9 @@ pub trait MtpTargetVerifier {
 /// [`SpecRng::from_clock`] the way the lower-level [`rejection_sample_draft`] does for
 /// its own `seed=None`. `MTP_VERIFY_DEFAULT_SEED` gives `mtp_verify_draft`'s probabilistic
 /// rejection sampling a fixed, deterministic-by-design default: identical inputs always
-/// produce identical accepted tokens, matching the same reproducibility contract
-/// [`crate::sampling::Sampler::with_seed`] documents for greedy generation. This value is
-/// arbitrary (mirrors [`SpecRng::new`]'s zero-seed fallback) and carries no other meaning.
+/// produce identical accepted tokens, matching the same reproducible-sampling contract
+/// [`crate::sampling::Sampler::with_seed`] documents. This value is arbitrary (mirrors
+/// [`SpecRng::new`]'s zero-seed fallback) and carries no other meaning.
 const MTP_VERIFY_DEFAULT_SEED: u64 = 0x853c_49e6_748f_ea9b;
 
 /// Verify a speculative MTP draft against the target model, with an explicit seed for the
@@ -1318,7 +1318,9 @@ const MTP_VERIFY_DEFAULT_SEED: u64 = 0x853c_49e6_748f_ea9b;
 /// - `Some(seed)`: probabilistic rejection sampling draws from an RNG seeded with `seed`.
 ///   Identical inputs and the same `seed` reproduce identical accepted tokens and
 ///   fallback token across runs — thread the caller's `GenerateConfig.seed` here to match
-///   the rest of a reproducible generation run.
+///   the rest of a reproducible generation run. Note [`SpecRng::new`] maps `seed == 0` to
+///   its internal fallback constant, so `Some(0)` behaves identically to that fallback
+///   seed rather than as a distinct zero-seeded stream (reproducibility still holds).
 /// - `None`: uses [`MTP_VERIFY_DEFAULT_SEED`], a fixed deterministic default — **not**
 ///   [`SpecRng::from_clock`]. This is documented compatibility behavior for
 ///   [`mtp_verify_draft`], the reverse of the lower-level [`rejection_sample_draft`], whose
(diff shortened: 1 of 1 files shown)


ohdearquant:

Code review round 1 findings addressed in af068f9aa (both MIN, no CRIT/MAJ):

1. **Determinism probes hardened to 8 independent pairs.** With two drafted positions at accept probability 0.5 and rejection stopping at the first failed draw, a single run-pair matches a reverted clock-seeded path by coincidence with probability 0.5² + 0.25² + 0.25² = 0.375. Both MTP-level tests (`mtp_verify_draft_with_seed_repeats_probabilistic_acceptance`, `mtp_verify_draft_default_seed_is_deterministic`) now compare 8 independently rebuilt pairs (the seeded variant across 8 distinct seeds), pushing a revert's survival probability to ~4×10⁻⁴. Mutation-verified: with the seed threading reverted, 6/6 first-run failures across both tests; restored, full `--lib` suite green (1359 passed).
2. **Seed-contract docs corrected.** The `Some(seed)` contract now notes that `SpecRng::new` maps `seed == 0` to its internal fallback constant (so `Some(0)` is reproducible but not a distinct zero stream), and the `MTP_VERIFY_DEFAULT_SEED` doc no longer describes `Sampler::with_seed` as a greedy-generation contract (it documents reproducible sampling).

Gates on af068f9aa: `cargo test -p lattice-inference --lib` 1359 passed / 0 failed, `cargo clippy --workspace --all-targets -- -D warnings` clean, `cargo fmt --check` clean.

ohdearquant:

`make bench-compare BASE=9997e54bb` (quick, fork point): zero regressions across 259 benches — 254/259 no change (p > 0.05), the 5 flagged at borderline p = 0.05 are all negative deltas (apparent improvements, machine noise). The changed code (`mtp_verify_draft` seed threading in `speculative.rs`) is on the MTP speculative-verify path, which no covered bench group exercises; the change replaces a clock-seeded RNG construction with a seeded one at identical cost.

PR #544, opened 2026-07-02
fix(inference): fail-closed integrity guard for Qwen embedding cache (#504 first slice)

### LINKED ISSUE #504: harden(serve): state/adapter/cache artifact integrity — signatures, compatibility hashes, provenance taint, quarantine
# harden(serve): state/adapter/cache artifact integrity — signatures, compatibility hashes, provenance taint, quarantine

As persistent-state features mature (state snapshots, cross-turn caching), this engine accumulates a
growing surface of loadable artifacts (GDN state snapshots, LoRA adapters, prompt-cache entries). Only
LoRA adapters currently get an integrity/provenance check before load.

**Threat model:** stale state loaded against the wrong model/tokenizer/adapter; a malicious state
artifact injected into a session; prompt-injection content cached as trusted recurrent state; an
adapter trained on poisoned input; a rollback that contaminates future state.

## 1. User-visible outcome
Prevents a corrupted, mismatched, or maliciously-crafted persisted state/adapter/cache artifact from
silently producing wrong or unsafe output — a prerequisite for shipping any persistent-state feature as
something other than an internal debug tool.

## 2. Prior art
Original threat-model synthesis; no external citation required. The pattern — treat loadable model/
session artifacts as untrusted input, verify provenance and compatibility before use — is standard
practice for any system deserializing externally-sourced state. A working precedent already exists
in-repo: the LoRA adapter manifest (`crates/tune/src/lora/manifest.rs`) carries `integrity_sha256`,
`base_model_rev`, `tokenizer_rev`, and an `AdapterStatus::Quarantined` state, enforced at load time in
`crates/tune/src/lora/loader.rs`. This issue extends the same discipline to GDN state snapshots and
prompt-cache artifacts, not re-inventing it for adapters.

## 3. Cheapest falsifying experiment
Build the minimum viable guard: a `StateArtifact` wrapper carrying (a) a content hash/signature, (b) a
model-id + tokenizer-rev + adapter-id compatibility tuple checked at load time, (c) a provenance/taint
flag set (`untrusted_email`, `tool_output`, `web_content`, `trusted_session`). Test: construct a
deliberately mismatched artifact (wrong model rev, tampered bytes, taint flag stripped) and confirm
load is rejected, not silently accepted.

## 4. Promote threshold
The hashing + compatibility-check + provenance-flag layer reduces attack success on a constructed
mismatched/tampered-artifact test suite to near-zero, while retaining >=70% of the snapshot/cache
speedup on benign workloads (a compatibility check is O(1), should not be the bottleneck).

## 5. Kill threshold
None for hashing/compatibility-checks/provenance flags — baseline infrastructure, mandatory before any
state/adapter/cache artifact ships as a loadable object outside a debug build (per #486's future
scope). The open-ended pieces — quarantine-on-drift heuristics, state-reset-after-high-risk-input
policy — remain killable if they add complexity without measurable attack-surface reduction.

## 6. Where the code lives
A new integrity-check module alongside the state/adapter loading paths: `crates/inference` for state
snapshots, `crates/tune` for the LoRA manifest (extend existing enforcement to newer artifact types).

## 7. Default-on guard
Quarantine/auto-reset ships default-off (e.g. `LATTICE_STATE_QUARANTINE`) until false-positive rate is
measured on benign workloads. Hashing/compatibility-checking has no off-switch by design — a disable-
able hash check is not a security boundary.

## 8. Gate level
**G1** (hashing/compatibility-check/provenance-flag layer — reject-on-mismatch, no behavior change on
valid input). **G2** (quarantine-on-drift and state-reset-after-high-risk-input policies — behavior-
changing, ship behind a flag).

---
Cross-refs: #486 (snapshot/branch/rollback — should land before or alongside it), #481 (GDN state as
first-class object, parent epic), #444 (governed LoRA manifest — the precedent this extends).

ohdearquant:

## What

First slice of #504 (persisted-artifact integrity): a fail-closed integrity + compatibility guard on the **Qwen embedding cache** — the one on-disk artifact with a real save/load pair and *no* load-time checks today.

On save, the cache is now wrapped in a manifest header:

```
[magic:8][manifest_len:u32 LE][manifest JSON][payload]
```

where the manifest carries `version`, `payload_sha256`, `base_model_rev`, `tokenizer_rev`, `embedding_dim`, and `entry_count`, and the payload is the existing record stream `[hash:u64, dim:u32, floats:f32*dim]` per entry.

On load, `parse_embedding_cache_file` rejects (returns `Err`, **never** load-and-warn) on any of: magic mismatch, version mismatch, manifest-length-cap overflow, truncated header/manifest, `payload_sha256` mismatch, `base_model_rev`/`tokenizer_rev`/`embedding_dim` mismatch, per-record dimension mismatch, non-finite float, truncated record, or trailing bytes after the declared entry count. Every error message names the field that mismatched. `self.cache` is mutated only after the whole file validates. A missing file is the legitimate "no artifact yet" case (`Ok(0)`). File reads are size-capped by a bounded `File::open + take(max_len + 1)` core (mirroring the LoRA `read_lora_file_bounded` precedent), so a hostile or corrupt file cannot drive an unbounded allocation even if it grows between the stat and the read.

## Why

Loadable on-disk artifacts should fail closed on corruption or cross-model reuse. The LoRA adapter path already has this (`crates/tune/src/lora/manifest.rs` + `loader.rs`: `integrity_sha256`, `base_model_rev`, `tokenizer_rev`, `AdapterStatus::Quarantined`); the Qwen embedding cache deserialized untrusted bytes straight into live state with zero integrity/compat/finiteness checks. This slice closes that one gap by mirroring the established LoRA pattern rather than inventing a new framework (FindExisting > Create).

## Persisted-artifact inventory (surface scan of `crates/inference` + `crates/tune`)

| Artifact kind | Save fn file:line | Load fn file:line | Existing integrity/compat check? |
|---|---:|---:|---|
| LoRA manifest JSON | crates/tune/src/lora/manifest.rs:124 | crates/tune/src/lora/manifest.rs:94 | Y — bounded manifest size (manifest.rs:97); UTF-8 + serde parse (manifest.rs:118); no manifest self-hash. |
| LoRA adapter safetensors (governed path) | crates/tune/src/lora/safetensors.rs:510 | crates/tune/src/lora/loader.rs:56 | Y — status pre-scan (loader.rs:67); relative-path confinement (loader.rs:113); bounded read (loader.rs:191); SHA-256 (loader.rs:197); rank/alpha/modules/model/id checks (loader.rs:214). |
| LoRA adapter safetensors (direct sibling API) | crates/tune/src/lora/mod.rs:145 | crates/tune/src/lora/mod.rs:135 | Partial — bounded read (safetensors.rs:494); safetensors parse (safetensors.rs:210); finite values (safetensors.rs:155); shape/rank consistency (safetensors.rs:313); no SHA-256/base_model_rev/tokenizer_rev/status gate. |
| **Qwen embedding cache binary** | **crates/inference/src/model/qwen.rs:671** | **crates/inference/src/model/qwen.rs:621** | **N (before this PR)** — record-boundary overflow checks only; no size cap, magic/version, integrity hash, model/tokenizer/dim compatibility, finite-value check, or trailing-byte rejection. **← this slice** |
| Tune registry filesystem weights.bin | crates/tune/src/registry/storage/backends.rs:274 | crates/tune/src/registry/storage/backends.rs:297 | Partial — model/path-traversal validation; raw backend load has no hash check; optional higher-level verified load (registry.rs:318). |
| Tune registry sqlite weights.bin | crates/tune/src/registry/storage/backends.rs:478 | crates/tune/src/registry/storage/backends.rs:537 | Partial — as filesystem backend; raw load unverified; optional verified load (registry.rs:318). |
| Tune registry safetensors bytes | crates/tune/src/registry/safetensors_io.rs:49 | crates/tune/src/registry/safetensors_io.rs:83 | Partial — parse + dtype + byte-alignment checks; no finite-value check, shape compatibility, or on-disk hash. |
| Tune registry multi-tensor safetensors bytes | crates/tune/src/registry/safetensors_io.rs:126 | crates/tune/src/registry/safetensors_io.rs:161 | Partial — parse + F32 filter + byte/shape element check; no finite-value check, shape compatibility, or on-disk hash. |
| Inference Q4 tensor file | crates/inference/src/weights/q4_weights.rs:626 | crates/inference/src/weights/q4_weights.rs:744 | Partial — magic/version + shape/payload allocation bounds; no content hash or model/tokenizer compatibility. |
| Inference F16 tensor file from quantizer | crates/inference/src/bin/quantize_q4.rs:575 | crates/inference/src/weights/q4_weights.rs:831 | Partial — magic/version + allocation bounds; no shape-product-vs-numel check, finite-value check, content hash, or compatibility. |
| Inference Q4 tensor file from quantizer | crates/inference/src/bin/quantize_q4.rs:504 | crates/inference/src/weights/q4_weights.rs:744 | Partial — same loader as Q4 tensor file; quantizer rejects non-finite source at convert time; no load-time hash/compatibility. |
| Inference merged_qkvz Q4 cache | crates/inference/src/forward/metal_qwen35.rs:3528 | crates/inference/src/forward/metal_qwen35.rs:14979 | Partial — temp-file+rename write, size-encoded filename, reuse size check, mmap Q4-header parse; no content hash or compatibility. |
| Quantization index JSON | crates/inference/src/bin/quantize_q4.rs:620 | crates/inference/src/forward/metal_qwen35.rs:14515 | N — absent/malformed/missing key treated as `None`; no integrity or compatibility enforcement. |

### Not persisted (in-memory only on current main — forward-looking for #504)

| Surface | Location | Detail |
|---|---:|---|
| GDN recurrent state snapshot | crates/inference/src/attention/gdn.rs:102 | In-memory clone/restore; no disk save/load in scope. |
| Cross-turn KV/GDN prefix cache | crates/inference/src/kv_cache/cross_turn.rs:38 | In-memory entry; no disk save/load in scope. |
| Paged KV prefix cache | crates/inference/src/kv_cache/prefix.rs:171 | In-memory map + shape validation; no disk save/load in scope. |
| Serve session state | crates/inference/src/bin/lattice_serve.rs:109 | In-memory per-job Metal state reset; no request/session disk save/load in scope. |
| Tune training checkpoint | crates/tune/src/train/loop/checkpoint.rs:87 | Serializable struct + docs only; no implemented file writer/reader in scope. |

**Verdict:** `REAL_UNGUARDED_ON_DISK_LOAD_EXISTS` — the Qwen embedding cache is a current, real, unguarded save/load pair (not speculative). LoRA governed path is already guarded; GDN/KV/serve/checkpoint are forward-looking for disk persistence.

## Scope — explicitly out of this slice (follow-up slices)

- **Signatures / quarantine machinery / provenance-taint** — deferred; this slice is content hash + compatibility fields + fail-closed reject only.
- **Legacy migration** — legacy raw record-stream files (no magic) are *rejected*, not migrated.
- Extending the guard to the other **partially-guarded** loaders above (registry raw weight loads, Q4/F16 tensor files, merged_qkvz cache, quantization index JSON) — each is its own slice.
- **Forward-looking persistence** (GDN snapshots, KV/prefix caches, serve session state, tune checkpoints) — no on-disk load path on main yet; guard lands when persistence does.

**Sibling-invocation-path check:** `cache_load`/`cache_save` are the only deserializer entry points for this artifact. Two callers, both routing through the now-guarded public API: the bench example (`crates/inference/examples/bench_profile.rs:93/100`) and the embed service (`crates/embed/src/service/native.rs` — which previously swallowed load errors silently and now logs a warning naming the path and error while remaining non-fatal, since the next `cache_save` regenerates the file). No independent deserializer bypasses the guard. LoRA loaders are a separate artifact family, already guarded. (An earlier revision of this description claimed the bench example was the sole caller; review caught the cross-crate embed-service caller.)

Each guard was verified load-bearing by reverse-applying it (via edit, not `git checkout`), touching the file to force a rebuild, and confirming the corresponding test fails, then restoring:

- Corrupt one payload byte → `payload_sha256` mismatch → load fails.
- Wrong `base_model_rev` / `tokenizer_rev` / `embedding_dim` → compatibility mismatch → load fails.
- Wrong magic / version → load fails.
- Truncated record / trailing bytes after declared `entry_count` → load fails.
- Non-finite float → load fails.
- **Bounded read:** neutralising the size cap (stat check + `take` bound both disabled) makes the oversized-file test fail; restoring the bounded `take(max_len + 1)` core makes it pass. This directly demonstrates the read cap closes the stat-to-read window, not just the stat fast-path.
- Intact file with correct compatibility fields → load succeeds.
- All reverts confirmed clean (`git diff` empty) after the protocol.

## Post-review hardening (third commit)

A review round surfaced five findings, all addressed in the third commit:

1. **Compat check was vacuous by default** — `base_model_rev`/`tokenizer_rev` defaulted to `"none"` with nothing populating them, so every real deployment compared `"none" == "none"`. `ModelInferenceConfig::load` now derives them as `sha256:<first 16 hex>` of the model dir's `config.json`/`tokenizer.json` when left at the sentinel; explicit values in `inference_config.json` still win; a missing source file degrades gracefully to `"none"`. A cache saved against one model/tokenizer revision now genuinely fails closed against another.
2. **Non-atomic save** — `cache_save` wrote with `std::fs::write`, so a crash mid-write could leave a truncated file that the fail-closed loader rejects forever. Now writes to a same-directory temp file and renames it over the target (mirroring the existing merged-weights cache pattern).
3. **Unchecked size-cap arithmetic** — `max_embedding_cache_file_len` and the bounded-read `take()` now use saturating arithmetic so absurd dimension values cannot wrap the cap into an under-rejecting small value.
4. **Embed-service caller swallowed load errors** — now logs a warning (path + error) on integrity failure, still non-fatal.
5. **Test gaps** — 10 new tests: six parser reject paths (version mismatch, truncated manifest, manifest cap, entry-count cap, per-record dim mismatch, truncated record), the public `cache_load` `Ok(0)` missing-file contract, the rev-derivation behavior (mutation-verified), a save/load round trip, and a no-leftover-temp-file check (mutation-verified).

## Post-review hardening, round 2 (fourth commit)

A second review round confirmed all five round-1 fixes and surfaced two residual findings, both addressed in the fourth commit:

1. **`base_model_rev` was not a model revision** — deriving it from `config.json` alone meant two checkpoints sharing config/tokenizer but differing in weights (e.g. an upstream re-upload that fixes weights with unchanged config) derived the same rev, so a stale cache still passed. `derive_base_model_rev` now folds in a **boundary-sampled weights fingerprint**: per weight shard (resolved via `model.safetensors.index.json` or the single `model.safetensors`), the filename, byte length, and first/last 1 MiB of content (shards ≤ 2 MiB hashed whole). This is documented precisely as a revision-drift fingerprint, **not** an adversarial-collision-resistant content hash — the integrity boundary for the cache payload remains `payload_sha256`, which hashes every payload byte. A mid-read failure degrades to the config-only derivation; missing `config.json` still leaves the sentinel.
2. **Temp file name was PID-only** — two threads in one process (e.g. two embed-service instances sharing the global cache path) saving concurrently could collide on the same `<name>.tmp.<pid>` and overwrite/rename over each other's temp file. Temp names now carry a process-global atomic sequence (`<name>.tmp.<pid>.<seq>`) and are opened with `File::create_new` (bounded 16-attempt retry, fail closed), so every save writes a call-unique file. An 8-thread concurrent-save test asserts the final file is always exactly one writer's intact content (noted in-source as a thread-interleaving-dependent guard rather than a deterministic one).

The same PID-only temp pattern exists in the merged-weights cache writer (`metal_qwen35::write_merged_qkvz`); that path is single-writer (model load) and a different artifact family, so it is left to its own slice rather than expanding this diff.

A third review round caught one residual case in the fingerprint (fifth commit): the resolver preferred the sharded index while `from_directory` loads the single `model.safetensors` first when both exist, so a mixed-layout directory could fingerprint files the loader never reads. The resolver now mirrors loader precedence exactly, with a mixed-layout regression test asserting the rev tracks the loaded file (flipping `model.safetensors` changes it; flipping the unused index shard does not).

## Gate outputs (CPU-only; `metal-gpu`/`wgpu-gpu` never enabled)

- `cargo test -p lattice-inference --lib` → **1386 passed, 0 failed, 7 ignored**.
- `cargo test -p lattice-embed --lib` → **254 passed, 0 failed**.
- `cargo clippy --workspace --all-targets -- -D warnings` → **clean (exit 0)**.
- `cargo fmt --check` → **clean (exit 0)**.
- `cargo check --workspace` → passes.

No `unwrap()`/`expect()` in library code (fixed-width slice conversions use `copy_from_slice`); no stubs; existing comments preserved. Commits reference #504 and stage only the changed files.

Does **not** close #504 (this is the first slice of a multi-slice effort). Not merged — opened for human review.


ohdearquant:

Review round addressed in 37a3b3af7 (five findings, two of them significant):

- **Vacuous compatibility check**: `base_model_rev`/`tokenizer_rev` defaulted to `"none"` and nothing populated them, so the manifest compat check compared `"none" == "none"` in every real deployment. `ModelInferenceConfig::load` now derives both as `sha256:<first 16 hex>` of the model dir's `config.json`/`tokenizer.json` when left at the sentinel (explicit config values still win; missing files degrade gracefully). Mutation-verified: reverting the derivation fails the new test.
- **Non-atomic save**: `cache_save` now writes to a same-directory temp file and renames over the target, so a crash mid-write can no longer leave a truncated file that the fail-closed loader would reject forever. Mutation-verified via the no-leftover-temp and round-trip tests.
- **Saturating size-cap arithmetic** in `max_embedding_cache_file_len` and the bounded-read `take()`.
- **Embed-service caller** (`crates/embed/src/service/native.rs`) no longer silently swallows `cache_load` errors — it warns with path + error, still non-fatal (next save regenerates). The PR body's earlier "sole caller" claim was wrong and has been corrected; this cross-crate caller was the miss.
- **Ten new tests** covering the six previously untested parser reject paths, the public missing-file `Ok(0)` contract, rev derivation, save/load round trip, and temp-file hygiene.

Gates on the new head: `cargo test -p lattice-inference --lib` 1376/0, `cargo test -p lattice-embed --lib` 254/0, `clippy --workspace --all-targets -D warnings` clean, `fmt --check` clean.

ohdearquant:

Second review round confirmed the round-1 fixes and surfaced two residual findings; both are addressed in e417f86cf:

1. **Weight-drift-blind `base_model_rev`** — the derived rev hashed `config.json` only, so checkpoints differing only in weights derived identical revs and a stale cache still passed compat. The derivation now folds in a boundary-sampled weights fingerprint (per-shard filename + length + first/last 1 MiB, resolved via the safetensors index or the single `model.safetensors`). Documented explicitly as a revision-drift fingerprint, not an adversarial content hash — `payload_sha256` remains the payload integrity boundary. Mutation-verified: flipping one weight byte (head, tail, either shard) or changing length changes the rev, and an end-to-end test shows `cache_load` then rejects naming `base_model_rev`.
2. **PID-only temp name** — concurrent same-process saves to the shared cache path could collide on `<name>.tmp.<pid>`. Temp names now carry a process-global atomic sequence and open with `File::create_new` (bounded retry, fail closed). An 8-thread concurrent-save test asserts the final file is always exactly one writer's intact content; under the reverted PID-only naming it failed 8/8 runs locally (noted in-source as an interleaving-dependent guard).

The same PID-only pattern in the merged-weights cache writer is single-writer and a different artifact family; left to its own slice.

Gates on e417f86cf: `lattice-inference` 1385/0 (7 ignored), `lattice-embed` 254/0, clippy `-D warnings` clean, fmt clean.

ohdearquant:

Third review round narrowed to one residual case, fixed in f5a4cb828: the fingerprint resolver preferred the sharded index while `from_directory` loads the single `model.safetensors` first when both files exist, so a mixed-layout directory could fingerprint files the loader never reads (and the resolver doc overstated its precedence claim). The resolver now mirrors loader precedence exactly, the doc is corrected, and a mixed-layout regression test asserts the rev tracks the loaded file: flipping a byte in `model.safetensors` changes it, flipping the unused index shard does not (that second assertion failed under the reverted index-first order, confirming it is load-bearing).

Gates on f5a4cb828: `lattice-inference` 1386/0 (7 ignored), `lattice-embed` 254/0, clippy `-D warnings` clean, fmt clean.

ohdearquant:

`make bench-compare BASE=9997e54bb` (quick, fork point): zero regressions across 259 benches — 253/259 no change (p > 0.05), the 6 flagged at borderline p = 0.05 are all negative deltas (apparent improvements, machine noise). The changed code is the embedding-cache load/save path and weight-file fingerprinting in `qwen.rs` (cold-start/model-load surfaces) plus a load-time warn in `lattice-embed`; no covered bench group exercises these, and the embed/inference compute kernels are untouched.

PR #549, opened 2026-07-02
fix: artifact-integrity hardening, #504 remaining slices (merged-Q4 cache, quantize_index.json, registry raw-load)

### LINKED ISSUE #504: harden(serve): state/adapter/cache artifact integrity — signatures, compatibility hashes, provenance taint, quarantine

ohdearquant:

Closes #504.

Completes the three remaining slices identified during the slice-1 (#544) review. Same fail-closed discipline slice 1 established: bounded reads, content hashing, reject-don't-warn, no behavior change on valid artifacts.

## Slice 1 — merged-Q4 cache content integrity

`MetalQwen35State::from_q4_dir`'s merge-on-first-load cache (`merged_qkvz_*.q4`) was validated by file size only, so a same-size corrupted or stale artifact loaded silently. `merged_qkvz_cache_is_valid` (relocated to `weights/q4_weights.rs`, gated `#[cfg(any(test, feature = "metal-gpu"))]` so the CPU suite exercises the real logic) now also compares a SHA-256 fingerprint of the merged file's payload against a fingerprint freshly derived from the current qkv/z source files via bounded reads. `merged_qkvz_expected_size` and both `write_merged_qkvz` payload-length computations are hardened against truncation underflow (`checked_sub`, descriptive error).

## Slice 2 — `quantize_index.json` fail-closed integrity

Sole runtime consumer (`read_quarot_seed_from_index`, feeding QuaRot rotation reconstruction) previously collapsed absent/malformed/no-key into one silent `None`. A corrupted index for an actually-rotated checkpoint would silently apply the wrong rotation and corrupt every downstream logit. Now: genuinely absent → `Ok(None)` (unchanged legacy fallback); present but oversized (16 MiB cap, stat-then-take bounded read), unreadable, non-JSON, or schema-mismatched → hard `Err` propagated by `from_q4_dir`. Typed `serde_json::from_slice::<QuantizeIndex>` against the producer-side schema in `quant/quarot/convert.rs` gives shape validation for free.

## Slice 3 — registry raw-load checksum bypass

`ModelRegistry::load_weights` performed zero integrity checking while `load_weights_verified` sat next to it — a silent bypass, since `register()` always records `weights_hash`. `load_weights` now verifies whenever a hash is on record (`TuneError::WeightIntegrityError` on mismatch); `load_weights_verified` is strictly stronger and additionally rejects hash-less records. The only unverified load left is a record explicitly registered without a hash.

## Tests

19 new tests across the three surfaces, all constructing deliberately tampered artifacts (same-size corrupted payload, same-size stale source, malformed/truncated/oversized/wrong-schema JSON, flipped-byte and truncated weight files, missing hash). Mutation-sensitivity verified per slice: guards reverse-applied → tests fail; restored → pass.

```
cargo test -p lattice-inference --lib          1415 passed; 0 failed
cargo test -p lattice-tune --lib               178 passed; 0 failed
cargo clippy (default + f16,metal-gpu) -D warnings   clean
```

## bench-compare

Not run: no hot-path change. All three guards run once at artifact-load time (O(1) or one linear hash of a file already being read); the load paths' steady-state inference code is untouched. Slice-1 fingerprinting adds a one-time hash of the qkv/z source files on cache-hit validation at model load.

### DIFF
--- a/crates/inference/src/forward/metal_qwen35.rs
+++ b/crates/inference/src/forward/metal_qwen35.rs
@@ -14418,19 +14418,6 @@ kernel void gdn_chunk_norm_silu_c32(
             }
         }
 
-        /// Read the QuaRot rotation seed from `quantize_index.json` per ADR-051.
-        ///
-        /// Returns `None` when the file is absent, malformed, or lacks a `quarot_seed`
-        /// key — all three are valid "no rotation" states for backwards compatibility
-        /// with older artifacts. The caller falls back to the legacy `config.json`
-        /// field on `None`.
-        fn read_quarot_seed_from_index(q4_dir: &std::path::Path) -> Option<u64> {
-            let path = q4_dir.join("quantize_index.json");
-            let bytes = std::fs::read(&path).ok()?;
-            let v: serde_json::Value = serde_json::from_slice(&bytes).ok()?;
-            v.get("quarot_seed").and_then(serde_json::Value::as_u64)
-        }
-
         /// Create a new [`MetalQwen35State`] by loading pre-quantized Q4/F16 files directly
         /// from a directory, without going through the full f32 weight representation.
         ///
@@ -15224,8 +15211,10 @@ kernel void gdn_chunk_norm_silu_c32(
                 // ADR-051 contract: `quarot_seed` lives in `quantize_index.json`. Fall back
                 // to the legacy `config.json` field (`quarot_rotation_seed`) for
                 // backwards compatibility with artifacts produced before the contract was
-                // formalized.
-                let index_seed = Self::read_quarot_seed_from_index(q4_dir);
+                // formalized. #504 remaining slice 2: a *present but malformed* index is a
+                // hard error (fail-closed) — only a genuinely absent index falls back.
+                let index_seed = crate::quant::quarot::convert::read_quarot_seed_from_index(q4_dir)
+                    .map_err(|e| format!("from_q4_dir: {e}"))?;
                 let seed_opt = index_seed.or(cfg.quarot_rotation_seed);
                 match seed_opt {
                     Some(seed) => Some(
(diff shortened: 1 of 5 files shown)


ohdearquant:

Review round-1 findings addressed in 3b874c4a0:

**MAJ (registry verification trusted caller-mutatable clone)** — `load_weights` and `load_weights_verified` now resolve the canonical record by `model.id` from the registry snapshot, verify against the canonical `weights_path`/`weights_hash`, and reject an argument that disagrees with the canonical record on either field. Blanking the clone's hash no longer re-opens the raw path, and an unregistered/deleted model is rejected with `ModelNotFound`. New regression tests cover the forged-hash clone, the hash-blanked clone, the unregistered model, and the honest-record-vs-tampered-disk case; the forged-clone test was proven mutation-sensitive (fails with the fix reverse-applied, passes restored).

**MED (`write_merged_qkvz` unbounded stat-then-read)** — the rebuild path now reuses `read_q4_payload_bounded` with the same `MAX_Q4_MERGE_PAYLOAD_LEN` cap as the validator, so a source file that grows after the size check cannot drive an unbounded allocation; the prior truncation guard is preserved inside the shared helper. New sparse-file test asserts an oversized source is rejected before any payload allocation and produces no merged artifact.

On the ADR-051 note: the object-without-`quarot_seed` `Ok(None)` behavior is intentional and unchanged — absent key still means unrotated per ADR-051; only present-but-malformed now fails closed.

ohdearquant:

Round-2 finding addressed in 1c59f752e (test-only): added `test_load_weights_rejects_mutated_clone_path`, which redirects a clone's `weights_path` at a byte-identical copy while leaving the hash canonical — so removal of the path predicate produces a wrongly-successful load the hash comparison cannot mask. Proven mutation-sensitive: fails with the `weights_path` half of the disagreement predicate removed, passes restored (64/64 registry tests green).

No implementation change in this commit, per the round-2 guidance to stop unless the follow-up diff changes the implementation.

PR #577, opened 2026-07-03
chore(workspace): align clippy msrv with rust-version 1.93 (#571)

### LINKED ISSUE #571: chore: align .clippy.toml msrv (1.85) with the declared rust-version (1.93)
Follow-up from #570 (the #568 fix).

The workspace now declares \`rust-version = "1.93"\` (verified end-to-end on 1.93.1 and gated by the new \`msrv\` CI job). \`.clippy.toml\` still pins \`msrv = "1.85.0"\`, which produces a benign but noisy driver warning on every clippy run:

```
warning: the MSRV in `clippy.toml` and `Cargo.toml` differ; using `1.85.0` from `clippy.toml`
```

The pin is deliberate for now: \`.clippy.toml\`'s msrv only sets the style-lint floor, and raising it to 1.93 activates \`clippy::manual_is_multiple_of\` at roughly 70 pre-existing sites across the workspace, which fail the \`-D warnings\` gate. Fixing those sites in #570 would have been scope creep on a user-facing build-break fix.

**Done means:**
- [ ] Convert the ~70 \`x % n == 0\` sites to \`x.is_multiple_of(n)\` (or selectively \`allow\` the lint with rationale if a site reads worse converted)
- [ ] Raise \`.clippy.toml\` msrv to 1.93 (or delete the file so clippy reads \`Cargo.toml\`'s \`rust-version\`)
- [ ] Remove the "intentionally stays at 1.85" caveat comment in the workspace \`Cargo.toml\`
- [ ] \`cargo clippy --workspace --all-targets --features f16,metal-gpu -- -D warnings\` clean, and the driver mismatch warning gone

Mechanical change, large-ish diff, zero runtime behavior. Good candidate for a quiet maintenance window.

ohdearquant:

Closes #571.

## What

Deletes `.clippy.toml` so clippy reads the MSRV from the workspace `rust-version = "1.93"` (single source of truth), removing the per-run driver warning `the MSRV in clippy.toml and Cargo.toml differ`. Also removes the now-stale "clippy msrv intentionally stays at 1.85" caveat from the workspace `Cargo.toml` comment.

Raising the effective clippy MSRV activates two lints previously suppressed by the 1.85 floor; both are closed here:

- **`clippy::manual_is_multiple_of`** — 36 sites converted (`inference` 28, `transport` 7, `tune` 1). Each site's divisor was checked for zero-safety before conversion (`%` panics on a zero divisor; `is_multiple_of(0)` returns `self == 0`). One deliberate exception: `crates/inference/src/vision/config.rs` keeps the `%` form under `#[allow]` with a rationale comment, because `spatial_merge_size` has no nonzero guard in `validate()` and the panic is load-bearing fail-closed validation.
- **`clippy::collapsible_if`** (let-chains; the issue didn't enumerate this one, but the `-D warnings` gate requires it) — 57 sites collapsed via `cargo clippy --fix` + `cargo fmt`. Workspace edition is 2024, so let-chains are well within the 1.93 floor.

## Gates

- `cargo clippy --workspace --all-targets --features f16,metal-gpu -- -D warnings` — clean, driver mismatch warning gone
- `cargo clippy --workspace --all-targets -- -D warnings` (default features) — clean
- `cargo fmt --check` — clean
- `cargo test --workspace --lib` — 1971 passed, 7 ignored, 0 failed

## bench-compare (quick mode, origin/main 7a1f4fb66 vs HEAD)

The A/B run flagged 5 FAILs — all in `lattice-embed` SIMD microbenchmarks (`simd_query_batch_dot_product`, `simd_normalize`, `int8_raw_dot_product`). These are false positives:

- The benched kernels live in `crates/embed/src/simd/{dot_product,normalize,quantized}.rs` — **files this PR does not touch**. The only embed change is one cosmetic let-chain collapse in `tier.rs` (1 hunk), outside the flagged hot paths.
- A targeted re-run of the flagged `rms_norm` group reported p = 0.10 (> 0.05, not significant).
- The machine was under concurrent load during the A/B run; this repo has a documented history of two-worktree quick-mode contention producing exactly this false-FAIL signature.

I'll re-run the flagged groups on an idle machine before merge and post the result as a PR comment. Zero runtime-behavior-changing code is touched by this PR — the diff is lint-driven syntax rewrites plus one comment block.


ohdearquant:
**Idle-machine bench re-run** (as promised in the PR body — round-1 numbers were collected under heavy concurrent load, box load 13–25; this run at load ~2.7).

All five rows that FAILed in round 1 collapsed to noise:

| Bench | Round 1 (loaded) | Idle re-run | 95% CI (idle) |
|---|---:|---:|---|
| `simd_query_batch_dot_product/simd_batch/768d_16c` | +23.20% FAIL | +1.06% | [+0.80%, +1.33%] |
| `simd_query_batch_dot_product/pair_loop/128d_256c` | +13.92% FAIL | +0.01% | [−0.82%, +0.84%] |
| `simd_query_batch_dot_product/pair_loop/768d_16c` | +8.95% FAIL | +0.81% | [−0.36%, +2.00%] |
| `simd_normalize/simd/1024` | FAIL | −2.75% | [−5.51%, +0.13%] |
| `int8_raw_dot_product/dot_product_i8_raw/129` | FAIL | +4.87% | [+4.63%, +5.12%] |

Full gate on the idle run: **0 FAIL**, 7 WARN, 38 improvements — a symmetric ±3–7% scatter in both directions, which is the expected quick-mode (`--quick`) noise floor. Per the gate rule, only CI-lower >7% fails.

The one remaining WARN worth a note, `int8_raw/129` (+4.87% on a 4.7 ns microbench): `dot_product_i8_raw` lives in `crates/embed/src/simd/quantized.rs`, which this PR does not touch. The only simd source file in the diff is `tier.rs`, where the change is a semantics-preserving let-chain collapse in an unrelated cosine fast-path. Sub-5ns odd-size microbenches are alignment/codegen-layout sensitive; this is jitter, not a regression introduced by a mechanical lint migration.

Conclusion: bench-compare confirms no real perf change, consistent with a lint-only diff.

PR #605, opened 2026-07-03
feat(cli): add lattice doctor preflight command (#586)

### LINKED ISSUE #586: feat(cli): lattice doctor — memory-fit + artifact-compatibility preflight
## Context

Nothing today answers "will this model fit in my RAM/VRAM?" before a load is attempted. Metal KV-cache memory scales directly with `max_cache_len` — buffer size is `max_cache_len * kv_dim * element_size` per layer (`crates/inference/src/forward/metal_qwen35.rs:5247`, `:5280`) — and `new_session`/`new_session_inner` (`metal_qwen35.rs:5962-5983`) only validate that `max_cache_len` is nonzero and does not exceed `max_position_embeddings`; there's no preflight estimate returned to the caller. `lattice_serve`'s `load_model` (`crates/inference/src/bin/lattice_serve.rs:254-281`) only discovers whether a model+context combination is feasible in the middle of allocating real Metal buffers — i.e. after most of the load work is already sunk.

## Acceptance criteria

- [ ] `lattice doctor <model>` (or `lattice info <model>`) command that, before attempting to load, reports: estimated weight memory, KV-cache memory as a function of context length, the max context length that fits on this machine, Metal-vs-CPU placement decision, detected quantization, and an actionable reason string for any unsupported artifact (architecture, tokenizer, dtype, quantization scheme, missing tensors)
- [ ] Exits nonzero with a clear message when the requested context length would not fit, instead of failing partway through a Metal allocation
- [ ] No model weights need to be fully loaded to produce the report (config + tensor index inspection is enough)

## Notes

De-risks future format additions (e.g. GGUF / model-pull workflows). Related to #91 (unified CLI) and #551 (serve's hard-coded context ceiling).

ohdearquant:

## Summary

- Adds `lattice doctor --model <dir> [--context N] [--tokenizer-dir DIR]`, a preflight command that reports weight memory, KV-cache bytes/token, max feasible context length, Metal-vs-CPU placement, detected quantization, tokenizer presence, and actionable reasons for any incompatibility — all from header/manifest inspection, never by loading tensor payload. Exits nonzero when the requested context doesn't fit or the artifact is broken.
- KV-cache math reuses `Qwen35Config::kv_bytes_per_token` (the same formula the real Metal allocator uses) and missing/dtype checks reuse `qwen_required_tensor_names` (the real loaders' own required-tensor list), rather than reimplementing either. Safetensors single-file-vs-index precedence mirrors `Qwen35Model::from_safetensors` exactly. Q4 directories prefer the `quantize_index.json` manifest over a raw scan specifically to avoid double-counting `merged_qkvz_*` runtime-cache files.
- All new code is one `mod doctor { ... }` plus a `Command::Doctor` CLI variant in `crates/inference/src/bin/lattice.rs`. No other file touched; no new dependency.

Closes #586.

## Test plan

- [x] 22 new unit tests: hand-computed memory math (cross-checked against `Qwen35Config::qwen35_0_8b()`'s documented KV identity), safetensors/Q4 inventory edge cases, end-to-end `build_report` scenarios via an injectable memory override
- [x] Full crate suite: 1425 lib + 85 bin tests, 0 failures (with and without `metal-gpu`)
- [x] `cargo clippy --workspace -- -D warnings` clean (both default and `metal-gpu`-featured)
- [x] `cargo fmt --check` clean
- [x] Ran the built binary against real local checkpoints (`qwen3.5-0.8b`, `qwen3.5-0.8b-q4`); KV bytes/token output matches hand arithmetic exactly; discovered and correctly excluded 8 on-disk Q4 files that are MTP draft-head weights absent from the manifest (validates the manifest-preferring design rather than undermining it)

## Follow-up: scale F16/BF16 weight bytes to resident f32 size

The CPU safetensors loader (`f32_weights.rs`) always materializes weights as owned f32, converting F16/BF16 on load. `lattice doctor` was feeding raw on-disk `byte_len` straight into the RAM budget, under-reporting resident memory by ~2x for any bf16/f16 safetensors checkpoint loaded on CPU (the common HF dtype) — optimistic in exactly the direction a preflight tool must not err.

Added `cpu_resident_bytes()`, scaling 2-byte dtypes (F16, BF16) to their 4-byte resident size; F32 and other dtypes are left unscaled (unsupported dtypes already block a "ready" verdict via the existing check). Q4 inventory untouched — that path stays quantized on Metal and was never affected by this bug.

Mutation-verified: reverting the scaling to identity made the new fixture test fail with the exact pre-scaling byte count; restoring it passes.

## Follow-up: account for Q4/Metal dequantization + duplication

`inspect_q4_dir` had the same class of bug as the CPU path above, but the Q4/Metal loader (`MetalQwen35State::from_q4_dir`) has a different and more varied set of expansions: `.f16` sidecars upcast to f32 (2x), `in_proj_a`/`in_proj_b` dequantized from Q4 to f16 Metal buffers for the CPU GDN recurrence path (3.2x — a `Q4Block` packs 32 weights in 20 bytes, f16 resident is 2 bytes/elem), `in_proj_qkv`/`in_proj_z` each mmap'd at their own size AND duplicated into a merged `in_proj_qkvz` runtime-cache buffer with no manifest entry of its own (2x each), and `embed_tokens` dequantized into a full f16 buffer for CPU embedding lookup AND separately mmap'd for the GPU logits GEMV (4.2x).

Added `q4_resident_bytes()`, a name-pattern classifier applied in both the manifest and no-manifest-fallback inventory paths, with direct unit coverage of every category. Also fixed two related report-accuracy gaps from the same review round: the "max feasible context length" language covered weights + KV cache only, silently omitting activation buffers and GDN recurrent-state scratch (reworded to say so explicitly), and the Q4/Metal context estimate ignored that `MetalChatBackend` hard-caps the KV cache at 4096 tokens regardless of the model's `max_position_embeddings` (added `effective_max_position_embeddings()` to cap the Metal-placement estimate, plus a report line showing the runtime cap separately from the model's architectural ceiling).

Mutation-verified: reverting either call site's fix, or the classifier's cap logic, made the corresponding test fail; restoring passed.

## Follow-up: classify MTP pre-FC norms and disclose MTP KV cache

A re-review found two further gaps, both specific to Q4/Metal checkpoints that include MTP (multi-token prediction) tensor files:

1. `q4_resident_bytes()` classified `mtp.pre_fc_norm_embedding.weight` and `mtp.pre_fc_norm_hidden.weight` at 1x (unadjusted on-disk bytes) instead of 2x. Both are `.f16` sidecars `load_mtp_q4_weights` upcasts to f32 via `load_f16_buf_as_f32`, same as `mtp.norm.weight` — but their normalized names end in `_embedding_weight`/`_hidden_weight`, not `_norm_weight`, so they fell through the existing `norm_weight` match. Added explicit suffix matches for both.
2. When a Q4 directory has MTP weights, `from_q4_dir` allocates a separate `MetalMtpCache` K/V buffer pair that `kv_bytes_per_token` never accounts for (it only covers the main model's full-attention KV cache), so the reported max feasible context length silently overstated what MTP checkpoints can actually fit. Rather than estimate the MTP cache's own bytes/token (its buffers are fixed f32 width regardless of the configurable main-model KV dtype — a second precise numeric claim not worth the added risk), added `has_mtp_tensors` detection (via an `mtp.`/`mtp_` prefix check during the existing inventory walk, in both the manifest and fallback paths) and an explicit disclosure line in the report when true.

Mutation-verified: reverting each of the four independent pieces (the classifier suffixes, both inventory-walk detection sites, and the Display disclosure block) one at a time was confirmed to make its guarding test fail, then restored to pass.

### bench-compare

`make bench-compare` (quick, `origin/main` `a00d7ebcd` vs `HEAD` `5b455a684`):

The gate reports **152 FAIL / 11 WARN / 35 improvement** out of 259 measured benchmarks — every group in `elementwise_cpu_bench` (`rms_norm`, `layer_norm`, `silu_inplace`, `gelu`, `add_bias_gelu`, `softmax_attention`, `elementwise_mul`) plus most of `lattice-embed`'s `simd` suite, with deltas as large as +645% (`softmax_attention/128`) alongside similarly large improvements scattered through the same benchmark families elsewhere in the report.

This PR's diff touches only `crates/inference/src/bin/lattice.rs`'s `doctor` module — a name-pattern classifier and struct-field addition with no runtime hot path — so there's no code path connecting it to any kernel in that list. Three things point at comparison noise rather than a caused regression, in order of how directly each was checked:

- **Direct control test.** `softmax_attention` is the largest reported FAIL (+645.34% / +636.95% on the two sizes). Re-running it twice in a row against the *same already-built* HEAD binary, zero code change in between, gave `-1.53%` (p=0.04, Criterion's own verdict: "Change within noise threshold") and `-0.03%` (p=0.92, "No change in performance detected"). The code did not get 645% slower; it didn't move outside Criterion's own noise floor at all.
- **Blanket, disjoint failure.** All seven `elementwise_cpu_bench` groups fail simultaneously and by a large margin. A one-file CLI diff can't plausibly cause a uniform regression across kernels that share no code with it or with each other.
- **Bidirectional swings within one family.** The `simd` suite has the same benchmark function swinging in opposite directions by similar magnitude a few size variants apart (e.g. one `int8_prepared_dot_product` size FAILs at +38% while a neighboring size WINs at -34%) — the signature of variance, not a directional effect.

The likely mechanism: `bench-compare.sh` builds BASE and HEAD in two independently-compiled worktree/target directories, then compares via a Criterion baseline copied from one into the other. This repo has seen that two-build comparison produce spurious deltas before; this run is a more extreme instance of the same pattern, not a new one.

Since the changed code has no runtime hot path, this isn't treated as blocking. A clean re-run (optionally scoped via `BENCH_GROUPS_INFERENCE`/`BENCH_GROUPS_EMBED`) is available on request if a maintainer wants independent confirmation before merge.

### DIFF
--- a/crates/inference/src/bin/lattice.rs
+++ b/crates/inference/src/bin/lattice.rs
@@ -1,10 +1,11 @@
-//! `lattice` CLI — interactive chat and HTTP serve subcommands.
+//! `lattice` CLI — interactive chat, HTTP serve, and preflight subcommands.
 //!
 //! # Usage
 //!
 //! ```text
 //! lattice chat --model /path/to/model [--max-tokens 256] [--temperature 0.7]
 //! lattice serve --model /path/to/model [--host 127.0.0.1] [--port 8080] [--max-tokens 256]
+//! lattice doctor --model /path/to/model [--context 4096]
 //! ```
 
 use clap::{Parser, Subcommand};
@@ -54,6 +55,23 @@ enum Command {
         /// Model identifier echoed in responses (defaults to the model path basename)
         #[arg(long)]
         model_id: Option<String>,
+        /// Directory containing tokenizer.json, when it is not shipped inside
+        /// --model (only needed for Q4 directories produced without a
+        /// co-located tokenizer; safetensors directories always ship one).
+        #[arg(long)]
+        tokenizer_dir: Option<String>,
+    },
+    /// Preflight check: memory fit and artifact compatibility, without
+    /// loading any model weights (config + tensor index inspection only).
+    Doctor {
+        /// Path to model directory (SafeTensors, or a native Q4 quantized
+        /// directory produced by `quantize_q4`)
+        #[arg(long)]
+        model: String,
+        /// Context length to check feasibility for. When omitted, only the
+        /// maximum feasible context length is reported.
+        #[arg(long)]
+        context: Option<usize>,
         /// Directory containing tokenizer.json, when it is not shipped inside
         /// --model (only needed for Q4 directories produced without a
         /// co-located tokenizer; safetensors directories always ship one).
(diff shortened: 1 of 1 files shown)


ohdearquant:
Pushed a follow-up fix (9ba1087a1) for the RAM-accounting issue from review: the CPU safetensors loader always materializes weights as owned f32 (converting F16/BF16 on load), but the doctor's weight-memory budget was summing raw on-disk bytes, under-reporting resident RAM by ~2x for bf16/f16 checkpoints on CPU. `cpu_resident_bytes()` now scales those two dtypes to their 4-byte resident size; Q4 (Metal, stays quantized) is unaffected. New test + mutation-checked (reverted the scaling, confirmed the test fails, restored it). Full suite (86 doctor/serve tests), clippy, and fmt all clean.

The remaining nice-to-haves (activation/scratch headroom margin, effective MAX_CACHE_LEN cap disclosure, config-less-directory fallback note, strict data_offsets value validation, dynamic KV display context) are left as-is for now -- happy to open a fast-follow issue for those if that's preferred over holding this PR open further.

Issue #663, opened 2026-07-05
feat(serve): lattice_serve does not validate the request's model field


ohdearquant:
## Problem

The two HTTP serve surfaces diverge on `model` field validation:

- `lattice serve` (the CLI's HTTP subcommand) rejects a request whose `model`
  field does not match the served model id, returning HTTP 400 with code
  `model_not_found` (`crates/inference/src/bin/lattice.rs`, `chat_completions`).
- `lattice_serve` (the standalone daemon) does not validate `model` at all —
  it accepts any value, or falls back to the loaded model id when the field is
  absent (`crates/inference/src/bin/lattice_serve.rs:1279`,
  `let model_id = req.model.clone().unwrap_or_else(|| s.model_id.to_string());`).

A client that sends a typo'd or stale `model` value against `lattice_serve`
gets a normal completion instead of an error naming the mismatch, which can
mask client-side configuration bugs during multi-model workflows.

## Proposal

Add the same fail-closed `model` field check to `lattice_serve`'s
`chat_completions` handler, returning HTTP 400 when `req.model` is present
and does not match `s.model_id`, matching `lattice serve`'s
`model_not_found` contract.

Found via the #654 capability-matrix executable-fixture harness, which
encodes this as an expected-deviation fixture pending this fix.

Issue #551, opened 2026-07-02
lattice_serve: derive MODEL_MAX_CONTEXT from model config instead of hard-coded 4096


ohdearquant:
> _Issue authored by Claude (Anthropic agent) on behalf of @ohdearquant._

## Problem

`crates/inference/src/bin/lattice_serve.rs` clamps `max_tokens` and `reasoning_budget` against a hard-coded constant:

```rust
const MODEL_MAX_CONTEXT: usize = 4096;
```

The clamp itself is correct and load-bearing (it prevents a capacity-overflow abort in the GPU worker, which would be a persistent denial of service since the worker thread owns the Metal state for the process lifetime). But the *value* is wrong to hard-code:

- A model with a larger context window (e.g. a 32K-context checkpoint) gets silently truncated to 4096 — any serving claim for larger models is dishonest until this is derived.
- A model with a smaller window would be over-admitted and still hit the fail-closed path in the generate loop, which works but produces a worse error at a later stage.

## Proposed fix

Derive the limit at startup from the loaded model's config (context length / KV-cache capacity actually allocated), store it in the server state, and use that everywhere the 4096 const is used today. Keep the fail-closed prompt+completion overflow check in the generate path as the second line of defense.

## Acceptance

- No hard-coded 4096 in `lattice_serve.rs`; the clamp bound comes from model config / allocated KV capacity.
- A test (or startup log assertion) showing the derived bound matches the model's actual context window for the default checkpoint.
- Overflow behavior remains fail-closed (reject before alloc, no abort).

ohdearquant:
Scope extension from PR #581 review: the new `lattice serve`/`lattice chat` Q4 path in `crates/inference/src/bin/lattice.rs` adds fresh call sites on the same hard-coded bound (`MetalChatBackend::MAX_CACHE_LEN = 4096` feeds both the Metal state's cache allocation and the served `max_context`). When fixing this, cover `lattice.rs` too, not just `lattice_serve.rs` — otherwise a 27B chat/serve session silently caps at 4096 context.

Related config-honesty item found in the same review: `load_q4_config` in `lattice.rs` falls back to a 27B default config when a Q4 checkpoint directory has no `config.json`. The fallback is loud (logged) but permissive — a 0.8B checkpoint without `config.json` would get a 27B config and fail confusingly downstream. Consider making the missing-config case a hard error as part of this pass.

PR #665, opened 2026-07-05
test(serve): executable endpoint capability matrix (#654)

### LINKED ISSUE #654: harden(serve): make the endpoint capability matrix executable
> _Issue authored by Codex (OpenAI agent) on behalf of @ohdearquant._

## Problem

Serve behavior keeps drifting across three sibling surfaces: `lattice serve`, standalone `lattice_serve`, and `chat_metal --json --serve`. The docs matrix now tracks this manually, but the recurring issue pattern is that features land in one path and sibling behavior is discovered later by review/audit.

Recent examples:

- #462/#619: prefix cache reached `chat_metal`, not either HTTP server.
- #585/#620: `logprobs` reached `lattice serve`; `lattice_serve` still has no fields and hardcodes `logprobs: None`.
- #589/#642: the matrix had to be manually source-audited because no gate encodes the expected parity/deviation.
- #641, filed from the matrix audit: `lattice_serve` silently coerces/drops malformed role/content where `lattice serve` rejects.

Current evidence:

- `docs/capability-matrix.md:70-100` is effectively a hand-maintained spec.
- `docs/capability-matrix.md:75`, `:94`, and `:97-99` show high-impact drift that is easy to regress.

## Proposal

Create an executable capability matrix test harness:

- Encode each request field/endpoint as `{surface, input, expected status/error/behavior}` fixtures.
- Exercise both HTTP handlers in-process where possible; for `chat_metal --json --serve`, factor request parsing enough to unit-test without spawning Metal.
- Require every matrix row to have either a test fixture or an explicit `not-testable-yet` justification.
- Make docs generation or docs lint compare `docs/capability-matrix.md` against the fixture IDs so stale rows fail closed.

ohdearquant:
## Summary

Closes #654. Turns `docs/capability-matrix.md` from a hand-maintained
claim into a fixture-backed contract, on all three serve surfaces:
`lattice serve` (in `bin/lattice.rs`), the standalone `lattice_serve`
daemon, and `chat_metal --json --serve`.

- **`bin/lattice.rs`**: extracted `validate_chat_request()` out of the
  `chat_completions` handler's pre-model validation cascade into a
  pure, directly-testable function — no behavior change, only the
  context-window check (which needs `state.model`) stays inline. Added
  7 fixture tests covering model mismatch, empty messages, non-user
  last message, unsupported-feature rejection ordering, stop-sequence
  handling, and logprobs resolution.
- **`bin/lattice_serve.rs`**: added one new fixture,
  `cm_lattice_serve_model_mismatch_accepted_expected_deviation`, that
  proves — rather than merely asserts in prose — that this surface
  does not yet validate the request's `model` field the way `lattice
  serve` does. That gap is real drift, tracked separately in #663; this
  PR encodes it as a passing "expected deviation" test so that fixing
  the drift without updating the fixture/doc together makes the test
  fail, rather than the doc silently going stale again. 29 pre-existing
  fixtures on this surface are otherwise untouched.
- **`bin/chat_metal.rs`**: extracted `parse_serve_request_line()` out of
  the `--json --serve` stdin loop into a pure function — same
  malformed/missing-prompt error strings, same default/override
  resolution, no behavior change. Added 6 fixture tests (defaults-only,
  per-request overrides, malformed JSON, missing prompt, zero
  reasoning-budget fallback, statelessness of stop-strings/history
  fields on this surface).
- **`scripts/check-capability-matrix.sh`** (new): cross-checks every
  fixture ID cited in the doc's new "Fixture manifest (#654)" section
  against a real `#[test]` fn across the three binaries, failing closed
  if a row cites a fixture that was renamed or removed. Wired into
  `scripts/lint-docs.sh` (`make lint-docs`), reusing the existing
  doc-lint mechanism rather than adding a new one.
- **`docs/capability-matrix.md`**: re-audited row-by-row against
  current `main` (post #656 fail-closed role/content-part/tool/
  response_format rejection, post #662 cross-turn KV/GDN prefix cache
  wiring) rather than trusting its own stale audit-commit claim.
  Several previously-documented drift rows are now symmetric across
  surfaces and were corrected; the `model`-field validation asymmetry
  is newly documented (#663). Added the Fixture manifest table the
  check script verifies, split by HTTP surfaces and the `chat_metal`
  stdin surface.
- **`.github/workflows/ci.yml`**: added a step that actually *runs*
  `cargo test --bin lattice_serve` and `--bin chat_metal` under
  `--features f16,metal-gpu` on the macOS leg. Previously these
  binaries were built and clippy'd there but their test suites were
  never executed by CI — meaning none of these fixtures, new or
  pre-existing, would have caught a regression.

## Test plan

- [x] `cargo test -p lattice-inference --bin lattice` — 124 passed
- [x] `cargo test -p lattice-inference --bin lattice_serve --features f16,metal-gpu` — 30 passed
- [x] `cargo test -p lattice-inference --bin chat_metal --features f16,metal-gpu` — 6 passed
- [x] `cargo clippy --workspace -- -D warnings` — clean
- [x] `cargo clippy -p lattice-inference --all-targets --features f16,metal-gpu -- -D warnings` — clean
- [x] `make lint-docs` — clean (includes the new fixture-check script)
- [x] `make bench-compare` — no computational hot path changed (only
      request-handling code in `bin/*.rs`, which criterion doesn't
      benchmark); run anyway per repo convention, no structural change.

This PR stays in **draft** — the endpoint-parity gap tracked as #663 is
intentionally left unfixed here; this PR only makes it visible and
gated.


ohdearquant:
Addressed round-1 design-review findings. New head: 37236f85a98c8e06a5daea9200c36b3472e012e9.

**1. MAJOR — refactor not order-preserving (fixed):** verified the pre-refactor order
directly against 3e0f74155 (context-window check before `stop` parsing). Restored it
by composing `validate_chat_request` + a new `check_context_window` + `parse_stop_strings`
inside one `prepare_chat_request` function, in that exact order, with `tokenize_len`/
`max_context` threaded through as thunks so the ordering itself is unit-testable without
a loaded model. Added `cm_serve_context_window_checked_before_stop_parsing`, which drives
`prepare_chat_request` end-to-end (not the sub-functions in isolation) — confirmed
mutation-sensitive by temporarily reverting the ordering and watching it fail, then
restoring it and watching it pass again.

**2. MEDIUM — fixture-check false positives (fixed):** rewrote `check-capability-matrix.sh`
to extract the actual set of `#[test]`/`#[tokio::test]`-attributed fn names from all three
binaries and check citations against that set, instead of a bare `grep "fn $tok("` that
any function or comment with a matching name could satisfy. Also scoped token extraction to
table rows only, since the section's prose legitimately backtick-references non-test fns by
name (e.g. `validate_chat_request`), which both the old whole-section scan and the tightened
check would otherwise misfire on. Added a `--selftest` mode that injects a table row citing a
fixture ID guaranteed not to exist and asserts the check fails on it; wired into
`make lint-docs` so this runs on every invocation. Verified portable to macOS bash 3.2 (ran
explicitly under `/bin/bash --version` 3.2.57 as well as the `/bin/sh` shebang) — no arrays,
`mapfile`, or `local`.

**3. MEDIUM — stale max-context doc row (fixed):** the row still described a hardcoded
`MODEL_MAX_CONTEXT` constant and cited #551 as open. #551 is closed; current source derives
the context window from the loaded model's config via `model_context_from_config`, falling
back to a constant only when no config is available. Updated the row to match current
source, and cited the new ordering fixture in the Fixture manifest's "Request precedence /
order" row.

Re-verified: `cargo test -p lattice-inference --bin lattice` (125 passed), `--bin
lattice_serve --features f16,metal-gpu` (30 passed), `--bin chat_metal --features
f16,metal-gpu` (6 passed); `cargo clippy --workspace -- -D warnings` and the metal-gpu
clippy leg both clean; `make lint-docs` clean (including the new selftest step). Staying
in draft.

PR #688, opened 2026-07-06
perf(inference): route tiny BERT attention GEMMs off Accelerate to in-process NEON

ohdearquant:
## Summary

Investigated whether routing the tiny per-head BERT attention GEMMs off Apple Accelerate to an in-process NEON kernel improves single-text embedding latency (all-MiniLM-L6-v2 / BGE-small on Apple Silicon CPU).

## Finding

The routing change is **not a net win**, and this PR is on hold pending disposition. The earlier body claimed a ~29% single-text speedup; that figure was within measurement noise on a shared machine and does not hold up under a controlled same-process micro-benchmark. Correcting it here.

Micro-benchmarks (same-process, high-iteration, CV < 2%) at the real single-text shapes show the in-process NEON kernel is shape-sensitive:

- `Q @ K^T` (k = head_dim = 32, a clean multiple of 16): NEON ~2.5x faster than Accelerate.
- `scores @ V` (k = seq_len ~ 15, remainder-heavy): NEON ~2.4x **slower** than Accelerate.

A single work-size threshold routes both to NEON, so the two effects roughly cancel and the per-head result is a small net regression. Routing only `Q @ K^T` recovers the win, but the attention GEMMs are only ~7% of the single-text encode, so the corrected best case is a ~3% single-text improvement on Apple Silicon, within noise.

## Broader result

A separate sweep confirmed Apple Accelerate wins every FFN / QKV / output-projection GEMM shape at every batch size from M=1 to M=128, by ~10x at the single-text regime (M ~ 15). Single-text embedding is Accelerate/AMX-bound: there is no in-process GEMM kernel that beats it. Batch throughput, where the fixed per-call dispatch cost amortizes, is where lattice's advantage shows (measured +30-60% versus the ONNX reference at batch sizes >= 8).

## Status

Held as draft. The attention numerical-parity tests and the mutation-sensitive route-decision test are sound. The routing policy itself needs either correction (`Q @ K^T`-only) or the change should be closed as a ~3% Apple-Silicon-only micro-optimization on a path that is already at parity with the reference.


ohdearquant:
Closing. A controlled same-process micro-benchmark (28 GEMM cells, two runs, CV under 2 percent, outputs cross-checked) shows Apple Accelerate wins every FFN, QKV, and attention GEMM shape at every batch size from M=1 to M=128, by about 10x at the single-text regime (M around 15), with no crossover. Single-text embedding is Accelerate/AMX-bound: no in-process GEMM kernel beats it, so routing the tiny attention GEMMs to an in-process NEON kernel is a net regression as written, and at best a ~3 percent Apple-Silicon-only micro-optimization even when narrowed to the clean-shape Q@K^T case. The honest position is parity with the ONNX reference on single text (both ride the same AMX substrate) and a decisive batch-throughput advantage (+30 to 60 percent at batch sizes 8 and above), where the fixed per-call dispatch cost amortizes. Keeping the branch for reference; not merging.

Issue #103, opened 2026-05-25
Qwen3-Embedding-0.6B forward-pass divergence (cosine 0.948 vs HF on whitespace/CJK inputs)


ohdearquant:
## Symptom

The HF parity regression test from the `embed-perf-quality` work measures Qwen3-Embedding-0.6B failing 3/5 cases against HF reference, with the worst at cosine **0.9477** (leading-whitespace input). On the "fox" input — where lattice tokens match HF token-for-token — vectors still diverge at cosine **0.9886**, max-abs-diff **1.41e-2**.

BGE, E5, all-MiniLM, and paraphrase-multilingual all reach min cosine ≥ 0.9998 across the same 5 inputs after that work's tokenizer + embed fixes. Qwen3-Embedding is the only outlier.

## Failing inputs

```
[qwen3-0.6b] "The quick brown fox jumps over the lazy dog."   cosine=0.988615  max_diff=1.41e-2
[qwen3-0.6b] "   leading whitespace and    multiple    spaces   "  cosine=0.947711  max_diff=3.47e-2
[qwen3-0.6b] "短い日本語のテストです。"                            cosine=0.989882  max_diff=1.74e-2
```

Passing inputs:
```
[qwen3-0.6b] "Pure Rust transformer inference engine."         cosine=0.997272 PASS
[qwen3-0.6b] "Café résumé naïve façade — Unicode test."        cosine=0.995420 PASS (barely)
```

## Important observation

For the "fox" input, **lattice's BPE tokenizer produces identical IDs to HF** (verified — fixture `qwen3_embedding_0_6b.json` records HF's tokenization, and `audit_tokenizer_parity` cases include this exact input). So the divergence is **not** in the tokenizer — it's in the forward pass between encoded IDs and the final pooled embedding.

Per `lattice/CLAUDE.md` quantitative bounds: bf16 vs f32 PPL delta is < 0.05 (negligible). A vector divergence of 1.4e-2 to 3.5e-2 is structurally larger than bf16 noise and points to a real algorithmic difference, not numerical drift.

## Hypotheses (ordered by historical precedent)

1. **RoPE convention mismatch** — Qwen3-Embedding's RoPE differs from Qwen3-Chat's RoPE in some subtle way (e.g., the per-head dimension, the theta base, or pair convention). The v0.2.3 stride-half vs interleaved RoPE bug had this exact symptom shape.
[...]

PR #699, opened 2026-07-06
perf(inference): extend fused add_bias+gelu to the LoRA-hook BERT FFN path

### LINKED ISSUE #675: perf(embed): wire existing add_bias_gelu fused kernel into BERT FFN (both forward paths)
From the four-lens CPU embedding optimization audit (source-verified).

**Problem**: a fused `add_bias_gelu` kernel already exists (crates/inference/src/forward/cpu/activation.rs:282, scalar fallback at :321) but the BERT FFN uses separate `add_bias` + `gelu` passes in BOTH the single-sequence path and the new batched path (crates/inference/src/model/bert.rs, both call sites) — two full read/write passes over the 4H intermediate activation where one suffices.

**Proposed change**: replace the add_bias-then-gelu pairs with the existing fused kernel at both call sites; grep the same files for any sibling call sites while there.

**Impact**: Medium (memory-bandwidth save on the largest activation in the layer). **Effort**: S — the kernel already exists and is tested. **Parity risk**: minimal (same math; HF-parity gate covers).

**Measure**: A/B single-text p50 + batch texts/s; expect a small consistent win, kill if noise.

Ref: ADR-070 (#671).

ohdearquant:
## Summary

Issue #675 asked for the existing fused `add_bias_gelu` kernel to be wired into the BERT FFN in place of separate `add_bias` + `gelu` passes. By the time this branch was cut, the no-adapter paths (`BertModel::encode` / `encode_batch`) already used the fused kernel (landed opportunistically in #678). The one remaining gap was `forward_with_hook`, which still ran the naive `add_bias` -> `lora.apply` -> `gelu` sequence whenever the caller could supply a real LoRA adapter (`CrossEncoderModel::score_with_hook`), because the code assumed the adapter add-in had to happen strictly between the bias-add and GELU.

That assumption isn't required for the value to land before GELU. `LoraHook::apply` only computes `output += scale * B @ (A @ x)`, an additive update that depends on the pre-FFN hidden state `x`, not on the current contents of the output buffer, so applying the adapter delta before the fused bias+GELU pass still puts it before GELU. This lets the fused kernel be used unconditionally, closing the remaining gap in #675.

**Reassociation, not bit-exactness.** f32 addition is not associative, so `(matmul + bias) + delta` and `(matmul + delta) + bias` are not guaranteed to be bit-identical (this can matter under extreme cancellation between terms of very different magnitude). This PR accepts that as a reassociation, not a bit-exact equivalence, because: real BERT FFN pre-activations are O(10-100) with O(1) biases and small LoRA deltas, far from any cancellation regime; there is no exact-output consumer on this adapter path; `embed_parity_vs_hf` empirically confirms parity (cosine similarity at least 0.9998) against the reference implementation on every shipping BERT-family model; and lattice's SIMD kernels already reassociate f32 sums elsewhere in this crate. The doc comments and this PR body previously said "same result" / "equivalent" without qualification; that was too strong and has been corrected.

## Change

One file, `crates/inference/src/model/bert.rs`:
- `forward_with_hook`'s FFN-intermediate stage now always does `lora.apply(...)` followed by the fused `add_bias_gelu(...)`, instead of branching on a `fuse_ffn_gelu: bool` between that and the old three-pass sequence.
- Removed the now-unnecessary `fuse_ffn_gelu` parameter and its two call sites.
- The ffn_intermediate composition (`lora.apply` then fused `add_bias_gelu`) is extracted into a small free function, `apply_ffn_intermediate_lora_and_activation`, so the ordering invariant can be unit-tested directly without a model.
- Updated the doc comments to state the reassociation argument plainly (see above), instead of overclaiming bit-exact equivalence.
- Dropped the now-unused `gelu` import from the top-level import list (`add_bias` is still used at the FFN-output stage, which is unaffected; the test module imports `gelu` separately for its own independent reference computations).

## Correctness

- `cargo clippy -p lattice-inference --tests -- -D warnings`: clean.
- `cargo fmt -p lattice-inference --check`: clean.
- `cargo test -p lattice-embed --test embed_parity_vs_hf`: 4 passed, 1 pre-existing ignore (Qwen3-Embedding, unrelated, see #103).
- `cargo test -p lattice-inference --lib bert --include-ignored` (with the model present locally, `LATTICE_INFERENCE_MODEL_DIR` set): 12 passed, 0 failed.

### Active-adapter regression tests (two, one default and one integration)

The existing HF-parity coverage only exercises the no-adapter path. This PR adds two tests that enforce the "delta lands before GELU" invariant instead of only asserting it in a comment:

- `test_ffn_intermediate_stage_lora_delta_lands_before_gelu` (default, runs in CI with no model): calls `apply_ffn_intermediate_lora_and_activation` directly against small, fixed synthetic buffers. Builds an independent reference using separate, unfused `matmul_bt` -> `add_bias` -> `+= delta` -> `gelu` calls and asserts the production function matches it within `1e-5`, and that the adapter measurably moves the output (`> 1e-3` vs no delta). This is the default CI gate for the invariant; the previous test alone was `#[ignore]`-gated on a downloaded model and would silently no-op in ordinary CI runs.
- `test_ffn_intermediate_lora_delta_lands_before_gelu` (`#[ignore]`, gated on `LATTICE_INFERENCE_MODEL_DIR`): end-to-end integration coverage through a real, single-layer-truncated `BertModel` and a real adapter hook reached the same way `CrossEncoderModel::score_with_hook` reaches it. A `CapturingDeltaHook` adds a nonzero constant delta to the `ffn_intermediate` projection output and records its input activation `x` (produced before the FFN block, so independent of where the hook fires), then builds the same kind of independent reference. Kept `#[ignore]`d because building a synthetic `BertModel` in-process would need a hand-built safetensors buffer (this crate has its own mmap-based safetensors reader, no external `safetensors` crate dependency) plus a synthetic `tokenizer.json`, which is more scaffolding than this fix warrants.

Both were verified mutation-sensitive: temporarily swapping the extracted function to the broken order (fused bias+GELU before the adapter add-in) makes both tests fail (max-abs-diff `0.0208` and `0.291` against their `1e-5`/`1e-4` thresholds); restoring the correct order makes both pass again.

## Performance

The `e2e_encode` / `e2e_encode_batch` Criterion benches route through `encode()`/`encode_batch()`, both of which already took the fused-kernel path before this change (per #678) or don't touch `forward_with_hook` at all. So these benches don't exercise the code path this PR actually changes. A same-binary re-run against a saved baseline confirmed this: results were not reproducible in direction (one benchmark flipped from an apparent 8.9% improvement to a 16.1% regression between two runs of the identical binary), consistent with the change being a no-op for those paths. No performance claim is made beyond removing the now-redundant runtime branch; the correctness argument above is the actual point of this PR, and it extends the FFN-fusion benefit already shipped for the no-adapter paths to the LoRA-adapter path as well.

Closes #675

## Test plan

- [x] `cargo clippy -p lattice-inference --tests -- -D warnings`
- [x] `cargo fmt -p lattice-inference --check`
- [x] `cargo test -p lattice-embed --test embed_parity_vs_hf`
- [x] `cargo test -p lattice-inference --lib bert --include-ignored` (with model present) — 12 passed, including both active-adapter regression tests

### DIFF
--- a/crates/inference/src/model/bert.rs
+++ b/crates/inference/src/model/bert.rs
@@ -453,10 +453,14 @@ impl BertModel {
         let seq_len = input.real_length;
         // `lora` here is caller-supplied and may be a real adapter (see
         // `CrossEncoderModel::score_with_hook`). The FFN fast path (#675) is
-        // still safe to take: `LoraHook::apply` only ever adds a delta that
+        // still taken here: `LoraHook::apply` only ever adds a delta that
         // depends on the pre-FFN hidden state, not on the bias-add's output,
         // so doing the adapter add-in before the fused bias+GELU pass gives
-        // the same result as adding it in between (addition commutes).
+        // the same value as adding it in between, up to floating-point
+        // summation-order rounding (f32 addition is not associative, so this
+        // is a reassociation, not an exact-value guarantee). See
+        // `forward_with_hook`'s doc comment for why that reassociation is
+        // accepted here.
         self.forward_with_hook(
             &input.input_ids[..seq_len],
             &input.attention_mask[..seq_len],
@@ -489,13 +493,28 @@ impl BertModel {
     /// Hook-aware internal full transformer forward pass.
     ///
     /// The FFN intermediate stage always uses the fused `add_bias_gelu`
-    /// kernel (#675) instead of separate `add_bias`/`gelu` passes. This is
-    /// safe with a real `lora` adapter too: `LoraHook::apply` only adds
+    /// kernel (#675) instead of separate `add_bias`/`gelu` passes, including
+    /// when a real `lora` adapter is active: `LoraHook::apply` only adds
     /// `scale * B @ (A @ x)` into the buffer, a delta that depends on the
-    /// pre-FFN hidden state (`x`) rather than on the bias-add's output, so
-    /// applying it before the fused bias+GELU pass is equivalent (up to
-    /// negligible f32 summation-order rounding) to the previous
-    /// bias-add-then-adapter-then-GELU ordering.
+    /// pre-FFN hidden state (`x`) rather than on the bias-add's output, so it
+    /// can be applied before the fused bias+GELU pass instead of between the
+    /// bias-add and GELU as before.
+    ///
+    /// This reorder changes floating-point summation order. Adding the bias
+    /// then the delta, versus adding the delta then the bias, is not
+    /// guaranteed to produce a bit-identical result, because f32 addition is
+    /// not associative (this can matter under extreme cancellation, e.g.
+    /// terms differing by many orders of magnitude). We accept this as a
+    /// reassociation, not a bit-exact equivalence, for these reasons:
+    ///
+    /// - Real BERT FFN pre-activations are O(10-100) with O(1) biases and
+    ///   small LoRA deltas, well outside any cancellation regime.
+    /// - There is no exact-output consumer on this adapter path.
+    /// - The `embed_parity_vs_hf` test empirically confirms parity against
+    ///   the reference implementation (cosine similarity at least 0.9998) on
+    ///   every shipping BERT-family model.
+    /// - lattice's SIMD kernels already reassociate f32 sums routinely
+    ///   elsewhere in this crate.
     fn forward_with_hook(
         &self,
         input_ids: &[u32],
@@ -592,11 +611,16 @@ impl BertModel {
                     intermediate_size,
                 );
                 // #675: adapter add-in before the fused bias+GELU pass (see
-                // `forward_with_hook`'s doc comment for why this ordering is
-                // equivalent to the previous add_bias-then-adapter-then-gelu
-                // sequence).
-                lora.apply(layer_idx, "ffn_intermediate", &hidden, ffn_intermediate);
-                add_bias_gelu(
+                // `forward_with_hook`'s doc comment for the accepted
+                // f32-reassociation argument versus the previous
+                // add_bias-then-adapter-then-gelu ordering). Extracted into
+                // `apply_ffn_intermediate_lora_and_activation` so this
+                // ordering can be unit-tested directly, without a model
+                // (see that function's doc comment).
+                apply_ffn_intermediate_lora_and_activation(
+                    lora,
+                    layer_idx,
+                    &hidden,
                     ffn_intermediate,
                     layer.ffn_intermediate_bias.data,
                     intermediate_size,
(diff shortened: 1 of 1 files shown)


ohdearquant:
Added a mutation-sensitive regression test (test_ffn_intermediate_lora_delta_lands_before_gelu) that exercises the reorder with a real (non no-op) adapter, addressing the gap that the previous commit's correctness argument was only backed by no-adapter parity tests.

The test uses a CapturingDeltaHook that adds a constant delta to the ffn_intermediate projection on a single-layer-truncated model, then builds an independent reference (separate matmul_bt/add_bias/gelu calls, hardcoding "delta lands before gelu") from the hook's captured input activation, which is unaffected by the ordering under test.

Mutation-sensitivity verified locally: swapping the ffn_intermediate site back to add_bias_gelu-then-lora.apply made the test fail with max-abs-diff 0.291 against the 1e-4 threshold. Restoring the correct order passes again. Full clippy/parity/unit gates re-run clean after the addition.

ohdearquant:
Addressed both review findings, reorder kept (not reverted).

Finding 1 (wording): corrected the doc comments and PR body to stop claiming the reorder produces "the same result" via commutativity. f32 addition is not associative, so this is a reassociation, not a bit-exact equivalence. Stated plainly why it is accepted: real BERT FFN activation magnitudes are far from any cancellation regime, there is no exact-output consumer on this path, embed_parity_vs_hf empirically confirms parity, and lattice already reassociates f32 sums elsewhere.

Finding 2 (default CI gate): investigated building a synthetic BertModel for a non-ignored test. Not feasible without significant scaffolding: BertWeights holds borrowed slices over a real, hand-parsed safetensors buffer (this crate has its own mmap-based reader, no external safetensors crate dependency), and from_bytes also needs a valid tokenizer.json. Building both synthetically is more plumbing than this fix warrants.

Took the kernel/composition-level route instead: extracted the ffn_intermediate composition (lora.apply then the fused add_bias_gelu) into a small free function, apply_ffn_intermediate_lora_and_activation, which forward_with_hook now calls. Added a new non-ignored test, test_ffn_intermediate_stage_lora_delta_lands_before_gelu, that calls this same function directly against small fixed synthetic buffers, needing no model, so it runs in CI by default. The existing model-backed test stays as ignored end-to-end integration coverage, with a comment explaining why it needs a real model.

Mutation-sensitivity re-verified after the extraction: reverting the function body to the broken order (delta after gelu) fails both the new default test (max-abs-diff 0.0208 against a 1e-5 threshold) and the existing model-backed test (max-abs-diff 0.291 against 1e-4); restoring the correct order passes both again.

All gates re-run clean: cargo fmt --check, cargo clippy --tests -D warnings, cargo test --lib bert --include-ignored (12 passed), cargo test embed_parity_vs_hf (4 passed, 1 unrelated ignore).

PR #708, opened 2026-07-07
perf(inference): skip terminal RMSNorm+lm_head on intermediate long-prefill chunks

ohdearquant:
## What

For a prompt longer than `max_prefill`, `forward_prefill_batched_chunk` is called once per chunk. In the last-position (`!all_positions`) mode, callers keep **only the final chunk's logits** — yet every chunk still ran the terminal tail: MTP pre-final capture, final RMSNorm, the full-vocab last-token `lm_head` GEMV, top-k, and their GPU readback. For all but the last chunk that output is discarded.

This adds an `emit_logits: bool` flag to `forward_prefill_batched_chunk` that gates **only** that terminal tail. Intermediate chunks still:
- commit their per-layer command buffer (advancing the KV cache and GDN recurrent state), and
- advance the session cursor via `set_position`,

then skip the discarded tail and return an empty `Vec`. The `emit_logits == true` path is byte-for-byte unchanged.

Both callers are threaded through:
- `forward_prefill_impl`: `true` for single/last chunk and for `all_positions`; `is_last` in the chunked last-position loop.
- `forward_prefill_from` (cross-turn prefix-reuse path): same, with `is_last = pos + chunk.len() == start_pos + total`.

## Why it's correct

The terminal tail does not feed the residual stream, KV cache, or GDN recurrent state — those are fully advanced by the per-layer loop that precedes it. Skipping it on a chunk whose logits are discarded changes nothing observable.

Guarded by a new gpu-locked, **mutation-sensitive** regression test (`metal_qwen35_chunked_prefill_emit_logits_skip_correctness`): on a 679-token prompt (> `max_prefill`, so 2 chunks) it asserts the chunked path's returned logits have length `vocab_size`, the argmax matches the `all_positions` final-position argmax, and the session cursor fully advances. Verified mutation-sensitive: forcing `is_last = false` (wrongly skipping the last chunk's tail) makes the test fail; restoring it passes (`argmax_ref = argmax_chunked = 271`, tokens = 679).

## Performance

This removes provably-discarded work; it is not a throughput claim. At n=4096 / max_prefill=512 the saving is ~7 full-vocab `lm_head` GEMVs, on the order of a few ms against a multi-second TTFT (below A/B resolution). There is no Criterion group that drives the real multi-chunk `forward_prefill` path, so no meaningful before/after number exists for it; `make bench-compare` (default CPU groups) is unaffected by this change and shows no change. The value here is the removed waste plus the correctness guard.

## Test

```
cargo test -p lattice-inference --features metal-gpu,f16 --release --lib \
  metal_qwen35_chunked_prefill_emit_logits_skip_correctness
# test result: ok. 1 passed  (argmax_ref=271, argmax_chunked=271, tokens=679)
```

### DIFF
--- a/crates/inference/src/forward/metal_qwen35.rs
+++ b/crates/inference/src/forward/metal_qwen35.rs
@@ -9483,7 +9483,8 @@ kernel void gdn_chunk_norm_silu_c32(
             );
             let max_prefill = self.session.max_prefill;
             if n <= max_prefill {
-                return self.forward_prefill_batched_chunk(token_ids, 0, all_positions);
+                // Only/last chunk — its logits are always the caller's answer.
+                return self.forward_prefill_batched_chunk(token_ids, 0, all_positions, true);
             }
             // Chunked batched prefill: each chunk is one command buffer (preserving the
             // n≤512 fast path within each chunk). GDN recurrent state threads across
@@ -9493,7 +9494,9 @@ kernel void gdn_chunk_norm_silu_c32(
                 let mut all_logits = Vec::with_capacity(n * vocab);
                 let mut start_pos = 0usize;
                 for chunk in token_ids.chunks(max_prefill) {
-                    all_logits.extend(self.forward_prefill_batched_chunk(chunk, start_pos, true));
+                    // Perplexity needs every chunk's per-position logits.
+                    all_logits
+                        .extend(self.forward_prefill_batched_chunk(chunk, start_pos, true, true));
                     start_pos += chunk.len();
                 }
                 all_logits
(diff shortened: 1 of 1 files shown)


ohdearquant:
`make bench-compare` (origin/main vs HEAD, default `elementwise_cpu_bench` + `simd` groups): no significant change — every group p > 0.05 except one borderline group at p = 0.05 / −1.4%, which is unrelated CPU-group variance from the two-worktree build (a known false-positive of the two-worktree A/B, disprovable by same-binary re-run). These default groups do not cover the Metal chunked-prefill path this PR touches; as noted in the description, there is no Criterion group that drives real multi-chunk `forward_prefill`, so the correctness guard (the new mutation-sensitive gpu-locked regression test) is the operative evidence, not an A/B number.

PR #709, opened 2026-07-07
feat(inference): group_size-specialized MAX_GRP for prefill attention occupancy

ohdearquant:
## Summary

Specializes the prefill attention kernel's threadgroup-memory footprint to each model's actual
group size (`num_attention_heads / num_key_value_heads`) instead of always allocating for the
fleet-maximum group size of 8.

The kernel `prefill_attention_batched_causal` sizes its threadgroup arrays by a compile-time
`MAX_GRP` constant, hardcoded to 8 so it could serve the largest group size in use. Models with a
smaller group size (e.g. group size 4) still paid for 8 rows of `q_s` / `score_s` / running
softmax state that were never populated. On Apple7/8 that inflated the kernel's threadgroup memory
past the ~16 KB co-residency threshold, forcing 1× occupancy where 2× was available.

## Change

- The kernel macro takes `MAX_GRP` as a parameter and is instantiated at 4, 6, and 8 (×
  f32/f16 KV) — six pipelines. The group-8 instantiation is byte-for-byte the previous kernel.
- At model load, the group size is known and fixed, so the correct variant is selected once (no
  per-dispatch branching) via `MAX_GRP = smallest of {4,6,8} ≥ group_size`.
- Output is **bit-identical** per model: the rows above the group size never held live data, so
  this is purely an occupancy/scheduling change with no arithmetic reordering.

Threadgroup memory by variant: group 4 → ~9 KB, group 6 → ~13 KB (both under the ~16 KB cliff →
2× occupancy), group 8 → ~17 KB (unchanged, 1×, unavoidable at that group size).

## Measurement (M2 Max, Qwen3.5-0.8B-Q4, idle GPU, same session)

Prefill TTFT through the actual production dispatch: `main` (group-8 kernel) vs this branch (which
auto-selects the group-4 variant for the 0.8B's group size of 4). Same harness, warmup + 3 measured
runs each, GPU serialized:

| prompt tokens | main / group-8 (ms) | this PR / group-4 (ms) | Δ |
|---:|---:|---:|---:|
| 2048 | 1204.3 | 1062.3 (std 1.9) | −11.8% |
| 4096 | 3179.2 | 2600.3 (std 3.5) | −18.2% |

The gain comes entirely from crossing back under the occupancy threshold; there is no change to
the computation. The branch numbers reproduce a raw MAX_GRP=4 kernel A/B within noise, confirming
the per-model selection adds no dispatch overhead.

## Correctness

- New GPU test loads the 0.8B model twice — once through normal production selection (group-4)
  and once with the prefill pipeline forced to the group-8 kernel — and compares `forward_prefill`
  on a fixed prompt: argmax and top-4 identical, `max_abs_diff = 0.0` (bit-identical, verified, not
  asserted).
- New unit test locks the group-size → variant mapping. Because the kernel's internal
  `group_size > MAX_GRP` check fails *closed* (writes no attention, silently) rather than
  panicking, selecting a variant below the model's group size would silently corrupt output. The
  mapping test guards exactly that: it must always round the group size **up**, never down.
- Group size 8 (e.g. the 2B model) selects the group-8 variant, which is the previous kernel
  unchanged — trivially bit-identical, a no-op control. A model whose Q4 checkpoint was available
  locally was used for the live parity leg; larger checkpoints not present locally are validated
  by the same code path in CI's e2e parity gate rather than asserted here.

`cargo clippy --workspace --features "f16,metal-gpu" -- -D warnings` clean (including `--tests`).

## Notes

Draft pending an independent review pass. The A/B above is the production dispatch measured this
session on an idle, serialized GPU.

### DIFF
--- a/crates/inference/src/forward/metal_qwen35.rs
+++ b/crates/inference/src/forward/metal_qwen35.rs
@@ -3302,7 +3302,10 @@ kernel void copy_kv_cache_batch_f16(
 // One threadgroup per (kv_head, query_token) processes all Q heads in the GQA group.
 //
 // KVT is the KV cache element type (float for f32 path, half for f16 path).
-#define DEFINE_PREFILL_ATTENTION_BATCHED_CAUSAL(NAME, KVT) \
+// GRP is the compile-time MAX_GRP specialization (4/6/8) — see the Rust-side
+// `prefill_maxgrp_suffix` helper for how a loaded model's group_size selects
+// among the g4/g6/g8 instantiations below.
+#define DEFINE_PREFILL_ATTENTION_BATCHED_CAUSAL(NAME, KVT, GRP) \
 kernel void NAME( \
     device const float* q          [[buffer(0)]], \
     device const KVT*   k_cache    [[buffer(1)]], \
@@ -3322,7 +3325,7 @@ kernel void NAME( \
     uint3 tgs3 [[threads_per_threadgroup]]) \
 { \
     constexpr uint HEAD_DIM    = 256; \
-    constexpr uint MAX_GRP     = 8; \
+    constexpr uint MAX_GRP     = GRP; \
     constexpr uint TILE_TOKENS = 256; \
     const uint lid = lid3.x; \
     const uint tgs = tgs3.x; \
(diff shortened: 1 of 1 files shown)


ohdearquant:
### Review status — ready to merge

**Codex round 2: APPROVE** (fix commit `1fe0a94b7`). The single round-1 medium is resolved: `prefill_maxgrp_suffix` now returns `Result` and fails closed (`Err`) on out-of-range group_size instead of silently rounding `>8` down to the g8 variant (which would hit the kernel's fail-closed `group_size > MAX_GRP` guard and write no attention). Both production call sites propagate with `?`; the doc no longer makes the false "stays correct if a caller skips validation" claim; the mapping unit test is now mutation-sensitive to both the round-down class (5→g6, 7→g8 boundaries) and the missing error path (0 and 9 assert `is_err()`).

**CI**: green at head `2df8e01` (up to date with main). The `q4 composed golden` leg on the first e2e-parity run was killed by a GitHub runner shutdown mid-job (`##[error]The runner has received a shutdown signal`) — an infra cancellation, not a correctness failure; the greedy-token `parity` gate itself passed, and main is green on that exact leg at the merged commit `4661d28`. Re-running the leg; auto-merge armed.

Local verify on the merged state: `cargo clippy --workspace --features "f16,metal-gpu" --tests -- -D warnings` clean; mapping test `prefill_maxgrp_suffix_rounds_up_never_down` passes.

PR #718, opened 2026-07-08
feat(inference): add query-likelihood reranking to the Qwen3.5 decoder

oceanwaves630:
## What

Adds `Qwen35Model::rerank(query, candidates) -> Vec<(usize, f32)>`: a query-likelihood reranker over the existing CPU log-likelihood path (`compute_token_nlls`). Each candidate is scored by the mean negative log-likelihood of the query tokens conditioned on the candidate prefix, and candidates come back sorted by descending relevance (higher = more relevant).

## Why

Reranking is a natural consumer of the decoder we already ship: score a small candidate set by how well each candidate predicts the query, with no new kernels. This wires the scoring path so a caller can reorder a retrieved candidate list by query-likelihood relevance.

## Design

- **Scoring**: for each candidate, tokenize `[candidate ++ query]`, run `compute_token_nlls`, and average the NLLs at the query-token positions. The query slice is `nlls[candidate_len - 1 .. candidate_len - 1 + query_len]` — the NLL at position `i` predicts token `i+1`, so the first query prediction is conditioned on the full candidate.
- **Fail-closed**: an empty candidate list returns `Ok(vec![])`; an empty query, an over-length query, or a candidate that reduces to empty after truncation returns a typed `InferenceError` naming the offending index. Over-length candidates are left-truncated so the query always stays inside the scoring window.
- **Deterministic**: a fresh KV/GDN state per call (order-independent), and a NaN-safe total order for ranking (NaN sorts last, ties broken by original index).
- **LoRA seam**: scoring flows through the model forward, so a composed adapter installed via `set_lora` is applied transparently — no rerank-side changes are needed when an adapter is present.

## Tests

11 tests: the query-position slice math against a direct-NLL reference, the fail-closed boundaries, left-truncation, NaN ordering, and an f16-gated checkpoint smoke + latency test.

## Honest limitations

- **Latency**: the CPU f32 path takes ~16s for 10 candidates on a real 0.8B checkpoint (release build). Fine for small candidate sets; a Metal-path rerank and/or shared-prefix reuse is a perf follow-up, not in this PR.
- **No `bench-compare`**: this adds a new method and touches no existing hot path, so there is no existing Criterion group to A/B against. Generation is unchanged, so `e2e-parity` is unaffected.
- Base-model reranking (no adapter) is a weak relevance signal on its own. The value lands when a task-specific adapter is composed through the seam; this PR is the plumbing.

### DIFF
--- a/crates/inference/src/model/qwen35/mod.rs
+++ b/crates/inference/src/model/qwen35/mod.rs
@@ -19,6 +19,7 @@ mod loading;
 mod model;
 mod moe;
 mod norm;
+mod rerank;
 mod sampling;
 pub(crate) mod stop_strings;
 mod weights;
--- a/crates/inference/src/model/qwen35/rerank.rs
+++ b/crates/inference/src/model/qwen35/rerank.rs
@@ -373,8 +373,9 @@ mod tests {
     /// sequence instead of just the query positions), the score computed for
     /// each candidate stops being "how well does this candidate predict the
     /// query" and the identity-token/random-model signal this test relies on
-    /// breaks — see `rerank_mutation_sensitive_to_query_slice_bounds` below,
-    /// which proves this by perturbing the same bounds directly.
+    /// breaks — see `rerank_score_matches_direct_query_position_nll_reference`,
+    /// which pins the exact contract by recomputing the query-position-only NLL
+    /// average independently and asserting equality.
     #[test]
     fn rerank_ranks_matching_candidate_first() {
         let model = build_model(test_config(), 0xA55A);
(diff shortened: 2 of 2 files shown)


ohdearquant (approved):
Rerank is a thin query-likelihood scorer composed over compute_token_nlls (score = negated mean per-token NLL); fail-closed, left-truncation preserves the query, NaN-safe ordering, and an independent per-position oracle test guards the slice. Rustdoc documents the within-call-only score comparability contract. CI green, branch up-to-date with main.

PR #720, opened 2026-07-08
docs(readme): fix stale version pins + add a Get-a-model step to CLI quick-start

oceanwaves630:
## What

Three funnel fixes to the top of the README, from a first-time-visitor read:

1. **Stale version pins.** The embeddings Quick Start and both GPU snippets pinned `lattice-embed = "0.4"`, but the published workspace is `0.5.0` and the crates.io badge on the same page shows current — the page disagreed with itself and a copy-paste installed a version behind. Bumped all three to `0.5`.

2. **Model-acquisition cliff.** Every CLI quick-start ends at `lattice chat --model ~/.lattice/models/qwen3.5-0.8b`, but the command that downloads a model to that path lived only in the Raspberry Pi subsection. A visitor on the primary macOS/Linux path ran chat, hit "no model at that path", and had no visible next step. Added a "Get a model" step to the CLI quick-start, before the first chat invocation.

3. **"No Python" clarity.** The new step notes that `huggingface_hub` is only the download transport, not a runtime dependency, so getting started doesn't read as contradicting the engine's no-Python positioning.

Plus a one-line release-checklist entry (`docs/RELEASE.md`) so the README pins track the tag on future minor releases.

## Scope

Docs only — no code, no hot path. `README.md` + `docs/RELEASE.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ All three can be used on their own.
 
 ```toml
 [dependencies]
-lattice-embed = "0.4"
+lattice-embed = "0.5"
 tokio = { version = "1", features = ["full"] }
 ```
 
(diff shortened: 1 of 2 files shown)


ohdearquant (approved):
Approved after reading the full diff myself — docs-only at this size doesn't need a codex leg, my read is the review. The Get-a-model step closes the flagship dead-end exactly as scoped, the transport-not-runtime sentence resolves the No-Python tension cleanly, and the RELEASE.md pin-check line is the right prevention for the pin drift. Good funnel work from forum finding to merged fix inside an hour.

Issue #171, opened 2026-05-31
perf(inference): lm_head two-stage block-top-k (eliminate 124K-threadgroup flood)


ohdearquant:
The `lm_head` GEMV (vocab 248,320) dispatches ~124K threadgroups and was the v0.3.0 regression culprit; it floods the Metal scheduler and drives most of the #77 contention loss. The sampler only needs argmax/top-k, not full logits.

### Tasks
- [ ] Stage 1: tiled vocab GEMV + **block-local** argmax (greedy) / top-k (≈970 tiles @ ROWS_PER_TG=256)
- [ ] Stage 2: global reduce of compact `(logit, token_id)` pairs → ~1K threadgroups total (≥100× reduction)
- [ ] exact top-k (local top-k of each tile contains any global top-k item); greedy fast path = block-argmax + global-argmax
- [ ] top-p: pragmatic top-k-first-then-softmax-over-K (gated, explicit) vs exact full-logit mode
- [ ] function constants for HIDDEN/ROWS_PER_TG/LOCAL_K

### Acceptance (ADR-064 gates)
- threadgroups ≤1,500 greedy / ≤2,000 top-k≤64; no full-logit allocation in production greedy path
- **greedy token agreement 100%** vs full-logit path on ≥10K positions; top-k set agreement 100%
- bench: `lm_head_full` vs `block_argmax` vs `block_topk_K{8,16,40,64}`, idle + load10
- expected: lm_head 1.1-1.8× faster idle, major contention-variance reduction

Ref: d5§3, #77. Gated by #167.

ohdearquant:
> _Comment authored by an AI agent on behalf of @ohdearquant._

Resolved on main by #559 with #628 hardening follow-up: lm_head_block_topk_f16/q4 two-stage block top-k pipelines are live (15 references in metal_qwen35.rs). Verified at source.

Issue #126, opened 2026-05-26
feat(inference): Metal Flash Attention for prefill path


ohdearquant:
## Summary

The current Metal prefill attention kernel tiles per KV-head only (one threadgroup per KV head). A proper FlashAttention-2-style kernel tiles across both Q and KV dimensions, keeping data in SRAM/L2 and dramatically reducing bandwidth for long prefill sequences.

## Evidence

- llama.cpp with --flash-attn beats MLX by 9 seconds at 8.5K context on M1 Max
- Draw Things Metal FA2: 43-120% latency reduction for image generation attention
- At context > 2K-4K, attention (not projection GEMMs) becomes the prefill bottleneck
- MLX effective throughput collapses to ~3 tok/s at 8.5K context without FA

## Current state

- metal_qwen35.rs fused_attention kernel: online-softmax, dispatches (kv_heads, 1, 1) threadgroups
- Works well for decode (M=1) via gemv_decode_m1 path
- For prefill with M=512: processes all seq_len rows per KV head — bandwidth-inefficient at long seq

## Scope

1. Implement FA2-style tiled kernel: tile Q dimension (BQ=64-128) and KV dimension (BKV=64-128)
2. Keep partial softmax accumulators in threadgroup memory
3. Handle causal masking within tiles
4. Integrate with both FlatKVCache and PagedKVCache read paths
5. Support GQA (grouped query attention) natively

## References

- FlashAttention-2 (Dao 2023)
- Draw Things Metal FA2 implementation (open source)
- MLX metal-flash-attention (mlx-mfa PyPI package)

ohdearquant:
**Scope update (still actionable, narrowed).** Verified against current `main` (2026-06-23): the batched prefill kernel `prefill_attention_batched_causal` (`crates/inference/src/forward/metal_qwen35.rs:2498`, landed via #228) now does **tiled online-softmax** over the KV dimension — `score_s[MAX_GRP * TILE_TOKENS]`, TILE_TOKENS=256, ~17KB threadgroup memory, running-max rescale. That covers the bandwidth concern for short-to-moderate prefill (no full score buffer; see #13, now closed).

[...]

PR #721, opened 2026-07-09
docs(adr): ADR-070 prefill GEMM optimization priority

oceanwaves630:
## What

Adds **ADR-070**, ranking the remaining Metal prefill performance levers by the engine's measured operator profile.

Prefill lags device-tuned engines at realistic context lengths, and that gap dominates end-to-end latency for agentic workloads (long prompt, short completion). Per-operator profiling shows **GEMM is 72–80% of prefill time**. By Amdahl's law that fixes the *operator* to attack: optimizing only the non-GEMM portion caps total speedup at ~1.3×, while optimizing GEMM caps at ~3.6× (Q4) / ~5.0× (f16) — so GEMM is the only operator whose measured share is large enough to attack the ~4× gap. The two closest GEMM tuning knobs are already resolved (double-buffering measured negative and closed; f32→half staging shipped at +29.6%) and the tile geometry already matches llama.cpp/MLX at BN=32. The ADR ranks from that evidence:

- **P1** — BN=64 GEMM N-tile widening: framed as the highest-priority *falsifiable experiment* on the dominant operator (not a measured-EV winner), with a BN=64-specific occupancy-headroom rationale and a falsifiable success criterion (≥+10% prefill @512, greedy parity, no PPL regression; a spill-induced flat result closes the lever).
- **P2** — dispatch-count reduction: norm+residual fusion (~48 dispatches/token removed), then wrapping the fused graph in an Indirect Command Buffer.
- **P3** — `lm_head` two-stage top-k reduction (#171).
- **Deferred** — Flash-Attention-2-on-Metal (#126), gated on a measured expected-value threshold (attention absolute time + plausible FA2 speedup vs the next-best GEMM/dispatch experiment), not scheduled speculatively.
- **Closed by measured evidence** — double-buffered threadgroup loads and GDN chunked-prefill tiling.
- **Out of scope (hardware exclusion, not a measured kill)** — tensor-core / non-`simdgroup_matrix` GEMM paths, reopen-able only on a future Apple family (Apple9 / M3+ cooperative-tensor) with a measured A/B.

## Structure

The Decision separates a **Measured evidence** table (each row tagged as a runtime measurement, an internal profiler result, or a source-level finding, with its most durable pointer — PR/commit/issue) from a **Prior / unvalidated-on-hardware** section. An external prior-art survey is in flight; its findings refine each priority's *implementation notes* and may reorder the GEMM *sub-levers*, but not the top-level ranking, which is fixed by the measured Amdahl profile. The kill records (double-buffering −20.7%, GDN tiling stage-2 −24%) are kept with numbers so closed levers are not re-tried.

## Scope

Docs only — a new file under `docs/adr/`. No code, no hot path.

### DIFF
--- a/docs/adr/ADR-070-prefill-gemm-optimization-priority.md
+++ b/docs/adr/ADR-070-prefill-gemm-optimization-priority.md
@@ -1,0 +1,132 @@
+# ADR-070: Prefill GEMM Optimization Priority (Metal)
+
+**Status**: Proposed
+**Date**: 2026-07-08
+**Crate**: lattice-inference
+
+> **Evidence basis**: The Decision below is grounded in this engine's own measured
+> profiling and A/B results (the **Measured** table). A parallel external prior-art
+> survey (llama.cpp `mul_mm` per-family tiling, MLX `steel_gemm`, Indirect Command
+> Buffers, Flash-Attention-2-on-Metal, roofline ceilings) is in flight; its findings
+> will refine the *implementation* of each priority below, not the *ordering*. The
+> ordering is fixed by Amdahl's law over the measured operator profile and does not
+> depend on the survey. Sections tagged **[prior, unvalidated on our hardware]** are
+> explicitly held apart so the survey folds in without a rewrite.
+
+## Context
+
+Prefill (processing the prompt before the first generated token) dominates
+end-to-end latency for agentic workloads (long prompt, short completion), and it
+lags llama.cpp and MLX by a wide margin at realistic context lengths. Decode is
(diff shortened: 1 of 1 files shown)


ohdearquant (approved):
Approved — signed ADR-070 content with codex R1-R3 fixes folded; re-read the corrected Amdahl framing at content level (GEMM-opt cap 3.6x Q4 / 5.0x f16, non-GEMM cap ~1.3-1.4x — now labeled correctly and it strengthens the GEMM-first argument). Decision order unchanged from what I signed. Auto-merge armed; P1 (BN=64) authorized to start per the arm agreement.

PR #722, opened 2026-07-09
docs(adr): ADR-070 fold — BN=64 GEMM measured-closed (M10), re-rank prefill levers

oceanwaves630:
## What

Follow-up to ADR-070 (#721): records the **BN=64 GEMM measurement**, re-ranks the remaining
prefill levers accordingly, folds the first tranche of the external prior-art survey, and fixes
a stale bench-harness reference.

## Changes

- **New measured row M10 — BN=64 Q4 GEMM N-tile widening: KILLED, −3.5%.** Widening the tiled
  Q4 GEMM's N-tile 32→64 (accumulators 8→16 per simdgroup) regressed prefill −3.5% @512 **and**
  @1024, re-measured independently (a second interleaved run agreed at −3.6%). Threadgroup memory
  was ample at 14.75 KiB, so this is **not** the M5 occupancy-memory cliff — the cause is
  register/accumulator pressure. Correctness bit-exact (greedy 706/706, PPL delta 0.0), which
  together with a consistent throughput drop confirms the wider kernel actually ran. This is the
  **second** tile-growth attempt to regress (after M5), so the productive direction is
  shrink/restructure, not grow.
- **Re-rank.** BN=64 was P1 (an unmeasured experiment); it is now closed by measurement, so it
  moves to the *Closed by measured evidence* bucket. The new **P1 is the GEMM structural
  sub-levers** (packed gate/up + packed QKV GEMMs, MLX-`steel_gemm`-style tile-family / split-K
  router) — GEMM stays the Amdahl-dominant prefill operator (M1: 72–80%), and with tile
  *geometry* now exhausted (M4/M5/M6/M10) the next GEMM experiment is structural, gated on a
  per-shape micro-bench. **Dispatch-count reduction becomes P2** (its prefill share is modest —
  the fixed dispatch cost amortizes across the prompt in one forward pass — but it is a ready,
  low-risk win that helps decode). `lm_head` two-stage stays P3; FA2 stays deferred.
- **Survey fold-in.** The external survey's first tranche independently reached the same verdicts
  on double-buffer (M5) and f32→half staging (M6) and the Amdahl profile, and had flagged BN=64's
  register-pressure kill condition as a live risk before it was measured — noted in the Closed
  bucket and the evidence-basis blockquote. Its remaining findings still fold into the
  per-priority implementation notes in place.
- **Stale reference fixed.** The prior P1 block cited a `bench_gemm_variants.rs` A/B harness that
  does not exist; that block is removed with the re-rank, so the ADR no longer references it.

## Structure preserved

The Measured table (each row a runtime measurement, internal profiler result, or source read,
with its durable pointer) stays separated from the Prior / unvalidated-on-hardware section. The
priority *ordering* remains fixed by the measured Amdahl operator profile, not by the survey.

## Scope

Docs only — one file under `docs/adr/`. No code, no hot path.

### DIFF
--- a/docs/adr/ADR-070-prefill-gemm-optimization-priority.md
+++ b/docs/adr/ADR-070-prefill-gemm-optimization-priority.md
@@ -7,9 +7,10 @@
 > **Evidence basis**: The Decision below is grounded in this engine's own measured
 > profiling and A/B results (the **Measured** table). A parallel external prior-art
 > survey (llama.cpp `mul_mm` per-family tiling, MLX `steel_gemm`, Indirect Command
-> Buffers, Flash-Attention-2-on-Metal, roofline ceilings) is in flight; its findings
-> will refine *how* each priority is implemented — and may reorder the GEMM *sub-levers* —
-> but not the top-level ranking. That the GEMM operator is attacked first is fixed by
+> Buffers, Flash-Attention-2-on-Metal, roofline ceilings) has partially landed — its first
+> tranche independently corroborated the M5/M6 GEMM results and the Amdahl profile; its
+> remaining findings will refine *how* each priority is implemented — and may reorder the
+> GEMM *sub-levers* — but not the top-level ranking. That the GEMM operator is attacked first is fixed by
 > Amdahl's law over the measured operator profile and does not depend on the survey.
 > Sections tagged **[prior, unvalidated on our hardware]** are explicitly held apart so
 > the survey folds in without a rewrite.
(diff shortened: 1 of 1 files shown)


ohdearquant (approved):
ADR fold verified against the measured evidence I ruled on: M10 row matches the live A/B record (-3.5% @512/@1024, bit-exact correctness proving the kernel ran, register-pressure attribution), P1 correctly re-ranked to structural sub-levers with per-shape micro-bench gates, closed-levers section and follow-ups consistent. Docs-only, codex R2 clean, CI green. Approved.

Issue #668, opened 2026-07-05
serve: relocate format_chat_template to a CPU-available module so the CPU serve path can drop render_prompt (#661 follow-up)


ohdearquant:
> _Issue authored by Claude (Anthropic agent) on behalf of @ohdearquant._

## Context
#667 unified the `lattice serve` Metal worker onto the engine's shared `format_chat_template`, resolving the Metal-path half of #661. The CPU/safetensors serve path in `crates/inference/src/bin/lattice.rs` still uses its own `render_prompt`, because `format_chat_template` (and the whole Metal worker) is `#[cfg(feature = "metal-gpu")]` — a CPU-only build cannot call it. `render_prompt` is currently pinned to the shared template by the `render_prompt_matches_shared_chat_template` anti-drift test, so the two renderers cannot silently diverge, but two renderers still exist.

## Remaining work
Relocate `format_chat_template` (and the `ChatMessage` formatting it depends on) into a module that is available in CPU-only builds, then have the CPU serve path render through the shared function and delete `render_prompt`. This removes the second renderer entirely; the `render_prompt_matches_shared_chat_template` anti-drift test then becomes unnecessary.

## Scope
Engine change — moving `format_chat_template` out of the `metal-gpu`-gated path. Deliberately deferred from #667, which was confined to `lattice.rs` and did not touch the engine's chat wrappers.

Issue #661, opened 2026-07-05
serve: unify 'lattice serve' Metal prompt rendering with the ChatMessage chat-template path


ohdearquant:
The `lattice serve` HTTP handler renders the message list into a ChatML prompt string with its own `render_prompt` (crates/inference/src/bin/lattice.rs:3135) and drives the Metal worker prompt-string-based, while `chat_metal` and `lattice_serve` go through `ChatMessage` + the engine's chat-completion wrappers (crates/inference/src/forward/metal_qwen35.rs:18218). Two independent renderers for the same template invite drift (template version bumps, im_end handling, tool-message formatting) and complicate cross-turn cache wiring, which plans reuse over the tokenized rendered prompt.

Proposed: move the `lattice serve` Metal path to `ChatMessage` + the engine chat-completion wrappers so both HTTP surfaces share one rendering implementation, then delete the bespoke renderer.

Out of scope for #462's serve cache wiring; filed as follow-up. Related: #619, #462.

ohdearquant:
> _Comment authored by Claude (Anthropic agent) on behalf of @ohdearquant._

Resolved by #667 (merged): the `lattice serve` Metal worker now renders prompts through the engine's shared `format_chat_template` on the `ChatMessage` path, so the Metal serve surface and the `chat_metal` / `lattice_serve` surfaces share one rendering implementation. Role/content validation is shared via `ValidatedRole::parse`, and an anti-drift test (`render_prompt_matches_shared_chat_template`) pins the surviving CPU renderer to the shared template.

The titled ask — unifying the Metal prompt rendering — is complete. One follow-on remains: the CPU/safetensors serve path still keeps its own `render_prompt` because `format_chat_template` is `metal-gpu`-gated and unavailable in CPU-only builds. Relocating `format_chat_template` to a CPU-available module so the CPU path can converge and `render_prompt` can be deleted is tracked in #668.

Closing as complete.

Issue #590, opened 2026-07-03
feat(inference): implement sparse-checkpoint replay for the cross-turn GDN cache (ReplayFromCheckpoint is a stub)


ohdearquant:
## Context

The cross-turn KV/GDN prefix-reuse planner (`crates/inference/src/kv_cache/cross_turn.rs`) defines three reuse modes in `PrefixReuseMode` (lines 77-87): `FullRefill`, `ExactAppend`, and `ReplayFromCheckpoint { checkpoint_len }`. Only the first two are reachable at runtime today.

`plan_prefix_reuse` (lines 119-179) already contains real logic for `ReplayFromCheckpoint` — lines 161-176 reuse state up to the highest exact checkpoint boundary at or below the shared token prefix — but the Metal runtime never gives it anything to find. `plan_cross_turn_reuse` in `crates/inference/src/forward/metal_qwen35.rs` (lines 16134-16144) always calls `plan_prefix_reuse(entry, metadata, prompt_ids, &[])` with an empty checkpoint slice, and its own doc comment explains why: "v1 never owns sparse GDN checkpoints, so it always plans with an empty checkpoint set — `PrefixReuseMode::ReplayFromCheckpoint` is therefore never produced by this call."

Even if a checkpoint were supplied, execution isn't implemented either: the `ReplayFromCheckpoint` arm of `restore_cross_turn_prefix` (lines 16151-16184) is a hard error:

```rust
PrefixReuseMode::ReplayFromCheckpoint { .. } => Err(InferenceError::PrefixCache(
    "restore_cross_turn_prefix: sparse GDN checkpoint replay is not implemented in v1".into(),
)),
```

Net effect: cache reuse only ever helps the one case where the new prompt is an exact append onto the single retained (latest) turn. Any conversation shape that shares a prefix with an *earlier* point than that — editing or regenerating an earlier turn, an agent branching to try an alternate continuation, or recovering after an interrupted turn — gets zero reuse and falls all the way back to `FullRefill` (full re-prefill), even though the planner is already written to handle it.

## Acceptance criteria

- [ ] Retain a small bounded ring of sparse GDN checkpoints per cross-turn slot (not just the single latest boundary), each tagged with its represented token length.
- [ ] Feed `plan_cross_turn_reuse` the slot's real checkpoint lengths instead of `&[]`.
[...]

Issue #491, opened 2026-06-30
bench(gdn-state): measure unified-memory bandwidth share of GDN-state vs KV vs weight traffic per decode step


ohdearquant:
# bench(gdn-state): measure unified-memory bandwidth share of GDN-state vs KV vs weight traffic per decode step

Before investing in recurrent-state compression (int4/int8/fp8 sweep, low-rank compression), establish
whether GDN state traffic is a meaningful fraction of per-token memory bandwidth at all, or noise next
to weight and KV traffic. This is the prerequisite measurement the rest of this family depends on.

## 1. User-visible outcome
None directly — this is a decision-gate measurement. It determines whether the int4/int8/fp8 state
precision work and low-rank state compression work (filed alongside this issue) are worth pursuing,
or whether effort should redirect to weight-bandwidth levers instead.

## 2. Prior art
No literature citation needed — this is an internal profiling question. GDN recurrence *compute* is
already known to be cheap (336us/step, the cheapest mixer measured) per prior internal benchmarking;
the open question is *bandwidth*, not compute. Builds on the byte-counter instrumentation proposed in
#422 (currently unimplemented).

## 3. Cheapest falsifying experiment
Extend #422's planned `bench_gdn_state` byte counters with a sweep against existing decode-step timing
instrumentation across context lengths (e.g. 512 / 2K / 8K) on Qwen3.5-0.8B. Report GDN-state bytes
moved per decode step as a percentage of total per-token memory traffic (weights + KV + state).

## 4. Promote threshold
State traffic share reaches low-double-digit percent or higher of per-token bandwidth at a realistic
context length -> justifies the state-precision and state-compression investigations.

## 5. Kill threshold
State traffic share stays single-digit percent at all measured context lengths -> deprioritize the
state-compression family; redirect effort to weight-bandwidth levers (#420, #421, #346, #423).

## 6. Where the code lives
`crates/inference` bench harness only (extension of the `bench_gdn_state` example proposed in #422).
No production inference-path changes.

## 7. Default-on guard
Not applicable — this is an offline benchmark binary, never reachable from `lattice_serve`.

## 8. Gate level
**G1** (measurement-primitive) — telemetry-only, zero output change.

---
Cross-refs: builds on #422 (byte counters), feeds the state-precision sweep and joint RAM-budget Pareto
issues filed alongside this one. Depends on nothing else in this batch.

ohdearquant:
> _Comment authored by an AI agent on behalf of @ohdearquant._

[...]

PR #723, opened 2026-07-09
docs(adr): ADR-071 — GDN recurrent state as a first-class serveable object (priority)

### LINKED ISSUE #462: Serve path re-prefills the entire conversation each turn (no cross-turn KV cache)

oceanwaves630:
## What

ADR-071 sequences the **#481 GDN-state family** (first-class serveable state object,
observability, snapshot/branch, compression) against **what the codebase already contains**.
Docs only — one new file under `docs/adr/`.

## Why this is a sequencing ADR, not a greenfield proposal

Source recon (verified against `origin/main`) found the machinery substantially shipped:

- **Cross-turn GDN checkpoint replay is merged** (issue #590 / PR #635, `53a726b14`, on `main`
  2026-07-04) and used by the CLI chat path.
- **Cross-turn prefix caching is already wired into both serve binaries** — `lattice_serve`
  (HTTP) via #662 and the `lattice serve` Metal worker via #666, both merged 2026-07-05.
  `lattice_serve.rs:989` calls `chat_completion_streaming_with_prefix_cache_and_cancel` on
  `CrossTurnSlotId::DEFAULT`. #462 is effectively delivered; its residual is the greedy-parity
  restore test (required to close it).
- **State-traffic observability is partially built** behind the `gdn-state-counters` feature;
  the counters are wired on the production decode path, and `examples/bench_gdn_state.rs` is the
  **#491 bandwidth-share decision gate** harness. So #491 is measure-and-record, not build.
- GPU state is not yet a first-class **type** (two parallel `Vec<Buffer>`) — that is the #481
  core refactor.

## Decision

- **P1** — run #491 and record the state-bandwidth fraction (near-free; the decision gate for
  all state-compression work; attach the measured f to the ADR as a new S-row same-day).
- **P2** — promote the GPU state to a first-class typed object (#481 core).
- **P3** — formalize observability modes (Off/Canary/Debug) on the typed object.
- **Serve residual (not a priority tier)** — cross-turn caching shipped (#662/#666); what
  remains under #462 is confirming/adding the greedy-parity restore test (the determinism
  guarantee ships as a test, not a prose claim; required to close #462). Renderer unification
  already merged (#667); only a separate minor CPU-path cleanup stays open (#668).

**Closed by arithmetic**: state compression for **decode speed** — roofline
`speedup <= 1/(1 - f + f/r)` caps it at <=8% even at f=10%, r=4, and the state fraction *shrinks*
with context (state traffic is config-fixed, unlike KV bytes/token). Storage/branch int8 is the
only surviving compression thesis and gates on #491. **Closed by measurement**: naive GDN
chunk-scan tile growth (the tiled-B64 variant measured -24%, #175).

## Structure

Source-verified / unit-test-pinned reality (e.g. decode state traffic = 40.4 MB/step, pinned by
`metal_qwen35.rs:28190`) is kept separate from the external prior-art survey, folded as
`[prior, unvalidated on our hardware]` data. The ordering is fixed by shipped-vs-not, not the survey.

## Review note

The first draft made serve-wiring P1 on a stale premise (a grep run against a stale local `main`);
codex review caught it, and this revision corrects the source claim, re-ranks accordingly, and
fixes a follow-on stale issue-ref (#661 was already merged via #667).

## Scope

Docs only. No code, no hot path.

### DIFF
--- a/docs/adr/ADR-071-gdn-state-object-priority.md
+++ b/docs/adr/ADR-071-gdn-state-object-priority.md
@@ -49,7 +49,7 @@ profiling with no merged artifact says so.
 | S1 | GDN state = **20,201,472 bytes (19.27 MiB)** active for Qwen3.5-0.8B (18 GDN layers × per-layer 1,122,304 B = conv 73,728 B + S-matrix 1,048,576 B). Fixed function of config. | unit-test-pinned | `metal_qwen35.rs:28190`; shape formula `:27938-27962` |
 | S2 | Per **decode step**, logical state traffic = **40,402,944 B (read+write, 20.2 MB each)** and is **context-invariant** — fixed by config, does not grow with context (unlike KV bytes/token). | unit-test-pinned + source-documented | `metal_qwen35.rs:28190-28191`; `examples/bench_gdn_state.rs:9-15` |
 | S3 | **Cross-turn GDN checkpoint replay is SHIPPED**: `ReplayFromCheckpoint` fully implemented via a `MetalGdnCheckpoint` ring (`CROSS_TURN_GDN_CHECKPOINT_CAP=3`), used by the CLI chat path. | source-verified (merged) | issue #590 / PR #635, commit `53a726b14` (on `main` 2026-07-04); `chat_metal.rs:426,:995` (call sites), `:488-496` (stats) |
-| S4 | **Cross-turn prefix caching is SHIPPED into both serve binaries**: `lattice_serve` (HTTP) via #662 and the `lattice serve` Metal worker via #666 (both merged 2026-07-05). `lattice_serve.rs:989` calls `chat_completion_streaming_with_prefix_cache_and_cancel` on `CrossTurnSlotId::DEFAULT`. #462 remains open only for the #661 renderer-unification tail and its greedy-parity acceptance test. | source-verified (merged) | #662, #666 (merged); `bin/lattice_serve.rs:989-990`; #462 (open, residual only) |
+| S4 | **Cross-turn prefix caching is SHIPPED into both serve binaries**: `lattice_serve` (HTTP) via #662 and the `lattice serve` Metal worker via #666 (both merged 2026-07-05). `lattice_serve.rs:989` calls `chat_completion_streaming_with_prefix_cache_and_cancel` on `CrossTurnSlotId::DEFAULT`. #462 remains open only for its greedy-parity / bit-exact acceptance test (the renderer unification asked for by #661 is already merged via #667). | source-verified (merged) | #662, #666 (merged); `bin/lattice_serve.rs:989-990`; #462 (open, residual only) |
 | S5 | GPU state is **not a first-class type** — two parallel `Vec<Buffer>` (`gdn_gpu_conv_bufs`, `gdn_gpu_s_matrices`) held by index discipline, no wrapping object/metadata. | source-read | #481 (tracking, open); `metal_qwen35.rs:1496-1497` |
 | S6 | **State-traffic observability is partially built** behind the `gdn-state-counters` feature: `GdnStateTrafficShape/Bucket/Report/Counters`, with the counter helpers **wired on the production decode path**, and `examples/bench_gdn_state.rs` documented as "the #491 bandwidth-share decision gate" harness. | source-verified | `Cargo.toml:36,188`; `metal_qwen35.rs:3579,3586` (helpers), `:5401,:5873,:5951` (decode-path attribution); `examples/bench_gdn_state.rs:9-15,196-203` |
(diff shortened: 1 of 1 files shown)


ohdearquant:
SPEC-GATE: SIGNED (Leo, Fable seat). This is the model form for a sequencing ADR: the Measured/source-verified table tags every row by epistemic status (runtime-measured / unit-test-pinned / source-read) with durable pointers, the recon corrected its own stale premise against merged code (S3: checkpoint replay shipped via #590/#635 — the harvest AND the seat's own 7/3 memory said otherwise), and the external survey's fraction estimate is correctly held apart from the repo-pinned numerator (S2) rather than laundered into a measured claim. The roofline closure of decode-speed compression is arithmetic-robust across the plausible f range and S2's context-invariance strengthens it with context growth.

Priority ordering approved as drafted: P1 #462 serve wiring (shipped engine, largest user-visible regression) → P2 typed state object → P3 run #491 → P4 observability modes. Closures recorded: decode-speed compression (roofline), low-rank/sketch (no need at KB scale), naive chunk-scan tile-growth (S9 measured −24%).

Two riders, binding:

R1: P3 (#491) is near-free and gates nothing upstream of it — run it EARLY, not strictly after P2. If it can run this week alongside P1, do that; attach the measured f to this ADR as a new S-row the same day it lands (the follow-up already promises this — the rider makes it same-day, so no downstream discussion argues from the survey's external estimate again).

R2: P1's bit-exact gate (restored logits bit-identical at a deterministic boundary) must appear as an explicit test in the #462 PR, not a claim in its body — same standard as the GPU-lock discipline, stated here so the impl PR review is mechanical.

Dependent implementation may proceed on merge, in P-order.

ohdearquant (approved):
Final read confirms the re-confirmed order landed verbatim: P1 = run #491 early/near-free with same-day S-row attachment (R1 folded), cross-turn caching correctly recorded as shipped in BOTH paths (independently verified at origin/main lattice_serve.rs:989 before granting the re-confirm). Codex r3 clean, docs-only CI green. Approved — auto-merge fires. Proceed: prune worktree, atlas enrich-ask, then B weight-quant ADR at the same bar.

Issue #703, opened 2026-07-07
feat(inference): online R3/R4 Hadamard rotation — recover QuaRot's symmetric-Q4 loss (offline v0 is net-negative)


ohdearquant:
## Motivation

Offline QuaRot v0 (R1/R2 rotations absorbed into weights at convert time) is currently a **net negative** versus plain asymmetric Q4 on the correct forward path.

Controlled measurement (Qwen3.5-0.8B, WikiText-2 raw test, window 512 / stride 256, seed `0xC0FFEE`):

| | Unrotated asym Q4 | QuaRot Q4 | Delta |
|---|---|---|---|
| Full corpus (310K tok) | **15.73** | 17.63 | **+1.90** (QuaRot worse) |
| 32K slice | 16.25 | 18.15 | +1.90 |

**Root cause.** The Hadamard rotation zero-centers weight distributions, which forces *symmetric* Q4 quantization. Symmetric Q4 has a worse per-block fidelity floor than the asymmetric Q4 we default to (measured relRMS ≈ 0.100 vs ≈ 0.080). Offline-only conversion has no runtime rotation to recover that loss, so on a numerically-correct forward path the rotation costs more than it saves.

**Historical note.** ADR-044 recorded QuaRot as a −1.61 PPL *win*. A code-bisection on identical corpus/slice/seed shows the delta sign is set by code version, not corpus: the ADR-044-era binary gives −0.81 (QuaRot better) and today's binary gives +1.90 (QuaRot worse) on the same file. That measurement predates the RoPE-pairing fix and six weeks of subsequent forward-path work; the forward fixes improved the unrotated baseline by 4.47 PPL but QuaRot by only 1.76. The May "win" was QuaRot compensating for forward-path bugs that have since been fixed. ADR-044's step-4 conclusion should carry a correction note.

## The missing mechanism: online R3/R4

The QuaRot paper defines four rotations. R1/R2 are offline (absorbed into weights — what v0 ships). **R3/R4 are applied online in the forward pass**: R4 is a fast Walsh–Hadamard transform before the down-projection (`down_proj`), R3 is a Hadamard in the attention path (before `o_proj` / on the value stream). These online rotations are what actually suppress activation outliers at inference time and let rotated low-bit quantization win — especially as model scale grows and activation outliers get worse.

Without R3/R4, offline v0 pays the symmetric-Q4 penalty with none of the online outlier-suppression upside.

## Scope (exploration)

[...]

Issue #684, opened 2026-07-05
FP8 weight support (E4M3): load and dequant non-expert weights


ohdearquant:
## Problem

lattice does not currently load FP8 weight tensors. `deepseek-ai/DeepSeek-V4-Flash-DSpark` stores its non-expert parameters (attention, router, embeddings) in FP8 (E4M3). Many recent large-model checkpoints also ship in FP8.

## Scope

- Add FP8 (E4M3, and E5M2 if present) tensor loading.
- Dequantize FP8 to the compute dtype on CPU and Metal.
- Wire it into the weight loader so an FP8 checkpoint loads and runs.

## Notes

Independently useful beyond this one model, since FP8 checkpoints are common. Part of the large-MoE offload epic.

Issue #683, opened 2026-07-05
FP4 block-format weight support (MXFP4 or NVFP4): dequant on CPU and Metal


ohdearquant:
## Problem

lattice supports integer Q4 quantization but not FP4 float block formats. Large-model release checkpoints increasingly ship expert weights in FP4. `deepseek-ai/DeepSeek-V4-Flash-DSpark` stores its MoE expert weights in FP4 (a 4-bit float format with a shared per-block scale, likely MXFP4 or NVFP4).

## Scope

- Identify the exact block format and scale layout used by the target checkpoint (block size, scale dtype, packing order).
- Add an FP4 dequantization path on CPU and Metal that decodes a block plus its scale to the compute dtype.
- Wire FP4 tensors into the weight loader so a checkpoint with FP4 tensors loads and runs.

## Notes

FP4 support is independently useful beyond this one model, since FP4 is becoming a common release format for large checkpoints. Part of the large-MoE offload epic.

Issue #616, opened 2026-07-03
quality(quant): stale perplexity.tsv contradicts ADR-044 (regenerate) + the QuaRot delta gate exists but is never run in CI


ohdearquant:
> _Issue authored by Claude (Anthropic agent) on behalf of @ohdearquant._

**Reframed after cross-checking ADR-044** (the original framing said QuaRot fails its acceptance gate — that is false on the authoritative number; see history below). QuaRot **passes** the ADR-044 gate on the real corpus: the post-4c full-corpus measurement (WikiText-2 raw test, 297,156 tokens, window 512 / stride 256) recorded in ADR-044 is QuaRot Q4 PPL 23.958276 vs unrotated Q4 PPL 25.572589, delta **−1.614** — QuaRot is materially better than naive Q4.

The two actual defects:

## 1. `docs/bench_results/perplexity.tsv` is a stale small-sample artifact that contradicts the ADR

The committed tsv shows `q4 19.266338` vs `q4-quarot 19.950512` at **2048 tokens** — below the old ~4K tokenizer truncation cap that ADR-044 step 4c explicitly fixed, so these rows predate the fix. On this stale sample the delta reads +0.684, which a naive reader (or a naive gate) would interpret as a QuaRot quality failure. The file also contains malformed hand-pasted stdout lines mixed into the tsv rows, and the lattice-vs-MLX rows use mismatched token counts (2048 vs 2041).

- [ ] Regenerate perplexity.tsv from the full-corpus harness run so the committed artifact matches ADR-044's canonical numbers (or delete it and point at the ADR)
- [ ] Same-token-count external baseline (MLX and/or llama.cpp) recorded alongside, replacing the mismatched rows

## 2. The delta gate exists but nothing runs it

`eval_perplexity` already implements the ADR-044 acceptance gate end to end (`--q4-dir` / `--quarot-q4-dir` / `--delta-threshold`, non-zero exit on `quarot − unrotated ≥ 0.5`). No workflow invokes it, so a real quantizer regression would land silently — the gate is decoration until CI (scheduled or engine-change-triggered) runs it and publishes the tsv artifact per run.

- [ ] A workflow runs `eval_perplexity` for q4 + quarot-q4 against a pinned corpus slice on the macOS runner, enforcing the delta gate fail-closed
- [ ] q8 row recorded alongside for trend data

## History

[...]

PR #724, opened 2026-07-09
docs(adr): ADR-072 — weight quantization priority (Metal)

oceanwaves630:

## What

ADR-072 ranks the **open weight-quantization levers** (#420 W3, #703 online-Hadamard, #683/#684
FP4/FP8, #421 SpinQuant, #423 policy) against what already ships. Docs only — one new file under
`docs/adr/`.

## Why this is a sequencing ADR, not a greenfield proposal

Source recon (verified against `origin/main`) found the core machinery shipped:

- **Q4** (group-32, v2 20 B/block = **5.0 effective bpw**, not 4.25–4.5) and **Q8** (per-row i8 on
  CPU; Q8_0 34 B/block on the Metal `gemm_q8_tiled` path) are shipped with Metal `gemm_q4_tiled` /
  `gemm_q8_tiled` kernels.
- **Offline QuaRot v0** (randomized Hadamard R1/R2) ships as a full pipeline wired into the serving
  path (`from_q4_dir` detects a QuaRot dir via the `quarot_seed` in `quantize_index.json`, not
  `ManifestFlavor::QuaRot`), with a CI composed-golden gate.
- **But offline QuaRot is a measured net-negative**: unrotated Q4 **16.589** vs QuaRot **19.007**
  PPL, delta **+2.418** (`docs/bench_results/perplexity.tsv`, #616-closed). The ADR-044-era "QuaRot
  better" was compensating for since-fixed RoPE bugs.
- The **Q4 PPL quality gate is armed and required** (golden 16.589, tol 0.05, `q4-ppl-quality` →
  `parity-gate.needs`).

## Decision

- **P1** — MLP-only W3 (#420). Highest leverage: decode is weight-bandwidth-bound and MLP GEMMs are
  ~35% of it; W3@group-32 ≈ 4.0 bpw vs Q4's 5.0 → ~20% MLP weight-traffic cut. Independent of the
  QuaRot net-negative. Validate via `Q3Block`@group-32 + the #616 `eval_perplexity` gate path + `make bench-compare`.
- **P2** — online R3/R4 Hadamard (#703). The lever that decides whether QuaRot is worth shipping at
  all; CPU/f16 equivalence gate first, lower-risk given the ADR-051 MTP runtime-rotation precedent.
- **P3** — FP8/FP4 dequant-on-read loaders (#684/#683). Mechanical, demand-gated; no compute payoff
  on Apple GPUs (no low-precision matrix unit).
- **Deferred** SpinQuant (#421) behind #703; **enabler** role-aware policy (#423) seeded by P1.
- **Closed by measurement**: offline QuaRot v0 as a quality tier (+2.418) — infrastructure stays
  shipped and gated, but it is not a recommended serving tier until #703 recovers it.

## Structure

Source/gate-pinned reality (e.g. the +2.418 QuaRot delta, the armed 16.589 gate) is kept separate
from the external prior-art survey, folded as `[prior, unvalidated on our hardware]`; the survey's
block-size arithmetic is corrected to the shipped group-32. Ordering is fixed by the measured decode
profile and the measured QuaRot result, not the survey.


ohdearquant (approved):
SPEC-GATE: SIGNED + approve press in one (codex r2 clean, gates green, content read end to end). This completes a trilogy of correctly-shaped sequencing ADRs: epistemic-tagged reality table, survey arithmetic corrected to the shipped group-32 format rather than laundered in, and the W4 net-negative honestly amending ADR-044 at priority level with the causal mechanism (old QuaRot win = compensation for since-fixed RoPE bugs) instead of quietly dropping it. P1 MLP-only W3 riding the unrotated baseline is right — it decouples the bandwidth win from the rotation question entirely. P2's CPU/f16-equivalence-before-Metal gate and P3's demand-gating are exactly the cheap-check-first discipline.

One binding rider: W7 (MLP ≈35% of decode) is internal profiling with no merged artifact and it is THE P1 leverage anchor — the first W3 PR must include the make bench-compare artifact as its public validation, and if the measured MLP share comes in materially below ~35%, the P1 rationale gets re-examined in that PR rather than ridden. Bench runs under the GPU flock (/tmp/lion-metal-gpu-test.lock), as always.

Dependent implementation may proceed on merge, in P-order.

Issue #252, opened 2026-06-23
test(inference): CI coverage for the f16 KV-cache Metal path (#238 follow-up)


ohdearquant:
## Context

PR #238 added an opt-in f16 KV cache for the Metal GQA decode path (`use_kv_f16` / `LATTICE_KV_F16`), **default-OFF**. Review verified the f16 path is correct (all 12 f32-pipeline bindings against the KV buffers are `if self.use_kv_f16` guarded; the MTP cache is a separate always-f32 allocation; `f32_to_f16` is correct round-to-nearest-even; measured Δ ≈ 4e-6 NLL is within f16-storage norms) and that it is byte-identical when off.

It flagged one gap: **no automated test exercises `use_kv_f16=true`.** A grep of `crates/inference/` finds no test setting `LATTICE_KV_F16` or `use_kv_f16: true`. The `e2e-parity` workflow builds the CPU path; `clippy` without `metal-gpu` never lints `metal_qwen35.rs`. So today the entire f16 path is verified only by a manual local run, not by anything reproducible in CI.

## Risk

Mergeable as-is because the path is default-OFF and byte-identical when off (worst case = a latent bug in a path no production caller enables). But before `use_kv_f16` is ever promoted to default or relied upon, it needs reproducible coverage — otherwise a future refactor of the Metal attention kernels could silently break the f16 branch with no signal.

## Proposed

Add a Metal-gated test (`#[cfg(all(target_os = "macos", feature = "metal-gpu"))]`, the same gating as the #234 bench harness so it only runs where Metal exists) that:

1. Runs greedy generation on a fixed prompt+model with `use_kv_f16 = false` and `use_kv_f16 = true`.
2. Asserts first-N greedy-token parity (the f16-vs-f32 KV difference should not change argmax for the first N tokens), or NLL within an f16-storage tolerance (~4e-6 observed, well under the 0.05 bf16/f16 literature bound from arxiv:2510.26788).

## Acceptance

- Test present and gated to the Metal feature.
- Passes on a Metal-capable runner.
- Wired so that promoting `use_kv_f16` to default (or a kernel refactor that breaks f16) would surface as a parity regression.

Source: review of the #234/#238/#245/#246/#248/#250 merge batch (2026-06-23). Follow-up to #238.

ohdearquant:
[...]

Issue #148, opened 2026-05-31
feat(kv_cache): f16 storage for PagedKVCache + Metal kernel updates


ohdearquant:
## Context

Follow-up from #121 (partially closed by PR #130).

PR #130 converted `FlatKVCache` to f16 storage. Two items from #121 remain:

1. **PagedKVCache f16** — `paged.rs` still stores `Vec<f32>` with `floats_per_page()` sizing for f32. Same f16 conversion pattern as FlatKVCache applies.
2. **Metal attention kernels** — `MetalKvCache` uses opaque `Buffer` objects. When Metal kernels read KV data, they should consume f16 directly instead of requiring f32 dequant.

## Scope

- Convert `PagedKVCache` internal storage from f32 to f16
- Update Metal attention kernels to read f16 KV buffers natively
- Bench: memory reduction + bandwidth improvement for PagedKVCache path

Depends on: Metal kernel infrastructure (#77 cluster)

ohdearquant:
**Progress (2026-06-23):** PR #238 added opt-in f16 KV (`use_kv_f16`/`LATTICE_KV_F16`, default-OFF) to the Metal GQA decode path (`MetalKvCache`). `PagedKVCache` (`kv_cache/paged.rs`) is still f32. Remaining: f16 storage for PagedKVCache + its Metal kernel updates.

ohdearquant:
Status update at `b3ce9df2ee`. Half of this landed; the half named in the title did not.

**Landed.** `FlatKVCache` stores f16 (`Vec<Vec<f16>>` for K and V, converting on write and read). `MetalKvCache` has an f16 allocation path — `MetalKvCache::new` takes `use_kv_f16` and allocates `half` Metal buffers when set. The Metal kernels have the matching f16 variants: `decode_attention_f16` and `decode_attention_flash_partial_f16` (`shaders/qwen35.metal:467`, `:600`), plus a narrowing copy kernel `copy_buf_offset_f16` (`:783`). Dispatch selects them: `dispatch_copy_offset_kv` and `dispatch_decode_attention` both branch on the f16 flag, and the batched prefill path uses the f16 variants too.

[...]

PR #725, opened 2026-07-09
docs(adr): ADR-073 — KV-cache quantization priority (Metal)

oceanwaves630:

## What

ADR-073 ranks the **open KV-cache quantization levers** (#154 f16 default-flip, #148 f16 paged-cache,
#118/#120/#122/#123 int8/int4 family) against what already ships. Docs only — one new file under
`docs/adr/`. Third sequencing ADR in the bundle set (siblings: ADR-071 GDN-state, ADR-072 weight-quant).

## Why this is a split verdict, not a clean ranking

Source recon (verified against `origin/main @ 454245434`) found this lane is **not** a single yes/no
like weight-quant was:

- **Rankable now on measured evidence**: f16 KV already shipped and was measured **decode-neutral**
  (PR #238, merged — 71.6 vs 72.8 tok/s @1k ctx, PPL Δ≈4e-6 on Qwen3.5-0.8B). It's a memory win, not
  a throughput one, because the model is **GDN-dominated** — only 6 of 24 layers carry a KV cache, so
  KV is **<1% of decode bandwidth**. PR #238's own words: *"The real decode lever is weight
  quantization, not KV dtype."*
- **Needs an experiment**: the int8/int4 family (#118/#120/#122/#123) has **zero lattice-measured**
  quality or throughput data. The central open question isn't just quality — it's whether KV bit-width
  moves decode throughput *at all* on a model whose KV is <1% of bandwidth.

## Prior-art refutation (the W4-amendment pattern)

The external survey's central premise — "KV ≈ 16% of decode bytes," which drives every expected-speedup
number — is **refuted**: it assumes a generic `Hkv=8, head_dim=128` shape. This engine is `Hkv=2`
(0.8B) / `Hkv=4` (27B), `head_dim=256`, 6-of-24 / 16-of-64 KV layers, measured <1% at 1k. Causal
mechanism (the GDN-hybrid architecture structurally caps KV's share) recorded at priority level.

## Decision

- **P1** — f16 KV default-flip (#154). Rankable now on the already-measured PR #238 result; only
  blocker is the CI parity-test gap (#252), an engineering completeness gate, not an empirical question.
- **P-low** — f16 for PagedKVCache (#148). Mechanical port of the proven pattern to the second cache;
  roadmap-gated (that engine serves zero production traffic).
- **P2 (gated)** — int8/int4 family behind a **minimal int8-K CPU experiment** + #491 (measure KV
  bandwidth share vs context). Quantize K-only/int8/full-attn-layers-only, CPU dequant-to-f32
  prototype, extend the existing `ppl_metal`/`eval_perplexity` harness. The experiment settles whether
  int8-K moves anything worth a fused Metal kernel before any kernel work starts.
- **Deferred**: int4 + WHT rotation (#123), fused kernel (#120), pre-RoPE K (#122) — all behind int8-K
  validation.

## Structure

Measured/source-verified reality (KV1–KV12, each tagged + `origin/main` pointer) is kept separate from
the external prior-art survey, folded as `[prior, unvalidated on our hardware]`. Ordering fixed by the
measured decode profile, not the survey. Apple-GPU no-low-precision-matrix-unit constraint carried
through (KV int8/int4 is storage/bandwidth only, no compute payoff — same as weights in ADR-072).

### DIFF
--- a/docs/adr/ADR-073-kv-cache-quantization-priority.md
+++ b/docs/adr/ADR-073-kv-cache-quantization-priority.md
@@ -107,8 +107,8 @@ gated behind an experiment because no lattice quality/throughput data exists for
 
 1. **P1 — f16 KV default-flip (#154).** Rankable on measured evidence exactly as ADR-072 ranked
    MLP-only W3 #1: on an already-measured, already-shipped-but-gated result. KV3 proved correctness
-   (Δ≈4e-6) and decode-neutrality, with a real memory win at long context (48→24 MiB @4096 on the
-   0.8B, 512→256 MiB on the 27B). The only blocker is KV5 (the CI parity-test gap) — an engineering
+   (Δ≈4e-6) and decode-neutrality, with a real memory win at long context (96→48 MiB @4096 on the
+   0.8B, 512→256 MiB on the 27B — the f32→f16 halving; int8 would halve the 0.8B figure again to 24 MiB). The only blocker is KV5 (the CI parity-test gap) — an engineering
    completeness gate, **not** an open empirical question; close it by extending #558's existing probe
    into the first-N greedy-parity test #252 originally specified, then flip `use_kv_f16` default to
    ON. Validate: the existing f32-vs-f16 A/B (`ppl_metal` / `bench_decode_ab`) already produces the
--- a/docs/adr/ADR-073-kv-cache-quantization-priority.md
+++ b/docs/adr/ADR-073-kv-cache-quantization-priority.md
@@ -1,0 +1,183 @@
+# ADR-073: KV-Cache Quantization Priority (Metal)
+
+**Status**: Proposed
+**Date**: 2026-07-09
+**Crate**: lattice-inference
+
+> **Evidence basis**: The Decision below is grounded in this engine's own shipped source and one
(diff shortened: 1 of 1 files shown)


ohdearquant (approved):
SPEC-GATE: SIGNED + approve press (codex r2 clean, content read end to end). The split-verdict structure is the honest shape for this evidence state: two levers ranked on measured data (P1 #154 default-flip riding KV3's measured neutrality+correctness; P-low #148 correctly roadmap-gated since the paged engine serves zero traffic), and the int8/int4 family gated behind a minimal experiment rather than ranked on data that does not exist. The refutation handling is the standard I asked repeated: the survey's 16%-KV premise refuted with the causal mechanism (GDN-hybrid caps KV to 6-of-24/16-of-64 layers, measured <1% at 1k) while honestly flagging the 8K-vs-1k mismatch and requiring re-derivation from our own long-context point rather than overclaiming. The P2 experiment scope (K-only, int8, full-attn layers only, CPU dequant shim, no rotation, no fused kernel) is exactly minimal, and naming #491 as prerequisite instrument ties the trilogy+1 together correctly. Riders inherited: GPU flock on all benches; the experiment's PPL number and the ≥1 long-context point are the public artifacts. Dependent work may proceed on merge, in P-order.

Issue #686, opened 2026-07-05
Epic: run large sparse MoE models that exceed RAM via expert offload (DeepSeek-V4-Flash-DSpark class)


ohdearquant:
## Goal

Track the engine mechanisms lattice needs to run a very large sparse Mixture-of-Experts model that does not fit in RAM, by streaming expert weights from disk. Motivating target: `deepseek-ai/DeepSeek-V4-Flash-DSpark`, roughly 284B total parameters with about 13B activated per token, shipped as FP4 expert weights plus FP8 non-expert weights, about 160 to 165GB on disk.

## Why this is tractable despite the size

Memory scales with the total parameter count, but compute (and the per-token working set) scales with the active count. Because only about 13B of 284B parameters are touched per token, the resident set can be kept far below the full 165GB: hold the attention, router, embeddings, norms, and KV cache in memory, plus a bounded cache of recently used experts, and stream the rest from disk on demand. The bottleneck becomes SSD bandwidth, not RAM capacity.

## Target hardware

Not near-term. The realistic first target is a 128GB-class Apple Silicon machine. On smaller machines (for example 32GB) even the always-resident non-expert weights do not fit, so this epic is groundwork for later, built incrementally.

## Sub-mechanisms

- [ ] MoE expert weight offloading: stream cold experts from disk, keep hot weights resident (#682)
- [ ] FP4 block-format weight support, MXFP4 or NVFP4 (#683)
- [ ] FP8 weight support, E4M3 (#684)
- [ ] Hybrid compressed attention, CSA plus HCA (#685)
- [ ] Million-token context KV management (gated on the attention design in #685)
- [ ] DeepSeek-V4 model config parsing and safetensors weight mapping, the loader

## Notes

Each of the offload, FP4, and FP8 pieces is independently useful for other large-model checkpoints, not only this one. The hybrid-attention piece is the least specified and depends on an architecture study that is in progress.

Issue #682, opened 2026-07-05
MoE expert weight offloading: run models larger than RAM by streaming cold experts from disk


ohdearquant:
## Problem

lattice currently loads all model weights into memory. That caps the largest model it can run at roughly the available RAM. Very large sparse MoE models (for example `deepseek-ai/DeepSeek-V4-Flash-DSpark`, about 284B total parameters, about 165GB on disk) exceed any consumer machine's memory, but activate only a small fraction of their experts per token, so most weights never need to be resident at once.

## Proposed mechanism

Run a model larger than RAM by streaming its expert weights from disk:

- Keep resident: attention, router, embeddings, norms, the KV cache, and a bounded LRU cache of recently used experts.
- Stream on demand: for each token, the router picks the top-k experts; load only those experts' weight tensors, from an mmap-backed or lazily read safetensors file, into the expert cache.
- Prefetch: kick off the expert loads as soon as the router decision is known, overlapping I/O with the resident-path compute.
- Budget: expose a configurable resident-memory cap for the expert cache.

The per-token working set is then on the order of the active-parameter size, so a 165GB model can run within a fraction of that resident.

## Scope

- An expert-weight residency manager (LRU cache plus a RAM budget).
- Lazy, mmap-backed expert tensor loading in the weight loader.
- Router-driven prefetch.
- Documentation of the expected tokens-per-second tradeoff as a function of SSD bandwidth.

Builds on the existing MoE routing path. Independently useful for any MoE that is larger than RAM. Part of the large-MoE offload epic.

ohdearquant:
This issue's mechanism is also exactly what's needed on the Q4-checkpoint loader path, not just the safetensors path — filing this as a cross-reference rather than a duplicate (dedup check against open issues before filing, per #876/#883's review).

#876 added a MoE FFN branch to `from_q4_dir` (`crates/inference/src/forward/metal_qwen35.rs`) so a Q4-quantized MoE checkpoint (`quantize_q4`, #874) can load at all. It currently uses the same "eager, fully-resident" design this issue is about to fix for the safetensors path: every layer's routed-expert tensors are mmap'd from their `.q4` file and immediately dequantized into a GPU-resident f16 buffer for the model's lifetime (`load_moe_ffn_q4` / `load_q4_mmap_dequant_f16`), with a cumulative-memory guard that fails closed rather than OOMing when the total exceeds `0.85 × recommendedMaxWorkingSetSize`.

[...]

PR #727, opened 2026-07-09
docs(adr): ADR-075 — MoE expert-offload priority (Metal)

oceanwaves630:

## ADR-075 — MoE expert-offload priority (Metal)

Fifth sequencing ADR in the bundle set, after ADR-071 (GDN-state), ADR-072 (weight-quant), ADR-073 (KV-quant), ADR-074 (MTP). Docs-only.

### Verdict: NEEDS-EXPERIMENT, reframed — gated, not ranked

The question as commissioned — *rank the open MoE expert-offload levers* — presupposes a working, measured, resident MoE serving path whose next optimization is offload. **That precondition is false on the current head**, and the reasons are the finding.

**Load-bearing facts (all verified against `origin/main @ c9ae21a9`):**

- Lattice ships exactly one MoE preset (`qwen36_35b_a3b`). The large model it actually **serves and has measured** (~4 tok/s) is `qwen36_27b`, which is **dense** — so no measured number this engine has ever produced touches MoE.
- The Metal MoE dispatch (ADR-053, merged PR #57) is wired into the live forward loop but has **zero test/bench/measurement** (`encode_moe_ffn` has one production call site, appears in no test), and runs **only at f16** — Q4/QuaRot, Q8-CPU, and Q8-NEON all three explicitly refuse MoE.
- The engine constructor takes a fully-loaded `&ModelWeights`, so the CPU loader materializes the full model as f32 (~120 GiB) before any Metal buffer is built — a **loader-sequencing ceiling** that makes the model not loadable on any consumer Apple Silicon today, independent of offload.

### Measured correction of a merged ADR (ADR-053)

ADR-053 quotes *"~15 GiB"* resident routed-expert weights and reasons its 32 GB feasibility from it — but it says *"At Q4 with the current config dimensions,"* and **Q4 was never implemented for MoE** (the three refusal points above). The only precision that runs is f16, so the real resident footprint is **60 GiB — 4× the ADR-053 figure**. This is recorded as an amendment here (ADR-053 stays an immutable record); its dispatch/routing design is unaffected, only its resident-memory arithmetic and the "borderline on 32 GB" claim are corrected (it's not borderline — it doesn't fit).

### Structure

A **Measured / source-verified reality** table (E1–E12, each tagged source-read / unit-test-pinned / runtime-measured with a durable pointer) plus a derived memory-arithmetic table, held apart from an **untrusted external prior-art survey** that is confirmed repo-blind (targets a nonexistent `crates/engine` crate and a GGUF loader lattice doesn't have) and scoped to a different model family (DeepSeek-V4-Flash). The survey's arithmetic is unranked until re-derived on lattice's real 6.0 MiB/expert f16 numbers; its one convergent conclusion (build and measure the resident dispatch before adding disk streaming) matches the source-read verdict.

### Decision

Gates the whole offload family (#682 the literal lever, #686 the epic, FP4 #683 / FP8 #684 adjacent) behind a minimal experiment:

- **Step 0:** confirm the loader-sequencing ceiling (instrument an actual MoE-preset load); if confirmed, fixing it (stream layer-by-layer) lowers the memory floor before any offload code.
- **Step 1:** drive a synthetic small MoE (4 layers, 16 experts) through the real `MetalQwen35Engine` for the **first-ever** measured MoE dispatch correctness + latency, under GPU flock, committed as an artifact.
- **Step 2:** only then re-derive the offload hit-rate arithmetic on real numbers, at which point #682's design becomes decidable.

Honest framing: this is **not** "REFUTED — wrong architecture" (MoE exists and is wired); it's "the lever is real, but the measurement floor beneath it does not exist yet." Includes a forward instruction (mirroring ADR-074's rider): when Step 1's bench lands, its numbers amend both this ADR's table and ADR-053's corrected figure with runtime-measured rows.

### Testing

Docs-only. No executable behavior changes.

### DIFF
--- a/docs/adr/ADR-075-moe-expert-offload-priority.md
+++ b/docs/adr/ADR-075-moe-expert-offload-priority.md
@@ -10,8 +10,9 @@
 > **NEEDS-EXPERIMENT, reframed**: the ADR-075 question as commissioned — *rank the open MoE
 > expert-offload levers* — presupposes a working, measured, resident MoE serving path whose next
 > optimization is offload. **That precondition is false on the current head.** The single MoE preset
-> is source-verified as not loadable on any consumer Apple Silicon machine today, its Metal dispatch
-> has zero test/bench/measurement, and it runs only at f16. This ADR also **records a measured
+> is source-verified as not loadable on the 32–128 GB M-series/M-Max machines today (loadability on
+> the 192–512 GB Ultra-class desktops is unproven pending measurement), its Metal dispatch has zero
+> test/bench/measurement, and it runs only at f16. This ADR also **records a source-derived
 > correction to a merged ADR**: ADR-053's headline "~15 GiB resident expert weights" figure is off by
 > 4× — the real resident footprint is **60 GiB at f16** — because the Q4 basis it assumed was never
 > implemented for MoE. A parallel external prior-art survey is folded as **[prior, unvalidated on our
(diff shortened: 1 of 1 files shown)


ohdearquant (approved):
SPEC-GATE: SIGNED (Leo). The reframe is the honest verdict: 'gated, not ranked' is correct when the resident path beneath every offload lever has never produced a token, and refusing to rank on a repo-blind survey scoped to a different model family is exactly right. The ADR-053 amendment is the house pattern done well — immutable record untouched, only the memory arithmetic and 32GB-feasibility corrected, with the causal mechanism (ADR-044 deferral enforced in the converter, no MoE quantizer ever landed) rather than a bare number. The codex-narrowed loadability claim (not loadable on 32-128 GB M-series/M-Max; Ultra-class unproven pending Step 0) survives my read at all three sites. Rider: Step 0's result lands as a new S-row on this ADR whichever way it goes — if from_safetensors turns out to stream internally, the 120 GiB peak claim gets corrected the same way this ADR corrected ADR-053. Proceed to G on merge.

Issue #59, opened 2026-05-23
Stable rerank API on lattice-inference with LoraHook injection


ohdearquant:
khive ADR-042 (Local re-rank via lattice-inference) consumes lattice-inference for cross-encoder rerank. Reading the current crate (`crates/inference/src/model/cross_encoder.rs`):

`CrossEncoderModel::score_batch(query: &str, candidates: &[&str]) -> Vec<f32>` already exists and is publicly re-exported via `crates/inference/src/lib.rs`. This covers the scoring surface — good.

**The genuine gap** is hook injection into the cross-encoder's forward pass. `score` and `score_batch` accept no `LoraHook` argument. khive needs the cross-encoder's underlying BERT (or equivalent) forward pass to honor a `LoraHook` so the brain pack can swap in profile-resolved adapters per call.

Concretely khive wants (any equivalent shape is fine — these are the consumer-side constraints, not a proposal of the exact public API):

1. **A way to inject `LoraHook` into the cross-encoder's BERT forward.** Either via a new method on `CrossEncoderModel` that takes `Option<&dyn LoraHook>`, or via a per-call setter on the underlying model.
2. **A way to dispatch by `RegisteredModel.id: Uuid`** (lattice ADR-029) — so khive can swap rerank model variants by registry id without rebuilding the binary. If `lattice-inference` doesn't own a registry-dispatch layer, that's khive's concern; please confirm so we wire it in `khive-pack-brain`.
3. **Latency budget**: ≤50ms / 32 (query, candidate) pairs on GPU, ≤200ms on CPU. Batching internal to lattice. khive submits one call per recall.

What khive will own on its side (do NOT add to lattice):
- An async wrapper if `score_batch` stays sync (khive runtime is tokio).
- A `RerankInput` struct or equivalent input shape — that's khive's call site preference, not a lattice API concern.

Open question: is the BertModel that backs `CrossEncoderModel` LoRA-injectable today? If yes, can `score_batch` be extended (or a new `score_batch_with_hook`) without breaking the existing API? If no, what's the design path for hook-aware cross-encoder inference?

This is the deepest unknown in khive ADR-042 — wiring this resolves the implementation path.

References:
- khive ADR-042 §1 (rerank stage), §3 (latency budget), §4 (hook resolution)
- lattice ADR-008 (LoraHook trait shape)
- `crates/inference/src/model/cross_encoder.rs` lines 50–73

PR #193, opened 2026-06-21
feat: LoRA backward training (GQA+GDN) + LatticeStudio macOS instrument app

ohdearquant:
## Summary

Two threads developed together on this branch:

1. **LoRA backward training engine** (`crates/inference`, `crates/tune`) — extends LoRA fine-tuning from a single lm_head trainer to a full-depth multi-layer backward tape through Qwen3.5's gated GQA **and** GatedDeltaNet layers, with `train_grad_full --json` metrics + PEFT adapter save.
2. **LatticeStudio macOS instrument app** (`apps/macos`) — a native SwiftUI panel that drives the lattice CLI binaries (train / quantize / generate) over a line-delimited `@@lattice {json}` event protocol. No in-process ML, no Python.

63 files, +18,547 / −52.

## Thread 1 — LoRA backward (inference + tune)

- **Surface-A (verified):** exact CPU backward through the 6 GQA layers + lm_head LoRA trainer. Materialised GQA forward+backward corrected for Qwen3.5 gated attention (gate + interleave + shifted q-norm).
- **RMSNorm correctness (verified):** backward tape and VJP now use shifted-gamma RMSNorm to match the real Qwen3.5 weights (`84029ae8`). A real-model forward-parity gate (`6d2da6e2`) diffs a loaded layer-23 against the trainer's forward — **TEST4 max abs 4.77e-6 < 1e-3, passing.**
- **Surface-B (UNVERIFIED — no correctness claim):** `f34ed3b1` adds LoRA weight gradients for the 18 GatedDeltaNet layers (the 5 GDN projections in_proj_qkv/z/b/a + out_proj). It is Option-gated to be **byte-identical to the base forward** when any param is None or rank==0, so it does not perturb inference. Weight-grad correctness is **pending a gradient-correctness verification run** (heavy real-model `train_grad_full --gradcheck` on Qwen3.5-0.8B + the new synthetic single/multi-head unit gradchecks). This PR makes no correctness claim for surface-B.
- **Streaming generation:** UTF-8-safe incremental detok (`ac223c44`) — flushes only complete codepoints per token so CJK/emoji don't stream as replacement chars.

## Thread 2 — LatticeStudio (apps/macos)

[...]

PR #732, opened 2026-07-09
docs(adr): ADR-079 adaptive micro-LoRA brain (composed train-govern-compose-route-consume loop)

oceanwaves630:
## Summary

ADR-079 is the synthesizing record for the composed multi-adapter LoRA system built one stage at a time across six independently-merged PRs. No prior document maps the loop as one system: seven existing LoRA ADRs each cover a single-adapter lifecycle slice or an orthogonal composition axis, and ADR-056's own appendix admits its designed training pipeline was never built.

The ADR names shipped / partial / gap per stage of the **TRAIN → GOVERN → COMPOSE → ROUTE → CONSUME** loop, every load-bearing row anchored to a merge commit + `origin/main` file:line.

### Composing PRs
- **#191** exact-gradient CPU backward trainer (TRAIN)
- **#444** governed manifest + fail-closed loader (GOVERN)
- **#443** decode-time weighted mixture blend, zero kernel change (COMPOSE)
- **#448** + **#453** RLOO + Fisher online router-refit (ROUTE)
- **#718** causal-LM query-likelihood rerank; ADR-057-D1 cross-encoder rerank (CONSUME)

### Two record corrections
1. The merged CPU backward trainer is **#191** (`7b46f6e77`), **not** commit `f34ed3b1`. The "surface-B" GDN LoRA weight-grad commits are not ancestors of `origin/main`: `f34ed3b1` was in the now-closed **PR #193**, and the current standalone **PR #202** (different head commits) is still OPEN — neither is on head. So today only GQA `q_proj`/`v_proj` adapters train; GDN LoRA weight-gradients do not exist in mergeable code.
2. **#718 is unrelated to #59.** #718 is the causal-LM reranker; #59 (cross-encoder `LoraHook` injection) was addressed by **PR #65** five weeks earlier. Two independent rerank-via-seam consumers now exist; the ADR names them as siblings.

### Named gaps (each with a re-entry trigger)
G1 GDN LoRA weight-grads (#202), G2 in-process route→refit→route loop-closure, G3 EWC anchor-pullback as a live anti-forgetting mode, G4 rerank-consumer disambiguation.

## Scope / review

- **Docs-only.** No code, no bench, no parity impact.
- Every reality row is source-read or PR/issue-state against `origin/main @ b0604722d`; the file:line anchors and the two corrections are the load-bearing claims to audit.
- **INDEX note:** this PR adds the ADR-079 row. ADR-071..078 are absent from `INDEX.md` (pre-existing — the bundle-sequencing ADRs were never registered); left for a separate housekeeping pass rather than scope-creeping this docs PR.


ohdearquant (approved):
SPEC GATE SIGNED: ADR-079. Composed-loop map with file:line pins at b0604722d, two load-bearing record corrections (GDN grads not on main — PR #202 open; #718 does not close #59 — that was #65), four named gaps each with re-entry triggers, S-row rider. Canonical map for Lane 1; supersedes nothing.

Issue #782, opened 2026-07-10
refactor(inference): C2 — shared HTTP serving module for lattice.rs/lattice_serve.rs (ADR-080 cluster C2 checklist)


ohdearquant:
## Context

ADR-080 (`docs/adr/ADR-080-consolidation-duplicated-contracts.md`, merged via #778) cluster **C2**: a shared serving module for the two OpenAI-compatible HTTP binaries. Sequenced second (after C1 #780, before C3/C4).

## Contract (per ADR-080)

`lattice.rs` and `lattice_serve.rs` stop maintaining independent copies of router construction, request normalization/validation, and response termination serialization. One shared serving module inside `crates/inference/src/bin/` (or a private `serve` sub-module of the crate — NOT a new crate) owns those; each binary keeps only backend-specific generation wiring (CPU/Metal dispatch vs daemon/queue model).

## Checklist

- [ ] Extract shared module: router construction + endpoint exposure (incl. making `/v1/models` uniformly available — today only `lattice_serve.rs` installs it)
- [ ] Shared OpenAI chat-request normalization + token-budget validation (extends the #656 unification slice)
- [ ] Shared completion-termination (`finish_reason`) serialization carrying the ENGINE-reported stop cause — fixes `lattice_serve.rs::spawn_worker` structurally discarding `stopped`/`stop_reason` (stop-string and EOS both currently serialize as `"stop"`) (#746)
- [ ] Disconnect cancellation as a CPU-generation-aware shared API so `lattice.rs` adopts the cancellation contract `lattice_serve.rs` already has (its own source comment documents it keeps generating after client disconnect) (#744)
- [ ] `max_tokens=0` fail-closed handling unified (#745, scope as narrowed during audit triage)
- [ ] Unified endpoint and error contracts across both binaries (the held finding this cluster resolves)

## Completion gates

- [ ] HTTP-level tests covering: request validation parity between binaries, engine stop-cause round-trip in both SSE and JSON, cancellation on disconnect for both binaries
- [ ] Mutation-sensitive: each behavioral fix's test fails with the fix reverted
- [ ] `make bench-compare` (binaries only — expect no kernel deltas; state so)
- [ ] Sibling grep: no remaining independent request-normalization or finish_reason serialization in either binary
- [ ] e2e-parity green

## Provenance

Evidence re-verified at 0699e60cc (ADR-080 C2). Related: #744, #745, #746, #656, ADR-080 (#778).

ohdearquant:
[...]

Issue #746, opened 2026-07-09
bug(inference): standalone serve worker hardcodes finish_reason "stop" regardless of actual stop cause


ohdearquant:
**Defect**: The standalone serve worker's `generate` closure returns only `(prompt_tokens, completion_tokens)` — no `stopped`/`stop_reason` information — so `Ev::Done` structurally carries no stop-reason field. Both the SSE terminal chunk and the non-streaming JSON body hard-code `"finish_reason": "stop"` unconditionally, so a generation that exhausts `max_tokens` without hitting EOS/stop is reported as `"stop"` on the standalone server, versus `"length"` on the unified server (whose `finish_reason_for` returns `"stop"` only when `GenerateOutput.stopped` is actually true).

**Evidence**:
- `crates/inference/src/bin/lattice_serve.rs:1028-1032` — `generate` closure signature returns `Result<(usize, usize), String>`, no stop information.
- `crates/inference/src/bin/lattice_serve.rs:1060-1069` — `Ok((prompt_tokens, completion_tokens)) => { let _ = job.tx.send(Ev::Done { prompt_tokens, completion_tokens }); }` — no stop-reason field.
- `crates/inference/src/bin/lattice_serve.rs:1354`, `:1406`, `:1493` — SSE terminal chunk and non-streaming JSON body all hard-code `"finish_reason": "stop"`.
- Contrast: `crates/inference/src/bin/lattice.rs:3183` — `finish_reason_for` returns `"stop"` only when `GenerateOutput.stopped` is true.

**Blast radius**: Every response from the standalone serve worker — a length-exhausted generation is misreported as `"stop"`, which affects client retry and truncation-detection logic.

**Suggested fix**: Propagate the real stop reason (EOS/length/stop-string) through `Ev::Done` and call a shared finish-reason serializer used by both binaries instead of hard-coding the value in the standalone worker.

Provenance: Audited at 13c8de8a3; re-verified at 0699e60cc.

ohdearquant:
Resolved by ADR-080 C2: PR #786 adds shared finish-reason mapping — the standalone serve worker now reports the actual stop cause instead of hardcoded "stop". Verified at origin/main @ 32697ed0e.

Issue #745, opened 2026-07-09
bug(inference): lattice_serve silently accepts max_tokens=0 instead of rejecting it


ohdearquant:
**Defect**: The standalone serve binary derives `max_new_tokens` from the request and clamps it against the model's context window, but never rejects a client-supplied `max_tokens: 0`. `.min(model_max_context.saturating_sub(1))` leaves a `0` value unchanged, so a zero-budget request is silently accepted and (presumably) produces zero completion tokens instead of a rejection, unlike the unified server's `validate_max_tokens`, which explicitly rejects `effective == 0` with an `invalid_max_tokens` error.

(Field parsing, `max_completion_tokens` support, and prompt-plus-completion context preflight in this binary have since been fixed elsewhere — this issue tracks only the remaining `max_tokens == 0` gap.)

**Evidence**:
- `crates/inference/src/bin/lattice_serve.rs:841-848` (`build_cfg`):
  ```
  let max_new_tokens = req.max_tokens.or(req.max_completion_tokens)
      .unwrap_or(d.max_tokens)
      .min(model_max_context.saturating_sub(1));
  ```
  No `== 0` rejection exists anywhere in this file.
- `crates/inference/src/bin/lattice.rs:2898-2930` (`validate_max_tokens`) — explicitly rejects `effective == 0` with `invalid_max_tokens`, unchanged.

**Blast radius**: Every chat-completions request to the standalone server with `max_tokens: 0` (or an equivalent zero-budget alias) — silently clamped through instead of rejected, unlike the unified server's contract for the same input.

**Suggested fix**: Add the same `== 0` rejection to the standalone server's request validation, ideally by sharing the unified server's `validate_max_tokens` logic rather than maintaining a second copy.

Provenance: Audited at 13c8de8a3; re-verified at 0699e60cc.

ohdearquant:
Resolved by ADR-080 C2: PR #786 adds shared zero-token rejection — max_tokens=0 is now rejected by both binaries via the shared serve contract. Verified at origin/main @ 32697ed0e.

Issue #744, opened 2026-07-09
bug(inference): unified server streaming ignores client disconnect, keeps generating


ohdearquant:
**Defect**: The unified server's CPU streaming callback discards `unbounded_send` failures and has no way to signal a stop, and the unified Metal path uses the callback-only `generate_streaming` API — neither backend observes a client disconnect and both continue generating to the token cap after the client goes away. The standalone worker binary, by contrast, checks a `CancelOnDrop` watch state and threads `should_cancel` through generation, stopping promptly at its documented checkpoints.

**Evidence**:
- `crates/inference/src/bin/lattice.rs:3519-3522` — comment explicitly documents the gap: "There is no true backpressure ... if the client disconnects mid-stream the producer keeps generating to the cap ... Per-token backpressure / disconnect-cancellation is a future refinement." The CPU branch closure still returns `()` and the Metal branch still calls `generate_streaming` (not a cancel-aware variant) at the same call sites.
- `crates/inference/src/bin/lattice_serve.rs:303` — the standalone worker checks `CancelOnDrop`/`should_cancel` state and threads it through generation.

**Blast radius**: Every streaming request to the unified server (`lattice` binary) — CPU requests run to the token cap after disconnect, and a long unified-Metal prefill cannot observe cancellation either.

**Suggested fix**: Route both binaries through a cancellable generation adapter, and add an equivalent cancellation-aware API for the CPU generation path so it can observe the same disconnect signal the standalone worker already handles.

Provenance: Audited at 13c8de8a3; re-verified at 0699e60cc.

ohdearquant:
Resolved by ADR-080 C2: PR #786 (shared HTTP serving contract) adds shared client-disconnect cancellation consumed by both lattice and lattice_serve streaming paths. Verified at origin/main @ 32697ed0e.

PR #786, opened 2026-07-10
refactor(inference): shared HTTP serving contract for lattice + lattice_serve (ADR-080 C2)

ohdearquant:
## Summary

Extracts the request/response contract shared by `lattice.rs` and `lattice_serve.rs` into
`crates/inference/src/serve/` (ADR-080 cluster C2), and fixes the three real bugs that
duplication had let drift apart:

- **#746** — `lattice_serve.rs` discarded the engine's actual stop cause and hardcoded
  `finish_reason: "stop"` in both the SSE and JSON responses, so a length-capped completion
  was misreported as an explicit stop. Root cause went one level deeper than the binary: the
  Metal engine's `ChatCompletionOutput` type (`crates/inference/src/forward/metal_qwen35.rs`)
  had no stop-cause field at all. Added `stopped: bool` to the struct and threaded it through
  all three construction sites, then wired both binaries onto the single shared
  `lattice_inference::serve::finish_reason(stopped)` mapping.
- **#744** — `lattice_serve.rs` had disconnect cancellation; `lattice.rs`'s CPU streaming path
  did not (its own comment called this "a future refinement") and kept generating to the token
  cap after the client left. Added a cancellation-aware `generate_streaming_with_cancel` to the
  CPU model (mirroring the Metal-side pattern that already existed), and hoisted the
  `CancelOnDrop` guard / `cancel_pair()` constructor into the shared module so both binaries'
  disconnect contract is one definition, not two.
- **#745** — `lattice_serve.rs` silently accepted `max_tokens: 0` and clamped it through to a
  zero-budget generation. `lattice_inference::serve::reject_zero_max_tokens` now rejects it with
  HTTP 400 before the resolved value reaches `GenerateConfig`, scoped narrowly to the zero case
  only — the two binaries' over-cap policies remain intentionally different (`lattice.rs`
  rejects over-cap requests; `lattice_serve.rs` clamps them to the model's context window) and
  are not unified by this change.

Each binary keeps its own router wiring and backend-specific generation dispatch (CPU/Metal
selection for `lattice.rs`, the daemon job queue for `lattice_serve.rs`); only the request/
response CONTRACT moved into the shared module.

Part of #782.

## Shared module

`crates/inference/src/serve/mod.rs` (new, gated behind the existing `serve` feature):

| Item | Purpose |
|---|---|
| `REQUEST_BODY_LIMIT_BYTES` | 1 MiB request body cap, shared constant |
| `ApiError` + `IntoResponse` | OpenAI error envelope `{"error":{"message","type","code","param"}}` |
| `finish_reason(stopped: bool)` | `"stop"` / `"length"` mapping from the engine's stop cause |
| `reject_zero_max_tokens(effective: usize)` | HTTP 400 on `max_tokens == 0` (#745) |
| `models_list_body(model_id, created)` | `GET /v1/models` response shape |
| `CancelOnDrop` / `cancel_pair()` | disconnect-cancellation guard + receiver pair (#744) |

## Per-fix verification (mutation-sensitive)

Each fix's test was confirmed to fail when the fix is reverted, then restored:

| Fix | Test | Reverted → | Restored → |
|---|---|---|---|
| #746 finish_reason (non-streaming) | `chat_completions_non_streaming_finish_reason_length_when_not_stopped` | `"stop"` (wrong) | `"length"` (correct) |
| #746 finish_reason (SSE) | `chat_completions_streaming_finish_reason_length_when_not_stopped` | hardcoded `"stop"` | `"length"` |
| #745 max_tokens=0 | `chat_completions_max_tokens_zero_400` | HTTP 500 | HTTP 400 |
| #744 cancel checkpoint (pre-prefill, CPU) | `generate_streaming_with_cancel_true_before_prefill_returns_interrupt` | `Some(Length)` | `Some(Interrupt)` |
| #744 cancel checkpoint (decode loop, CPU) | `generate_streaming_with_cancel_mid_decode_stops_early_fast_path` | ran to completion | stopped early |

## Sibling-invocation-path grep (post-extraction)

Grepped both binaries for remaining independent copies of the contracts this PR centralizes:

- `finish_reason`/`"stop"`/`"length"` literals: only intentional residue remains —
  `lattice_serve.rs`'s `Ev::Failed` arm and `lattice.rs`'s `StreamMsg::Failed` arm both still
  hardcode `"stop"` for the generation-*failure* path (HTTP 500, not a completion), matching
  each other and pre-existing; not part of this contract.
- `.route(...)` construction: both binaries independently wire their own `Router` (by design —
  only the contract is shared, not router construction itself); route paths/methods now match
  1:1 (`/health`, `/v1/models`, `/v1/chat/completions`).
- `ApiError`/`ErrorBody`/`ErrorDetail`: zero remaining local definitions in either binary.
- **Found and fixed as part of this PR** (not left as residue): both binaries had their own
  private `REQUEST_BODY_LIMIT_BYTES` constant, and `lattice_serve.rs` had kept a private
  `CancelOnDrop` struct + built its cancel channel via a bare `watch::channel(false)` instead of
  `serve::cancel_pair()`. Both are now single-sourced from the shared module.
- `lattice.rs`'s CPU streaming glue (`chat_completions`'s handler wiring the `cancel_guard`
  into the SSE stream's `flat_map` closure) has no dedicated HTTP-level disconnect test — only
  the underlying `generate_streaming_with_cancel` engine primitive and the shared
  `cancel_pair`/`CancelOnDrop` are unit/mutation-tested. The wiring itself relies on standard
  Rust closure-capture drop semantics and was assessed as low-risk, but this is a known
  coverage gap, disclosed rather than silently left out.

## Tests

- `cargo test -p lattice-inference` (full suite, both binaries, doctests): all green.
- `cargo clippy -p lattice-inference --bin lattice_serve --bin lattice --features "serve,f16,metal-gpu" --tests -- -D warnings`: clean.

## Bench

`make bench-compare` (quick mode, `origin/main` @ `b16d6fdf4` vs this branch @ `162691a98`).
This PR touches no kernels — the changes are HTTP-layer contract extraction and CPU-side
cancellation checkpoints in the decode loop's control flow, not compute.

```
### `local-compare` — perf regression report

❌ 4 FAIL (regression >7.0% confirmed by 95% CI)
⚠ 29 WARN (regression 3.0-7.0% confirmed)
🚀 68 confirmed improvement
```

All 4 FAIL + 29 WARN entries are in `lattice-embed`'s `simd_normalized_cosine_fast_path` /
`simd_batch_cosine_normalized_query` / `simd_prepared_query_normalized_cosine` /
`int8_raw_dot_product` groups — the known cross-build measurement noise tracked in #714, and
outside this PR's diff (`crates/embed/` is untouched here). The `lattice-inference`
`elementwise_cpu_bench` group this PR's crate actually owns (`rms_norm`, `layer_norm`,
`silu_inplace`, `gelu`, `add_bias_gelu`, `softmax_attention`, `residual_add_layer_norm`) shows
zero FAIL/WARN entries — every one of its ~15 benchmarks confirmed as `🚀 WIN` (i.e. HEAD
measured faster, all in the -3% to -11% range), consistent with this PR changing no kernel
code and the delta being ordinary build/measurement noise between the two separate `origin/main`
and branch builds, not a real speedup.

(One follow-up commit landed on this branch after the bench run, deduplicating a constant and a
guard type between the two binaries — no kernel or benchmarked code path touched; re-running
bench-compare against it would not change this verdict.)

---

# Mutation-verification log (codex rounds 1-4)

## Round 2 (codex review #1: APPROVE-WITH-FIXES, 2 majors + 2 mediums)

Addressed all four findings from codex's round-1 review.

### Major 1 — complete the extraction / align the error contract

Both binaries independently owned request types, validation cascades, handlers, and routers,
and had drifted: oversized body (413 vs 400), zero `max_tokens` (`invalid_max_tokens` vs erased
to `invalid_request`), `GET /` (daemon only), and `DefaultBodyLimit` vs manual `to_bytes`.

- Aligned status + error code for every check that exists identically on both binaries:
  `unsupported_feature`, `invalid_role`, `invalid_messages`, `invalid_request_body`,
  `invalid_max_tokens`, `context_length_exceeded`, `request_body_too_large`, `internal_error`.
  New lattice_serve-only code `content_part_limit_exceeded` (#649 DoS hardening, no lattice.rs
  analog) is documented, not silently mismatched.
- `lattice_serve.rs`'s `chat_completions` oversized-body branch now returns 413
  `request_body_too_large` (was 400 `invalid_request`), matching `lattice.rs`.
- Fixed a real bug found while aligning: `build_cfg`'s `max_tokens: 0` error branch was being
  re-wrapped through `err_response(..., err.message())`, discarding `build_cfg`'s already-correct
  `invalid_max_tokens` code. Now propagates the original `ApiError` directly.
- `GET /` added to `lattice.rs`'s router (both binaries expose it now), built from the shared
  `lattice_inference::serve::root_body()`.
- Reject-vs-clamp over-cap policy is kept as a *documented* intentional divergence, not unified
  (see `reject_zero_max_tokens`'s doc comment and the parity table's
  `max_tokens_over_cap_reject_vs_clamp` case).
- Extracted `lattice_serve.rs`'s inline router construction (previously only inside `run()`) into
  a standalone `pub fn router(state: AppState) -> Router`, mirroring `lattice.rs`'s existing
  `router()`, so both are independently testable via `tower::ServiceExt::oneshot`.
- **New**: `lattice_inference::serve::CHAT_COMPLETIONS_PARITY_CASES` (`crates/inference/src/serve/mod.rs`)
  — a shared, table-driven fixture set (8 cases) of request bodies + expected `(status, code)` per
  binary, with explicit `divergence_reason` for the one genuinely-intentional-divergence case.
  Both `lattice.rs` and `lattice_serve.rs` gained a
  `chat_completions_matches_shared_parity_table` test that drives every case through that
  binary's real `Router` via `oneshot` and asserts status + full error-envelope `code`.

### Major 2 — HTTP-level client-disconnect test (lattice.rs)

Added `chat_completions_streaming_disconnect_stops_generation`
(`crates/inference/src/bin/lattice.rs`, `#[cfg(feature = "test-utils")] mod disconnect_cancellation`):
drives the real `chat_completions` handler with a tiny all-zero-weight CPU model
(`lattice_inference::model::qwen35::test_support::tiny_zero_model`, new, gated behind a new
`test-utils` Cargo feature since bin targets can't see this crate's `#[cfg(test)]`-only
internals), reads two real SSE frames via `http_body_util::BodyExt::frame` (role chunk, then a
genuine content-delta chunk — asserted on the frame's actual JSON payload shape, not just its
arrival), then drops the response body to simulate a disconnect.

### Medium — post-prefill CPU cancellation mutation-pin

Added `generate_streaming_with_cancel_true_after_prefill_returns_interrupt`
(`crates/inference/src/model/qwen35/generation.rs`): `should_cancel` returns `true` only on its
SECOND call (the post-prefill checkpoint), asserting `generated_tokens == 0`, empty text,
`StopReason::Interrupt`, and zero `on_token` callback invocations.

### Medium — `docs/capability-matrix.md`

Updated the stale route/disconnect-cancellation/summary/fixture-manifest rows: `GET /`,
`GET /v1/models`, and disconnect cancellation are now `supported` on `lattice serve` (previously
"not yet"/"partial"), citing the shared `serve/mod.rs` helpers and the new fixtures. `make
lint-docs` (deno fmt + `check-capability-matrix.sh`) passes clean.

## Mutation log (round 2, verified this session)

| Fix | Test | Mutation applied | Reverted → | Restored → |
|---|---|---|---|---|
| Disconnect cancellation, HTTP-level composition (lattice.rs) | `chat_completions_streaming_disconnect_stops_generation` | removed `let _cancel_guard_tied_to_stream_lifetime = &cancel_guard;` from `body_stream`'s `flat_map` closure | FAILED — second SSE frame was the finish chunk (`finish_reason: "length"`, no `content`), not a content delta | PASSED — second frame carries real `delta.content`, `finish_reason: null` |
| Post-prefill cancellation checkpoint (generation.rs) | `generate_streaming_with_cancel_true_after_prefill_returns_interrupt` | short-circuited the post-prefill `if should_cancel() { ... }` guard to `if false && should_cancel()` | FAILED — `generated_tokens == 1` (assertion `left == right` failed: `1 != 0`) | PASSED — `generated_tokens == 0`, `on_token` never called |

Both mutations were applied, confirmed red, reverted, and confirmed green again in the same
session — no mutation markers remain in the tree (`grep -rn "MUTATION TEST" crates/` is clean).

## Verification (round 2)

- `cargo test -p lattice-inference --bin lattice --features "serve,f16,metal-gpu,test-utils"`:
  130 passed, 0 failed.
- `cargo test -p lattice-inference --bin lattice_serve --features "serve,f16,metal-gpu"`:
  38 passed, 0 failed.
- `cargo test -p lattice-inference --lib --features "serve,f16,metal-gpu,test-utils"`: all
  `serve::tests::*` and `model::qwen35::generation::tests::generate_streaming_with_cancel_*`
  green.
- `cargo clippy -p lattice-inference --features "serve,f16,metal-gpu,test-utils" --tests -- -D warnings`: clean.
- `cargo fmt --check`: clean.
- `make lint-docs`: clean (`=== Doc Lint Passed ===`).
- `cargo test -p lattice-inference --features "serve,f16,metal-gpu,test-utils"` (full crate,
  lib + all bins): in progress / see final report.

CI wiring: `.github/workflows/ci.yml`'s `feature-matrix` job now runs
`cargo test -p lattice-inference --bin lattice --features f16,metal-gpu,test-utils` (was missing
`test-utils`, which would have silently excluded both new `lattice.rs` test modules from ever
running in CI).

## Correction (round 3): two round-2 claims above did not hold up

Codex's round-2 review ran its own reverse mutations in a disposable worktree and found the
following claims in this addendum were **false**, verified failures rather than opinions:

- **"context_length_exceeded" was NOT fully aligned.** The parity-case table above lists it as an
  aligned status/code, with no streaming qualification. In fact `lattice_serve.rs`'s streaming
  branch only discovered a prompt-plus-budget overflow via `Ev::Rejected` *after* already
  committing to HTTP 200 SSE, terminating with `finish_reason: "length"` instead of the 400
  `context_length_exceeded` the non-streaming path (and `lattice.rs`'s equivalent `stream: true`
  request) returned. This was a genuine second same-input divergence this addendum failed to
  disclose.
- **The parity table's mutation-sensitivity was overstated.** Every case in
  `CHAT_COMPLETIONS_PARITY_CASES` targeted only `POST /v1/chat/completions`, and none sent an
  oversized body. Codex proved directly: removing `lattice.rs`'s `.route("/", get(root))`
  entirely left `chat_completions_matches_shared_parity_table` green, and restoring
  `lattice_serve.rs`'s old 400/`invalid_request` oversized-body mapping also left its parity test
  green. The table could not have caught either regression it claimed to guard against.
- **The disconnect test's mutation log implied more than it proved.** The row above is accurate
  for the mutation it names (guard-capture removal), but the test does not independently prove
  post-drop cancellation *reaches the generator* rather than merely stopping via `on_token`'s
  failed-send path. Codex proved this by replacing the CPU `should_cancel` predicate with
  `move || false` (callback path left intact): the test stayed green in 0.02s, because the failed
  channel send alone stops the loop just as fast as a correctly wired cancellation would.

### Major 1 — streaming context-overflow status now aligned

Added the "preflight before committing SSE" fix in `lattice_serve.rs`'s `chat_completions`: the
streaming branch now `await`s the worker's FIRST event before calling `Sse::new(..)` at all.
`Ev::Rejected`/`Ev::Failed` on that first event get the same un-committed 400/500 treatment the
non-streaming branch already had; anything else (`Ev::Delta`/`Ev::Done`) is threaded into the SSE
stream's state machine as an already-consumed "pending" event so it isn't dropped. `lattice.rs`
already got this right structurally (`prepare_chat_request`'s context-window check runs
unconditionally before the `req.stream` branch); a new HTTP-router-level test
(`streaming_context_overflow` module) now pins that too, so a future regression at the
composition layer — not just inside `prepare_chat_request` — is caught.

New tests: `chat_completions_streaming_context_overflow_returns_400_before_committing_sse` in
both `lattice_serve.rs` (fakes the worker's `Ev::Rejected` reply via the existing
`test_app_state_with_jobs()` fixture) and `lattice.rs` (drives a real over-window request through
the real router via `tower::ServiceExt::oneshot`). Mutation-verified: reverting the daemon's
pre-`Sse::new()` preflight (passing every first event through unconditionally, the old behavior)
made the daemon's new test fail (`left: 200, right: 400`); restored, it passes again.

### Medium 2 — parity table extended with method/path + oversized-body coverage

`ParityCase` gained `method`/`path` fields and a `CaseBody` enum (`Fixed`/`Oversized`, the latter
generating a body over `REQUEST_BODY_LIMIT_BYTES` at test time rather than embedding a >1MiB
string literal). Added two new shared cases: `get_root_route_exposed` (`GET /`, 200/200) and
`oversized_body_over_limit` (`POST /v1/chat/completions` with a >1MiB body, 413/`request_body_too_large`
on both binaries). Both binaries' `chat_completions_matches_shared_parity_table` tests now build
requests from `case.method`/`case.path`/`case.body.build()` instead of hardcoded
POST/`/v1/chat/completions`.

Mutation-verified with codex's exact two named mutations: removing `lattice.rs`'s
`.route("/", get(root))` now makes its parity test fail (`case 'get_root_route_exposed': expected
status 200, got 404`); restoring `lattice_serve.rs`'s old 400/`invalid_request` oversized-body
mapping now makes its parity test fail (`case 'oversized_body_over_limit': expected status 413,
got 400`). Both reverted and confirmed green again.

### Medium 3 — post-drop generator-side cancellation probe

Added `ModelBackend::CpuFakeGenerate` to `lattice.rs` (gated `#[cfg(all(feature = "test-utils",
test))]`): wraps a real tiny model for `tokenize_len`/`max_context`/`tokenizer` but substitutes
the CPU streaming generation call with an injected closure, so a test can observe `should_cancel`
being polled by the exact production composition (`chat_completions`'s CPU streaming arm)
independently of `on_token`.

New test `chat_completions_streaming_disconnect_cancellation_reaches_generator_post_drop`
(`post_drop_cancellation_probe` module): the fake generator sends one real delta, then
synchronizes with the test via two handshake channels — `checkpoint1` (sent the instant
`chat_completions(..).await` returns, before any frame is read or the body is dropped) and `go`
(sent only after `drop(body)`) — taking a `pre_drop_cancelled` reading right after `checkpoint1`
and a bounded should-cancel-only poll loop right after `go`. This two-handshake design was
necessary: an earlier single-poll-and-time-it version of this test raced against
`chat_completions`'s own return and non-deterministically read `false` even under the
guard-removal mutation, because an unused `cancel_guard`'s drop timing isn't ordered relative to
the generator thread's scheduling without an explicit synchronization point.

Mutation-verified independently, 5/5 runs each direction:
- Guard-capture removal: `pre_drop_cancelled` reads `true` (impossible under correct wiring,
  since `cancel_guard` is still alive at the `checkpoint1` synchronization point) — test fails.
- `should_cancel` predicate replaced by `move || false` (callback path intact): the post-`go` poll
  loop exhausts its budget and reports `ExhaustedWithoutCancel` — test fails (times out at the
  outer 10s await, ~28s wall time under this mutation since the internal loop still runs to its
  own ~20s budget in the background).

Both mutations reverted and confirmed green again (5/5 consecutive runs).

### Minor 4 — workspace dependencies

Moved `http-body-util` and `tower` from `crates/inference/Cargo.toml`'s local
`[dev-dependencies]` version pins to the root `Cargo.toml`'s `[workspace.dependencies]`,
referenced via `.workspace = true`; `tower`'s `util` feature stays at the crate use site.

### Minor 5 — doc staleness

- `serve/mod.rs`: removed the dead `ParityCase::same`/`::diverging` rustdoc links (methods that
  were never implemented; the table uses direct struct literals).
- `lattice_serve.rs`'s router doc comment: corrected to say the oversized-body divergence is
  ENFORCEMENT MECHANISM only — both binaries return 413 `request_body_too_large` today.
- `lattice.rs`'s disconnect-test comments (two locations): corrected `finish_reason: "interrupt"`
  to `finish_reason: "length"` (the shared `finish_reason()` mapping has no `"interrupt"` string;
  an interrupted/cancelled result reports `"length"`, matching what the guard-removal mutation
  actually observed).
- `test_support.rs:49`: backtick-quoted `` `"<unk>"` `` (was raw, triggering an unclosed-HTML-tag
  rustdoc warning).

`cargo doc -p lattice-inference --features "serve,f16,metal-gpu,test-utils" --no-deps` shows zero
warnings for `serve/mod.rs` and `test_support.rs`; `lattice.rs`/`lattice_serve.rs` each retain
exactly one pre-existing, unrelated warning (a stale `Q4Block` intra-doc link and a stale
`CancelOnDrop` intra-doc link respectively, both present on `origin/main` before this PR and
outside this PR's changed line ranges).

## Verification (round 3)

- `cargo test -p lattice-inference --bin lattice --features "serve,f16,metal-gpu,test-utils"`:
  132 passed, 0 failed (130 round-2 baseline + 2 new: streaming-context-overflow,
  post-drop-cancellation-probe).
- `cargo test -p lattice-inference --bin lattice_serve --features "serve,f16,metal-gpu"`:
  39 passed, 0 failed (38 round-2 baseline + 1 new: streaming-context-overflow).
- `cargo clippy -p lattice-inference --bin lattice --features "serve,f16,metal-gpu,test-utils" --tests -- -D warnings`: clean.
- `cargo clippy -p lattice-inference --bin lattice_serve --features "serve,f16,metal-gpu,test-utils" --tests -- -D warnings`: clean.
- `cargo fmt --check`: clean.
- `cargo doc -p lattice-inference --features "serve,f16,metal-gpu,test-utils" --no-deps`: clean
  on all round-2/round-3 changed files (see Minor 5 above).
- Full-crate background run from round 2 (`cargo test -p lattice-inference --features
  "serve,f16,metal-gpu,test-utils"`, all lib + bins) completed during this round: 1758 passed, 4
  failed. All 4 failures are pre-existing, unrelated Metal-precision/harness tests
  (`f16_kv_metal_path_executes_and_reports_capability` — requires `LATTICE_KV_F16=1` env var not
  set by this invocation; `metal_generate_streaming_honors_stop_string_and_excludes_stop_text`,
  `metal_qwen35_chunked_prefill_length_1_final_chunk_matches_step_loop`,
  `metal_qwen35_prefill_all_logits_long_prompt_no_longer_panics` — Metal chunked-prefill numeric
  parity, in `forward::metal_qwen35`, no relation to this PR's `serve`/`bin` changes). None touch
  `serve/`, `bin/lattice.rs`, or `bin/lattice_serve.rs`.

## Correction (round 4): two round-3 fixes did not hold up under codex's own reverse mutations

Codex's round-3 review (0 blockers/majors, 2 mediums, verdict APPROVE-WITH-FIXES) ran its own
disposable-worktree mutations against the round-3 fixes above and found two gaps this addendum
did not disclose:

- **Medium 3 above ("post-drop generator-side cancellation probe") pinned a copy, not the
  production predicate.** The real `ModelBackend::Cpu` arm and the test-only `CpuFakeGenerate`
  arm each independently constructed `move || *cancel_rx.borrow()`. Codex mutated ONLY the real
  `Cpu` arm's predicate to `move || false` (leaving `CpuFakeGenerate`'s own copy and the callback
  path untouched) and the round-3 post-drop test — which only ever constructs
  `CpuFakeGenerate` — stayed green. The test proved the fake arm's copy was wired correctly; it
  never observed the real `Cpu` arm's predicate at all.
- **Major 1's round-3 test ("streaming context-overflow status now aligned") verified the
  response mapping but not same-input real-router parity.** The daemon-side test called
  `chat_completions` directly with a request that would NOT overflow in production (`max_tokens`
  defaulting to 100 against a 4096-token `model_max_context`) and a fake task injecting
  `Ev::Rejected` directly, while `lattice.rs`'s companion test genuinely overflowed a real
  1024-token tiny model through its real router. The round-3 doc comment calling these
  "identical requests" was wrong. Codex proved directly: removing `spawn_worker`'s
  `check_prompt_fits_window` call left both the daemon's focused handler test and the pure
  `check_prompt_fits_window` unit tests green — no test exercised the real worker's window
  enforcement end to end.

### Medium 1 — unified the CPU streaming cancellation predicate

Extracted `spawn_cpu_style_streaming_generation` (`lattice.rs`), which constructs the
`on_token`/`should_cancel` composition exactly once. The real `ModelBackend::Cpu` arm now builds
an `Arc<dyn Fn(...)>` that delegates to `cpu_model.generate_streaming_with_cancel` and passes it
into this shared helper; the test-only `ModelBackend::CpuFakeGenerate` arm passes its injected
closure into the SAME helper. There is no longer a separate predicate construction site for a
mutation to leave unmutated — both arms observe the same `should_cancel` closure.

Mutation-verified with codex's exact named mutation (replace the shared `should_cancel` predicate
with `move || false`, `on_token` unchanged): the post-drop cancellation test
(`chat_completions_streaming_disconnect_cancellation_reaches_generator_post_drop`) failed with a
10s timeout / `ExhaustedWithoutCancel`, exactly as it would have under codex's round-3 mutation
against the real `Cpu` arm specifically — because after this fix there is only one call site left
to mutate, and it's the one both the real arm and the fake-arm-driven test route through. Reverted
and confirmed green again.

### Medium 2 — real-router same-input overflow parity for the daemon

Extracted `enforce_prompt_window` (`lattice_serve.rs`): tokenizes + calls
`check_prompt_fits_window`, replacing the inline block previously duplicated only inside
`spawn_worker`'s real Metal closure. `spawn_worker` now calls this function; a new test module
`real_router_overflow_parity` also calls it, via a REAL worker thread running the actual
`run_worker_loop` with a real (tiny, test-only) `BpeTokenizer` from
`test_support::tiny_zero_model` — no Metal engine involved, since an overflowing request never
reaches generation. New test
`chat_completions_streaming_context_overflow_matches_lattice_real_router` drives
`lattice_inference::serve::OVERFLOW_PARITY_REQUEST_BODY` — a new shared fixture both binaries'
tests now build from (`OVERFLOW_PARITY_CONTEXT_WINDOW` = 1024, matching `lattice.rs`'s tiny test
model's fixed context window) — through the daemon's real `router()`.

The existing handler-focused daemon test's doc comment and assertion message were corrected to
stop claiming "identical request body" / same-input parity; it is documented as pinning only the
response-mapping contract in isolation, with the new `real_router_overflow_parity` test doing the
actual same-input real-router parity work. `lattice.rs`'s `streaming_context_overflow` test was
also switched to build its request from the same shared `OVERFLOW_PARITY_*` constants instead of
independently-typed literals.

Mutation-verified with codex's exact named mutation (`enforce_prompt_window` unconditionally
returns `Ok(prompt_len)`, dropping its `check_prompt_fits_window` call): the new
`real_router_overflow_parity` test failed (`left: 200, right: 400`) — the real worker accepted
the overflowing request and the daemon committed a 200 SSE response. Reverted and confirmed green
again.

## Verification (round 4)

- `cargo fmt --check`: clean.
- `cargo clippy -p lattice-inference --bin lattice --bin lattice_serve --features
  "serve,f16,metal-gpu,test-utils" --tests -- -D warnings`: clean.
- `cargo test -p lattice-inference --bin lattice --features "serve,f16,metal-gpu,test-utils"`:
  132 passed, 0 failed (unchanged count — round 4 modified existing tests' internals, not test
  count, on the `lattice.rs` side).
- `cargo test -p lattice-inference --bin lattice_serve --features "serve,f16,metal-gpu,test-utils"`:
  40 passed, 0 failed (39 round-3 baseline + 1 new: `real_router_overflow_parity`).
- `cargo test -p lattice-inference --bin lattice_serve --features "serve,f16,metal-gpu"` (no
  `test-utils`, matching CI's non-test-utils leg): 39 passed, 0 failed — confirms the new
  `real_router_overflow_parity` module is correctly gated behind `test-utils` and does not affect
  the default test-utils-less build.
- `make lint-docs`: clean (`=== Doc Lint Passed ===`).
- `cargo doc -p lattice-inference --features "serve,f16,metal-gpu,test-utils" --no-deps`: zero new
  warnings; the same two pre-existing, unrelated warnings from round 3 (`Q4Block`, `CancelOnDrop`
  intra-doc links) remain, both predating this PR.
- Both mutations (shared `should_cancel` → `move || false`; `enforce_prompt_window` →
  unconditional `Ok`) applied, confirmed red, reverted, confirmed green again in this session — no
  mutation markers remain in the tree.


ohdearquant:

## Review of PR #786 (round 1)

The three named bug fixes are correct, and I found no blocker in the generation or stop-cause logic. Four fixes remain before merge: the PR does not yet complete ADR-080 C2's shared router/request/error contract, the disclosed `lattice.rs` HTTP disconnect gap is an explicit #782 completion gate, the CPU post-prefill checkpoint is not mutation-pinned, and the capability matrix now describes behavior that this PR removes.

### Prior review context

- No prior reviews or review threads exist on PR #786.
- This round reviewed the full `origin/main...HEAD` diff at `3e7bf65a7709001ca41047a07b327d6c54425193` against `origin/main` at `b16d6fdf4b730581d47989f73e107f8605add022`.

### What this round examined

- Behavior preservation in both binaries, including validation order, intentionally different over-cap policies, endpoint sets, request-size handling, and SSE framing.
- CPU and Metal cancellation ownership/checkpoints, callback interruption, prefill/decode coverage, and `CancelOnDrop` lifetime.
- All three `ChatCompletionOutput` construction sites and stop-cause propagation through daemon worker events into SSE and JSON.
- Current tests, claimed mutation sensitivity, current-head GitHub checks, and the ADR-080/#782 completion contract.

### Findings (4 total: 0 blockers, 2 majors, 2 mediums)

#### Majors

1. [`crates/inference/src/bin/lattice_serve.rs:1242`](crates/inference/src/bin/lattice_serve.rs#L1242) The shared module does not yet unify the router/request/error contract required by ADR-080 C2.
   - **Current**: `serve/mod.rs` shares an error type, a body-limit constant, and several leaf helpers, but both binaries still independently own their request types, validation cascades, handlers, and router construction (`lattice.rs:2664,3016,3315,3821`; `lattice_serve.rs:221,278,796,849,1766`). This is the structure ADR-080 explicitly says to extract at [`docs/adr/ADR-080-consolidation-duplicated-contracts.md:147`](docs/adr/ADR-080-consolidation-duplicated-contracts.md#L147).
   - **Issue**: the remaining copies already expose materially different contracts. An oversized body maps to HTTP 413 plus `request_body_too_large` in `lattice.rs` (`lattice.rs:3467-3470`) but HTTP 400 plus generic `invalid_request` in `lattice_serve.rs` (`lattice_serve.rs:1242-1254,1578-1594`). A zero-token request similarly uses `invalid_max_tokens` in `lattice.rs` but is erased to `invalid_request` by the daemon's `err_response`. Route sets also do not match the PR body's “1:1” claim: the daemon still exposes `GET /` at `lattice_serve.rs:1767`, while `lattice.rs:3821-3827` does not. The 1 MiB numeric limit is shared, but its HTTP enforcement contract is not. Axum also documents that `DefaultBodyLimit` is extractor-local, which is materially different from the daemon's manual `to_bytes` mapping ([Axum `DefaultBodyLimit`](https://docs.rs/axum/latest/axum/extract/struct.DefaultBodyLimit.html)).
   - **Fix**: complete the shared router/request-normalization/error-response layer required by #782, parameterizing only the documented intentional differences (especially reject-vs-clamp for over-cap `max_tokens`). Align same-input status/error-code behavior, decide the `GET /` contract explicitly, and add cross-binary parity tests for route exposure, malformed/oversized bodies, and common validation errors.
   - **Test**: a table-driven test that sends identical bodies through both routers and compares status plus the full `error` object, with explicit expected-deviation entries for intentionally different policies.
   - **Confidence**: derived from the two live router/handler implementations and ADR-080's extraction contract; route sets were also observed with `rg -n '\.route\(' ...`.

2. [`crates/inference/src/bin/lattice.rs:3548`](crates/inference/src/bin/lattice.rs#L3548) Close the disclosed HTTP-level disconnect test gap before merge.
   - **Current**: cancellation depends on `cancel_guard` being captured by `body_stream` at `lattice.rs:3655-3660`, so dropping the Axum response body must simultaneously drop the stream receiver and flip the watch value consumed by the blocking CPU/Metal job. The primitive and guard have unit tests, but this composition has none.
   - **Issue**: this is the one layer where a closure-capture/lifetime refactor can silently leave generation detached from the response. ADR-080 names HTTP disconnect handling as a wire-risk not covered by e2e parity and requires HTTP cancellation tests at [`docs/adr/ADR-080-consolidation-duplicated-contracts.md:285`](docs/adr/ADR-080-consolidation-duplicated-contracts.md#L285); issue #782 repeats the requirement for both binaries. The PR body's disclosure is good, but it does not waive that gate.
   - **Fix**: add a `lattice.rs` handler/router test that starts a streaming request, polls enough of the body to establish the stream, drops the response body, and proves the generation-side cancellation signal fires and no later delta is accepted. A small injected/fake backend is preferable to a model/GPU test.
   - **Test**: the test above should fail if the guard is removed from the SSE stream capture or its receiver is no longer threaded into `generate_streaming_with_cancel`.
   - **Confidence**: derived from the stream ownership path and the explicit ADR/#782 completion gate. Tokio's watch contract confirms `send`/closure behavior when receivers disappear ([Tokio `watch::Sender`](https://docs.rs/tokio/latest/tokio/sync/watch/struct.Sender.html)).

#### Mediums

3. [`crates/inference/src/model/qwen35/generation.rs:545`](crates/inference/src/model/qwen35/generation.rs#L545) The post-prefill CPU cancellation checkpoint is correct but not mutation-sensitive.
   - **Current**: source correctly polls before prefill, after prefill, and at both decode-loop tops. The five new tests cover first poll (`n == 1`), decode polls (`n >= 3`), and callback refusal, but no test returns true only on poll 2. Removing the post-prefill check leaves all five green.
   - **Issue**: cancellation during prefill would then sample/emit a token before the next loop checkpoint. That is a real regression class in the user-prioritized prefill contract and misses ADR-080's mutation-sensitive-guard rule at `docs/adr/ADR-080-consolidation-duplicated-contracts.md:274-278`.
   - **Fix**: add a CPU test whose predicate is true only on its second call and assert zero generated tokens, empty text, `Interrupt`, and zero callback invocations; mutation-check it with only the post-prefill guard removed.
   - **Test**: `generate_streaming_with_cancel_true_after_prefill_returns_interrupt` (or equivalent).
   - **Confidence**: derived from the checkpoint call sequence and observed by running all five current `generate_streaming_with_cancel_` tests.

4. [`docs/capability-matrix.md:56`](docs/capability-matrix.md#L56) Update the serving capability documentation in this PR.
   - **Current**: the matrix still says `lattice serve` has no `/v1/models` route (`:56, :91, :117, :173`) and no CPU/independent disconnect cancellation (`:80, :162`). Both statements become false at this PR's head, and the cited source line numbers are stale.
   - **Issue**: this is the repository's authoritative cross-surface contract table; leaving it stale will misdirect follow-up work and contradict the shipped API.
   - **Fix**: update the route, cancellation, fixture-manifest, and divergence-summary rows. Keep the documented intentional differences, including stop/logprobs support and reject-vs-clamp token caps.
   - **Test**: `make lint-docs` plus the capability-matrix checker used by CI.
   - **Confidence**: derived from the current router/cancellation source and the unchanged document.

- `generate_streaming_with_cancel` checks cancellation before and after CPU prefill and before work in both decode branches. Callback refusal stops immediately, does not flush trailing bytes, returns `StopReason::Interrupt`, and does not make another callback.
- `lattice_serve` retains its pre-existing dequeue, prefill-like, decode, and send-failure cancellation coverage. Hoisting `CancelOnDrop`/`cancel_pair` does not change its ownership order.
- The `lattice.rs` guard is retained through the role chunk, content chunks, terminal chunk, and `[DONE]`; normal completion drops it only after the stream finishes. The missing item is regression coverage, not a derived lifetime bug.
- All three `ChatCompletionOutput` constructors copy the underlying `stopped` value (`metal_qwen35.rs:12680,14497,15908`). EOS, stop-token, and stop-string termination all set `stopped=true` and serialize as OpenAI `"stop"`; length exhaustion stays false and serializes as `"length"`. Both daemon response modes consume the propagated flag.
- `lattice.rs` still rejects an over-cap token request, while `lattice_serve` still clamps it. SSE event order/framing is unchanged apart from the corrected terminal reason.
- Current-head GitHub CI, app-binary, e2e-parity, Q4 PPL/composed-golden, feature-matrix, MSRV, semver, and gate jobs were green when inspected.
- Observed local gates:
  - `cargo fmt --all -- --check` — pass.
  - `cargo clippy --workspace -- -D warnings` — pass.
  - `cargo test -p lattice-inference --features serve` — pass (1,600 library tests, 123 `lattice` tests, integrations and doctests).
  - `cargo test -p lattice-inference --bin lattice --features 'serve,f16,metal-gpu'` — 128/128 pass.
  - `cargo test -p lattice-inference --bin lattice_serve --features 'serve,f16,metal-gpu'` — 35/35 pass.
  - Reverse-mutation spot check: removing the daemon zero-token guard made `chat_completions_max_tokens_zero_400` fail with 500 vs 400; hardcoding both terminal reasons made both `length_when_not_stopped` tests fail. These PR-body mutation claims are genuine.

### What I did NOT check

- I did not run a real TCP client disconnect against a live CPU or Metal model; that missing automated HTTP composition test is finding 2.
- I did not run local Metal generation or `make e2e-parity`; I inspected the current-head remote parity/golden results instead.
- I did not rerun `make bench-compare`; I reviewed the PR-body report and verified that the follow-up commit only deduplicates the body-limit constant and guard type.
- I did not run the full workspace test suite; validation was the inference package with serving features plus workspace clippy.

- Re-review after the shared contract/parity tests and HTTP disconnect test land. Focus only on same-input error/status parity, route exposure, and mutation sensitivity; another broad round should not be necessary if those fixes are narrow.

Domain utility: MEDIUM — the cancellation/channel briefing helped structure the ownership audit, while repository source, ADR-080, and issue #782 supplied the decisive contract.

ohdearquant:

## Review of PR #786 (round 2)

The implementation closes most of round 1, including the actual guard wiring and the post-prefill checkpoint. It is not ready to approve unchanged: one same-input streaming validation path still has a client-visible status/error divergence that the addendum calls aligned, and the two new regression surfaces omit the exact mutations named in round 1.

### Prior review context

- Round 1 found two majors and two mediums: incomplete cross-binary router/error parity, no HTTP-level disconnect regression test, no mutation pin for the CPU post-prefill checkpoint, and stale capability-matrix rows.
- This round intentionally focused on those four findings plus the new CI/test-fixture seams, at head `85826aa8a63b5237042a1079b8a14286fd1f9141` versus `origin/main` `b16d6fdf4b730581d47989f73e107f8605add022`.
- There are no prior GitHub reviews or review threads on PR #786.

### What this round examined

- Same-input status/error behavior, route exposure, and the eight shared parity cases through both real routers.
- The complete `chat_completions -> body_stream -> cancel_guard/cancel_rx -> generate_streaming_with_cancel` CPU path.
- Guard-removal, receiver-unthreading, post-prefill-checkpoint, route-removal, and oversized-body reverse mutations in a disposable detached worktree.
- The updated capability matrix against the current routers and cancellation source.
- `test-utils` feature gating, the tiny model fixture, dev dependencies, and the CI job that enables the feature.

### Findings (5 total: 0 blockers, 1 major, 2 mediums, 2 minors)

#### Major

1. [`crates/inference/src/bin/lattice_serve.rs:1520`](crates/inference/src/bin/lattice_serve.rs#L1520) Streaming context-window rejection is still a second same-input wire divergence, contrary to the addendum and parity-table contract.
   - **Current**: `lattice serve` completes `prepare_chat_request` and `check_context_window` before branching on `stream` (`lattice.rs:3490-3504`), so an over-context streaming request returns HTTP 400 with `context_length_exceeded`. `lattice_serve` cannot discover the prompt overflow until its worker sends `Ev::Rejected`; by then the SSE response is already HTTP 200, and `Phase::Body` emits a terminal chunk with `finish_reason: "length"` (`lattice_serve.rs:1520-1538`). Only the non-streaming daemon path maps that event to HTTP 400/`context_length_exceeded` (`lattice_serve.rs:1613-1635`).
   - **Issue**: the addendum lists `context_length_exceeded` among the aligned status/code checks without a streaming qualification, `CHAT_COMPLETIONS_PARITY_CASES` describes over-cap reject-vs-clamp as the sole intentional divergence, and the capability-matrix divergence summary does not mention this path. A client sending the same validly parsed `stream: true` body can therefore receive a structured 400 from one binary and a committed 200 SSE termination from the other.
   - **Fix**: either perform the daemon's context preflight before committing SSE so the status/error contract truly aligns, or explicitly document this as another intentional divergence in the addendum and capability matrix and encode it in the shared parity fixtures. The first option matches the stated contract; the second is the minimum honest description if worker ownership makes preflight impractical in this PR.
   - **Test**: drive a `stream: true` prompt-plus-budget overflow through both real routers with equal context limits and assert the intended status/body behavior.
   - **Confidence**: derived from the two current handler phase orders and the daemon's explicit “status code is already committed to 200 SSE” branch.

#### Mediums

2. [`crates/inference/src/serve/mod.rs:249`](crates/inference/src/serve/mod.rs#L249) The shared parity table omits two of round 1's three concrete drift cases.
   - **Current**: zero-token parity is present, but none of the eight cases sends a body over `REQUEST_BODY_LIMIT_BYTES`, and every case targets only `POST /v1/chat/completions`; no case exercises `GET /` or the route set. The capability matrix itself acknowledges that the new routes are not router-fixtured (`docs/capability-matrix.md:188`), and has no fixture-manifest row for `GET /`.
   - **Issue**: these are not theoretical gaps. In the disposable worktree, removing only `lattice.rs`'s `.route("/", get(root))` left `chat_completions_matches_shared_parity_table` green. Restoring the daemon's old oversized mapping (400/`invalid_request` instead of 413/`request_body_too_large`) also left its parity-table test green. The table therefore cannot prevent two regressions it claims to close.
   - **Fix**: extend the shared case representation to carry method/path and a generated-body variant, then add `GET /`/route exposure and a `REQUEST_BODY_LIMIT_BYTES + 1` case. Separate shared route/body-limit tables are also fine if both binaries consume the same fixtures.
   - **Test**: the two reverse mutations above must each make its corresponding real-router test fail. Keep the existing zero-token and documented over-cap cases.
   - **Confidence**: observed by running both mutated parity tests; both passed.

3. [`crates/inference/src/bin/lattice.rs:5387`](crates/inference/src/bin/lattice.rs#L5387) The disconnect test proves guard retention before the response is returned, but does not prove post-drop cancellation reaches the generator.
   - **Current**: the test correctly validates that frame two is a real content delta, then executes `drop(body)` and returns without observing the generation task, its stop reason, a cancellation probe, or absence of later work. Removing the guard capture fails the delta assertion, so that half is genuinely mutation-sensitive.
   - **Issue**: replacing only the CPU predicate at `lattice.rs:3600` with `move || false` left `chat_completions_streaming_disconnect_stops_generation` green. The failed channel send still makes `on_token` return false after the body is dropped, masking the unthreaded independent receiver. This is the exact second mutation round 1 required the HTTP composition test to catch. The comment that fast test completion is “practical proof” is not an oracle: Tokio documents that started `spawn_blocking` work cannot be aborted and runtime shutdown waits for it to finish, but eventual completion does not identify which cancellation path stopped it ([Tokio `spawn_blocking`](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html)).
   - **Fix**: add a post-drop generation-side probe. A test-only backend/fixture that emits one real delta, then enters a phase that polls only `should_cancel` and signals completion would isolate this wiring cleanly. After dropping the response body, await that completion under a timeout and assert `Interrupt`/no further work.
   - **Test**: the test must fail independently when (a) the guard capture is removed and (b) the `cancel_rx` predicate is replaced by `|| false` while the callback path remains intact.
   - **Confidence**: observed in a detached head worktree: guard-removal failed on frame two as claimed; receiver-unthreading passed in 0.02s.

#### Minors

4. [`crates/inference/Cargo.toml:210`](crates/inference/Cargo.toml#L210) The new `http-body-util` and `tower` dev dependencies are version-pinned locally (`:210`, `:216`) instead of being declared in root `[workspace.dependencies]` and referenced with `.workspace = true`, contrary to the repository dependency convention. Move both declarations to the workspace table and retain `tower`'s `util` feature at the crate use site.

5. [`crates/inference/src/serve/mod.rs:215`](crates/inference/src/serve/mod.rs#L215) New/stale documentation in the reviewed seams does not match the code.
   - `ParityCase::same` and `ParityCase::diverging` are linked but do not exist (`serve/mod.rs:215-217`).
   - The daemon router comment still says its body-limit mechanism has a different status code from `lattice.rs`, although both now return 413 (`lattice_serve.rs:1798-1803`).
   - The disconnect-test comment predicts `finish_reason: "interrupt"`, but the shared boolean mapping emits `"length"`; the guard-removal mutation actually observed `"length"` (`lattice.rs:5303-5314`).
   - The public test fixture writes `"<unk>"` as raw rustdoc HTML (`test_support.rs:49`).
   - **Fix**: correct/remove the stale links and wording and quote `<unk>` as code. `cargo doc` should introduce no warnings from these changed files.
   - **Confidence**: observed with `cargo doc -p lattice-inference --features serve,f16,metal-gpu,test-utils --no-deps`; all four changed-file warnings/references above are reproducible.

- Both current routers expose `GET /`, `GET /health`, `GET /v1/models`, and `POST /v1/chat/completions`. `lattice_serve::router(AppState)` is separated from process startup and both parity tests use `tower::ServiceExt::oneshot`, whose contract is to ready and invoke the real `Service` once ([Tower `ServiceExt::oneshot`](https://docs.rs/tower/latest/tower/trait.ServiceExt.html#method.oneshot)).
- The actual oversized-body implementation is aligned today: both paths enforce 1 MiB and produce 413/`request_body_too_large`. Axum confirms that `DefaultBodyLimit` applies through body-consuming extractors such as `Json`, while direct body polling needs its own limit; that matches the two current enforcement mechanisms ([Axum `DefaultBodyLimit`](https://docs.rs/axum/latest/axum/extract/struct.DefaultBodyLimit.html)).
- Zero-token, malformed JSON, empty messages, shared unsupported features/content, and role errors now preserve their specific shared error codes. The daemon no longer re-wraps `build_cfg`'s `invalid_max_tokens` error.
- The over-cap reject-vs-clamp behavior remains intact. The daemon parity fixture's 500 is explicitly identified as the no-worker harness artifact after validation; `build_cfg_clamps_to_runtime_context` separately pins the clamp.
- The production CPU stream currently captures `cancel_guard`, threads `cancel_rx` into `generate_streaming_with_cancel`, and also returns channel-send failure from `on_token`. Finding 3 is a regression-oracle failure, not a derived production wiring bug at this head.
- `generate_streaming_with_cancel_true_after_prefill_returns_interrupt` returns true only on poll two and asserts zero generated tokens, empty text, `StopReason::Interrupt`, and zero callbacks. Disabling only the post-prefill guard made it fail with `generated_tokens == 1`; this round independently reproduced the addendum's mutation claim.
- The capability matrix's updated route and cancellation rows match the current source semantics. Its omission is the streaming context-status divergence in finding 1, not the newly added route/cancellation claims themselves.
- `test_support` is gated at the module boundary by `#[cfg(feature = "test-utils")]`; `test-utils` is absent from default features. `cargo check -p lattice-inference --no-default-features` passed, so the fixture does not enter normal/default-disabled library builds. CI's macOS lattice-bin test explicitly enables the feature, so both new gated test modules run there.
- The current working tree remained source-clean throughout; all reverse mutations were confined to `/tmp/lattice-pr786-r2-mut-85826aa8`.
- Observed local gates:
  - `cargo fmt --all -- --check` — pass.
  - `git diff --check origin/main...HEAD` — pass.
  - `cargo clippy -p lattice-inference --features serve,f16,metal-gpu,test-utils --tests -- -D warnings` — pass.
  - `cargo test -p lattice-inference --bin lattice --features serve,f16,metal-gpu,test-utils` — 130/130 pass.
  - `cargo test -p lattice-inference --bin lattice_serve --features serve,f16,metal-gpu` — 38/38 pass.
  - `make lint-docs` — pass; all cited capability fixtures resolved.
  - Targeted parity, disconnect, and post-prefill tests — pass at unmodified head.
- Current-head GitHub CI, feature-matrix, app-binaries, MSRV, semver, e2e parity, Metal attention, embed drift, bench-compile, and Q4 PPL jobs were green when rechecked. `q4 composed golden` was still pending.

### What I did NOT check

- I did not run a real TCP client disconnect against a live production-sized CPU or Metal model; the HTTP composition was tested in-process with the tiny CPU model.
- I did not run the full workspace or full inference library test suites in this round. I ran both complete affected bin suites, the targeted library test, feature-enabled clippy, and relied on current-head remote CI for the broader matrix.
- I did not rerun local Metal generation, `make e2e-parity`, or `make bench-compare`; the corresponding current-head remote gates were green where applicable, and these four fix commits do not change kernels.
- I did not wait for the still-pending `q4 composed golden` job.

One short follow-up is warranted after fixes. Check only: streaming context-overflow status/error treatment, oversized/root parity mutation sensitivity, and a post-drop receiver-threading oracle. No further broad review should be necessary if those changes remain narrow.

Domain utility: MEDIUM — the streaming/channel-contract material helped separate guard lifetime from generator-side cancellation, while repository source and direct reverse mutations supplied the decisive evidence.

Verdict: APPROVE-WITH-FIXES

ohdearquant:

## Review of PR #786 (round 3, narrow follow-up)

### Prior review context

- Round 1 identified parity and cancellation-contract gaps requiring HTTP-level, mutation-sensitive coverage.
- Round 2 concluded `APPROVE-WITH-FIXES` with one major (streaming context overflow), two mediums (parity-table mutation sensitivity and post-drop cancellation oracle), and two minors (workspace dev-dependencies and documentation).
- This round checks only those five fixes at head `e4f9dba976e63d7ac9e92c114d17908f78f4ec4a` against `origin/main` `b16d6fdf4b730581d47989f73e107f8605add022`.

### What this round examined

- Streaming context-overflow preflight ordering, normal SSE behavior, and parity coverage.
- Root-route and oversized-body real-router mutation sensitivity.
- Deterministic post-drop cancellation mutation sensitivity and two-handshake race behavior.
- Workspace dev-dependency placement, changed-file rustdoc, and production exclusion of `CpuFakeGenerate`.

### Findings (2 total: 0 blockers, 0 majors, 2 mediums, 0 minors)

#### Mediums

1. [`crates/inference/src/bin/lattice.rs:3621`](crates/inference/src/bin/lattice.rs#L3621) The post-drop DI seam mutation-tests a copy of the CPU cancellation predicate, not the production predicate.
   - **Current**: the real `ModelBackend::Cpu` arm builds `move || *cancel_rx.borrow()` at `:3637`, while the test-only `CpuFakeGenerate` arm independently rebuilds the predicate at `:3687`; the new oracle constructs only `CpuFakeGenerate` at `:5606`.
   - **Issue**: in a detached disposable worktree, I replaced only the real CPU arm's predicate at `:3637` with `move || false`, leaving the callback path and fake arm intact -- the exact round-2 mutation -- and `chat_completions_streaming_disconnect_cancellation_reaches_generator_post_drop` still passed. The original disconnect test also remained green. The new test therefore proves that the copied fake-arm predicate is wired, but it does not pin the production CPU call site whose regression it was requested to catch.
   - **Fix**: factor the `on_token`/`should_cancel` construction into one helper used by both the real CPU generation call and the injected fake call, or inject the fake generator behind the real CPU arm without duplicating the predicate. Re-run the mutation at that single shared predicate and require the post-drop test to fail while `on_token` remains unchanged.
   - **Test**: `serve::tests::post_drop_cancellation_probe::chat_completions_streaming_disconnect_cancellation_reaches_generator_post_drop` must go red when the production/shared predicate alone becomes `move || false`.
   - **Confidence**: observed with the exact reverse mutation; the new test passed in 0.01s and the original disconnect test passed in 0.02s.

2. [`crates/inference/src/bin/lattice_serve.rs:3073`](crates/inference/src/bin/lattice_serve.rs#L3073) The context-overflow tests verify the response mapping, but not the requested same-input real-router parity path.
   - **Current**: the daemon test calls `chat_completions` directly with a short body, `max_tokens` defaulting to 100, and a 4096-token context, then has a fake task inject `Ev::Rejected`. That request would not overflow in production. The `lattice.rs` companion at `:5809` instead sends `max_tokens: 1024` through the real router backed by a real 1024-token tiny model, so the body genuinely overflows. The comment at `lattice.rs:5829` calls these identical requests, but they are not.
   - **Issue**: this is enough to mutation-pin the new pre-`Sse::new` `Ev::Rejected -> 400/context_length_exceeded` mapping, but it does not exercise one equal-limit overflow request through both routers as round 2 requested. In particular, removing the production worker closure's `check_prompt_fits_window` call at `lattice_serve.rs:1065` would leave both the direct handler test and the pure helper test green.
   - **Fix**: use one shared over-window request fixture and drive the daemon's real router plus a test worker seam that executes the production tokenize/window-check wrapper before emitting its first event. Keep the current focused handler test if useful, but do not label it same-input parity by itself.
   - **Test**: removing the production worker-side `check_prompt_fits_window` invocation must make the daemon overflow parity test fail; the corresponding `lattice.rs` real-router test should consume the same request parameters and effective context limit.
   - **Confidence**: derived from the two fixture bodies/state configurations and the worker call graph; observed tests confirm the mapping but do not traverse the daemon router-to-worker preflight.

- The daemon phase order did change: it awaits `rx.recv()` and handles first-event `Rejected`/`Failed` at `lattice_serve.rs:1460-1505`, before `Sse::new(..).into_response()` at `:1643-1645`. A first `Delta` or `Done` is retained in `pending` and replayed by `Phase::Body` at `:1507-1532`, so the preflight does not drop the first generation event.
- Unmodified targeted streaming-overflow, normal streaming, post-drop, and both real-router parity tests pass. In the disposable worktree I temporarily strengthened the normal daemon SSE test to require `role chunk < content "hi" < length finish chunk`; it passed with HTTP 200, directly confirming normal commit and first-delta replay. Axum's official source shows that `Sse::into_response` constructs the event-stream response from the supplied stream, which supports treating the preceding match as the pre-commit boundary ([Axum SSE source](https://docs.rs/axum/latest/src/axum/response/sse.rs.html#89-105)).
- Removing only `lattice.rs`'s `.route("/", get(root))` makes its real-router parity test fail on `get_root_route_exposed` (expected 200, observed 404).
- Restoring only the daemon's old oversized-body 400/`invalid_request` mapping makes its real-router parity test fail on `oversized_body_over_limit` (expected 413, observed 400).
- Removing the SSE stream's `cancel_guard` capture makes the new post-drop test fail deterministically with `pre_drop_cancelled: true`. The first handshake is sent only after `chat_completions(..).await` returns, and the second only after `drop(body)`; the blocking generator receives each before reading the corresponding predicate. This establishes the required ordering without sleeps as the oracle. Rust's channel documentation confirms sends become available to the receiver in send order and `recv` blocks until a message is available ([Rust `std::sync::mpsc::channel`](https://doc.rust-lang.org/stable/std/sync/mpsc/fn.channel.html)).
- `http-body-util` and `tower` now have workspace versions in root `Cargo.toml`; the inference crate uses `.workspace = true` and keeps `tower`'s `util` feature locally.
- `CpuFakeGenerate` and all of its match arms are guarded by `#[cfg(all(feature = "test-utils", test))]`. `cargo check -p lattice-inference --bin lattice --features serve,f16,metal-gpu,test-utils` succeeds as a non-test build, so that enum seam is absent there; the separately feature-gated public tiny-model fixture remains intentionally opt-in.
- The four round-2 documentation defects are corrected. Ordinary `cargo doc -p lattice-inference --features serve,f16,metal-gpu,test-utils --no-deps` succeeds without warnings from those fixes. Its remaining warnings are pre-existing (including the daemon's old `CancelOnDrop` link, whose introducing commit is already on `origin/main`), not regressions from the five round-3 commits.
- Observed clean-head gates after discarding all mutation artifacts:
  - `cargo fmt --all -- --check` -- pass.
  - `git diff --check origin/main...HEAD` -- pass.
  - `cargo clippy -p lattice-inference --bin lattice --bin lattice_serve --features serve,f16,metal-gpu,test-utils --tests -- -D warnings` -- pass.
  - `cargo test -p lattice-inference --bin lattice --features serve,f16,metal-gpu,test-utils` -- 132/132 pass.
  - `cargo test -p lattice-inference --bin lattice_serve --features serve,f16,metal-gpu` -- 39/39 pass.
  - `make lint-docs` -- pass, including the capability-matrix fixture check.
- Current-head required GitHub CI, feature matrix, app binaries, MSRV, semver, E2E parity, embed drift, WASM parity, bench compile, and Q4 PPL gates were green when checked. The informational Metal attention and Q4 composed-golden legs were still pending.
- All mutations were confined to `/tmp/lattice-pr786-r3-mut.WauFHY`, which was restored and removed. The review worktree's tracked sources remained unchanged.

### What I did NOT check

- I did not perform a broad re-review outside the user-specified round-3 seams.
- I did not run a live TCP client against a production-sized CPU/Metal model. The disconnect and SSE checks were in-process with the tiny/fake backends.
- I did not benchmark wall-clock time to first content token. Source inspection and the temporary event-order assertion show that the already-received first delta is replayed without loss; HTTP headers and the role chunk now necessarily wait for the worker's first event because that is the new preflight boundary.
- I did not run the full workspace suite, local E2E parity, Metal generation, or benchmarks in this narrow fix-only round. Current remote gates covered the broader affected matrix where noted.
- I did not wait for the two informational GitHub jobs that were still pending.

One short final round is useful after the two test-seam fixes. Re-run only (1) the production/shared `move || false` cancellation mutation and (2) removal of the daemon worker's prompt-window check against the same-input real-router overflow fixture. If both go red and no broader code changes land, stop reviewing.

Domain utility: MEDIUM -- the concurrency-oracle guidance usefully emphasized checking which exact production predicate the injected test executes, while direct mutations supplied the decisive evidence.

Verdict: APPROVE-WITH-FIXES

ohdearquant:

## Review of PR #786 (round 4, final narrow check)

### Verdict: APPROVE

### Prior review context

- Rounds 1–3 are recorded in `pr_review.md`, `pr_review_r2.md`, and `pr_review_r3.md`.
- Round 3 ended `APPROVE-WITH-FIXES` on two medium test-fidelity gaps and set a precise stop condition: both requested mutations must make their regression tests fail, with no broader production-behavior change.
- This final round is intentionally limited to those two mutations and quick clean-head sanity checks.

### What this round examined

- The single shared CPU-style streaming cancellation predicate and its real/fake callers.
- The daemon's extracted prompt-window helper, its production caller, and the real-worker router test.
- Same-request and same-effective-context-limit parity between the two binaries' companion tests.
- The three post-round-3 commits, formatting, targeted clippy, and the affected clean-head tests.

### Findings (0 total: 0 blockers, 0 majors, 0 mediums, 0 minors, 0 nits)

None.

### Mutation evidence

1. **Production/shared cancellation predicate: red as required.**
   - At clean head, the sole executable CPU-style `move || *cancel_rx.borrow()` is inside `spawn_cpu_style_streaming_generation` at `crates/inference/src/bin/lattice.rs:3544`. The real `ModelBackend::Cpu` arm calls that helper at `:3691`; `CpuFakeGenerate` calls the same helper at `:3733`. An `rg` of the file found no second executable copy that the mutation could skip.
   - In detached worktree `/tmp/lattice-pr786-r4-mut.06b180c`, I changed only that shared predicate to `move || false`; `on_token` remained unchanged.
   - Observed command: `cargo test -p lattice-inference --bin lattice --features serve,f16,metal-gpu,test-utils serve::tests::post_drop_cancellation_probe::chat_completions_streaming_disconnect_cancellation_reaches_generator_post_drop -- --exact --nocapture`.
   - Observed result: exit 101, 0 passed / 1 failed. The test panicked at its post-drop timeout with `should_cancel never observed the disconnect`; the test binary reported the failure after 28.70s.

2. **Daemon worker prompt-window check: red as required.**
   - `enforce_prompt_window` contains the production check at `crates/inference/src/bin/lattice_serve.rs:1038`. Its two relevant callers are the real `spawn_worker` closure at `:1088` and the real-worker parity closure at `:3198`; the latter runs the real `run_worker_loop` on a real thread with the tiny model's real `BpeTokenizer`.
   - In the same detached worktree, after restoring mutation 1, I neutralized only the `check_prompt_fits_window` branch inside `enforce_prompt_window`.
   - Observed command: `cargo test -p lattice-inference --bin lattice_serve --features serve,f16,metal-gpu,test-utils imp::tests::real_router_overflow_parity::chat_completions_streaming_context_overflow_matches_lattice_real_router -- --exact --nocapture`.
   - Observed result: exit 101, 0 passed / 1 failed in 0.01s. The real router returned HTTP 200; the test expected HTTP 400.

- Clean review head and GitHub PR head both resolve to `06b180caada9894711a49470c32241b6bfff8937`.
- Since round-3 head `e4f9dba976e63d7ac9e92c114d17908f78f4ec4a`, exactly the three named fix commits landed: `c8a8d3321`, `c333b6c5a`, and `06b180caa`. Their diff is confined to `crates/inference/src/serve/mod.rs`, `crates/inference/src/bin/lattice_serve.rs`, and `crates/inference/src/bin/lattice.rs`; there were no broader code changes.
- Diff reading shows no production-behavior change from the two extractions:
  - `spawn_cpu_style_streaming_generation` hoists the prior CPU `spawn_blocking`, unchanged `on_token`, cancellation predicate, generator call, and result handling into one function. The production CPU arm supplies the same `generate_streaming_with_cancel` call; the test arm supplies only its injected generator.
  - `enforce_prompt_window` hoists the production worker's prior tokenize + `check_prompt_fits_window` sequence verbatim, including the same error prefix and early return. `spawn_worker` ignores the returned length just as it did before extraction.
- Same-input/effective-limit parity is real. Both tests build the exact JSON body from `OVERFLOW_PARITY_REQUEST_BODY` at `crates/inference/src/serve/mod.rs:55` (`max_tokens: 1024`, `stream: true`). The daemon passes `OVERFLOW_PARITY_CONTEXT_WINDOW` (1024) to its real router at `lattice_serve.rs:3226`; the `lattice.rs` companion uses the tiny CPU model plus the shared 4096 request cap at `lattice.rs:5864`. That model's context is the RoPE maximum, constructed from `max_position_embeddings: 1024` at `crates/inference/src/model/qwen35/test_support.rs:96` and `:158`.
- Observed clean-head gates after both mutations were restored and the disposable worktree was removed:
  - `cargo fmt --all -- --check` — pass.
  - `cargo clippy -p lattice-inference --bin lattice --bin lattice_serve --features serve,f16,metal-gpu,test-utils --tests -- -D warnings` — pass. I touched the two clean source files first to force recompilation rather than reuse mutation artifacts.
  - The restored post-drop cancellation test named above — pass, 1/1.
  - The restored daemon real-router overflow parity test named above — pass, 1/1; observed HTTP 400.
  - `cargo test -p lattice-inference --bin lattice --features serve,f16,metal-gpu,test-utils serve::tests::streaming_context_overflow::chat_completions_streaming_context_overflow_returns_400_before_committing_sse -- --exact --nocapture` — pass, 1/1.
  - `git diff --check origin/main...HEAD` — pass.
- Both mutations were confined to the detached disposable worktree. It was restored to a clean status and removed; the review worktree's tracked source content remains unchanged.

### What I did NOT check

- This is not a broad re-review of PR #786; it follows the round-3 stop condition.
- I did not rerun the workspace-wide test suite, local E2E parity, Metal generation, or benchmarks. Those areas were outside this final fix-only scope.
- I did not reassess findings closed in rounds 1–3 beyond the two specified seams.

Stop reviewing. Both required mutations go red, the restored tests and clean-head gates go green, and no broader code changes landed. Another round is not useful unless the PR changes again.

Domain utility: LOW — the async-cancellation briefing reinforced that cooperative cancellation is only real at the exact polled predicate, but the local call graph and mutation results supplied the decisive evidence.

Verdict: APPROVE

oceanwaves630 (approved):
Reviewed the test-utils DI seam and the streaming preflight ordering at the head commit. All CpuFakeGenerate sites (enum variant, four match arms including the non-streaming one) are gated behind cfg(all(feature = "test-utils", test)), so the seam is excluded even from a test-utils release build; the test_support module is feature-gated and opt-in by design. The chat handler awaits the worker's first event and returns uncommitted 400/500 for Rejected/Failed before constructing the SSE response, with a first Delta/Done carried into the stream state so no token is dropped. CI green, mutation checks documented in the PR body.

Issue #783, opened 2026-07-10
refactor(inference): C3 — backend-neutral decode policy, apply-or-fail-closed (ADR-080 cluster C3 checklist)


ohdearquant:
## Context

ADR-080 (`docs/adr/ADR-080-consolidation-duplicated-contracts.md`, merged via #778) cluster **C3**: backend-neutral decode policy. Sequenced third (after C1 #780, C2).

## Contract (per ADR-080)

Decode-policy controls (`stop_strings` matching state, reasoning-budget accounting, logprobs formatting) live in ONE policy struct in `crates/inference/src/model/qwen35/generation.rs` wrapping a backend-specific per-step forward callback. Every decode loop either drives through the wrapper or explicitly REJECTS the unsupported `GenerateConfig` field at its entry point — "apply or fail closed", the same contract grammar/logprobs already have on quantized/batch paths. Silent omission is the defect.

## Checklist — the 10-site decode-loop family

- [ ] Policy struct + per-step callback API in `model/qwen35/generation.rs` (both canonical CPU entry points route through it; they already apply full policy — refactor-only)
- [ ] `forward/cpu_f16.rs` — apply or reject `stop_strings`/`reasoning_budget`
- [ ] `forward/cpu_q8.rs` — same
- [ ] `forward/neon_forward.rs` — same
- [ ] `forward/batch_prefill.rs` — same
- [ ] `forward/metal_qwen35.rs` ×4 entry points (direct, streaming, MTP, multimodal-adjacent) — route or reject; #657 already landed stop_strings on four Metal loops, close the residue
- [ ] Canonical first-wins/repetition-aware greedy selection shared by MTP and self-speculative paths (held finding)
- [ ] MTP-style fast paths that genuinely cannot support a field reject the config explicitly instead of ignoring it

## Completion gates

- [ ] Per-backend tests: config with stop_strings/reasoning_budget on each entry point either takes effect (asserted) or errors (asserted) — never silently ignored; mutation-verified
- [ ] `make bench-compare` decode groups — policy wrapper must not regress decode throughput
- [ ] Sibling grep for remaining independent decode loops not routed/rejecting
- [ ] e2e-parity green

## Provenance

Evidence re-verified at 0699e60cc (ADR-080 C3). Related: #657 (merged), ADR-080 (#778).

ohdearquant:
Closed by PR #787 (merged): backend-neutral decode policy, apply-or-fail-closed (ADR-080 C3). PR lacked a closing keyword.

PR #778, opened 2026-07-09
docs(adr): ADR-080 — consolidation of duplicated numeric-contract helpers (audit 20260709)

ohdearquant:
> _PR description authored by Claude (Anthropic agent) on behalf of @ohdearquant._

## Summary

Adds ADR-080, proposing consolidation of four duplicated numeric-contract families in
`lattice-inference` identified by an internal duplication audit (run 2026-07-09):

- **C1 — attention-softmax**: shared fail-closed row-finalizer contract (zeroed row on
  non-positive denominator / NaN / +inf; no floor-clamping of `-inf` masks).
- **C2 — http-serve**: shared serving module for router construction, request
  normalization, and termination (`finish_reason`) serialization across `lattice.rs` and
  `lattice_serve.rs`.
- **C3 — sampling-decode**: backend-neutral decode-policy wrapper (stop-strings,
  reasoning-budget, logprobs) with explicit fail-closed rejection for unsupported fields.
- **C4 — matmul-gemm**: one checked, release-active GEMM argument validator shared across
  CPU/WGPU/Metal wrapper entry points, plus a distinct BitNet ternary-matvec validator.

**Decision**: consolidate within `lattice-inference` (no new crate) — every site selected for consolidation
across all four clusters lives in `lattice-inference` (fann/embed sites were surveyed and excluded). Backend-specific kernels
(CPU scalar/SIMD, Metal, WGPU) stay separate; only the shared validation/fail-closed
contract is extracted per cluster.

The ADR resolves 12 unfiled held findings from the audit plus 6 already-filed issues (18 total)
(#739-#741, #744-#746) via four cluster-scoped consolidation motions rather than
one-off single-site fixes, closing the recurring "same contract reimplemented at every
new call site" root cause. Full evidence tables, divergence details, and sequencing
(C1 → C2 → C3 → C4, softmax first) are in the ADR body.

Audit provenance: run at `13c8de8a3`, findings re-verified against `origin/main @
0699e60cc` (51 commits later) — see the ADR's Provenance section. Related issue range:
#739-#777.

## Test plan

- [x] Docs-only change — `make bench-compare` N/A (no source touched)
- [x] `docs/adr/INDEX.md` updated with the ADR-080 row (Proposed status, deps on
      ADR-058/064/066)
- [ ] Codex review of the ADR diff for factual consistency with `origin/main`
- [ ] Founder sign-off before dependent implementation PRs (C1-C4) begin

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #787, opened 2026-07-10
refactor(inference): backend-neutral decode policy, apply-or-fail-closed (ADR-080 C3)

ohdearquant:
## Contract

ADR-080 cluster C3: backend-neutral decode policy, apply-or-fail-closed.

Decode-policy controls (stop-string matching, reasoning-budget accounting,
logprobs formatting) live in one policy struct, `DecodePolicy`, in
`model/qwen35/generation.rs`. Construction (`DecodePolicy::init`) and every
per-step transition (`DecodePolicy::transition`) are the only two ways any
caller can drive the policy: `init` builds the struct, records the
prefill-derived first token's logprob in the same call, and constructs the
policy's private `StopMode` (`Disabled` / `Streaming` / `FullScan`) from the
real `gen_cfg.stop_strings` — the only place a `StopMode` value is ever
produced, so which stop-check behavior applies (or whether it applies at
all) is fixed by the config the policy was built from, not by anything a
call site writes. `transition` takes a backend-specific per-step forward
callback plus raw per-token I/O primitives for the stop check
(`decode_delta`, a `text`/`token_logprob_end_offsets` buffer pair, and
`emit_confirmed`) — each backend keeps `forward_step`, grammar
masking/advance, sampling, and EOS/stop-token checking entirely to itself,
and hands `transition` the token its own pipeline just sampled plus three
backend callbacks (grammar-advance, EOS/stop-token check, push into its own
token vectors) and those I/O primitives, which `DecodePolicy::stop_check`
alone uses to decide the stop outcome by dispatching on `self.stop_mode`.
`transition` returns whether to stop (grammar/EOS/stop-string/interrupt) or
the emitted token plus whether the answer-budget window has closed, having
already applied the reasoning-budget override, logprobs recording, and
reasoning-end capture in the process. Every decode loop either drives
through `init`+`transition`, or explicitly rejects the unsupported
`GenerateConfig` field at its entry point via the existing
`check_grammar_not_set` / `check_logprobs_not_set` precedent, extended here
with sibling `check_stop_strings_not_set` / `check_reasoning_budget_not_set`
/ `check_mtp_not_requested` guards. Silent omission — a caller sets a field,
gets no error, and the field is quietly ignored — is the defect this
cluster closes.

Three held findings from ADR-080 are also resolved here:

1. **Reasoning-budget bookkeeping and logprobs recording were hand-duplicated
   six times.** CPU `decode_loop`, `decode_loop_with_stops`, both branches of
   `generate_streaming_with_cancel` (all four in `model/qwen35/generation.rs`),
   and both Metal streaming loops (`generate_streaming` /
   `generate_streaming_with_cancel` and
   `generate_streaming_with_prefix_cache_and_cancel_inner`, in
   `forward/metal_qwen35.rs`) each independently reimplemented identical
   `think_close_id` resolution, `thinking_closed` / `reasoning_end_len`
   tracking, the `decode_cap` / `force_close_think` calls, the answer-budget
   break condition, and per-token logprobs recording — confirmed
   byte-for-byte identical across the CPU/Metal boundary. All six now drive
   the same `DecodePolicy::init` / `DecodePolicy::transition` calls instead.
2. `generate_greedy_mtp` / `generate_greedy_self_spec` each carried their own
   inline `max_by` argmax, which keeps the **last** tied index on equal
   logits rather than the first — inconsistent with the rest of the crate's
   tie-break convention (issue #280 class). Both now call one shared
   `crate::sampling::argmax_f32_first_wins`; a second surviving last-wins
   `max_by` in `mtp_forward_one`'s draft-token selection (missed in the
   initial sweep) is fixed the same way.
3. `mtp_route_active` / `self_spec_route_active` already excluded
   `stop_strings` and `reasoning_budget` from their fast-path eligibility but
   not `repetition_penalty` — a config with a non-identity penalty could
   silently enter a fast path that selects via raw argmax and never applies
   it (an output-correctness gap, not just perf: MTP/self-spec never call
   `sample_token`). Both predicates now exclude `repetition_penalty != 1.0`
   as well.

`DecodePolicy`'s constituent operations (budget override, emitted
bookkeeping, logprobs recording, reasoning-end capture, answer-budget check)
are private to `model/qwen35/generation.rs` — `init`/`transition` are the
only way any of the six sites, three of them in a different module
(`forward/metal_qwen35.rs`), can reach them, so a call site can no longer
apply a subset and silently skip another. Stop-string matching itself is
owned by `StopMode`, a private enum on `DecodePolicy` with no public
constructor other than `DecodePolicy::init`: `Streaming` wraps
`StopStringMatcher`'s incremental byte-holdback semantics (a partial match
must never reach the caller's confirmed-text sink) for streaming callers,
`FullScan` wraps the non-streaming full-text rescan
(`earliest_stop_match` / `earliest_stop_match_from`), and `Disabled` covers
an empty `gen_cfg.stop_strings` — genuinely different consumption shapes of
the same shared components, selected once from the real config rather than
by a closure a call site could supply. A caller cannot construct, name, or
swap a `StopMode` at all (the type isn't reachable outside
`generation.rs`); omitting the stop check entirely is a compile error
(`transition`'s I/O-primitive parameters are non-optional), not a silent
gap.

## Per-site disposition

| Site | Verdict | Notes |
|---|---|---|
| `model/qwen35/generation.rs` — `Qwen35Model::generate` (CPU canonical, direct) | routed via `DecodePolicy::init`/`transition` | `decode_loop` / `decode_loop_with_stops` both drive the one atomic transition |
| `model/qwen35/generation.rs` — `generate_streaming` / `generate_streaming_with_cancel` (CPU canonical, streaming) | routed via `DecodePolicy::init`/`transition` | both inline loops (fast-path and stop-string-path) drive the transition; grammar/sampling/detokenizer logic untouched |
| `forward/cpu_f16.rs` — `generate_f16` | rejects | previously silently omitted both fields |
| `forward/cpu_q8.rs` — `generate_q8` | rejects | previously silently omitted both fields |
| `forward/neon_forward.rs` — `generate_q8_neon` | rejects | previously silently omitted both fields |
| `forward/batch_prefill.rs` — `Qwen35Model::generate_with_batch_prefill` | rejects | previously silently omitted both fields |
| `forward/metal_qwen35.rs` — `MetalQwen35State::generate` (direct) | mixed | `stop_strings` already routed (#657); `reasoning_budget` + `logprobs` now reject |
| `forward/metal_qwen35.rs` — `generate_streaming` / `generate_streaming_with_cancel` | routed via `DecodePolicy::init`/`transition` | drives the same transition as the CPU loops (re-exported `pub(crate)`, gated identically to its Metal-only consumer) |
| `forward/metal_qwen35.rs` — `generate_streaming_with_prefix_cache_and_cancel_inner`, wrapped by `generate_streaming_with_prefix_cache_and_cancel` | routed via `DecodePolicy::init`/`transition`, rejects logprobs + active MTP | the public wrapper now runs `check_logprobs_not_set` + `check_mtp_not_requested` before any cache/state mutation, so a rejection never evicts a warm cross-turn entry |
| `forward/metal_qwen35.rs` — `generate_greedy_mtp` (MTP fast path) | exempt-with-reason | draft/verify loop can't support stop-strings, reasoning-budget, non-identity repetition-penalty, or logprobs; route predicate now excludes all four, falls through to `generate` (which itself rejects/honors `logprobs`/`enable_mtp`). Draft-token selection now uses the shared first-wins argmax |
| `forward/metal_qwen35.rs` — `generate_greedy_self_spec` (self-spec fast path) | exempt-with-reason | same as MTP |
| `forward/metal_qwen35.rs` — `generate_multimodal` | mixed | `stop_strings` already routed; `reasoning_budget` now rejects via `multimodal_generate_preflight` |
| `generate.rs` (legacy `Qwen3`/`QwenModel`) | exempt-with-reason | different config schema entirely — no `stop_strings`/`reasoning_budget` fields exist on it |
| `bin/*.rs` (all CLI tools) | routed (transitively) | all call into the entry points above; no independent decode loop found |

Sibling grep for completion: `grep -rl "sample_token(\|GenerateConfig" crates/inference/src` enumerated every file touching the Qwen3.5 decode-policy surface; every hit is accounted for above. Sibling grep for the argmax fix: `grep -n "max_by\b" crates/inference/src/forward/metal_qwen35.rs` returns only the fixed `mtp_forward_one` site (replaced) and two `#[cfg(test)]`-gated oracle helpers in sibling test modules — no other production last-wins argmax remains.

## Codex round-2 review — 4 findings, all fixed

Codex round-2 (head `09a33838d`) returned REJECT: 1 blocker, 1 major (a held
architecture-contract fork escalated for a ruling before implementation), 2
mediums. All four are fixed in this PR's current head.

- **Blocker** — `gen_cfg.logprobs` was not gated on the MTP/self-spec fast
  paths, and the direct Metal `generate()` entry point had no
  `check_logprobs_not_set` preflight at all: all three silently returned
  empty `token_logprobs` instead of rejecting the request. Fixed by adding
  `logprobs.is_none()` to `mtp_route_active` / `self_spec_route_active`, and
  a `check_logprobs_not_set` preflight at the top of `generate()`.
- **Medium** — the prefix-cache path's public wrapper caught every `Err`
  from its inner implementation and unconditionally called `reset_state()`
  plus evicted the cache slot, so a preflight-only rejection (e.g.
  `logprobs`) destroyed a valid pre-existing warm cross-turn entry. Fixed by
  hoisting the preflight checks to the public wrapper, before the
  destructive `Err`-recovery match.
- **Medium** — the same prefix-cache path accepted
  `GenerateConfig::enable_mtp` but never read or rejected it, unlike the
  direct Metal entry point. Fixed with a new sibling guard,
  `check_mtp_not_requested`, called alongside the logprobs guard in the
  hoisted wrapper location.
- **Major** (Leo's ruling: extend the design, do not amend ADR-080, scoped
  to exactly two additions) —
  1. `DecodePolicy::new` → `DecodePolicy::init`, which now records the
     prefill-derived first token's logprob as part of construction, killing
     three duplicated freestanding `record_logprob` call sites that
     previously lived outside the policy.
  2. `DecodePolicy::transition` gained a sixth, mandatory `stop_check`
     closure parameter. Omitting it is a compile error (verified:
     `E0061: this method takes 9 arguments but 8 arguments were supplied`).
     All six call sites now supply this closure explicitly (the
     no-stop-strings fast paths supply `|_, _| StopCheckOutcome::Continue`
     rather than omitting the parameter).

Every fix is mutation-verified (reverse-apply in place → confirm red →
restore → confirm green) or backed by the compiler-error observation for
the two structural guarantees (this and the round-1 `E0624` privacy proof).
Full per-finding mutation log kept locally (gitignored
`.khive/codex_reviews/`, not part of this diff).

## Codex round-3 review — 1 major + 2 mediums, all fixed

Codex round-3 (head `69d14b7ed`) returned APPROVE-WITH-FIXES: every
round-1/round-2 output-bug concern was confirmed closed (route-predicate
logprobs gating, direct-Metal ordering, prefix-cache ordering/MTP-guard,
`init`'s first-token logprob ownership, the full `GenerateConfig` field
re-sweep, CI green). What remained was hardening the round-2 `stop_check`
closure parameter itself: it could still compile as a trivial
`|_, _| StopCheckOutcome::Continue` for **any** configuration, including one
with real `stop_strings` set — the API could not structurally distinguish
"a real matcher adapter" from "a no-op."

- **Major** — `DecodePolicy::transition`'s `stop_check` closure parameter
  was an arbitrary outcome-producing callback: nothing prevented a call site
  from claiming `Continue` without actually checking anything, for a config
  that genuinely has stop strings. Fixed by replacing the closure with a
  private `StopMode` enum (`Disabled` / `Streaming(StopStringMatcher)` /
  `FullScan { stop_strings, max_stop }`) owned by `DecodePolicy` as a new
  `stop_mode` field, constructed exactly once via the private
  `StopMode::for_config(stop_strings, streaming)` inside `DecodePolicy::init`
  (which gained a trailing `streaming: bool` parameter) — no call site can
  construct, pass in, or swap a `StopMode` independently of the config the
  policy was built from; the type isn't even nameable outside
  `generation.rs` (verified: `error[E0603]: module 'generation' is private`
  when referenced from `forward/metal_qwen35.rs`). `transition` now takes
  only raw I/O primitives (`decode_delta`, a `text: &mut String` buffer, a
  `token_logprob_end_offsets` buffer, and `emit_confirmed`) and a new private
  `stop_check` method drives the actual decision from `self.stop_mode`. A
  new `pub(crate) check_initial_stop` covers the prefill-derived first
  token's check (before the decode loop) using the *same* `stop_mode`
  instance `transition` keeps driving — critical for `Streaming`'s
  byte-holdback continuity across the first/second-token boundary, a real
  correctness risk in a naive two-matcher design caught during this fix. A
  new `finish_stop` mirrors the natural-end tail flush the same way. All 6
  `transition` call sites (CPU `decode_loop`, `decode_loop_with_stops`, both
  branches of CPU `generate_streaming_with_cancel`, both Metal streaming
  functions) converted; `StopStringMatcher` lost its lifetime parameter
  (owns a cloned `Vec<String>` instead of borrowing) so it could be embedded
  directly in `StopMode::Streaming` without threading a lifetime through
  every `DecodePolicy` call site.
- **Medium** — `crate::sampling::record_logprob` was `pub(crate)` and
  imported directly into `generation.rs`, so any sibling decode site could
  call it and independently append to `token_logprobs`, recreating the
  duplicate-choreography bug this cluster closes for the other four
  constituent methods. Fixed by deleting the freestanding function: the
  pure computation (`compute_step_logprobs`) is now `pub(crate)`, but the
  `token_logprobs.push(...)` mutation lives only inside the already-private
  `DecodePolicy::record_logprob`. Verified: a probe call to
  `crate::sampling::record_logprob(...)` from `forward/metal_qwen35.rs` now
  fails with `error[E0425]: cannot find function 'record_logprob' in module
  'crate::sampling'` — the function no longer exists to be called from
  anywhere outside `DecodePolicy`.
- **Medium** — the PR's bench-compare claim was wrong: `elementwise_cpu_bench`
  (the only `lattice-inference` group `make bench-compare` runs by default)
  times standalone norm/activation/softmax kernels directly and never
  constructs a `DecodePolicy`, calls `transition`, or enters a generation
  loop, so its clean result did not support the claim that it was "the
  group this refactor could affect." No existing bench in
  `crates/inference/benches/` drives `Qwen35Model::generate`/
  `generate_streaming` either. Building a new Criterion harness was judged
  out of this round's scope (new bench infrastructure is itself new
  reviewable surface); corrected below instead of crediting
  `elementwise_cpu_bench`.

All three fixes are mutation/compile-fail-verified (reverse-apply →
observe red/compile-error → restore → confirm green). Mutation log updated
locally (gitignored `.khive/codex_reviews/`, not part of this diff).

## Testing

- 27 new/changed sites across both codex rounds, each covered by a
  dedicated unit test, all mutation-verified.
- New round-2 tests: `mtp_route_blocked_by_set_logprobs`,
  `self_spec_route_blocked_by_set_logprobs`,
  `metal_generate_plain_rejects_logprobs_config`,
  `generate_streaming_with_prefix_cache_rejects_logprobs_request_preserves_warm_entry`
  (warms a real cross-turn entry, makes a rejected `logprobs` call, asserts
  the same token ids are still cached afterward),
  `generate_streaming_with_prefix_cache_rejects_active_mtp_request`,
  `init_records_the_prefill_tokens_logprob_before_any_transition_call`
  (uses `max_new_tokens: 1` so `transition` never executes, isolating that
  the sole logprob can only have come from `init`).
- Full `cargo test -p lattice-inference --lib --features f16,metal-gpu -- --test-threads=1`, run under the GPU lock after this PR's final head: **1795 passed; 2 failed; 20 ignored** (4041.85s). Both failures are confirmed pre-existing and unrelated:
  - `f16_kv_metal_path_executes_and_reports_capability` — gated on unset `LATTICE_KV_F16=1`, same confirmed-pre-existing case from the prior round.
  - `metal_qwen35_chunked_prefill_long_prompt_matches_step_loop` — failed only the test's own loose logit-value regression guard (`max_abs_diff=0.848629` vs its `<0.5` bound); **argmax matched exactly** (271==271). The test's own in-source comment ("KNOWN ISSUE (2026-06-03)") documents this as a pre-existing, nondeterministic, occupancy-sensitive GPU race in `gdn_recurrence_fused`/`decode_attention`, confirmed present verbatim on `origin/main` before any change in this PR. Neither `forward_prefill`, `forward_step`, nor either kernel is touched by this PR's diff.
- Targeted Metal test run (route predicates, prefix-cache rejection tests, direct-generate rejection tests): 15 passed, 0 failed.
- Default-feature `cargo test -p lattice-inference --lib` (no Metal required): 1629 passed, 0 failed, 18 ignored.
- `cargo clippy --workspace -- -D warnings` and `cargo fmt --check`: clean under both default features and `--features f16,metal-gpu`.
- Greedy-token output is byte-identical on every routed path: the redesign only replaces local bookkeeping and separate method calls with atomic calls in the same order — no sampling, forward, grammar, or stop-condition logic changed. e2e-parity (greedy token match vs HF transformers) is CI-gated and confirms this on the canonical CPU path.
- **Round-3 (this head, `34161ce7f`):** `cargo test -p lattice-inference --lib --features f16` (default, no Metal): **1632 passed, 0 failed, 18 ignored** (124.09s). Full `forward::metal_qwen35::` module under the GPU lock, `--test-threads=4`: **236 passed, 3 failed** (3840.11s) — triaged per this repo's flaky-vs-deterministic protocol (re-run solo, `--test-threads=1`, idle GPU): `f16_kv_metal_path_executes_and_reports_capability` is the same env-gated (`LATTICE_KV_F16` unset) pre-existing case noted above; `metal_qwen35_chunked_prefill_length_1_final_chunk_matches_step_loop` and `metal_qwen35_prefill_all_logits_long_prompt_no_longer_panics` both went **green solo** (2 passed, 0 failed, 140.97s) — same argmax-matches-but-tight-logit-guard contended-GPU signature as the round-1/2 KNOWN-ISSUE case above, not a regression; neither test's code path is touched by this round's diff. Targeted new-adapter tests (12 Metal stop-string/route tests, 3 new CPU `DecodePolicy`/`record_logprob` tests): all green. `cargo clippy -p lattice-inference --lib` and `cargo clippy --workspace`, both `--features f16` and `--features f16,metal-gpu`, `-D warnings`: clean, all 4 combinations. `cargo fmt --check`: clean.

## Bench-compare

**Correction (codex round-3 medium):** an earlier version of this section
credited `elementwise_cpu_bench`'s clean Qwen3.5-adjacent group as evidence
this refactor doesn't regress performance. That was wrong: that bench times
standalone norm/activation/softmax kernels directly and never constructs a
`DecodePolicy`, calls `transition`, or runs a generation loop at all — it
cannot observe anything this PR changes. No bench in
`crates/inference/benches/` drives `Qwen35Model::generate` /
`generate_streaming` (the closest, `inference_bench.rs`, is a BERT-only
encode/layer-op bench). Building a new Criterion harness for
`DecodePolicy::transition`/orchestration overhead was judged out of this
round's scope. **This refactor is therefore unmeasured by the default
`make bench-compare` suite** — it is a like-for-like call-order
restructuring (same branches, same allocations, same per-step work moved
behind method calls instead of inline), so no throughput change is
expected, but that expectation is architectural reasoning, not a
measurement, and should be read as such.

`make bench-compare`, base `origin/main` @ 65865ecb5, head @ 69d14b7ed (this PR, rounds 1-2): **14 FAIL / 14 WARN / 59 WIN**. Every FAIL/WARN row is either (a) one of `crates/embed`'s `simd`/`int8`/`tier_prepared_batch_sizes` benches — a crate this PR touches zero lines of, with the same bench families showing both large FAILs and large WINs on sibling size/shape variants of the identical op (the documented contended-machine-noise signature; 3-4 sibling bench jobs were queued/contending on this machine's heavy lane during the run window), or (b) `residual_add_layer_norm/unfused/hidden384_seq128` (+5.88% WARN), a BERT CPU norm kernel this PR's diff never touches. Neither category exercises the decode-policy code path this PR actually changes (see correction above) — they are cited only as evidence nothing in `crates/embed`/BERT-CPU regressed, not as evidence about `DecodePolicy` itself.

**Round-3 re-run attempted, contended — not usable as evidence.** `make bench-compare` was re-run twice against this round's head (`34161ce7f`) but both runs landed while unrelated fleet activity was hammering this shared machine (a separate lambda's `kkernel-bench` process + a large `rustfmt` sweep, load average 36-44 at the time). The second attempt showed +90% to +277% "regressions" exclusively in `crates/embed`'s `simd`/`int8` benches — the same untouched crate/bench-family as the round-1/2 noise above, but at a magnitude an order of magnitude larger than anything in this PR's diff could plausibly cause, confirming machine contention rather than a real signal. No clean, uncontended re-run was obtained this round. This does not change the substance of the correction above: the default suite still does not exercise `DecodePolicy`/`transition`, so this refactor remains unmeasured by `make bench-compare` regardless of contention.

## Scope note

The ADR-080 C3 checklist's first line item — a shared policy struct wrapping
a per-step callback in `model/qwen35/generation.rs` — is implemented as
`DecodePolicy::init` + `DecodePolicy::transition`, described above. The
defect this cluster exists to close (silent field omission, plus the
duplicated-bookkeeping drift risk) is fully closed by this PR, including
every instance codex's round-1 and round-2 reviews found in the initial
implementation.

Refs: #783, #778 (ADR-080)

### DIFF
--- a/crates/inference/src/forward/batch_prefill.rs
+++ b/crates/inference/src/forward/batch_prefill.rs
@@ -26,7 +26,8 @@ use crate::forward::cpu::{elementwise_mul, matmul_bt, silu_inplace};
 use crate::model::qwen35::{
     AttentionWeights, CommonLayerWeights, DenseFfnWeights, FeedForwardWeights, ForwardScratch,
     FullAttentionLayerWeights, KvCache, Qwen35Model, check_grammar_not_set, check_logprobs_not_set,
-    decode_tokens, qwen35_rms_norm, resize, sample_token, should_stop_token,
+    check_reasoning_budget_not_set, check_stop_strings_not_set, decode_tokens, qwen35_rms_norm,
+    resize, sample_token, should_stop_token,
 };
 use crate::model::qwen35_config::{GenerateConfig, GenerateOutput, Qwen35Config};
 use crate::stop_reason::StopReason;
@@ -432,6 +433,10 @@ impl Qwen35Model {
         // Same rationale for logprobs capture, which is also not wired into this
         // generate loop (#585).
         check_logprobs_not_set(gen_cfg)?;
+        // Same rationale for stop_strings matching and reasoning-budget forcing,
+        // neither of which is wired into this generate loop (ADR-080 C3, #783).
+        check_stop_strings_not_set(gen_cfg)?;
+        check_reasoning_budget_not_set(gen_cfg)?;
         // Context preflight. apply_partial_rope indexes the precomputed cos/sin table
         // without bounds checks, so a position at or past max_context() is an out-of-bounds
         // slice access — a release panic, not a clean error. Mirror the same total-token
@@ -2072,6 +2077,60 @@ mod tests {
             matches!(result, Err(InferenceError::InvalidInput(_))),
             "generate_with_batch_prefill must fail closed with InvalidInput when grammar is \
              set (#397/#398); got {result:?}"
+        );
+    }
+
+    /// `generate_with_batch_prefill` must reject a `GenerateConfig` that sets
+    /// `stop_strings` with a typed `InvalidInput` error before sampling any token
+    /// (ADR-080 C3, #783).
+    ///
+    /// Mutation sensitivity: removing `check_stop_strings_not_set` causes the
+    /// function to proceed through the full prefill + decode path with the tiny
+    /// weights, returning `Ok(...)`. The `matches!` assert then fails.
+    #[test]
+    fn generate_with_batch_prefill_rejects_stop_strings_config_before_sampling() {
+        use crate::error::InferenceError;
+
+        let cfg = tiny_test_config();
+        let model = build_random_model(cfg, 0x1234_5678_abcd_ef01);
+
+        let gen_cfg = GenerateConfig {
+            stop_strings: vec!["</s>".to_string()],
+            ..Default::default()
+        };
+
+        let result = model.generate_with_batch_prefill("a b c", &gen_cfg);
+        assert!(
+            matches!(result, Err(InferenceError::InvalidInput(_))),
+            "generate_with_batch_prefill must fail closed with InvalidInput when \
+             stop_strings is set (ADR-080 C3, #783); got {result:?}"
+        );
+    }
+
+    /// `generate_with_batch_prefill` must reject a `GenerateConfig` that sets
+    /// `reasoning_budget` with a typed `InvalidInput` error before sampling any
+    /// token (ADR-080 C3, #783).
+    ///
+    /// Mutation sensitivity: removing `check_reasoning_budget_not_set` causes the
+    /// function to proceed through the full prefill + decode path with the tiny
+    /// weights, returning `Ok(...)`. The `matches!` assert then fails.
+    #[test]
+    fn generate_with_batch_prefill_rejects_reasoning_budget_config_before_sampling() {
+        use crate::error::InferenceError;
+
+        let cfg = tiny_test_config();
+        let model = build_random_model(cfg, 0x1234_5678_abcd_ef01);
+
+        let gen_cfg = GenerateConfig {
+            reasoning_budget: Some(16),
+            ..Default::default()
+        };
+
+        let result = model.generate_with_batch_prefill("a b c", &gen_cfg);
+        assert!(
+            matches!(result, Err(InferenceError::InvalidInput(_))),
+            "generate_with_batch_prefill must fail closed with InvalidInput when \
+             reasoning_budget is set (ADR-080 C3, #783); got {result:?}"
         );
     }
 
--- a/crates/inference/src/forward/cpu_f16.rs
+++ b/crates/inference/src/forward/cpu_f16.rs
@@ -891,6 +891,10 @@ pub fn generate_f16(
     // Same rationale for logprobs capture, which is also not wired into this
     // generate loop (#585).
     crate::model::qwen35::check_logprobs_not_set(gen_cfg)?;
+    // Same rationale for stop_strings matching and reasoning-budget forcing,
+    // neither of which is wired into this generate loop (ADR-080 C3, #783).
+    crate::model::qwen35::check_stop_strings_not_set(gen_cfg)?;
+    crate::model::qwen35::check_reasoning_budget_not_set(gen_cfg)?;
 
     // Context preflight. The RoPE cos/sin tables are indexed unchecked in
     // forward_step_f16 (`rope.cos_at(base + i)`), so a position at or past the
@@ -1884,6 +1888,90 @@ mod tests {
             matches!(result, Err(InferenceError::InvalidInput(_))),
             "generate_f16 must fail closed with InvalidInput when grammar is set (#397/#398); \
              got {result:?}"
+        );
+    }
+
+    /// `generate_f16` must reject a `GenerateConfig` that sets `stop_strings` with a
+    /// typed `InvalidInput` error before sampling any token (ADR-080 C3, #783).
+    ///
+    /// Mutation sensitivity: removing the `check_stop_strings_not_set` call makes the
+    /// function proceed past the guard and attempt to forward with empty weights,
+    /// producing a panic or a non-`InvalidInput` error — this assert fails either way.
+    #[test]
+    fn generate_f16_rejects_stop_strings_config_before_sampling() {
+        use crate::error::InferenceError;
+        use std::collections::HashMap;
+
+        let mut vocab: HashMap<String, u32> = HashMap::new();
+        for (i, c) in ["h", "e", "l", "o"].iter().enumerate() {
+            vocab.insert((*c).to_string(), i as u32);
+        }
+        let merges = vec![
+            ("h".to_string(), "e".to_string()),
+            ("he".to_string(), "l".to_string()),
+        ];
+        let tokenizer = BpeTokenizer::from_vocab_and_merges(vocab, merges).unwrap();
+
+        let cfg = Qwen35Config::qwen35_2b();
+        let rope = RopeTable::new(cfg.rope_dim(), 8, cfg.rope_theta);
+        let weights = F16ModelWeights {
+            embed_tokens: vec![],
+            final_norm: vec![],
+            layers: vec![],
+        };
+
+        let gen_cfg = GenerateConfig {
+            stop_strings: vec!["</s>".to_string()],
+            ..Default::default()
+        };
+
+        let result = generate_f16(&weights, &cfg, &tokenizer, &rope, "hello", &gen_cfg);
+        assert!(
+            matches!(result, Err(InferenceError::InvalidInput(_))),
+            "generate_f16 must fail closed with InvalidInput when stop_strings is set \
+             (ADR-080 C3, #783); got {result:?}"
+        );
+    }
+
+    /// `generate_f16` must reject a `GenerateConfig` that sets `reasoning_budget` with
+    /// a typed `InvalidInput` error before sampling any token (ADR-080 C3, #783).
+    ///
+    /// Mutation sensitivity: removing the `check_reasoning_budget_not_set` call makes
+    /// the function proceed past the guard and attempt to forward with empty weights,
+    /// producing a panic or a non-`InvalidInput` error — this assert fails either way.
+    #[test]
+    fn generate_f16_rejects_reasoning_budget_config_before_sampling() {
+        use crate::error::InferenceError;
+        use std::collections::HashMap;
+
+        let mut vocab: HashMap<String, u32> = HashMap::new();
+        for (i, c) in ["h", "e", "l", "o"].iter().enumerate() {
+            vocab.insert((*c).to_string(), i as u32);
+        }
+        let merges = vec![
+            ("h".to_string(), "e".to_string()),
+            ("he".to_string(), "l".to_string()),
+        ];
+        let tokenizer = BpeTokenizer::from_vocab_and_merges(vocab, merges).unwrap();
+
+        let cfg = Qwen35Config::qwen35_2b();
+        let rope = RopeTable::new(cfg.rope_dim(), 8, cfg.rope_theta);
+        let weights = F16ModelWeights {
+            embed_tokens: vec![],
+            final_norm: vec![],
+            layers: vec![],
+        };
+
+        let gen_cfg = GenerateConfig {
+            reasoning_budget: Some(16),
+            ..Default::default()
+        };
+
+        let result = generate_f16(&weights, &cfg, &tokenizer, &rope, "hello", &gen_cfg);
+        assert!(
+            matches!(result, Err(InferenceError::InvalidInput(_))),
+            "generate_f16 must fail closed with InvalidInput when reasoning_budget is set \
+             (ADR-080 C3, #783); got {result:?}"
         );
     }
 
--- a/crates/inference/src/forward/cpu_q8.rs
+++ b/crates/inference/src/forward/cpu_q8.rs
@@ -743,6 +743,10 @@ pub fn generate_q8(
     // Same rationale for logprobs capture, which is also not wired into this
     // generate loop (#585).
     crate::model::qwen35::check_logprobs_not_set(gen_cfg)?;
+    // Same rationale for stop_strings matching and reasoning-budget forcing,
+    // neither of which is wired into this generate loop (ADR-080 C3, #783).
+    crate::model::qwen35::check_stop_strings_not_set(gen_cfg)?;
+    crate::model::qwen35::check_reasoning_budget_not_set(gen_cfg)?;
 
     // Context preflight. The RoPE cos/sin tables are indexed unchecked in
     // full_attention_step_q8 (`rope.cos_at(position * half + i)`), so a position
@@ -1595,6 +1599,93 @@ mod tests {
             matches!(result, Err(InferenceError::InvalidInput(_))),
             "generate_q8 must fail closed with InvalidInput when grammar is set (#397/#398); \
              got {result:?}"
+        );
+    }
+
+    /// `generate_q8` must reject a `GenerateConfig` that sets `stop_strings` with a
+    /// typed `InvalidInput` error before sampling any token (ADR-080 C3, #783).
+    ///
+    /// Mirrors `generate_q8_rejects_grammar_config_before_sampling`'s structure: the
+    /// guard fires before any weight dereference, so empty weight vecs are sufficient.
+    ///
+    /// Mutation sensitivity: removing the `check_stop_strings_not_set` call makes the
+    /// function proceed past the guard and attempt to forward with empty weights,
+    /// producing a panic or a non-`InvalidInput` error — this assert fails either way.
+    #[test]
+    fn generate_q8_rejects_stop_strings_config_before_sampling() {
+        use crate::error::InferenceError;
+        use std::collections::HashMap;
+
+        let mut vocab: HashMap<String, u32> = HashMap::new();
+        for (i, c) in ["h", "e", "l", "o"].iter().enumerate() {
+            vocab.insert((*c).to_string(), i as u32);
+        }
+        let merges = vec![
+            ("h".to_string(), "e".to_string()),
+            ("he".to_string(), "l".to_string()),
+        ];
+        let tokenizer = BpeTokenizer::from_vocab_and_merges(vocab, merges).unwrap();
+
+        let cfg = Qwen35Config::qwen35_2b();
+        let rope = RopeTable::new(cfg.rope_dim(), 8, cfg.rope_theta);
+        let weights = Q8ModelWeights {
+            embed_tokens: vec![],
+            final_norm: vec![],
+            layers: vec![],
+        };
+
+        let gen_cfg = GenerateConfig {
+            stop_strings: vec!["</s>".to_string()],
+            ..Default::default()
+        };
+
+        let result = generate_q8(&weights, &cfg, &tokenizer, &rope, "hello", &gen_cfg);
+        assert!(
+            matches!(result, Err(InferenceError::InvalidInput(_))),
+            "generate_q8 must fail closed with InvalidInput when stop_strings is set \
+             (ADR-080 C3, #783); got {result:?}"
+        );
+    }
+
+    /// `generate_q8` must reject a `GenerateConfig` that sets `reasoning_budget` with
+    /// a typed `InvalidInput` error before sampling any token (ADR-080 C3, #783).
+    ///
+    /// Mutation sensitivity: removing the `check_reasoning_budget_not_set` call makes
+    /// the function proceed past the guard and attempt to forward with empty weights,
+    /// producing a panic or a non-`InvalidInput` error — this assert fails either way.
+    #[test]
+    fn generate_q8_rejects_reasoning_budget_config_before_sampling() {
+        use crate::error::InferenceError;
+        use std::collections::HashMap;
+
+        let mut vocab: HashMap<String, u32> = HashMap::new();
+        for (i, c) in ["h", "e", "l", "o"].iter().enumerate() {
+            vocab.insert((*c).to_string(), i as u32);
+        }
+        let merges = vec![
+            ("h".to_string(), "e".to_string()),
+            ("he".to_string(), "l".to_string()),
+        ];
+        let tokenizer = BpeTokenizer::from_vocab_and_merges(vocab, merges).unwrap();
+
+        let cfg = Qwen35Config::qwen35_2b();
+        let rope = RopeTable::new(cfg.rope_dim(), 8, cfg.rope_theta);
+        let weights = Q8ModelWeights {
+            embed_tokens: vec![],
+            final_norm: vec![],
+            layers: vec![],
+        };
+
+        let gen_cfg = GenerateConfig {
+            reasoning_budget: Some(16),
+            ..Default::default()
+        };
+
+        let result = generate_q8(&weights, &cfg, &tokenizer, &rope, "hello", &gen_cfg);
+        assert!(
+            matches!(result, Err(InferenceError::InvalidInput(_))),
+            "generate_q8 must fail closed with InvalidInput when reasoning_budget is set \
+             (ADR-080 C3, #783); got {result:?}"
         );
     }
 
--- a/crates/inference/src/forward/metal_qwen35.rs
+++ b/crates/inference/src/forward/metal_qwen35.rs
@@ -558,6 +558,13 @@ mod mtp_resolve_tests {
 /// `gen_cfg.stop_strings.is_empty()` is load-bearing: MTP's draft/verify loop
 /// has no string-stop awareness, so any non-empty `stop_strings` must route
 /// around it to the plain loop (which does check stops after every token).
+///
+/// `gen_cfg.reasoning_budget.is_none()` is the same kind of load-bearing
+/// route-around (ADR-080 C3, #783): MTP's draft/verify loop has no
+/// budget-forcing awareness (no `decode_cap` / `force_close_think` wiring),
+/// so any set `reasoning_budget` must route around it to the plain loop,
+/// which either applies budget forcing or rejects the config via
+/// `check_reasoning_budget_not_set`.
 #[cfg(any(test, all(target_os = "macos", feature = "metal-gpu")))]
 fn mtp_route_active(
     mtp_present: bool,
@@ -572,12 +579,13 @@ fn mtp_route_active(
         && !use_compact
         && gen_cfg.grammar.is_none()
         && gen_cfg.stop_strings.is_empty()
+        && gen_cfg.reasoning_budget.is_none()
 }
 
 /// Whether `generate()` should take the GDN-first self-speculative greedy
 /// branch. Mirrors the `use_self_spec` computation inline in
-/// `MetalQwen35State::generate` exactly -- same `stop_strings.is_empty()`
-/// route-around rationale as [`mtp_route_active`].
+/// `MetalQwen35State::generate` exactly -- same `stop_strings.is_empty()` /
+/// `reasoning_budget.is_none()` route-around rationale as [`mtp_route_active`].
 #[cfg(any(test, all(target_os = "macos", feature = "metal-gpu")))]
 fn self_spec_route_active(
     gdn_checkpoints_present: bool,
@@ -593,6 +601,7 @@ fn self_spec_route_active(
         && !use_compact
         && gen_cfg.grammar.is_none()
         && gen_cfg.stop_strings.is_empty()
+        && gen_cfg.reasoning_budget.is_none()
         && num_active_linear_attention_layers > 0
 }
 
@@ -661,6 +670,43 @@ mod route_predicate_tests {
             !self_spec_route_active(true, true, &gen_cfg, false, 1),
             "self-spec must not activate when stop_strings is non-empty -- \
              same string-stop-unaware draft/verify loop as MTP"
+        );
+    }
+
+    /// ADR-080 C3 (#783): a set `reasoning_budget` must route around MTP even
+    /// though every other MTP precondition is satisfied -- MTP's draft/verify
+    /// loop has no budget-forcing (`decode_cap` / `force_close_think`)
+    /// awareness, mirroring the `stop_strings` route-around above.
+    ///
+    /// Mutation sensitivity: removing the `reasoning_budget.is_none()` clause
+    /// from `mtp_route_active` makes this assertion fail (the route stays
+    /// active), which would silently drop the caller's reasoning budget once
+    /// MTP takes over decoding.
+    #[test]
+    fn mtp_route_blocked_by_set_reasoning_budget() {
+        let gen_cfg = GenerateConfig {
+            reasoning_budget: Some(64),
+            ..greedy_gen_cfg(vec![])
+        };
+        assert!(
+            !mtp_route_active(true, true, &gen_cfg, false),
+            "MTP must not activate when reasoning_budget is set -- its \
+             draft/verify loop has no budget-forcing awareness"
+        );
+    }
+
+    /// Sibling to `mtp_route_blocked_by_set_reasoning_budget` for the
+    /// self-speculative route (ADR-080 C3, #783).
+    #[test]
+    fn self_spec_route_blocked_by_set_reasoning_budget() {
+        let gen_cfg = GenerateConfig {
+            reasoning_budget: Some(64),
+            ..greedy_gen_cfg(vec![])
+        };
+        assert!(
+            !self_spec_route_active(true, true, &gen_cfg, false, 1),
+            "self-spec must not activate when reasoning_budget is set -- same \
+             budget-forcing-unaware draft/verify loop as MTP"
         );
     }
 }
@@ -7855,14 +7901,11 @@ mod inner {
                 };
             }
 
-            // Greedy argmax from prefill logits.
-            let pending_first = prefill_logits
-                .iter()
-                .copied()
-                .enumerate()
-                .max_by(|(_, a), (_, b)| a.partial_cmp(b).unwrap_or(std::cmp::Ordering::Equal))
-                .map(|(i, _)| i as u32)
-                .unwrap_or(0);
+            // Greedy argmax from prefill logits. Shared first-wins helper
+            // (ADR-080 C3, #783) rather than a hand-rolled `max_by`, which
+            // keeps the LAST tied maximum instead of the engine-wide-contract
+            // first tied maximum.
+            let pending_first = crate::sampling::argmax_f32_first_wins(prefill_logits);
 
             let is_stop = |id: u32| id == cfg.eos_token_id || gen_cfg.stop_token_ids.contains(&id);
 
@@ -8100,15 +8143,10 @@ mod inner {
                 };
             }
 
-            let argmax_logits = |logits: &[f32]| -> u32 {
-                logits
-                    .iter()
-                    .copied()
-                    .enumerate()
-                    .max_by(|(_, a), (_, b)| a.partial_cmp(b).unwrap_or(std::cmp::Ordering::Equal))
-                    .map(|(i, _)| i as u32)
-                    .unwrap_or(0)
-            };
+            // Shared first-wins argmax helper (ADR-080 C3, #783) rather than a
+            // hand-rolled `max_by` closure, which kept the LAST tied maximum
+            // instead of the engine-wide-contract first tied maximum.
+            let argmax_logits = crate::sampling::argmax_f32_first_wins;
 
             let is_stop = |id: u32| id == cfg.eos_token_id || gen_cfg.stop_token_ids.contains(&id);
 
@@ -8453,6 +8491,16 @@ mod inner {
                     token_logprobs: vec![],
                 });
             }
+
+            // Reject reasoning_budget before any prefill/state work (ADR-080 C3,
+            // #783): budget-forcing (`decode_cap` / `force_close_think`) is not
+            // wired into this decode loop, unlike `generate_streaming` /
+            // `generate_streaming_with_cancel` below, which do apply it. Without
+            // this guard the field would be silently ignored, producing
+            // unbudgeted output despite a reasoning budget being requested.
+            // `stop_strings` does not need an equivalent guard here: this loop
+            // already applies it via `stop_text_state` further down.
+            crate::model::qwen35::check_reasoning_budget_not_set(gen_cfg)?;
 
             // Fail-closed: a prompt longer than the KV cache would trip the
             // forward_prefill length assertion (n <= max_cache_len) and panic the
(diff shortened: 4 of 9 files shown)


ohdearquant:

## Review of PR #787 (round 2)

### Verdict: REJECT

The repetition-penalty fixes are real and mutation-sensitive, but the requested full-matrix audit found another output-affecting omission: both speculative predicates still admit `logprobs`, and both fast paths return an empty `token_logprobs` vector. The one-owner extraction also still narrows ADR-080 by leaving stop-string policy and all three initial-token logprob calls outside `DecodePolicy`.

### Prior review context

- Round 1 at `a9d435c88` found two blockers, one major, and two mediums.
- This round reviewed committed head `09a33838d` against `origin/main` `27b9047ee`.
- The fix commit touches only `metal_qwen35.rs`, `generation.rs`, and the `qwen35` re-export needed by those changes; its hunks are confined to the five reported fixes, their tests, and explanatory comments. I found no unrelated source change in that commit.
- GitHub had no review threads. At the final status read recorded below, current-head required CI was green; one non-aggregate q4 job remained pending.

### What this round examined

- Every round-1 finding and the complete `GenerateConfig` field matrix at the prefix-cache, MTP, and self-spec entry points.
- The accepted ADR-080 C3 ownership contract, issue #783, the current PR body, the full fix-commit diff, all six `transition` call sites, and the three prefill/initial-token logprob sites.
- Sibling searches for dense `max_by` argmax copies across `crates/inference/src`.
- Clean-head fmt, workspace clippy, targeted tests, two reverse mutations, and two added falsification tests in a detached disposable worktree.

### Findings (4 total: 1 blocker, 1 major, 2 mediums)

#### Blockers

1. [`crates/inference/src/forward/metal_qwen35.rs:584`] MTP and self-spec still silently accept and discard `logprobs`.
   - **Current**: `mtp_route_active` (`:584-592`) and `self_spec_route_active` (`:609-618`) gate grammar, stop strings, reasoning budget, and repetition penalty, but not `gen_cfg.logprobs`. With `logprobs: Some(0)`, both predicates can return true. `generate_greedy_mtp` and `generate_greedy_self_spec` then unconditionally return `token_logprobs: vec![]` at `:8211` and `:8524`. Routing around the fast paths alone would not close this: the direct Metal `generate` fallback also unconditionally returns an empty vector at `:8918` and has no `check_logprobs_not_set` preflight.
   - **Issue**: This is the same apply-or-fail-closed defect as round-1 blocker #1, now on both speculative routes: a caller requests an output field, receives `Ok`, and the requested output is silently absent. It directly contradicts the PR body's claim that the full silent-omission family is closed.
   - **Fix**: Call `check_logprobs_not_set(gen_cfg)` at the start of direct Metal `generate`, before prefill/session mutation, and add `gen_cfg.logprobs.is_none()` to both fast-path predicates so their eligibility contracts are complete. Add direct-entry and per-predicate rejection tests.
   - **Test**: In the disposable worktree I added `logprobs: Some(0)` to the otherwise-active predicate fixture and asserted both routes were inactive. Current head failed immediately because `mtp_route_active(...)` returned true; source inspection establishes the identical self-spec omission and the unconditional empty outputs.
   - **Confidence**: Observed by the added falsification test and derived from the two unconditional return values.

#### Majors

2. [`docs/adr/ADR-080-consolidation-duplicated-contracts.md:178`] The new transition closes partial use of five methods, but still does not satisfy the accepted one-owner contract.
   - **Current**: The five constituent methods are genuinely private (`generation.rs:1078-1138`), and all six decode loops call the one public(crate) `transition` (`generation.rs:702`, `:865`, `:1305`, `:1415`; `metal_qwen35.rs:13294`, `:15841`). That part of round-1 major #3 is closed. However, ADR-080 requires stop-string matching state, reasoning accounting, and logprobs formatting in the policy struct (`ADR-080:178-185`). The implementation explicitly leaves stop matching outside the API (`generation.rs:1013-1024`), so a site can call `transition` yet omit the stop matcher independently. The prefill-derived first token also still bypasses the policy through duplicated `record_logprob` calls at `generation.rs:235`, `generation.rs:603`, and `metal_qwen35.rs:13143`.
   - **Issue**: The PR prose argues that streaming and non-streaming need different stop-consumption shapes; that is technically valid, but it does not justify silently narrowing ownership. Mode-specific matcher adapters/state can still sit behind one owner. The current API does not make stop-policy omission impossible, and its “one owner” claim excludes the initial generated step. This remains a contract decision, not an implementation detail that comments can override.
   - **Fix**: Either (a) extend the policy/transition design with mode-specific stop handling and an initialization/first-step transition that owns the initial logprob record, or (b) amend ADR-080 and issue #783 explicitly before merge, then make the PR body state the narrower contract. The existing accepted ADR cannot be declared complete as written.
   - **Test**: Mutation-test omission of the single complete policy transition, including stop processing and the initial token, at each routed entry point. The current positive logprob test covers the returned vector but does not prove the initial step is policy-owned.
   - **Confidence**: Derived from the accepted ADR, the struct surface, all six loop call sites, and the three surviving initial-token calls.

#### Mediums

3. [`crates/inference/src/forward/metal_qwen35.rs:15319`] The prefix-cache rejection still mutates/evicts pre-existing cache state.
   - **Current**: The guard itself is correctly first in the inner function (`:15364`), but the public wrapper catches every `Err`, calls `reset_state()`, and removes the slot (`:15328-15334`). The committed regression test starts with no entry and only asserts it remains `None` (`:26693`, `:26727`), so it cannot prove the stated “before any cache/state mutation” property for a live cache.
   - **Issue**: An invalid `logprobs` request now fails closed, but it also destroys a valid cross-turn entry and session state even though no generation work began. This is observable cache behavior and contradicts both the source comment and the round-two acceptance condition.
   - **Fix**: Hoist the config preflight to the public wrapper before its destructive error-recovery block, or distinguish preflight `InvalidInput` from errors that can leave restored/prefilled state inconsistent. Add a test that warms a cache entry, makes the rejected call, and proves the same entry survives.
   - **Test**: I added exactly that warm-entry test in the disposable worktree. It failed because the post-error wrapper removed the entry.
   - **Confidence**: Observed by the added falsification test.

4. [`crates/inference/src/model/qwen35_config.rs:675`] The prefix-cache entry point silently ignores `enable_mtp`.
   - **Current**: `GenerateConfig::enable_mtp` requests MTP or defers to `LATTICE_MTP` (`:675-678`). Direct Metal `generate` resolves it at `metal_qwen35.rs:8691-8699`; the prefix-cache function accepts the same config but never reads or rejects the field.
   - **Issue**: This is the remaining non-output field exposed by the requested complete prefix-cache matrix audit. MTP is intended to preserve greedy output, so this is not blocker-level output corruption, but the API still silently ignores an explicit route request.
   - **Fix**: If cross-turn prefix caching cannot compose with MTP, reject an active MTP request before cache/state mutation and document that incompatibility. If `enable_mtp` is only a best-effort hint, say that in the field and entry-point rustdoc instead of presenting it as an enable switch.
   - **Confidence**: Derived from the complete field definition and all production reads of `enable_mtp`.

### Round-1 finding disposition

- **Blocker 1 — prefix-cache logprobs**: The silent `Ok` is fixed and the guard-removal mutation goes red, but the no-mutation part is not closed; see medium #3.
- **Blocker 2 — repetition penalty**: Closed. Both predicates now require identity penalty; their tests and the output-discriminating case pass, and removing both clauses makes exactly the two predicate tests fail.
- **Major 3 — one owner**: Partially closed. Method privacy and six-loop transition routing are correct; stop ownership and initial-token logprobs remain outside the owner; see major #2.
- **Medium 4 — `mtp_forward_one` argmax**: Closed. It uses `argmax_f32_first_wins` at `metal_qwen35.rs:5521`. The sibling search found only test-oracle `max_by` uses; no production dense-logit last-wins copy remains. Rust's official `Iterator::max_by` documentation confirms that equal maxima select the last element: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.max_by.
- **Medium 5 — testing characterization**: Closed. The current PR body now says the two numeric prefill failures passed in isolated head/base runs and characterizes the full-run behavior as load-sensitive/transient; it no longer claims deterministic pre-existing failures.

- `cargo fmt --all -- --check` and `git diff --check origin/main...HEAD` passed at committed head.
- `cargo clippy --workspace -- -D warnings` passed.
- The nine `route_predicate_tests` passed, including both repetition tests and the output-level winner-flip test.
- `transition_records_one_logprob_per_generated_token`, the prefix-cache logprobs rejection test, and both first-wins tie tests passed.
- Removing the prefix guard made its regression test fail with the original `Ok(... token_logprobs: [])` behavior.
- Removing both repetition-penalty clauses made the MTP and self-spec predicate tests fail; the other seven route tests stayed green.
- The fix commit contains no unrelated file or hunk outside the reported round-1 remediations, their tests, and supporting comments/re-export.

### Current-head CI

- Passed at the 2026-07-10T11:47:19-04:00 final status read: the aggregate `ci-gate`, macOS/Ubuntu x86/ARM CI, both feature matrices, MSRV, bench-compile, app binaries, cargo-deny, semver, e2e parity, q4 PPL, Metal attention parity, embed/wasm drift gates, actionlint, typos, rustdoc lint, and unwrap lint.
- `q4 composed golden` remained pending, but the required aggregate `ci-gate` was green and there were no failed current-head checks.

### What I did NOT check

- I did not rerun the 50+ minute full `f16,metal-gpu` library suite or `make bench-compare`; the PR body reports those runs, while this review ran the focused regressions and checked current GitHub jobs.
- I did not independently recreate the reported E0624 compiler transcript. Source visibility is conclusive that the five methods are module-private, and the Metal module can reach only `new`, `cap`, and `transition` through the crate-private re-export.
- I did not post this review, push, commit, or modify source in the requested worktree. All test additions and reverse mutations were confined to `/tmp/lattice-pr787-r2-mutation`.

- Another round is necessary. Recheck direct Metal `generate` plus both speculative predicates for `logprobs`, verify invalid prefix-cache requests preserve live cache state, and require an explicit ADR decision on stop-string/initial-step ownership. Stop once those items are closed and all current-head checks are green.

Domain utility: MEDIUM — the composed Rust visibility/invariant guidance helped distinguish genuine method privacy from complete contract ownership; the decisive evidence came from ADR-080, current source, and falsification tests.

Verdict: REJECT

ohdearquant:

## Review of PR #787 (round 3)

The round-2 output bugs are closed: the speculative predicates reject `logprobs`, direct Metal rejects before prefill/session mutation, prefix-cache rejection preserves a warm entry, and active MTP is rejected there. `DecodePolicy::init` also owns the prefill-derived token's actual logprob record. The remaining issue is narrower but still part of the maintainer-approved Option A contract: the API makes a ninth closure syntactically mandatory, but does not make a real stop adapter mandatory, and the logprob helper remains callable outside the owner. The submitted benchmark comparison also does not exercise any changed policy code.

### Prior review context

- Round 1 at `a9d435c88` found two blockers, one major, and two mediums.
- Round 2 at `09a33838d` found one blocker, one major, and two mediums.
- The maintainer chain then selected Option A: ADR-080's one-owner contract stands. This round therefore treats initialization ownership and mode-specific stop adapters behind the owner as requirements, not as an open design choice.
- This round reviewed committed head `69d14b7ed95eae9ccc0ceb7566c3c025a67c02a8` against `origin/main` `65865ecb5` and the round-2 head.

### What this round examined

- The complete `09a33838d..8e1ccf4ad` fix diff and every changed/adjacent decode-policy, route-predicate, preflight, cache-recovery, configuration, and regression-test section.
- All `DecodePolicy::init` and `transition` call sites, every surviving `record_logprob` reference, and the full `GenerateConfig` field matrix at direct Metal, MTP, self-spec, and prefix-cache entry points.
- The PR body, ADR-080 C3, review history, GitHub review threads, current-head checks, merge-parent scope, local mutation log, benchmark harness scope, and the June 3 GPU-race provenance.
- Targeted current-head tests, a compile-fail mutation in a detached disposable worktree, fmt, workspace/default clippy, and `f16,metal-gpu` clippy.

### Findings (3 total: 1 major, 2 mediums)

#### Majors

1. [`crates/inference/src/model/qwen35/generation.rs:1042`] The mandatory closure does not enforce the ruled stop-adapter ownership contract.
   - **Current**: `transition` requires an arbitrary `FnMut` at `:1305`, but the API cannot distinguish a real matcher adapter from a no-op. The rustdoc explicitly blesses a trivial `Continue` closure at `:1047-1049` and `:1111-1114`, and the canonical CPU `decode_loop` passes exactly `|_, _| StopCheckOutcome::Continue` at `:1464`. The actual non-streaming matcher remains caller choreography at `:1582-1617`; Metal likewise keeps an optional matcher outside `DecodePolicy` and passes a closure that branches on it.
   - **Issue**: Removing the argument is now a compile error, but replacing a real matcher with `|_, _| Continue` still compiles. That is the precise escape hatch the Option A ruling excluded: a site can call the owner transition while omitting configured stop matching. The adapters are callback bodies owned by each call site, not mode-specific adapters behind the single owner. Today the unconditional closure is reached only through the `stop_strings.is_empty()` branch, so I did not derive a current output bug; this is a major contract/invariant failure because the extraction still cannot prevent the original silent-omission class.
   - **Fix**: Put a private stop-mode adapter behind `DecodePolicy` (for example, a policy-owned enum with `Disabled` constructible only when `gen_cfg.stop_strings` is empty, plus streaming-holdback and full-scan variants). Let the caller supply only backend sinks/buffers needed by that adapter. `transition` must drive the adapter state and should not accept an arbitrary outcome-producing closure that can claim `Continue` without checking.
   - **Test**: Keep the E0061 omission check, and add a mutation test that replaces an active-stop adapter with the disabled/no-op mode; construction or compilation must fail, rather than merely relying on an output test to catch the omission later.
   - **Confidence**: Observed by removing argument 9 in detached worktree `/tmp/lattice-pr787-r3.KVZYyl` and running `cargo check -p lattice-inference --lib` (`E0061: this method takes 9 arguments but 8 arguments were supplied`); derived from the accepted ruling and the compiling production no-op closure.

#### Mediums

2. [`crates/inference/src/sampling.rs:844`] The first-token duplicates are gone, but `record_logprob` is still crate-visible and can be reintroduced outside the owner.
   - **Current**: The three old free calls are genuinely absent, all four construction sites use `DecodePolicy::init`, and the policy's associated method is private at `generation.rs:1242`. However, that method is only a wrapper over `crate::sampling::record_logprob`, which remains `pub(crate)` here and is imported directly into `generation.rs`. Rust defines `pub(crate)` as visible throughout the crate, so any sibling decode site can call it and recreate the duplicate choreography the ruling said should be unrepresentable ([Rust Reference: visibility and privacy](https://doc.rust-lang.org/stable/reference/visibility-and-privacy.html#pubin-path-pubcrate-pubsuper-and-pubself)).
   - **Issue**: `init` correctly makes the first record unavoidable when constructing a policy, but privacy does not make an extra freestanding record impossible. The PR body's structural claim is therefore stronger than the code.
   - **Fix**: Move the recording operation behind a module-private policy implementation/capability so decode callers cannot name it. If the probability computation must remain shared, expose computation separately but keep mutation of `token_logprobs` private to `DecodePolicy`.
   - **Test**: Add a compile-fail privacy check from a sibling decode module, analogous to the E0624 check used for the policy's constituent methods.
   - **Confidence**: Derived from the current visibility and a crate-wide `git grep`; only policy-owned production calls exist now, but the helper is callable from every crate module.

3. [`scripts/bench-compare.sh:51`] The reported clean inference benchmark does not measure the changed decode-policy paths.
   - **Current**: `make bench-compare` runs only `elementwise_cpu_bench` for `lattice-inference` (`:51`, `:62-64`, `:99-100`). That bench imports and times standalone norm/activation/softmax kernels (`crates/inference/benches/elementwise_cpu_bench.rs:1-14`); it never constructs `DecodePolicy`, invokes `transition`, or enters a generation loop. Its 0 FAIL / 0 WARN result therefore cannot support the PR body's statement that it is the group this refactor could affect. The noisy FAIL/WARN rows in untouched `embed` and the BERT norm kernel are reasonably discounted under the known contention, but that does not create coverage of the changed hot path.
   - **Issue**: This is a per-token decode-loop refactor, so the repository's benchmark discipline needs either a policy/generation benchmark or an explicit statement that the default comparison has no relevant group. A clean unrelated kernel suite is not affirmative performance evidence.
   - **Fix**: Correct the PR body and run a relevant A/B benchmark: a small `DecodePolicy::transition`/CPU generation harness would isolate orchestration overhead; the existing cross-turn benchmark exercises the prefix-cache loop but is model/GPU-heavy. If no stable relevant harness exists, say so and treat the refactor as unmeasured rather than crediting `elementwise_cpu_bench`.
   - **Test**: No correctness test; attach base/head benchmark output from the relevant harness with the same machine/lock discipline.
   - **Confidence**: Derived from the benchmark script and benchmark source. I did not rerun `make bench-compare`; the 14 FAIL / 14 WARN / 59 WIN values are PR-body/local-artifact claims, not this round's observed output.

- Round-2 blocker closure: `mtp_route_active` and `self_spec_route_active` include `gen_cfg.logprobs.is_none()` at `metal_qwen35.rs:599` and `:628`. Running `cargo test -p lattice-inference --lib --features f16,metal-gpu route_blocked_by_set_logprobs -- --test-threads=1` observed 2 passed, including the exact `Some(0)` falsification fixtures.
- Direct Metal ordering: `check_logprobs_not_set` is at `metal_qwen35.rs:8669`, before context rejection, `reset_state`, compact-route mutation, prefill, or session mutation. The only prior work is RNG setup/tokenization and zero-token early return. The targeted `metal_generate_plain_rejects_logprobs_config` test passed under the GPU lock.
- Prefix-cache ordering and preservation: both guards are at `metal_qwen35.rs:15409-15410`, before the inner call and the destructive `Err` arm at `:15421-15427`. Running the `rejects_logprobs` filter under the GPU lock observed 4 passed; the warm-entry regression saves token IDs before rejection and asserts exact equality afterward at `:26933-26965`.
- Prefix-cache MTP guard: `check_mtp_not_requested` resolves both `Some(true)` and `None` plus `LATTICE_MTP`; `generate_streaming_with_prefix_cache_rejects_active_mtp_request` passed under the GPU lock.
- Initialization ownership today: all old `DecodePolicy::new` references and the three freestanding first-token calls are gone. `init` records at `generation.rs:1172`. Running `init_records_the_prefill_tokens_logprob_before_any_transition_call` observed 1 passed with `max_new_tokens: 1`, so the single record cannot have come from `transition`.
- Every configured-stop loop I audited contains real matching work: CPU non-streaming uses detokenize + bounded `earliest_stop_match_from`; CPU streaming uses `StopStringMatcher` byte holdback; both Metal streaming loops use their optional `StopStringMatcher` plus sink interruption. The finding is that the API still permits substituting a no-op, not that these active branches currently omit their matchers.
- Full `GenerateConfig` re-sweep: I found no remaining silent field omission at direct Metal, MTP, self-spec, or prefix-cache. `max_new_tokens`, sampling fields, seed, stop-token IDs, grammar, string stops, and reasoning/logprob policy are consumed or routed; speculative routes admit only greedy, grammar-free, stop-string-free, budget-free, identity-penalty, logprob-free configs; prefix-cache explicitly rejects active MTP and logprobs. Fields irrelevant under the admitted mode (for example seed/top-p under deterministic greedy selection, or `enable_thinking` without a reasoning budget) do not change that mode's result.
- The full-suite failure characterization is historically real. Commit `32606fa52` was authored on 2026-06-03 specifically to document the nondeterministic, occupancy-sensitive, argmax-stable chunked-prefill race; the same comment and test exist on `origin/main`, and this PR does not change `forward_prefill`, `forward_step`, `gdn_recurrence_fused`, or `decode_attention` logic. The current comment's `ADR-065` pointer is stale after later ADR numbering reuse, but the commit provenance and in-source characterization are genuine.
- Testing evidence is internally consistent: 1795 passed + 2 failed + 20 ignored equals the 1817 feature-suite tests enumerated by the current binary. I observed the targeted regressions above; I did not relabel the full-suite run as personally observed.
- `cargo fmt --all -- --check`, `git diff --check origin/main...HEAD`, `cargo clippy --workspace -- -D warnings`, and `cargo clippy -p lattice-inference --features f16,metal-gpu -- -D warnings` all passed locally on committed head.
- Current-head GitHub CI is fully green, including all three platform jobs, both feature matrices, MSRV, semver, bench compile, e2e parity, Metal attention parity, q4 PPL, q4 composed golden, and aggregate gates. No review threads exist.
- Merge scope is clean: `69d14b7ed` has parents `8e1ccf4ad` and `65865ecb5`; `8e1ccf4ad..69d14b7ed` is exactly origin/main's docs-only ADR-number reconciliation across five docs files. No source change entered through the merge.

### What I did NOT check

- I did not rerun the 4041-second full `cargo test -p lattice-inference --lib --features f16,metal-gpu -- --test-threads=1` suite. The PR's 1795/2/20 result is credited as supplied evidence, with targeted tests and current-head CI observed independently.
- I launched the long-prompt race test once under the GPU lock, but the command's output channel detached before the final test summary; I therefore do not count that run as observed pass/fail evidence. The pre-existing characterization above is based on current/base source and the original June 3 commit.
- I did not rerun `make bench-compare`, and I did not run a replacement policy/generation benchmark.
- I did not independently replay all 27 claimed reverse mutations; I ran the requested predicate/cache/init tests and the missing-argument compile mutation.
- I did not reread unrelated unchanged regions of the roughly 28K-line Metal implementation; I reviewed the full round-3 diff, every changed/adjacent function, all policy/config call sites, and crate-wide sibling searches relevant to the stated contract.

- One focused follow-up round is useful after the owner API no longer accepts an arbitrary no-op stop result, logprob mutation is private to the owner, and the benchmark statement/evidence is corrected. Stop after that round if the fix is scoped and current-head gates remain green.

Domain utility: MEDIUM — invariant-carrying API guidance helped distinguish “argument required” from “valid adapter enforced,” while repository source, the maintainer ruling, and mutation results supplied the decisive evidence.

Verdict: APPROVE-WITH-FIXES

ohdearquant:

## Review of PR #787 (round 4)

The three round-3 implementation findings are closed. The owner now derives a private stop mode from the real configuration, no outcome-producing stop callback remains, logprob-vector mutation is private to `DecodePolicy`, and the benchmark section plainly says the changed path is unmeasured. One medium artifact-consistency fix remains: the PR's opening contract and several current rustdoc lines still describe the removed mandatory callback. Current-head CI is also not yet merge-green: the q4 PPL job is an observed setup-only infrastructure failure, but its replacement success is not visible yet and another required e2e job is still running.

### Prior review context

- Round 3 at `69d14b7ed` found one major and two mediums: the arbitrary no-op stop callback, crate-visible logprob mutation, and unrelated benchmark evidence.
- This focused final round reviewed the single fix commit `69d14b7ed..34161ce7f` against exactly those findings.

### What this round examined

- The complete fix commit and all six `DecodePolicy::transition` call sites: CPU streaming with/without stop strings, CPU `decode_loop`, CPU `decode_loop_with_stops`, direct Metal streaming, and Metal prefix-cache streaming.
- Disposable compile probes for the old no-op callback, attempted `StopMode::Disabled` construction beside non-empty stop strings, the deleted free `record_logprob`, and direct access to private `DecodePolicy::record_logprob`.
- The PR benchmark wording, fix-commit scope, relevant CPU/Metal tests, fmt/clippy, review threads, and current-head GitHub job metadata.

#### Mediums

1. [`crates/inference/src/model/qwen35/generation.rs:1005`] The current contract documentation still describes the callback that this fix removed.
   - **Current**: `StepOutcome` says the “mandatory `stop_check` callback” reported `Stopped`/`Interrupted` at `:1005-1017`, and the `DecodePolicy` overview still says callers supply four callbacks including “the stop-string check” at `:1059-1064`. The PR body's opening Contract likewise says `transition` takes a mandatory stop-check callback and four callbacks; that is current-contract prose, not the explicitly historical round-2 section. Smaller stale references remain at `generation.rs:776-777` and `:1834-1836`.
   - **Issue**: Head now intentionally accepts no outcome-producing stop callback. Leaving the public review artifact and owner rustdoc on the rejected API shape contradicts the implementation and can send the next maintainer back toward the exact escape hatch this commit closes.
   - **Fix**: Update the PR opening Contract and these current rustdoc/comments to describe `StopMode`, `decode_delta`, and `emit_confirmed`; retain callback language only inside clearly historical round-2 discussion.
   - **Test**: No runtime test; PR-description/code-doc consistency review.
   - **Confidence**: Observed in the current PR body returned by `gh pr view 787` and derived from the committed `transition` signature at `generation.rs:1467-1481`.

- The round-3 major is closed. `StopMode` is private at `generation.rs:1121`; `StopMode::for_config` selects `Disabled` only for an empty slice and otherwise selects `Streaming` or `FullScan` at `:1144-1157`; `DecodePolicy::init` derives it directly from `gen_cfg.stop_strings` at `:1189-1215`. `transition` owns the stop decision at `:1467-1517` and takes raw decode/sink primitives, not a result-producing stop callback.
- Disposable probe evidence in detached worktree `/tmp/lattice-pr787-r4.AU4qxK`: adding the old trailing `|_, _| StopCheckOutcome::Continue` to a production call failed with E0061; attempting `crate::model::qwen35::generation::StopMode::Disabled` from Metal beside a config containing `"STOP"` failed with E0603. The latter establishes the caller boundary: only the owner module can name the variant, and the only production construction site is `for_config`.
- Streaming byte holdback and non-streaming full scan both live behind the owner. `stop_check` dispatches `Streaming(StopStringMatcher)` and `FullScan { stop_strings, max_stop }` at `generation.rs:1330-1395`; `check_initial_stop` preserves the same matcher's first-token state at `:1398-1424`; `finish_stop` owns streaming tail flush at `:1520-1560`.
- Every transition call site now supplies only decoding/output primitives and cannot return `Continue`: CPU fast streaming `generation.rs:734`, CPU stop-string streaming `:911`, CPU no-stop decode loop `:1680`, CPU full-scan decode loop `:1793`, direct Metal streaming `metal_qwen35.rs:13347`, and prefix-cache Metal streaming `:15865`. The round-3 offender is now the `DecodePolicy`-owned disabled dispatch; its caller-side closures at `generation.rs:1698-1701` cannot choose a stop outcome.
- The round-3 logprob medium is closed. The former free `crate::sampling::record_logprob` is gone; `sampling.rs:763` exports only pure `compute_step_logprobs`. The only production `token_logprobs.push(TokenLogprob { ... })` is inside private `DecodePolicy::record_logprob` at `generation.rs:1295-1310`, called only by `init` and `transition`.
- The claimed privacy property is discriminating. A sibling probe calling the removed free function failed with E0425. A sibling probe constructing a policy and directly calling `policy.record_logprob(...)` failed with E0624; relaxing that method back to `pub(crate)` would make that probe compile, so it tests the exact visibility boundary rather than incidental behavior.
- The benchmark correction is honest. The PR now says `elementwise_cpu_bench` cannot observe this change, no existing generation bench drives `DecodePolicy`, and the refactor is “unmeasured by the default `make bench-compare` suite.” Its no-regression expectation is explicitly labeled architectural reasoning rather than measurement; no “closest plausible group” implication remains.
- Fix-commit scope is clean. `34161ce7f` changes only `generation.rs`, `stop_strings.rs`, `sampling.rs`, and the two Metal owner call sites. The `StopStringMatcher` borrow-to-owned change is necessary to store it inside `StopMode`; the remaining edits are the adapter conversion, logprob-helper removal, associated call-site rewiring/comments, and tests. I found no unrelated production behavior.
- Observed local gates at committed head: `cargo fmt --all -- --check`; `cargo clippy --workspace -- -D warnings`; and `cargo clippy --workspace --features f16,metal-gpu -- -D warnings` all passed. `model::qwen35::generation::` passed 23/23 with 4 ignored; `model::qwen35::stop_strings::` passed 28/28 with 1 ignored. The Metal streaming filter passed 6/6, the stop-reason filter passed 7/7, and the route-predicate filter passed 11/11. Metal tests acquired `/tmp/lion-metal-gpu-test.lock` through their built-in `gpu_test_lock` guard.
- GitHub metadata confirms the q4 PPL failure is infrastructure-only: head SHA is exactly `34161ce7f`; the job ran 41 seconds, failed at `dtolnay/rust-toolchain@1.94.1`, and skipped build, artifact generation, and the PPL gate itself. All completed CI/feature/MSRV/semver/e2e/Metal-attention jobs are green. At the observation time, `q4 composed golden` was still running and the q4 PPL rerun was not yet visible as a replacement check, so `mergeStateStatus` remained `BLOCKED`.
- No GitHub review threads exist.

### What I did NOT check

- Per instruction, I did not rerun the unchanged full Metal suite; round 3's full-suite result remains the evidence for that unchanged code.
- I did not rerun `make bench-compare` because the corrected claim is that its default groups do not exercise this code.
- I did not rerun all workspace tests or the full `lattice-inference` library suite; I ran the touched owner/matcher groups and targeted Metal groups.

- No further code-review round is useful. Correct the stale current-contract prose and require the actual-head q4 PPL rerun plus all remaining required checks to turn green before merge; those are mechanical artifact/merge-gate verifications, not a new review angle.

Domain utility: MEDIUM — invariant-carrying API and module-privacy guidance sharpened the compile probes, while committed source, mutation results, and GitHub job metadata supplied the decisive evidence.

Verdict: APPROVE-WITH-FIXES

oceanwaves630 (approved):
Approving per the documented gate trail: four review rounds, enforcement properties compile-probed, final head is the doc-only fix.

Issue #445, opened 2026-06-29
refactor(tune): extract shared CPU tape/Adam from train_grad_full into a library module


ohdearquant:
## Problem

`train_micro_lora` (`crates/tune/src/lora/train.rs`) duplicates the CPU forward/backward tape and the Adam optimizer loop from the `train_grad_full` binary (`crates/tune/src/bin/train_grad_full.rs`). The duplication exists because the binary's helper functions are private and cannot be imported from library code.

The two copies are line-equivalent today (verified during the #444 review — no algebra divergence), but maintaining two copies of the training math is a drift risk: a future fix or change to one path can silently diverge from the other.

## Proposal

Extract the shared training core into a single library-private module (e.g. `crates/tune/src/lora/train_core.rs`):

- the tape helpers: `forward_full`, `nll_and_grads`, `lm_head_logits`, `rmsnorm_seq`, `position_nll`, `shifted`
- the Adam update loop

Have both `train_micro_lora` and the `train_grad_full` binary call the shared module. Keep CLI-only concerns (data loading, logging, eval, gradcheck) in the binary.

## Context

Follow-up from the governed-manifest + micro-LoRA trainer PR (#444). Tracked as a deliberate deferral so the PR stays scoped to the governance/safety fixes; this is a maintainability cleanup with no behavior change.

Issue #262, opened 2026-06-23
gdn: asymmetric value/key heads (num_value_heads > num_key_heads) silently produces wrong output


ohdearquant:
## Summary

`gated_delta_net_step` only correctly handles **symmetric** GDN head configs (`num_value_heads == num_key_heads`). For any **asymmetric** config (`num_value_heads > num_key_heads` — the general Qwen3-Next case, and the code's own default), it silently writes only the first `num_key_heads` value heads and feeds stale buffer data through the output projection, yielding wrong logits with no error.

**Not triggered by the shipped qwen3.5-0.8b** (configured 16k/16v symmetric — that's why e2e-parity is green). This is a latent gap that would silently corrupt output for larger Qwen3-Next models and for any config that omits the value-heads field.

## Evidence (crates/inference/src/attention/gdn.rs)

- State is sized for **value** heads: `s_matrices: vec![0.0; value_heads * key_dim * value_dim]` (`GatedDeltaNetState::new`, ~L87, via `cfg.linear_num_value_heads()`).
- Output width is **value** heads: `linear_output_dim() = linear_num_value_heads() * linear_value_head_dim` (qwen35_config.rs:439).
- But the step loop iterates **key** heads: `let num_heads = cfg.linear_num_key_heads;` (gdn.rs:193), `for h in 0..num_heads { ... }` (L267 recurrence, L333 gated RMSNorm).
- Net effect for `value_heads > key_heads`: steps 7-8 write `gated_norm_buf[0 .. key_heads*value_dim]`; `out_proj` then reads `gated_norm_buf[..output_dim]` = `[.. value_heads*value_dim]` (L346-347). The tail `[key_heads*value_dim .. value_heads*value_dim]` is **stale/uninitialized scratch** → wrong output.

## The footgun default

`linear_num_value_heads()` returns `self.linear_num_value_heads.unwrap_or(32)` (qwen35_config.rs:378) while `linear_num_key_heads` defaults to 16. So a config.json that omits the value-heads field gets **32 value / 16 key** — asymmetric — and silently hits the corruption path. (The hardcoded `qwen35_2b()` / qwen3.5-0.8b defaults set `Some(16)`, dodging it.)

## Misleading test coverage

[...]

PR #792, opened 2026-07-10
feat(tune): GDN LoRA weight gradients via train_core (port of #202)

ohdearquant:
## What

Re-homes PR #202's GDN LoRA weight-gradient surface (surface-B) onto the `train_core.rs` shared module introduced by #445/#488, so both training binaries that share `train_core` benefit from it, instead of the feature living only in the old monolithic `train_grad_full.rs` that PR #202 was written against.

## Why

PR #202 added LoRA weight-gradient support for GatedDeltaNet (GDN) layers — previously only GQA/attention layers had trainable weight gradients; GDN layers only propagated the activation gradient (`dx`) through, with no way to attach or train a LoRA adapter on them. Since #202 was opened, #445/#488 refactored the trainer internals (`Dims`, `GdnDims`, `MixerKind`, `LoraParams`, `forward_full`, `nll_and_grads`, `apply_adam_updates`, etc.) out of the monolithic binary into `crates/tune/src/lora/train_core.rs`, shrinking `train_grad_full.rs` from 1236 to 755 lines. #202 no longer applies as a rebase; this PR re-implements its algorithmic core against the current architecture.

## Ported (from #202, algorithm unchanged)

- `crates/inference/src/attention/gdn_backward.rs`: `gdn_forward_save` now accepts 10 optional LoRA weight slices (one A/B pair per GDN projection: `in_proj_qkv`, `in_proj_z`, `in_proj_b`, `in_proj_a`, `out_proj`) plus rank/scale — byte-identical to the prior no-LoRA path when omitted (all `None`, rank 0). `gdn_backward` now returns `GdnGrads { grad_a_*, grad_b_*, dx }` instead of writing `dx` into a caller-supplied buffer.
- `crates/tune/src/lora/mod.rs`: `LoraAdapter::validate_against` now recognises `in_proj_b`/`in_proj_a` (previously only `in_proj_qkv`/`in_proj_z`/`out_proj` were recognised — a saved GDN adapter's beta/alpha LoRA layers would have failed validation on load).
- `crates/inference/examples/diff_gdn_layer.rs`: mechanical call-site update for the new `gdn_forward_save` signature.

## Re-homed onto train_core (new relative to #202, adapting its slot-per-layer design onto train_core's structure)

- `crates/tune/src/lora/train_core.rs`: new `GdnLoraParams` (10-array LoRA container, doubling as the gradient container the same way the existing `LoraParams`/`Grads` pattern does), `forward_full`/`eval_chain_nll`/`nll_and_grads` extended with a `gdn_loras` slice + GDN slot count, and a new `apply_gdn_adam_updates` optimizer step mirroring `apply_adam_updates`. GQA surface-A is untouched (verified via lib test suite + clippy).
- `crates/tune/src/bin/train_grad_full.rs`: GDN layers in the materialised range now get their own LoRA slot (mirroring the existing GQA slot bookkeeping), wired through the TBV check (chain NLL vs real model), gradcheck mode (adds a second finite-difference loop over the 10 GDN LoRA arrays per slot, alongside the existing GQA loop), the training loop, and `--save` (adds `in_proj_qkv/z/b/a/out_proj` to the PEFT safetensors export).
- `crates/tune/src/lora/train.rs`: the library `train_micro_lora` entry point (GQA-only) passes empty GDN LoRA arrays — behavior is unchanged (byte-identical no-LoRA GDN path), it simply had to accept the new function signatures.

`train_grad_layer23.rs` is unaffected and not wired: it has its own self-contained single-layer implementation and only ever materialises layer 23, which is a GQA layer, so there is no GDN layer for it to attach LoRA to.

## Deliberately NOT ported

- `ops.rs`/`tape.rs` shifted-gamma changes: #202 baked `(1 + gamma)` into `rms_norm_forward`/`rmsnorm_backward` themselves. Main already resolves the same shifted-gamma requirement at the call site — `train_core::shifted()` pre-adds 1 to gamma before calling those primitives with plain-gamma semantics. Porting #202's primitive-level diff verbatim would double-shift gamma (a real numerical bug). Left as-is.
- `lora_forward_parity_test.rs` (917-line new test file): a diagnostic investigating exactly the plain-vs-shifted-gamma question above, predating the discovery that main solves it differently. Its premise ("tape.rs uses plain gamma, that's a bug") no longer holds, so porting it would test a stale hypothesis.

## Codex round-1 fix (this round)

Round 1 came back **REJECT** with one blocker and one medium; both are fixed here.

**Blocker — alpha/beta were dimensioned per KEY head, but the model defines them per VALUE head.** `in_proj_b`/`in_proj_a` (and their gates `a_log`/`dt_bias`) must have `linear_num_value_heads()` rows to match the authoritative f16 weight loader (`f16_weights.rs`) and the shipping fused forward (`gdn_fused.rs`), which project and index them per value head. The ported code instead projected to `num_kh` and indexed gates through `kh = h / ratio`, silently reducing distinct value-head adjoints into shared key-head slots. This is invisible on Qwen3.5-0.8B (`num_kh == value_heads == 16`) but wrong on the 4B (16 key / 32 value) and 27B (16 key / 48 value) shapes — the known #262 asymmetric-heads trap, now load-bearing because it propagated into LoRA export/validation.

Fixed by re-dimensioning by `value_heads` throughout, keeping key-head sharing *only* for the Q/K projections (`kh = h / ratio` remains correct and unchanged there):
- `crates/inference/src/attention/gdn_backward.rs`: `GdnSaved`'s `beta_raw`/`alpha_proj`/`beta`/`g` buffers, the per-value-head accumulators in `gdn_backward` (renamed `d_alpha_vh`/`d_beta_raw_vh`, no cross-head reduction — alpha/beta are genuinely one-per-value-head), and the LoRA B-factor allocations in `gdn_forward_save`/`gdn_backward`.
- `crates/tune/src/lora/train_core.rs`: `GdnLoraParams::zeros`'s `b_b`/`b_a` sizing (`gd.num_kh` → `gd.value_heads`).
- `crates/tune/src/bin/train_grad_full.rs`: PEFT export `d_out` for `in_proj_b`/`in_proj_a` (`gdn_dims.num_kh` → `gdn_dims.value_heads`).
- `crates/tune/src/lora/mod.rs`: `LoraAdapter::validate_against`'s expected dims for `in_proj_b`/`in_proj_a` (`config.linear_num_key_heads` → `config.linear_num_value_heads()`), plus two new asymmetric-config regression tests (`test_validate_against_gdn_asymmetric_value_heads_35b_a3b` at value_heads=32, `test_validate_against_gdn_asymmetric_value_heads_27b` at value_heads=48) that assert the correct dim validates and the old key-head dim (16) is rejected — `qwen35_0_8b` can't distinguish these because it's symmetric.

Note: this dimension bug pre-dates this PR — confirmed via `git show origin/main:.../gdn_backward.rs` that the same `num_kh`-scoped alpha/beta indexing already existed on main before this port, as a latent bug in the (previously LoRA-less, weight-grad-less) backward path. It became load-bearing only once wired into LoRA export/validation here.

Added the exact test class codex named as the one that "would have caught it": `source_parity_forward_save_vs_fused_asymmetric_heads`, a LoRA-off comparison of `gdn_forward_save` (training path) against the shipping `gated_delta_net_step_fused` (inference path) on a genuinely asymmetric 2-key/4-value-head fixture. This is deliberately an *independent* oracle — unlike the finite-difference gradcheck, which shares the same (previously buggy) shape assumption on both its analytic and numerical sides and so could not have caught this class by construction. TBV: ran this exact test against the pre-fix code (`git show efe20fd46:...` — this PR's own prior head) and confirmed it fails hard (`matmul_bt: b too short for n*k`); it passes clean on the fixed code.

**Medium — the gradcheck's aggregate `n_tested >= 10` threshold hid that 4 of the 10 arrays (`grad_a_b`, `grad_b_b`, `grad_a_a`, `grad_b_a` — exactly the alpha/beta ones) had zero entries above the finite-difference noise floor in both fixtures.** Replaced the aggregate-only gate with per-array coverage tracking (`PerArrayCoverage`, one `(n_tested, max_rel)` pair per named array) and an assertion requiring `n_tested >= 1` for *every* array, not just the sum. Getting real (non-fixture-hacked) coverage on the alpha/beta arrays required two things: (1) widening the fixture's `a_log` range and increasing `lora_scale`/`seq_len` to strengthen the genuinely-indirect scalar-gate signal path, and (2) fixing a real precision bug in the finite-difference oracle itself — `fd_lora_weight_entry`'s `perturbed_loss` accumulated in f32 (unlike its sibling `fd_grad_inputs_linear`, which correctly used f64), causing catastrophic-cancellation error growth once scale/seq_len were pushed high enough to lift the alpha/beta signal. Switched it to f64 accumulation (forward stays f32, matching production).

Mutation-tested the fixed alpha/beta path directly (scaling one A-side alpha accumulation term by 0.5): both gradcheck tests now correctly fail (`max_rel≈0.50` on `grad_a_a`), confirming the per-array gate is sensitive to the exact class codex flagged, not passing vacuously. Reverted after confirming.

## Gradient verification (updated, per-array — supersedes the prior "across all 10 arrays" claim)

Real per-array coverage from both final gradcheck runs (`n_tested`, `max_rel`, all comfortably under the `1e-2` gate):

```
gradcheck_gdn_lora_weight_grads (single-head):
  grad_a_qkv  : n_tested=7  max_rel=2.087e-4
  grad_b_qkv  : n_tested=6  max_rel=1.382e-4
  grad_a_z    : n_tested=7  max_rel=3.170e-5
  grad_b_z    : n_tested=8  max_rel=7.218e-5
  grad_a_b    : n_tested=3  max_rel=1.835e-3
  grad_b_b    : n_tested=2  max_rel=2.210e-4
  grad_a_a    : n_tested=2  max_rel=1.766e-3
  grad_b_a    : n_tested=2  max_rel=3.071e-4
  grad_a_out  : n_tested=8  max_rel=1.920e-4
  grad_b_out  : n_tested=6  max_rel=3.977e-5

gradcheck_gdn_lora_weight_grads_multi_head (asymmetric, num_kh=2/value_heads=4):
  grad_a_qkv  : n_tested=7  max_rel=4.100e-4
  grad_b_qkv  : n_tested=7  max_rel=7.769e-5
  grad_a_z    : n_tested=7  max_rel=5.993e-4
  grad_b_z    : n_tested=7  max_rel=9.411e-5
  grad_a_b    : n_tested=6  max_rel=4.378e-3
  grad_b_b    : n_tested=8  max_rel=2.244e-3
  grad_a_a    : n_tested=4  max_rel=5.272e-3
  grad_b_a    : n_tested=4  max_rel=2.886e-3
  grad_a_out  : n_tested=7  max_rel=3.144e-5
  grad_b_out  : n_tested=7  max_rel=3.309e-6
```

Both tests pass with every array individually asserted `n_tested >= 1` (no aggregate escape hatch).

Added `train_core.rs` unit tests for the re-homed plumbing (`GdnLoraParams::zeros` shape/overflow checks, `apply_gdn_adam_updates` sign-of-update check) and mutation-verified the optimizer-step test by disabling one array's update and confirming the test catches it (then restored).

## Gates

- `cargo fmt --check` — clean
- `cargo clippy --workspace --all-targets --features train-backward,safetensors,inference-hook -- -D warnings` — clean
- `cargo test -p lattice-inference -p lattice-tune --features train-backward,safetensors,inference-hook` — all green (1623 lib tests in `lattice-inference` incl. the source-parity test, 264 in `lattice-tune` lib incl. the 2 asymmetric `validate_against` regressions, 2 new `train_grad_full` binary-level unit tests for the round-2 shape-aware constructors, plus all other bin/integration/doctest targets)
- `cargo build --workspace` (default features) — clean, no new default-build warnings

## Bench (ADR-058)

This PR touches `crates/inference/` (`gdn_backward.rs`), so ADR-058's before/after gate applies even though the change is additive-only to the LoRA argument surface (all pre-existing, non-LoRA call sites pass `None`/`0`/`0.0` unchanged). Two runs, both disclosed:

**Run 1 (invalidated — sibling-lane contention):**

```
bash /Users/lion/projects/firm/resources/scripts/with-heavy-lane.sh "lattice:792-bench" -- make bench-compare
```

106 FAIL / 12 WARN / 113 confirmed-improvement, but the pattern was whole-crate and symmetric across benchmark groups this diff never touches: every `lattice-embed` SIMD/cosine-distance benchmark showed up to +386% "regression," while every `lattice-inference` `elementwise_cpu_bench` benchmark (`gelu`, `layer_norm`, `rms_norm`, `silu_inplace`, `softmax_attention`, `residual_add_layer_norm`, `memory_size`) showed up to -88% "improvement" — uniformly in one direction per crate, unconnected to the actual code changed. Concurrent heavy-lane jobs from sibling worktrees (`lattice:c3-r2-metal-final`, `lattice:c3-r2-full-suite`, and others) were running during this measurement window. Discarded as contention-corrupted, not written up as a regression or improvement claim.

**Run 2 (standing — clean re-run):**

```
bash /Users/lion/projects/firm/resources/scripts/with-heavy-lane.sh "lattice:792-bench-rerun2" -- bash -c "cd /Users/lion/projects/khive/lattice-pr202port && make bench-compare"
```

Base=`origin/main` (`65865ecb5`), Head=`HEAD` (`efe20fd46`). The whole-crate symmetric pattern is gone: **zero FAIL/WARN entries in `lattice-inference`'s `elementwise_cpu_bench` group** — every one of `gelu`/`layer_norm`/`rms_norm`/`silu_inplace`/`softmax_attention`/`residual_add_layer_norm`/`memory_size` sits within ±2%, consistent with noise, and this is the group that actually exercises the crate this PR modifies. `lattice-embed`'s unrelated SIMD/cosine-distance group (a crate this PR does not touch at all) still shows 9 FAIL / 29 WARN / 13 confirmed-improvement, single-digit-to-high-teens percent, mixed direction — ordinary microbenchmark jitter on microsecond-scale ops, not the systemic whole-crate corruption seen in run 1.

Net: no bench evidence of a regression in the code this PR actually changes. Not re-run this round (this round's fix is a dimension/coverage correction, no algorithmic/perf-shaped change to the touched benchmark group).

## Codex round-2 fix

Round 2 came back **REJECT** with one surviving blocker (everything else reconfirmed clean: value-head re-dimensioning internally consistent across `GdnSaved`/forward/backward/VJPs, the parity test mutation-sensitive, per-array counts reproducing exactly, `f64` oracle-only, save→load→validate round-trips passing for 32 and 48 value heads).

**Blocker** — `crates/tune/src/bin/train_grad_full.rs` still allocated `b_b`/`b_a` with `gdn_dims.num_kh * rank` at four sites (the gradcheck-mode and training-mode LoRA-array constructors), independently of the round-1 fix to `GdnLoraParams::zeros`. Since `forward_full` feeds those slices into the corrected `gdn_forward_save` (which now expects `value_heads` rows), the binary's real gradcheck/training path would have broken on any asymmetric config, and the PEFT export would label `d_out = value_heads` while the stored buffer was shorter. Round-1's `validate_against` tests missed this because they construct correctly-sized adapters directly rather than going through the binary's own constructors.

Fixed by:
- Adding `GdnLoraParams::shaped` (`crates/tune/src/lora/train_core.rs`) — a shape-aware constructor that is now the single source of truth for every GDN LoRA array's length; `zeros()` is reimplemented on top of it.
- Extracting `train_grad_full.rs`'s two inline LoRA-array constructors into named, testable functions (`gradcheck_gdn_loras`, `zero_b_gdn_loras`), both routed through `::shaped`, so this class of drift is now a compile-time-shared-code property instead of a grep-and-hope one.
- Adding unit tests for both extracted functions against an asymmetric `GdnDims{num_kh: 2, value_heads: 4}` fixture, asserting `b_b`/`b_a` are sized by `value_heads` (`gradcheck_gdn_loras_sizes_b_b_and_b_a_by_value_heads`, `zero_b_gdn_loras_sizes_b_b_and_b_a_by_value_heads`).
- Final `num_kh` grep sweep across all four codex-cited files: zero remaining key-head-scoped alpha/beta sizing. The one surviving `gdn_dims.num_kh` reference (`train_core.rs:346`) is `GdnSaved::new`'s `num_key_heads` parameter — a legitimate Q/K-sharing use, verified against `GdnSaved::new`'s signature (takes `num_key_heads` and `value_heads` as distinct params).

Branch refreshed onto `origin/main` (merge, no conflicts — docs-only diff) before this push, per the strict up-to-date-and-green merge gate.

All gates green at the merged head: fmt, clippy `-D warnings`, full `lattice-inference` + `lattice-tune` test suite including the new binary-level unit tests (1623 + 264 + 2 passed, 0 failed).

## Follow-up

Once this merges, #202 should be closed in favor of it (do not close #202 from this PR — leaving that to the owning maintainer).

Refs #202

### DIFF
--- a/crates/inference/examples/diff_gdn_layer.rs
+++ b/crates/inference/examples/diff_gdn_layer.rs
@@ -99,7 +99,25 @@ fn main() {
         eps,
     );
     let mut mat_out = vec![0.0f32; seq_len * hidden];
-    gdn_forward_save(&normed, gdn_w, cfg, &mut saved, &mut mat_out);
+    gdn_forward_save(
+        &normed,
+        gdn_w,
+        cfg,
+        &mut saved,
+        &mut mat_out,
+        None,
+        None,
+        None,
+        None,
+        None,
+        None,
+        None,
+        None,
+        None,
+        None,
+        0,
+        0.0,
+    );
 
     assert_eq!(mat_out.len(), real_out.len(), "output length mismatch");
     let (mut max_diff, mut argmax) = (0.0f32, 0usize);
--- a/crates/inference/src/attention/gdn_backward.rs
+++ b/crates/inference/src/attention/gdn_backward.rs
@@ -1,9 +1,10 @@
 //! Reverse-mode backward (VJP) for a single GatedDeltaNet token sequence.
 //!
 //! Scope: computes grad_input (d_loss/d_input for each token) for a sequence
-//! processed through one GDN layer.  Parameter (weight) gradients are out of
-//! scope here — only input-gradients are needed for LoRA gradient flow through
-//! the 18 frozen GDN layers to reach lower GQA layers.
+//! processed through one GDN layer, AND (when LoRA is present) the LoRA weight
+//! gradients for each of the 5 GDN projections (qkv, z, b/beta, a/alpha, out).
+//! This is "surface-B" LoRA: mirrors the GQA surface-A implementation in
+//! backward/attention_gqa.rs, using the same lora_vjp primitive from ops.rs.
 //!
 //! # Forward recap (matches `gdn_fused.rs` hot path)
 //!
@@ -65,7 +66,33 @@
 //! eliminate catastrophic cancellation at eps = 1e-3.
 
 use crate::attention::gdn::{sigmoid, softplus};
+use crate::backward::ops::lora_vjp;
 use crate::model::qwen35_config::Qwen35Config;
+
+/// Gradients of the GDN layer w.r.t. LoRA params for all 5 projections.
+///
+/// Naming and layout mirror AttnGrads in backward/attention_gqa.rs.
+/// grad_a_* shape: [rank * d_in], grad_b_* shape: [d_out * rank].
+/// Fields are zero-length Vec when LoRA is absent (no-LoRA forward).
+pub struct GdnGrads {
+    /// in_proj_qkv LoRA: d_in=hidden, d_out=qkv_dim
+    pub grad_a_qkv: Vec<f32>,
+    pub grad_b_qkv: Vec<f32>,
+    /// in_proj_z LoRA: d_in=hidden, d_out=output_dim
+    pub grad_a_z: Vec<f32>,
+    pub grad_b_z: Vec<f32>,
+    /// in_proj_b (beta sigmoid) LoRA: d_in=hidden, d_out=num_kh
+    pub grad_a_b: Vec<f32>,
+    pub grad_b_b: Vec<f32>,
+    /// in_proj_a (alpha decay) LoRA: d_in=hidden, d_out=num_kh
+    pub grad_a_a: Vec<f32>,
+    pub grad_b_a: Vec<f32>,
+    /// out_proj LoRA: d_in=output_dim, d_out=hidden
+    pub grad_a_out: Vec<f32>,
+    pub grad_b_out: Vec<f32>,
+    /// Input gradient (always present): [seq_len, hidden_size]
+    pub dx: Vec<f32>,
+}
 
 /// Saved activations for one forward pass over a sequence.
 ///
@@ -140,6 +167,52 @@ pub struct GdnSaved {
     /// SiLU(z) per value-head: [seq_len, value_heads, value_dim]
     /// Stored because backward through gated_rms_norm needs it.
     pub silu_z: Vec<f32>,
+
+    // ---- LoRA caches (populated only when LoRA is present) ----
+    // h_* = A_proj @ x for each projection; shape [seq_len * lora_rank].
+    // Empty when LoRA absent.
+    /// LoRA rank used (0 = no LoRA)
+    pub lora_rank: usize,
+    /// LoRA scale factor
+    pub lora_scale: f32,
+
+    /// h_qkv = A_qkv @ x: [seq_len, lora_rank]
+    pub h_qkv: Vec<f32>,
+    /// h_z = A_z @ x: [seq_len, lora_rank]
+    pub h_z: Vec<f32>,
+    /// h_b = A_b @ x: [seq_len, lora_rank]  (beta sigmoid projection)
+    pub h_b: Vec<f32>,
+    /// h_a = A_a @ x: [seq_len, lora_rank]  (alpha decay projection)
+    pub h_a: Vec<f32>,
+    /// h_out = A_out @ gated_buf: [seq_len, lora_rank]  (output projection input)
+    pub h_out: Vec<f32>,
+    /// gated_buf (input to out_proj + LoRA_out): [seq_len, output_dim]
+    /// Needed for the out_proj LoRA backward (x side of that linear).
+    pub gated_buf: Vec<f32>,
+
+    // ---- LoRA weight matrices (cloned from forward params for backward) ----
+    // Storing these avoids threading 10 extra args through gdn_backward.
+    // Empty Vec when LoRA absent.
+    /// A_qkv: [rank, hidden]
+    pub lora_a_qkv: Vec<f32>,
+    /// B_qkv: [qkv_dim, rank]
+    pub lora_b_qkv: Vec<f32>,
+    /// A_z: [rank, hidden]
+    pub lora_a_z: Vec<f32>,
+    /// B_z: [output_dim, rank]
+    pub lora_b_z: Vec<f32>,
+    /// A_b: [rank, hidden]
+    pub lora_a_b: Vec<f32>,
+    /// B_b: [num_kh, rank]
+    pub lora_b_b: Vec<f32>,
+    /// A_a: [rank, hidden]
+    pub lora_a_a: Vec<f32>,
+    /// B_a: [num_kh, rank]
+    pub lora_b_a: Vec<f32>,
+    /// A_out: [rank, output_dim]
+    pub lora_a_out: Vec<f32>,
+    /// B_out: [hidden, rank]
+    pub lora_b_out: Vec<f32>,
 }
 
 impl GdnSaved {
@@ -197,6 +270,26 @@ impl GdnSaved {
             o_heads: vec![0.0; seq_len * value_heads * value_dim],
             rms_vals: vec![0.0; seq_len * value_heads],
             silu_z: vec![0.0; seq_len * value_heads * value_dim],
+            // LoRA caches: start empty; populated in gdn_forward_save when LoRA present.
+            lora_rank: 0,
+            lora_scale: 0.0,
+            h_qkv: Vec::new(),
+            h_z: Vec::new(),
+            h_b: Vec::new(),
+            h_a: Vec::new(),
+            h_out: Vec::new(),
+            gated_buf: Vec::new(),
+            // LoRA weight matrices: start empty; cloned from params in gdn_forward_save.
+            lora_a_qkv: Vec::new(),
+            lora_b_qkv: Vec::new(),
+            lora_a_z: Vec::new(),
+            lora_b_z: Vec::new(),
+            lora_a_b: Vec::new(),
+            lora_b_b: Vec::new(),
+            lora_a_a: Vec::new(),
+            lora_b_a: Vec::new(),
+            lora_a_out: Vec::new(),
+            lora_b_out: Vec::new(),
         }
     }
 }
@@ -211,15 +304,54 @@ impl GdnSaved {
 /// `inputs`:      [seq_len, hidden_size]
 /// `weights`:     frozen GDN layer weights
 /// `cfg`:         model config
-/// `norm_weight`: gamma for gated RMSNorm [value_dim]
 /// `saved`:       output struct, must be pre-allocated via `GdnSaved::new`
 /// `outputs`:     [seq_len, hidden_size] — written in-place
+///
+/// Optional LoRA parameters for the 5 GDN projections (qkv, z, b, a, out).
+/// When all five pairs are `Some`, the forward adds `scale * (x @ A^T) @ B^T`
+/// to each projection output and caches `h = A @ x` for the backward.
+/// When any pair is `None`, the forward is byte-identical to the no-LoRA path.
+#[allow(clippy::too_many_arguments)]
+/// The ten LoRA weight-matrix slices bound for one GDN forward, in fixed order:
+/// (a_qkv, b_qkv, a_z, b_z, a_b, b_b, a_a, b_a, a_out, b_out). Factored out to keep
+/// the `lora_bound` binding under clippy's type-complexity threshold (this module is
+/// `train-backward`-gated, so default clippy never lints it — see the app-bins gate).
+type LoraBound<'a> = (
+    &'a [f32],
+    &'a [f32],
+    &'a [f32],
+    &'a [f32],
+    &'a [f32],
+    &'a [f32],
+    &'a [f32],
+    &'a [f32],
+    &'a [f32],
+    &'a [f32],
+);
+
 pub fn gdn_forward_save(
     inputs: &[f32],
     weights: &crate::attention::gdn::GatedDeltaNetWeights,
     _cfg: &Qwen35Config,
     saved: &mut GdnSaved,
     outputs: &mut [f32],
+    // LoRA params for in_proj_qkv: a=[rank,hidden], b=[qkv_dim,rank]
+    lora_a_qkv: Option<&[f32]>,
+    lora_b_qkv: Option<&[f32]>,
+    // LoRA params for in_proj_z: a=[rank,hidden], b=[output_dim,rank]
+    lora_a_z: Option<&[f32]>,
+    lora_b_z: Option<&[f32]>,
+    // LoRA params for in_proj_b (beta): a=[rank,hidden], b=[num_kh,rank]
+    lora_a_b: Option<&[f32]>,
+    lora_b_b: Option<&[f32]>,
+    // LoRA params for in_proj_a (alpha): a=[rank,hidden], b=[num_kh,rank]
+    lora_a_a: Option<&[f32]>,
+    lora_b_a: Option<&[f32]>,
+    // LoRA params for out_proj: a=[rank,output_dim], b=[hidden,rank]
+    lora_a_out: Option<&[f32]>,
+    lora_b_out: Option<&[f32]>,
+    lora_rank: usize,
+    lora_scale: f32,
 ) {
     use crate::forward::cpu::matmul_bt;
 
@@ -238,6 +370,80 @@ pub fn gdn_forward_save(
     let buf_len = kernel_size.saturating_sub(1);
     let q_total = num_kh * key_dim;
 
+    // Bind all ten LoRA slices at once.  If any pair is None or rank==0 the entire
+    // LoRA path is skipped and the forward is byte-identical to the no-LoRA path.
+    // Option<&[f32]> is Copy, so destructuring copies the refs — no heap allocation.
+    let lora_bound: Option<LoraBound> = if lora_rank > 0 {
+        match (
+            lora_a_qkv, lora_b_qkv, lora_a_z, lora_b_z, lora_a_b, lora_b_b, lora_a_a, lora_b_a,
+            lora_a_out, lora_b_out,
+        ) {
+            (
+                Some(a_qkv),
+                Some(b_qkv),
+                Some(a_z),
+                Some(b_z),
+                Some(a_b),
+                Some(b_b),
+                Some(a_a),
+                Some(b_a),
+                Some(a_out),
+                Some(b_out),
+            ) => Some((a_qkv, b_qkv, a_z, b_z, a_b, b_b, a_a, b_a, a_out, b_out)),
+            _ => None,
+        }
+    } else {
+        None
+    };
+
+    // Reset LoRA cache state unconditionally. `saved` is a reusable buffer; a
+    // prior LoRA-bound call would otherwise leave stale rank/matrices/h_* here.
+    // gdn_backward gates its LoRA-gradient path on saved.lora_rank, so a leftover
+    // nonzero rank after a no-LoRA call computes phantom gradients. The Some branch
+    // below repopulates these; the None path now stays truly LoRA-free.
+    saved.lora_rank = 0;
+    saved.lora_scale = 0.0;
+    saved.h_qkv.clear();
+    saved.h_z.clear();
+    saved.h_b.clear();
+    saved.h_a.clear();
+    saved.h_out.clear();
+    saved.gated_buf.clear();
+    saved.lora_a_qkv.clear();
+    saved.lora_b_qkv.clear();
+    saved.lora_a_z.clear();
+    saved.lora_b_z.clear();
+    saved.lora_a_b.clear();
+    saved.lora_b_b.clear();
+    saved.lora_a_a.clear();
+    saved.lora_b_a.clear();
+    saved.lora_a_out.clear();
+    saved.lora_b_out.clear();
+
+    // Initialise LoRA cache buffers on the saved struct when LoRA is active.
+    // Also clone the weight matrices so gdn_backward doesn't need them threaded through.
+    if let Some((a_qkv, b_qkv, a_z, b_z, a_b, b_b, a_a, b_a, a_out, b_out)) = lora_bound {
+        saved.lora_rank = lora_rank;
+        saved.lora_scale = lora_scale;
+        saved.h_qkv = vec![0.0f32; seq_len * lora_rank];
+        saved.h_z = vec![0.0f32; seq_len * lora_rank];
+        saved.h_b = vec![0.0f32; seq_len * lora_rank];
+        saved.h_a = vec![0.0f32; seq_len * lora_rank];
+        saved.h_out = vec![0.0f32; seq_len * lora_rank];
+        saved.gated_buf = vec![0.0f32; seq_len * output_dim];
+        // Clone weight matrices for backward use.
+        saved.lora_a_qkv = a_qkv.to_vec();
+        saved.lora_b_qkv = b_qkv.to_vec();
+        saved.lora_a_z = a_z.to_vec();
+        saved.lora_b_z = b_z.to_vec();
+        saved.lora_a_b = a_b.to_vec();
+        saved.lora_b_b = b_b.to_vec();
+        saved.lora_a_a = a_a.to_vec();
+        saved.lora_b_a = b_a.to_vec();
+        saved.lora_a_out = a_out.to_vec();
+        saved.lora_b_out = b_out.to_vec();
+    }
+
     // Rolling conv buffer (shared across time, updated per step)
     let mut conv_buf_live = vec![0.0f32; qkv_dim * buf_len];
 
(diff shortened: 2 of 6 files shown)


ohdearquant:

## Review of PR #792 (round 1)

### Verdict: REJECT

Two correctness/validation defects remain: one blocker in the GDN alpha/beta head dimension and one medium test-coverage defect. The passing gradchecks are real for the terms they exercise, but they do not establish the PR body's claimed coverage of all ten arrays.

### Prior review context

- No prior GitHub review rounds or inline review threads were present at head `efe20fd46`.
- PR #202 explicitly left its gradient check pending. This round therefore treated PR #792's finite-difference evidence as untrusted until rerun and mutation-tested.

### What this round examined

- Read the full six-file PR diff and the changed source, adjacent GDN inference/weight-loader code, RMSNorm primitives, Adam implementation, PEFT safetensors serializer/parser, PR #202 context, repository policies, and ADR-058.
- Enumerated every workspace caller of `gdn_forward_save`, `gdn_backward`, `forward_full`, `eval_chain_nll`, and `nll_and_grads`.
- Ran the two GDN LoRA weight gradchecks at unmodified head, then independently scaled `grad_b_qkv` by 0.5 in a disposable worktree and reran them.
- Instrumented the disposable gradcheck to report how many sampled entries in each of the ten arrays actually crossed the `1e-4` noise floor.
- Checked shifted-gamma composition, GQA-path diff, empty-GDN behavior, Adam argument/state parity, PEFT naming/shape consistency, local fmt/clippy/tests/builds, and GitHub status checks at the exact head.

### Findings (# total: 1 blocker, 0 majors, 1 medium)

#### Blockers

1. [`crates/inference/src/attention/gdn_backward.rs:509`](crates/inference/src/attention/gdn_backward.rs#L509) GDN alpha/beta are implemented per key head, but the model and shipping forward define them per value head
   - **Current**: `gdn_forward_save` projects `in_proj_b`/`in_proj_a` to `num_kh`, indexes the resulting gates through `kh = h / ratio`, and `gdn_backward` reduces all value-head alpha/beta adjoints into key-head slots (`crates/inference/src/attention/gdn_backward.rs:995-1009`, `crates/inference/src/attention/gdn_backward.rs:1170-1205`). The same dimension is propagated into `GdnLoraParams` (`crates/tune/src/lora/train_core.rs:45`, `crates/tune/src/lora/train_core.rs:166-169`), export (`crates/tune/src/bin/train_grad_full.rs:965-982`), and validation (`crates/tune/src/lora/mod.rs:242-243`).
   - **Issue**: the authoritative weight loader requires `in_proj_a`, `in_proj_b`, `A_log`, and `dt_bias` to have `linear_num_value_heads()` rows (`crates/inference/src/weights/f16_weights.rs:851`, `crates/inference/src/weights/f16_weights.rs:1026-1048`), and the shipping fused forward projects and indexes them per value head (`crates/inference/src/attention/gdn_fused.rs:421-448`, `crates/inference/src/attention/gdn_fused.rs:486-503`). The distinction is observable in supported configs: Qwen3.5-4B is 16 key heads / 32 value heads and 27B is 16 / 48 (`crates/inference/src/model/qwen35_config.rs:277-280`, `crates/inference/src/model/qwen35_config.rs:323-326`); the official upstream [4B config](https://huggingface.co/Qwen/Qwen3.5-4B/blob/main/config.json) and [27B config](https://huggingface.co/Qwen/Qwen3.5-27B/blob/main/config.json) independently confirm those values. The new code therefore trains a different forward graph on asymmetric models, exports too-short B factors, and then incorrectly blesses those shapes in `validate_against`. The multi-head fixture is self-consistent with this wrong graph, so its finite-difference oracle cannot catch the mismatch.
   - **Fix**: size/project `in_proj_a` and `in_proj_b`, their saved activations, gates, LoRA B factors, backward adjoints, export dimensions, and validation dimensions by `value_heads`. Keep key-head sharing only for Q/K slices. Add a source-parity regression using an asymmetric 2-key/4-value fixture and explicit 4B/27B validation-shape tests.
   - **Test**: compare LoRA-off `gdn_forward_save` against the scalar/shipping GDN forward for an asymmetric-head fixture, then finite-difference all ten arrays on that corrected graph. Add `validate_against` cases requiring d_out 32 and 48 for `in_proj_a/b`.
   - **Confidence**: derived from the shipping forward, fail-closed weight-loader shapes, and supported config presets. Strongest rebuttal considered: the binary defaults to 0.8B, where key/value head counts both equal 16. That does not make the generic Qwen3.5 validation/export surface correct, and the trainer accepts other model directories/configs without enforcing a symmetric-head restriction.

#### Mediums

1. [`crates/inference/src/attention/gdn_backward.rs:2115`](crates/inference/src/attention/gdn_backward.rs#L2115) The gradcheck's aggregate threshold leaves four advertised arrays completely unchecked
   - **Current**: the test aggregates one `n_tested` across all arrays and requires only `n_tested >= 10` (`crates/inference/src/attention/gdn_backward.rs:2129-2154`, `crates/inference/src/attention/gdn_backward.rs:2180-2188`). The PR body says the oracle passes “across all 10” arrays and explains the fixture change as lifting alpha/beta signal.
   - **Issue**: disposable instrumentation showed per-array counts of `6,5,6,6,0,0,0,0,7,6` for the tuned single-head fixture and `7,2,1,2,0,0,0,0,3,4` for the multi-head fixture. Arrays 4-7 are `grad_a_b`, `grad_b_b`, `grad_a_a`, and `grad_b_a`; none is compared at all. The 0.5→2.0 scale and 4→6 sequence tuning increased aggregate signal but did not make the claimed alpha/beta gradients testable.
   - **Fix**: track coverage and max relative error per array, require at least one (preferably several) above-floor entries for every array, and redesign the fixture/loss until alpha/beta terms cross the floor without relying on an aggregate escape hatch.
   - **Test**: mutation-test one alpha/beta A term and one B term after the per-array gate is added. My independent `grad_b_qkv *= 0.5` mutation already failed both tests at `max_rel≈0.500`, confirming broad sensitivity for a covered term but not repairing the alpha/beta blind spot.
   - **Confidence**: observed by instrumenting and running the disposable worktree test serially with `--nocapture --test-threads=1`.

- **Unmodified numerical tests**: observed both `attention::gdn_backward::tests::gradcheck_gdn_lora_weight_grads` tests pass at head.
- **Independent mutation**: observed both tests fail when `grad_b_qkv` accumulation is scaled by 0.5, with maximum relative errors `5.001e-1` and `5.003e-1`. No source in the review worktree was modified.
- **No-LoRA caller coverage**: workspace search found production callers only in `train_core.rs` and `diff_gdn_layer.rs`; all pre-existing no-LoRA paths pass ten `None`s plus rank/scale `0/0.0`. Test callers are likewise updated. No call site was missed.
- **No-LoRA arithmetic**: with `lora_bound=None`, every added projection/out branch is skipped and the pre-existing arithmetic/order remains unchanged. Reusable LoRA caches are cleared before the call, preventing stale-cache phantom gradients. The shipping inference path is `gdn_fused`, not `gdn_forward_save`, so it pays no new allocation or inner-loop branch.
- **Shifted gamma**: residual/final norms call `shifted()` once when building `LayerW`/`Head`; `rms_norm_forward` and `rmsnorm_backward` consume that plain weight consistently. GDN's internal `linear_attn.norm_weight` is passed directly in both `gdn_forward_save` and `gdn_backward`, so it is not shifted. Gamma is applied exactly once on each path.
- **GQA / empty-GDN path**: the GQA forward/backward bodies are unchanged apart from surrounding signatures/return tuple. `train_micro_lora` assigns GDN layers no slot, passes an empty GDN slice and zero GDN slot count, and reaches the explicit all-None GDN branch.
- **Adam**: `apply_gdn_adam_updates` uses the same `AdamState::step` arguments as `apply_adam_updates` (including `eps` placement and zero/no weight decay). `AdamState` maintains bias-correction timesteps per key, so adding ten tensors does not advance GQA tensors' counters.
- **PEFT names/round-trip logic**: export names exactly match inference hook names. The safetensors writer emits unknown-to-GQA modules under `self_attn`, while the parser intentionally takes the last module segment. A disposable test containing all five 0.8B GDN modules observed save → load → `validate_against` pass, and a legacy q/v-only GQA adapter passed the same round trip. The head-count blocker above still makes asymmetric-model alpha/beta shapes wrong.
- **Benchmark discipline**: ADR-058 is superseded as a CI design but retained as `make bench-compare` provenance; repository policy still requires the run. The PR discloses both attempts, exact SHAs/commands, does not claim the contention-corrupted first run, and reports no FAIL/WARN in the touched inference benchmark group on the clean rerun. The residual mixed-direction embed results are disclosed and come from an untouched crate. For this training-only additive surface, that is an honest and proportionate performance record.
- **Formatting and lint**: observed `cargo fmt --all -- --check`, `git diff --check`, and `cargo clippy --workspace --all-targets --features train-backward,safetensors,inference-hook -- -D warnings` pass at head.
- **Feature/build/test matrix**: observed `cargo test -p lattice-inference -p lattice-tune --features train-backward,safetensors,inference-hook` pass in full (inference: 1622 passed/18 ignored in the main lib target; tune: 262 passed/1 ignored, plus all bin, integration, and doctest targets). Observed `cargo build --workspace` pass with default features, so the new training surface creates no default-build warning/error.
- **CI at head**: GitHub reports `ci-gate`, Ubuntu/macOS/ARM CI, both feature-matrix jobs, MSRV, bench-compile, app-binaries, e2e parity, Q4 PPL/composed golden, and parity-gate successful for `efe20fd46`.

### What I did NOT check

- I did not run Metal/GPU tests, as requested.
- I did not run the real-model `train_grad_full --gradcheck`/TBV path because it requires a local Qwen3.5 checkpoint and dataset; the unit-level oracle and source-parity analysis above were sufficient to expose the blocker.
- I did not rerun `make bench-compare`; I audited the PR's disclosed commands/results against the training-only diff and ADR/repository policy.

- Another round is useful after the alpha/beta dimension is corrected and each of the ten arrays has a per-array finite-difference coverage gate. The next round should concentrate on asymmetric-head forward parity and save → load → validate behavior for 4B/27B shapes.

Domain utility: MEDIUM — the benchmark-noise guidance helped calibrate the two-run disclosure, while the decisive findings came from repository-specific GDN shape and gradient-oracle analysis.

Verdict: REJECT

ohdearquant:

## Review of PR #792 (round 2)

### Verdict: REJECT

### Prior review context

- Round 1 at `efe20fd46` rejected the PR for one blocker: alpha/beta were dimensioned by key heads instead of value heads.
- Round 1 also found one medium: an aggregate finite-difference coverage gate allowed four alpha/beta arrays to have zero entries above the oracle noise floor.
- This focused reconfirm reviews only `efe20fd46..a18bf7147` and the two named remediations.

### What this round examined

- Source-level dimensioning and indexing for alpha/beta across inference saved state/backward, tune allocations/export, and adapter validation.
- Asymmetric forward parity, mutation sensitivity, asymmetric validation rejection, and save-load-validate behavior.
- Per-array finite-difference counts and the scope of f64 oracle accumulation.
- Remediation diff scope, PR-body evidence, local gates, and GitHub CI at the exact head.

### Findings (# total: 1 blocker)

#### Blockers

1. [`crates/tune/src/bin/train_grad_full.rs:505`](crates/tune/src/bin/train_grad_full.rs#L505) The asymmetric alpha/beta fix missed both live `train_grad_full` LoRA constructors
   - **Current**: gradcheck mode still constructs `b_b` and `b_a` with `gdn_dims.num_kh * rank` at lines 505 and 507; training mode repeats the same key-head-sized allocations at lines 790 and 792.
   - **Issue**: `forward_full` passes these slices into the corrected `gdn_forward_save`, which now projects alpha/beta to `value_heads` rows. On 4B/27B-class asymmetric configurations these matrices are too short, so the binary's real gradcheck/training path cannot consume the very shapes this round claims to support. Export now labels the learned matrices with `d_out = value_heads`, which also disagrees with their allocated storage length. This is the exact surviving `num_kh`-sized alpha/beta buffer the round-1 next-focus sweep was meant to eliminate.
   - **Fix**: use `gdn_dims.value_heads * rank` for `b_b` and `b_a` in both constructors; preferably route both through a single shape-aware constructor so the already-correct `GdnLoraParams::zeros` logic cannot drift from randomized/training initialization again.
   - **Test**: add an asymmetric `GdnDims { num_kh: 2, value_heads: 4, ... }` constructor/forward test covering the binary's nonzero gradcheck initialization and zero-B training initialization, or extract the constructors to testable library helpers.
   - **Confidence**: derived from the committed constructor sites and the corrected `gdn_forward_save` contract; the production training constructor is reached at lines 783-868. Strongest rebuttal considered: the default 0.8B config has symmetric heads, but the PR explicitly claims generic 4B/27B asymmetric support and `train_grad_full` loads dimensions from the selected model config.

### Evidence log

- Exact reviewed head: `a18bf7147e8d25cb127fbf19153cc7778725759f`.
- Remediation diff: four files, 415 insertions / 131 deletions. Its changes are confined to value-head re-dimensioning, per-array/oracle tests, validation tests, and explanatory comments; no unrelated feature or production numerical change was found.
- `source_parity_forward_save_vs_fused_asymmetric_heads` passed at the exact head (1/1).
- Both new asymmetric `validate_against` regressions passed: d_out 32 and 48 validate, while each test also asserts the old d_out 16 is rejected.
- A disposable combined regression saved, loaded, and validated `in_proj_b`/`in_proj_a` adapters for both 32- and 48-value-head configs; all four round trips passed. This agrees with the traced format contract: save emits B as `[layer.d_out, rank]` (`safetensors.rs:627-628`) and load reconstructs `d_out = b_shape[0]` (`safetensors.rs:281-352`).
- The PR body's named per-array counts reproduce exactly for all ten arrays in both fixtures:
  - single head: `7,6,7,8,3,2,2,2,8,6`;
  - asymmetric 2-key/4-value: `7,7,7,7,6,8,4,4,7,7`.
  Every count is nonzero and each fixture enforces `n_tested >= 1` per array.
- `fd_lora_weight_entry` casts forward outputs and coefficients to f64 only inside the `#[cfg(test)]` finite-difference loss closure (`gdn_backward.rs:1960-2063`). The production forward/backward remain f32; the only production numerical changes in this round are the required value-head projection/index axes.
- Source sweep found surviving key-head allocations for alpha/beta B factors at `train_grad_full.rs:505,507,790,792`.
- Mutation verification passed: in a disposable worktree, changing only `GdnSaved::beta_raw` back from `seq_len * value_heads` to `seq_len * num_key_heads` made `source_parity_forward_save_vs_fused_asymmetric_heads` fail with an out-of-range slice at `gdn_backward.rs:516`; the unmodified head passes.
- GitHub `ci-gate`, Ubuntu/macOS/ARM CI, feature matrices, MSRV, semver, bench-compile, app binaries, embed drift, Metal informational parity, WASM parity, and Q4 PPL passed at the exact head. Only the top-level e2e parity and Q4 composed golden legs were still pending at last inspection. The draft branch is also behind `main` and must be refreshed before merge under the repository's strict merge gate.

- The inference-side correction is internally consistent: `GdnSaved` alpha/beta buffers, forward projections and gate indexing, backward adjoints, LoRA B-gradient allocations, and VJPs all use `value_heads`; Q/K offsets alone retain `kh = h / ratio` sharing.
- `GdnLoraParams::zeros`, PEFT export metadata, and `validate_against` use `value_heads` as claimed.
- The per-array test gate replaces the aggregate escape hatch and independently requires coverage for every named array. The retained aggregate count is redundant but no longer permits an untested array.
- The f64 change is an oracle-precision change only, not a production tolerance or arithmetic change. Rust's official `Iterator::sum` documentation confirms the iterator item/output type determines the available f64 sum implementation: https://doc.rust-lang.org/std/iter/trait.Sum.html.
- `cargo fmt --all -- --check`, `git diff --check`, the exact source-parity test, both gradchecks, and both asymmetric validation tests passed locally at head.
- Heavy-lane `cargo clippy --workspace --all-targets --features train-backward,safetensors,inference-hook -- -D warnings` passed.
- Heavy-lane `cargo test -p lattice-inference -p lattice-tune --features train-backward,safetensors,inference-hook` passed all library, binary, integration, and doctest targets (main libraries: inference `1623 passed / 18 ignored`, tune `264 passed / 1 ignored`).
- Heavy-lane `cargo build --workspace` with default features passed.
- The PR body now makes the “all 10 arrays” claim with per-array evidence, and the reproduced counts match it exactly.

### What I did NOT check

- No Metal/GPU execution was performed, as requested.
- I did not run the real-model `train_grad_full` gradcheck/training path because it requires a local asymmetric checkpoint and dataset; source tracing is sufficient to establish that its B-factor slices are too short.
- Two GitHub e2e-parity legs were still pending at the final status poll; no completed check was red.

- One short reconfirm is useful after the four `train_grad_full` allocations are corrected. Repeat the `num_kh` alpha/beta sweep and exercise an asymmetric constructor/forward path; no broader math round is indicated unless that fix changes more than those shapes/tests.

Domain utility: MEDIUM — numerical-oracle guidance helped distinguish the f64 finite-difference improvement from production arithmetic, while the decisive blocker came from repository-specific shape/caller tracing.

Verdict: REJECT

ohdearquant:

## Review of PR #792 (round 3)

### Verdict: APPROVE

### Prior review context

- Round 1 found alpha/beta dimensioned by key heads instead of value heads and an aggregate finite-difference coverage escape hatch.
- Round 2 confirmed the inference/tune remediation but found four surviving key-head allocations in the two live `train_grad_full` LoRA constructors.
- This short reconfirm examines only that blocker’s remediation, regression sensitivity, bug-class sweep, remediation/merge scope, and required gates.

### What this round examined

- The shared `GdnLoraParams::shaped` shape derivation and all three callers: `zeros`, gradcheck initialization, and zero-B training initialization.
- Both new asymmetric binary-constructor tests before and after mutating the shared `b_b` dimension back to key heads.
- The full-PR alpha/beta `num_kh` sweep and the surviving `GdnSaved::new` call.
- `a18bf7147..7976ed0e1` commit/tree scope, PR-body claims, branch freshness, local gates, and exact-head GitHub checks.

### Findings (# total: 0)

No blocker, major, medium, minor, or nit was found in the requested reconfirm scope.

### Evidence log

- Exact reviewed head: `7976ed0e149fe0b39c99a2ad2fae293514c2762f`.
- [`crates/tune/src/lora/train_core.rs:176`](crates/tune/src/lora/train_core.rs#L176) is the one shared shape-aware constructor. Its only alpha/beta B-factor sizing expressions are `checked(gd.value_heads, rank, ...)` at lines 199 and 202; [`zeros`](crates/tune/src/lora/train_core.rs#L208) delegates to it rather than repeating sizes.
- [`crates/tune/src/bin/train_grad_full.rs:207`](crates/tune/src/bin/train_grad_full.rs#L207) and [`crates/tune/src/bin/train_grad_full.rs:250`](crates/tune/src/bin/train_grad_full.rs#L250) are the two extracted constructors. Both call `GdnLoraParams::shaped`; the live gradcheck and training paths call those helpers at lines 574 and 849. No inline `b_b`/`b_a` sizing remains in the binary.
- The two asymmetric `num_kh=2, value_heads=4, rank=3` constructor tests pass at head: `2 passed / 0 failed`. In a disposable worktree, changing only shared `b_b` sizing from `gd.value_heads` to `gd.num_kh` made both tests fail (`left: 6`, `right: 12`; cargo exit 101). The worktree was removed afterward.
- The repeated full-PR/source sweep found zero key-head-scoped alpha/beta sizing survivors. [`crates/tune/src/lora/train_core.rs:346`](crates/tune/src/lora/train_core.rs#L346) passes `gdn_dims.num_kh` as the first head-count argument to [`GdnSaved::new`](crates/inference/src/attention/gdn_backward.rs#L223), whose signature is explicitly `(seq_len, num_key_heads, value_heads, ...)`; this is the legitimate Q/K-sharing parameter claimed by the author.
- `a18bf7147..7976ed0e1` consists of fix commit `6b68486f0` plus merge commit `7976ed0e1`. The fix changes only `train_grad_full.rs` and `train_core.rs`. The merge’s first-parent patch is byte-for-byte identical to main commit `65865ecb5` relative to its parent `27b9047ee` (patch comparison exit 0), so it contains only main’s five ADR/index changes and no hidden feature-side edits.
- The PR body’s round-2 section accurately names the shared constructor, both extracted helpers/tests, the surviving legitimate `num_kh` call, and the local gate counts.
- Freshness was re-fetched at `2026-07-10T17:34:48-04:00`: `origin/main=65865ecb5`, remote branch/head=`7976ed0e1`, `git merge-base --is-ancestor origin/main HEAD` exited 0, and `HEAD...origin/main` was `4 0`. The branch is up to date with main.
- Local gates observed at head:
  - `cargo fmt --all -- --check` and `git diff --check`: pass.
  - Heavy-lane `cargo clippy --workspace --all-targets --features train-backward,safetensors,inference-hook -- -D warnings`: pass.
  - Heavy-lane `cargo test -p lattice-inference -p lattice-tune --features train-backward,safetensors,inference-hook`: pass across library, binary, integration, and doctest targets. Main libraries: inference `1623 passed / 18 ignored`; tune `264 passed / 1 ignored`; `train_grad_full` `2 passed / 0 failed`.
- GitHub exact-head CI at the final `2026-07-10T17:55:34-04:00` poll: every completed check is green, including required `ci-gate`, `app-binaries-gate`, `cargo-deny`, `typos`, and `actionlint`; Ubuntu/macOS/ARM CI, feature matrices, MSRV, semver, bench-compile, app binaries, Q4 PPL/composed golden, WASM parity, and embed drift also pass. `e2e parity` and informational Metal parity remain in progress, so required `parity-gate` has not yet rendered. Branch protection must continue to hold the merge until it does.

- The round-2 blocker is fixed through real sharing rather than three copied size tables.
- Both extracted binary construction modes are covered by asymmetric, mutation-sensitive tests.
- The fix/merge/PR-body scope matches the remediation claim.
- Local formatting, lint, and requested crate suites pass; the branch is current with main.

### What I did NOT check

- I did not observe final conclusions for the still-running e2e and informational Metal parity legs before the review deadline. No completed check was red, but strict branch protection must prevent merge until required `parity-gate` renders green.
- No Metal/GPU execution or real-model training run was performed locally, as requested.
- No broader math round or benchmark rerun was performed because the remediation changes constructor shape centralization and tests only.

- Stop after this reconfirm; no further code-review round is indicated. The remaining action is mechanical: wait for exact-head `parity-gate` to render green before merge.

Domain utility: LOW — the suggested Rust performance domain was not material to this shape-correctness reconfirm; repository-specific source tracing and mutation evidence were decisive.

Verdict: APPROVE

oceanwaves630 (approved):
Approving per the documented gate trail: two independent audit legs at the current head; asymmetric-heads blocker closed with shape-aware constructor and full FD coverage.

Issue #791, opened 2026-07-10
chore(inference): delete undispatched legacy Metal attn_scores/attn_softmax/attn_context kernels (ADR-080 C1)


ohdearquant:
## Summary

The legacy Metal kernels `attn_scores` / `attn_softmax` / `attn_context` (`crates/inference/src/forward/shaders/flash_attention.metal:127-260`) are dead code: no `make_pipeline` call references them (the 7 instantiated pipelines are matmul_bt, rms_norm, fused_attention, fused_qk_norm_rope, silu_mul, copy_buf, add_buf), no dispatch site exists, and the MSL compile test covers only the live kernel names. ADR-080 C1 already states the disposition: delete (dead code, not a live divergence to fix).

## Notes for the deletion PR

- The dead `attn_softmax` uses a finite max sentinel (`local_max = -1e30f`) and the same `sum_val > 0.0f ? ... : 0.0f` multiply-through-zero trap as the live kernel bug #789 — dead does not mean correct; do not resurrect without applying the C1 contract.
- The similarly-named WGPU pipeline fields under `forward/gpu/inner/` belong to a different backend (see #790) and are NOT part of this deletion.

## Provenance

ADR-080 C1 (#780) checklist residue; static dispatch-site evidence verified at origin/main 27b9047ee.

ohdearquant:
Closed by PR #794 (merged 2026-07-11): fail-closed Metal fused_attention finalize + deletion of the undispatched attn_scores/attn_softmax/attn_context kernels. PR lacked a closing keyword.

Issue #790, opened 2026-07-10
bug(inference): WGPU attention_softmax floor-clamps NaN denominator instead of failing closed


ohdearquant:
## Summary

The WGPU `attention_softmax` kernel floor-clamps its softmax denominator, converting a NaN sum into a finite reciprocal instead of zeroing the row — the same defect class as the CPU `batch_prefill` floor-clamp fixed in #785. NaN input rows fail open and propagate.

## Location

`crates/inference/src/forward/gpu/inner/shaders.rs:341` (WGSL kernel body):

```wgsl
let inv_sum = 1.0 / max(sum_scratch[0u], 1e-20);
```

There is no pre-exp NaN scan (no analog to `softmax_row::row_fails_closed_pre_exp`); the kernel always computes forward and only clamps the reciprocal. WGSL `max()` drops a NaN operand, so `max(NaN, 1e-20) = 1e-20` produces a finite `inv_sum` — but the numerator lanes are already NaN, and `NaN * finite = NaN`.

## Reachability

`forward::gpu` is gated behind the `wgpu-gpu` feature and, at origin/main 27b9047ee, is not wired into any production binary — the only consumer is `examples/bench_gpu.rs`. It is a real, compiling, publicly-constructible path (`GpuModelState::new`/`forward`), but not on the serving path today. Triage accordingly (lower priority than #789, same contract).

## Reproduction (deterministic, 3/3 identical runs)

Same isolation technique as #789: tiny synthetic model, single corrupted weight `q_proj_weight[11,5] = NaN`. Repro source: `repro_wgpu_softmax_nan.rs` (runs as an example with `--features "f16,wgpu-gpu"`).

- GPU output: 160/160 elements NaN
- CPU fail-closed reference (crate's own `softmax_row` helpers): 0/160 NaN

## Fix shape

Port the ADR-080 C1 row contract to the WGSL kernel: pre-exp NaN/+inf detection and whole-row zero by assignment on the fail-closed conditions; remove the floor-clamp (a non-positive or non-finite denominator must zero the row, not clamp the reciprocal). Mutation-sensitive regression test.

## Provenance

ADR-080 C1 GPU residue tracked on #780 (CPU side closed by #785). Reproduced at origin/main 27b9047ee.

Issue #739, opened 2026-07-09
bug(inference): Qwen3.5 f32/F16/batched-prefill attention rows don't fail closed on NaN


ohdearquant:
**Defect**: The Qwen3.5 f32 CPU decode path (`compute_attention_context`), the F16 sibling, and the batched-prefill online-softmax recurrence all divide by `sum_exp`/`running_sum` without a finiteness or positivity guard. A cached K containing NaN makes one QK score (and therefore `sum_exp`) NaN, and the NaN propagates through V/context/logits unguarded. This contradicts the fail-closed contract the Q8 and NEON siblings already implement (explicit `if sum_exp > 0.0 { .. } else { zero the row }`).

**Evidence**:
- `crates/inference/src/model/qwen35/forward.rs:393-399` — unconditional `let inv_sum = 1.0 / sum_exp;`, no guard.
- `crates/inference/src/forward/cpu_f16.rs:369-375` — byte-identical unconditional-divide pattern.
- `crates/inference/src/forward/batch_prefill.rs:1085-1088` — `running_sum.max(f32::MIN_POSITIVE)` floors a NaN denominator to a finite value via `f32::max`'s non-NaN-operand semantics, but the V accumulator (`acc[d] += weight * v[...]`) is already NaN before the floor runs, so the floored denominator does not fix anything.
- Contrast: `crates/inference/src/forward/cpu_q8.rs:417-432` — explicit `if sum_exp > 0.0 { .. } else { fill 0.0 }` guard, citing prior fixes for the same class of bug.
- `crates/inference/src/forward/neon_forward.rs:653` — NEON sibling uses the same guarded rule as Q8.

**Blast radius**: Every normal `Qwen35Model` generate/streaming/perplexity forward step (f32 CPU path), the F16 generation mode, and the batched-prefill path (public API, reachable directly).

**Suggested fix**: Extract a single allocation-free fail-closed online-softmax finalizer (finite-check + zero-row fallback) and use it uniformly in the f32, F16, Q8, NEON, and batched-prefill implementations instead of maintaining four independent copies of the same normalization step.

Provenance: Audited at 13c8de8a3; file unchanged between SHAs.

ohdearquant:
[...]

Issue #611, opened 2026-07-03
bug(inference): Metal generation fails OPEN on empty grammar mask — CPU's finite-logit guard missing on all Metal paths


ohdearquant:
> _Issue authored by Claude (Anthropic agent) on behalf of @ohdearquant._

## Problem

The Metal generation paths apply grammar masks to logits but never perform the fail-closed finite-logit check the CPU paths do. Verified on main: `metal_qwen35.rs` calls `engine.mask_logits(...)` at its prefill/decode mask sites (e.g. `crates/inference/src/forward/metal_qwen35.rs:10351` area, plus the streaming and prefix-cache streaming variants) with no check that at least one finite logit survived. The generic and Qwen3.5 CPU paths guard exactly this: `has_finite_logit` (`crates/inference/src/generate.rs:383`) and return `InvalidGrammarState` on an empty mask.

Consequence: on Metal, an all-invalid grammar mask (grammar bug, exhausted PDA state, or pathological schema) continues into sampling over all-`-inf` logits instead of failing closed. Depending on the sampler's NaN/all-masked handling this either produces an arbitrary token that violates the grammar or degenerate output — the same fail-open family as the PDA over-accept class, on the highest-throughput path.

History context: grammar enforcement was wired path-by-path (#397, then the empty-mask fail-closed fix landed for CPU) — the Metal siblings never got the second half.

## Acceptance

- [ ] Every Metal mask site (plain, cancellable streaming, prefix-cache streaming; prefill and decode) checks for a surviving finite logit and fails closed with the same error contract as CPU
- [ ] Mutation-sensitive test: a grammar state producing an all-invalid mask on the Metal path must yield the error, not a token
- [ ] Softmax fail-closed invariants (all-(-inf) row handling) re-verified on the Metal sampler as defense-in-depth

PR #794, opened 2026-07-10
fix(inference): fail-closed Metal fused_attention finalize + delete dead kernels

### LINKED ISSUE #789: bug(inference): Metal fused_attention zero-row branch multiplies NaN by zero instead of assigning zero (fail-open)
## Summary

The live Metal `fused_attention` kernel's zero-row branch cannot actually zero a NaN accumulator: it multiplies by zero instead of assigning zero, and IEEE-754 defines `NaN * 0.0 == NaN`. A NaN entering attention (e.g. one corrupted Q lane) propagates through the whole downstream network instead of failing closed per the ADR-080 C1 row contract.

## Location

`crates/inference/src/forward/shaders/flash_attention.metal:596-599`:

```c
const float inv_l = (l_i > 0.0f) ? (1.0f / l_i) : 0.0f;
O4[out_base4] = o_frag * inv_l;
```

The `l_i > 0.0f` comparison correctly detects the NaN denominator (comparison is false, `inv_l = 0.0f`), but `o_frag` is also NaN by then, and `NaN * 0.0f = NaN`. Contrast the CPU contract (`softmax_row::finalize_row`, merged in #785), which zeroes by assignment (`row.fill(0.0)`).

This kernel is the only attention kernel instantiated (`make_pipeline("fused_attention")`) and is on the live serving path via the Qwen Metal backend.

## Reproduction (deterministic, 3/3 identical runs)

Tiny synthetic model (hidden=32, heads=4, kv_heads=2, head_dim=8, seq_len=5, 1 layer) with a single corrupted weight `q_proj_weight[11,5] = NaN` — isolates the softmax finalize as the only place the NaN could be cured. Repro source: `repro_metal_fused_attention_nan.rs` (runs with `--features "f16,metal-gpu"` under the machine GPU lock).

- Metal output: 160/160 elements NaN
- CPU fail-closed reference (crate's own `softmax_row` helpers): 0/160 NaN

Mechanism: `p_ij = exp(NaN - m_new) = NaN` regardless of the max; `l_i` and `o_frag` both go NaN; the finalize multiply-through-zero cannot recover; `o_proj` spreads the single corrupted head into every output element.

## Fix shape

Fail-closed finalize by assignment: detect the bad denominator (and NaN accumulator) and write `0.0` to the output lanes directly, mirroring `finalize_row`'s semantics. Regression test must be mutation-sensitive (revert to multiply-through-zero → red).

## Provenance

ADR-080 C1 GPU residue tracked on #780 (CPU side closed by #785). Reproduced at origin/main 27b9047ee.

ohdearquant:
# PR A: Metal fused_attention fail-closed finalize + dead kernel deletion (ADR-080 C1)

Fixes #789, #791.

## Fix round (2026-07-10): codex round-1 REJECT resolved

Round 1 rejected on a blocker (zero-row vs typed-error contract conflict) and a
major (the two `forward::metal::` tests were never run by any CI job and
silently skip-passed with no Metal device). Both are resolved in this round;
see the round-1 review for full detail.

**Blocker — ruling, not new plumbing.** Zero-row (this PR's existing kernel
behavior: zero the bad row by direct assignment, forward returns `Ok`) is the
canonical fail-closed semantic for **internal** attention softmax — a row that
feeds a matmul inside the forward pass and is never itself host-visible. It is
not weakened to accommodate this PR; it was already ADR-080 C1's contract and
already the behavior of every other canonical site (CPU GQA, CPU decode). The
conflict the review found was against ADR-066 D2 invariant #1's literal
wording ("normalizes to Σ=1.0±ε or returns `InvalidInput`"), which was written
for the **public-boundary** softmax class (#611 sampling/serve) and had never
been explicitly scoped away from internal attention softmax. `docs/adr/ADR-066-output-correctness-gate-architecture.md`
is amended in this PR's diff (D2 invariant #1 + a new "Amendment
(2026-07-10)" subsection) to state the scope split explicitly and to name the
tradeoff without hedging: zeroing swallows numerical *defects* (a NaN produced
by an upstream bug becomes a silent zero row, not a raised error), and
ADR-066's own differential/quality-gate layers (D1 L2/L3) remain the detection
surface for that class — this amendment does not read as "NaN is fine," it
scopes which layer is responsible for catching it. No host-visible error
plumbing is added to the Metal kernel or `MetalForwardPass`.

**Major — CI enforcement.** Both `forward::metal::` tests
(`msl_template_assembles_and_compiles`,
`fused_attention_fails_closed_on_nan_q_lane`, and the new
`fused_attention_fails_closed_on_inf_q_lane`) now honor
`LATTICE_METAL_TEST_ENFORCE=1`: a missing Metal device or a failed
`MetalForwardPass::new` construction panics instead of silently returning
`Ok`/skip, matching the existing q4 and grammar fail-closed convention
(`ci.yml`). `.github/workflows/ci.yml` gets one new macOS-only step,
`inference — metal-gpu fused_attention fail-closed gate`, that runs
`cargo test -p lattice-inference --features f16,metal-gpu --lib
forward::metal::` under that env var and greps the exact count
(`3 passed; 0 failed`) so a zero-match filter cannot pass silently.

**Test coverage — all three invalid-row classes from the review, as far as
the architecture allows.** The regression is now two tests sharing one
fixture builder (`run_fused_attention_with_corrupted_q_lane`), both
mutation-sensitive against the same finalize guard:

- **NaN** (`fused_attention_fails_closed_on_nan_q_lane`, unchanged from round
  1): one `q_proj_weight` entry set to `f32::NAN`.
- **+inf** (`fused_attention_fails_closed_on_inf_q_lane`, new): the identical
  construction with the entry set to `f32::INFINITY` instead; asserts
  `!is_finite()` rather than `!is_nan()` since the sign of the surviving
  accumulation can land on either `+inf` or `-inf` depending on the
  activation's sign at each query position.
- **All-`(-inf)`**: **not covered by a direct kernel-dispatch fixture, by
  design.** `fused_attention` takes no external mask buffer, and the kernel
  always includes the causal self-key (`Q_i · K_i`), which stays finite for
  any finite Q/K pair — there is no way to drive a genuine all-`-inf` score
  row through this live entry point without injecting `-inf` directly into
  the score buffer, i.e. bypassing the boundary this suite is scoped to. The
  finalize guard's arithmetic branch is identical for this class and the
  +inf class above (`l_i` non-positive or non-finite triggers the same
  zero-assignment branch), so the +inf regression already exercises the
  shared fail-closed code path. This is documented inline in
  `forward/metal.rs` next to the two tests, matching what the round-1 review
  itself concluded ("I did not add a direct all-`-inf` kernel-dispatch
  fixture... the missing caller-visible error is already observed for NaN
  and +inf").

**Mutation-sensitivity (this round, both classes)**: reverted the finalize in
a disposable worktree back to `O4[out_base4] = o_frag * inv_l;` (no `isfinite`
guard), `touch`ed `flash_attention.metal`, reran both tests under the GPU
lock — RED (both panicked, 160/160 non-finite elements each). Restored the
fix, `touch`ed again, confirmed the shader source was byte-identical to this
branch's copy, reran — GREEN (0/160 non-finite, both tests).

**Scope of this round**: tests, CI, and ADR text only. `flash_attention.metal`
is untouched in this round — verified via `git diff --stat` showing zero
changes to that file; the live kernel behavior from round 1 is unchanged.

## What changed (round 1)

**#789** — `flash_attention.metal`'s `fused_attention` finalize computed the
correct reciprocal (`inv_l = (l_i > 0.0f) ? (1.0f / l_i) : 0.0f`) but then
applied it with a multiply (`O4[out_base4] = o_frag * inv_l`). Under
IEEE-754, `NaN * 0.0f == NaN`, so a row whose accumulator `o_frag` was
already NaN (from a NaN Q lane) stayed NaN even though `inv_l` correctly
evaluated to `0.0f`. Fixed by assigning the literal `0.0f` directly to the
output lanes when the denominator is non-positive or non-finite, mirroring
`attention::softmax_row::finalize_row`'s `row.fill(0.0)` semantics — an
assignment, never a multiply-through.

**#791** — Deleted the dead `attn_scores`, `attn_softmax`, and `attn_context`
kernels from `flash_attention.metal`. None of the 7 pipelines instantiated
via `make_pipeline` in `forward/metal.rs` (`matmul_bt`, `rms_norm`,
`fused_attention`, `fused_qk_norm_rope`, `silu_mul`, `copy_buf`, `add_buf`)
references them, and no dispatch site exists. Added an explicit negative
assertion to `msl_template_assembles_and_compiles` so the three names cannot
silently resurface.

## Regression test

`fused_attention_fails_closed_on_nan_q_lane` and (added this fix round)
`fused_attention_fails_closed_on_inf_q_lane` (`forward/metal.rs`), adapted
from the filed repro (`repro_metal_fused_attention_nan.rs`): a tiny
1-layer/4-head model with one entry of `q_proj_weight` corrupted (isolates
one head's Q vector at every position; K/V/embeddings/residual stay clean).
Runs under the machine-wide GPU flock, honors
`LATTICE_METAL_TEST_ENFORCE=1` (see fix-round section above).

- Before the fix: 160/160 output elements non-finite (both classes).
- After the fix: 0/160 non-finite (both classes).

## Bench-compare

`make bench-compare` (default groups: `elementwise_cpu_bench` + embed `simd`),
BASE `origin/main` (65865ecb5) vs HEAD (0d3214eee):

```
❌ 21 FAIL (regression >7.0% confirmed by 95% CI)
⚠ 6 WARN (regression 3.0-7.0% confirmed)
🚀 114 confirmed improvement
```

Not a real regression, for two independent reasons:

1. **Structurally unreachable.** This PR's diff touches exactly two files:
   `flash_attention.metal` (a shader source string, only compiled into a
   Metal pipeline and dispatched when the `metal-gpu` feature is enabled)
   and `forward/metal.rs`'s `#[cfg(test)]` module (test-only, excluded from
   every `[[bench]]` binary). `default-features` for `lattice-inference` is
   `["std", "download", "serve"]` — no `metal-gpu`. Neither of the two
   changed files is linked into, referenced by, or reachable from the
   `elementwise_cpu_bench` or embed `simd` Criterion targets that
   `make bench-compare` actually runs (confirmed by grepping
   `crates/inference/benches/` for `flash_attention`/`fused_attention`: zero
   hits). A CPU-only benchmark binary that does not contain the changed code
   cannot be regressed by it.
2. **Bidirectional noise signature.** The same run reports 114 *improved*
   benchmarks of the same shape and magnitude (small SIMD/elementwise ops,
   `Δ` in the tens-of-percent, sub-microsecond absolute times) as the 21
   FAILs — e.g. `elementwise_mul/4096` FAILs at +202% while
   `int8_quantization/quantize/1024` WINs at -4.4% in the same run, neither
   touched by this diff. This is the documented machine-contention pattern
   (several sibling worktrees were building/benching concurrently on this
   host during the run); a genuine regression would not present as a
   scatter of wins and fails across unrelated microbenchmarks in both
   directions.

The actual correctness gate for this change is the mutation-sensitive
regression tests above (RED before the fix, GREEN after) plus the existing
Metal attention/parity suite (12 passed, 2 ignored-as-slow, run under
`forward::metal::` and `test_decode_attention` filters) — `make
bench-compare`'s CPU-only groups have no path to the modified GPU kernel and
are included here only because the task asked for the measurement, not
because they can gate a Metal-only change. This fix round adds no new
kernel code, so the disclosure is unchanged from round 1.

## Sibling-invocation-path sweep

Grepped every `.metal`/`.wgsl` shader source under
`crates/inference/src/forward/shaders/` and `forward/gpu/inner/shaders.rs`,
plus every CPU-side call site sharing the `inv_l`/`inv_sum`/denominator-guard
shape, for the same two bug classes: (a) a `max()`-based reduction silently
dropping a NaN operand, and (b) a finalize that multiplies a possibly-NaN
numerator through a correctly-computed zero reciprocal instead of assigning
zero directly.

| Site | Verdict |
|---|---|
| `attention/flash_causal.rs::normalize_head_output` | Already correct — explicit `bad[row]` fail-closed assignment (`fill(0.0)`) precedes the `inv_l` multiply path; the multiply only runs when the row is known-good. Pre-existing ADR-080 C1 fix. |
| `forward/batch_prefill.rs` (`row_poisoned` guard) | Already correct — `if row_poisoned \|\| !running_sum.is_finite() \|\| running_sum <= 0.0 { fill(0.0) } else { multiply }`; assignment precedes multiply. Landed as #739. |
| `forward/shaders/qwen35.metal::decode_attention`/`_f16`, `decode_attention_flash_partial`/`_reduce` (live Qwen3.5 decode path) | Already correct for the reported bug class — `out[...] = dn > 0.0f ? acc/dn : 0.0f;` is a ternary that assigns the literal `0.0f` to the *output* directly on the false branch, never re-evaluating `acc * (1/dn)`. Distinct edge case noted below (not fixed here, out of scope). |
| `forward/shaders/decode_reference.metal::decode_attention_reference` | **Found the same multiply-through-zero shape** (`inv_sum = (sum_exp > 0.0f) ? (1.0f/sum_exp) : 0.0f; ... acc += exp(...) * inv_sum * v_cache[...];`) and the same finite-sentinel (`-1e30f`) max-fold as the dead `attn_softmax`. Verdict: **out of scope, not fixed** — this kernel is `#[cfg(test)]`-only (`MSL_DECODE_REFERENCE` in `metal_qwen35.rs`'s test module), used solely as a frozen correctness oracle for a parity test against the live flash decode kernel, never dispatched in any production binary. Flagging for the maintainers in case the reference itself should eventually be hardened, but leaving it alone here since touching a frozen parity oracle changes what the parity test is actually proving. |
| `forward/gpu/inner/shaders.rs::attention_softmax` (WGSL) | Fixed in PR B (#790), not this PR. |
| `forward/shaders/gemm*.metal`, `gemm_gemm.rs` | Different operation family (GEMM argument validation, ADR-080 C4) — out of scope for the softmax cluster. |

**Adjacent observation (not a defect in the filed issues' scope):** the live
`decode_attention` family's fail-closed ternary is keyed off the softmax
denominator `dn`/`l_s[qi]`, which only goes non-finite when a NaN reaches the
*score* path (Q or K corruption). A hypothetical corruption confined to `V`
only (clean Q/K, poisoned V) would leave `dn` finite while `acc` accumulates
NaN via `score * v`, producing a NaN output through a *finite* denominator —
a different bug class (NaN propagation through a valid weight, not a
softmax-denominator fail-open) that neither #789 nor #790 describe and that
this PR does not attempt to fix. Noting it for the record in case a future
audit wants to scope it.

## Checks

- [x] `cargo clippy --workspace -- -D warnings`
- [x] `cargo clippy -p lattice-inference --all-targets --features f16,metal-gpu -- -D warnings`
- [x] `cargo build -p lattice-inference --features f16,metal-gpu`
- [x] `cargo test -p lattice-inference --features "f16,metal-gpu" --lib forward::metal::` (GPU lock,
      `LATTICE_METAL_TEST_ENFORCE=1`) — 3 passed (msl-assembly + NaN + +inf), 0 failed
- [x] Mutation sensitivity (this round, disposable worktree): revert finalize → RED (both new/updated
      tests, 160/160 non-finite) → restore → GREEN (0/160), byte-identical shader confirmed after restore
- [x] `make bench-compare` — see analysis above; not gate-blocking (structurally unreachable change,
      unchanged from round 1 since no kernel code changed this round)
- [x] `actionlint .github/workflows/ci.yml` — clean

## Round 2 update (head 3a305fa19)

Review round 2 returned APPROVE-WITH-FIXES with one medium, applied: the new CI gate no longer pipes its gating command through `tee` — the test writes its log via redirection and gates on its own exit status (explicit fail-cat-exit), per the repository's CI authoring rule. Source comments also drop review-round narration in favor of the invariant/issue references. `actionlint` and fmt clean; no kernel or test-behavior change (the 3-test enforced group is unchanged).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -245,13 +245,14 @@ jobs:
           LATTICE_METAL_TEST_ENFORCE: '1'
         run: cargo test -p lattice-inference --features f16,metal-gpu --lib q4 -- --nocapture
       # ADR-080 C1 / #789: the `forward::metal::` module's assembled-MSL check and the
-      # NaN/+inf fused_attention fail-closed regressions (codex round-1 major on PR
-      # #794) silently early-returned when no Metal device was present and were never
-      # actually run by any required workflow — the `q4` filter above does not match
-      # this module. LATTICE_METAL_TEST_ENFORCE=1 makes a missing device or failed
-      # pass construction a hard test failure instead of a skip, matching the Q4 and
+      # The NaN/+inf fused_attention fail-closed regressions (#789) silently
+      # early-returned when no Metal device was present and were never actually run
+      # by any required workflow — the `q4` filter above does not match this module.
+      # LATTICE_METAL_TEST_ENFORCE=1 makes a missing device or failed pass
+      # construction a hard test failure instead of a skip, matching the Q4 and
       # grammar steps' convention; the exact-count grep guards against a silently
-      # zero-matching filter the same way those steps do.
+      # zero-matching filter the same way those steps do. The gating command writes
+      # its log directly (no pipe) so its exit status gates on its own.
       - name: inference — metal-gpu fused_attention fail-closed gate (macOS only)
         if: runner.os == 'macOS'
         env:
@@ -259,7 +260,11 @@ jobs:
         run: |
           set -euo pipefail
           log="${RUNNER_TEMP:-/tmp}/metal-fused-attention-failclosed-test.log"
-          cargo test -p lattice-inference --features f16,metal-gpu --lib forward::metal:: -- --nocapture 2>&1 | tee "$log"
+          if ! cargo test -p lattice-inference --features f16,metal-gpu --lib forward::metal:: -- --nocapture >"$log" 2>&1; then
+            cat "$log"
+            exit 1
+          fi
+          cat "$log"
           grep -Eq '^test result: ok\. 3 passed; 0 failed; 0 ignored; 0 measured; [0-9]+ filtered out;' "$log"
       # #611: Metal generation paths applied grammar masks via `mask_logits`
       # but never checked for a surviving finite logit before sampling — a
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -244,6 +244,23 @@ jobs:
         env:
           LATTICE_METAL_TEST_ENFORCE: '1'
         run: cargo test -p lattice-inference --features f16,metal-gpu --lib q4 -- --nocapture
+      # ADR-080 C1 / #789: the `forward::metal::` module's assembled-MSL check and the
+      # NaN/+inf fused_attention fail-closed regressions (codex round-1 major on PR
+      # #794) silently early-returned when no Metal device was present and were never
+      # actually run by any required workflow — the `q4` filter above does not match
+      # this module. LATTICE_METAL_TEST_ENFORCE=1 makes a missing device or failed
+      # pass construction a hard test failure instead of a skip, matching the Q4 and
+      # grammar steps' convention; the exact-count grep guards against a silently
+      # zero-matching filter the same way those steps do.
+      - name: inference — metal-gpu fused_attention fail-closed gate (macOS only)
+        if: runner.os == 'macOS'
+        env:
+          LATTICE_METAL_TEST_ENFORCE: '1'
+        run: |
+          set -euo pipefail
+          log="${RUNNER_TEMP:-/tmp}/metal-fused-attention-failclosed-test.log"
+          cargo test -p lattice-inference --features f16,metal-gpu --lib forward::metal:: -- --nocapture 2>&1 | tee "$log"
+          grep -Eq '^test result: ok\. 3 passed; 0 failed; 0 ignored; 0 measured; [0-9]+ filtered out;' "$log"
       # #611: Metal generation paths applied grammar masks via `mask_logits`
       # but never checked for a surviving finite logit before sampling — a
       # grammar that blocked every token silently emitted token 0 instead of
--- a/crates/inference/src/forward/metal.rs
+++ b/crates/inference/src/forward/metal.rs
@@ -825,6 +825,28 @@ mod inner {
         use super::*;
         use crate::weights::{QwenLayerWeights, Tensor1D, Tensor2D};
 
+        /// ADR-066 D3.1 / ADR-080 C1 (codex round-1 major, PR #794): a capability-gated
+        /// Metal test must report a distinct skip or fail closed when
+        /// `LATTICE_METAL_TEST_ENFORCE=1` is set, never silently early-return `ok` —
+        /// otherwise a runner that loses its Metal device (or never had one) greens every
+        /// gate that depends on this module without ever executing the Metal path.
+        /// Mirrors the `forward::metal_qwen35` enforce convention already used by the Q4
+        /// and grammar fail-closed CI steps.
+        fn metal_test_device(context: &str) -> Option<Device> {
+            match Device::system_default() {
+                Some(device) => Some(device),
+                None => {
+                    let enforce = std::env::var_os("LATTICE_METAL_TEST_ENFORCE").is_some();
+                    assert!(
+                        !enforce,
+                        "LATTICE_METAL_TEST_ENFORCE=1 but no Metal device present ({context})"
+                    );
+                    eprintln!("[METAL_TEST_SKIP] context={context} reason=no_metal_device");
+                    None
+                }
+            }
+        }
+
         /// Guards the `flash_attention.metal` extraction: the file loaded via
         /// `include_str!` and run through `msl_source_for`'s token substitution must
         /// still produce a complete, compilable shader library. Mutation-sensitive —
@@ -839,8 +861,9 @@ mod inner {
                 "unsubstituted placeholder token remains in assembled MSL"
             );
             // Compile the assembled shader and confirm every pipeline function resolves.
-            // Skips silently if the host exposes no Metal device (e.g. non-GPU CI).
-            let Some(device) = Device::system_default() else {
+            // Fails closed under LATTICE_METAL_TEST_ENFORCE=1 if no Metal device is present
+            // (see `metal_test_device`); otherwise reports a distinct skip.
+            let Some(device) = metal_test_device("msl_template_assembles_and_compiles") else {
                 return;
             };
             let opts = CompileOptions::new();
@@ -874,22 +897,24 @@ mod inner {
             }
         }
 
-        /// ADR-080 C1 (#789) regression: the fused_attention finalize must fail
-        /// closed by ASSIGNMENT when the online-softmax denominator is
-        /// non-positive or non-finite (e.g. a NaN score from a corrupted Q
-        /// lane), not by multiplying the (possibly already-NaN) numerator
-        /// through a zeroed reciprocal -- `NaN * 0.0f == NaN` under IEEE-754,
-        /// so a `* inv_l` finalize cannot recover a poisoned row.
+        /// ADR-080 C1 (#789) regression fixture: corrupts a single entry of
+        /// `q_proj_weight` with `corrupt_value` so one attention head's Q vector
+        /// carries that value at every query position (K/V/embeddings/residual stay
+        /// completely clean), then runs the live `MetalForwardPass::forward` boundary
+        /// and returns its output. Shared by the NaN and +inf invalid-row regressions
+        /// below (ADR-066 D2 invariant #1's non-finite-row classes) so both exercise
+        /// the identical construction and the identical finalize fix.
         ///
-        /// Mutation-sensitive: reverting the fix in flash_attention.metal back
-        /// to `O4[out_base4] = o_frag * inv_l;` makes this test fail (160/160
-        /// NaN elements instead of 0), confirmed at PR time.
-        #[test]
-        fn fused_attention_fails_closed_on_nan_q_lane() {
+        /// Returns `None` only when there is no Metal device and
+        /// `LATTICE_METAL_TEST_ENFORCE` is unset (see `metal_test_device`); with the
+        /// enforce var set, a missing device or failed pass construction panics
+        /// instead of skipping.
+        fn run_fused_attention_with_corrupted_q_lane(
+            context: &str,
+            corrupt_value: f32,
+        ) -> Option<Vec<f32>> {
             let _lock = gpu_test_lock();
-            let Some(_device_probe) = Device::system_default() else {
-                return;
-            };
+            let _device_probe = metal_test_device(context)?;
 
             let config = QwenConfig {
                 vocab_size: 8,
@@ -922,7 +947,7 @@ mod inner {
             let corrupt_in_idx = 5usize;
 
             let mut q_proj_flat = test_random_vec(&mut rng, q_dim * hidden);
-            q_proj_flat[corrupt_out_idx * hidden + corrupt_in_idx] = f32::NAN;
+            q_proj_flat[corrupt_out_idx * hidden + corrupt_in_idx] = corrupt_value;
             let k_proj_flat = test_random_vec(&mut rng, kv_dim * hidden);
             let v_proj_flat = test_random_vec(&mut rng, kv_dim * hidden);
             let o_proj_flat = test_random_vec(&mut rng, hidden * q_dim);
@@ -1014,13 +1039,50 @@ mod inner {
                     .copy_from_slice(&embed_tokens_flat[tok * hidden..(tok + 1) * hidden]);
             }
 
-            let Ok(mut pass) = MetalForwardPass::new(&config, &weights, 8) else {
-                return; // no Metal device on this host
+            let pass = MetalForwardPass::new(&config, &weights, 8);
+            let mut pass = match pass {
+                Ok(pass) => pass,
+                Err(err) => {
+                    let enforce = std::env::var_os("LATTICE_METAL_TEST_ENFORCE").is_some();
+                    assert!(
+                        !enforce,
+                        "LATTICE_METAL_TEST_ENFORCE=1 but MetalForwardPass::new failed \
+                         ({context}): {err}"
+                    );
+                    eprintln!(
+                        "[METAL_TEST_SKIP] context={context} reason=forward_pass_construct_failed \
+                         error={err}"
+                    );
+                    return None;
+                }
             };
 
             let metal_out = pass
                 .forward(&hidden_input, seq_len)
                 .expect("Metal forward should not itself error");
+
+            Some(metal_out)
+        }
+
+        /// ADR-080 C1 (#789) regression, NaN class: the fused_attention finalize must
+        /// fail closed by ASSIGNMENT when the online-softmax denominator is
+        /// non-positive or non-finite (here: a NaN score from a corrupted Q lane), not
+        /// by multiplying the (possibly already-NaN) numerator through a zeroed
+        /// reciprocal -- `NaN * 0.0f == NaN` under IEEE-754, so a `* inv_l` finalize
+        /// cannot recover a poisoned row.
+        ///
+        /// Mutation-sensitive: reverting the fix in flash_attention.metal back to
+        /// `O4[out_base4] = o_frag * inv_l;` makes this test fail (160/160 NaN
+        /// elements instead of 0), confirmed at PR time (round 1) and reconfirmed
+        /// after the fix-round refactor (round 2).
+        #[test]
+        fn fused_attention_fails_closed_on_nan_q_lane() {
+            let Some(metal_out) = run_fused_attention_with_corrupted_q_lane(
+                "fused_attention_fails_closed_on_nan_q_lane",
+                f32::NAN,
+            ) else {
+                return;
+            };
 
             let nan_count = metal_out.iter().filter(|v| v.is_nan()).count();
             assert_eq!(
@@ -1032,6 +1094,56 @@ mod inner {
                 metal_out.len()
             );
         }
+
+        /// ADR-080 C1 / ADR-066 D2 invariant #1, +inf class: the same corrupted-Q-lane
+        /// construction as the NaN test above, but with the poisoned weight entry set
+        /// to `f32::INFINITY` instead of `f32::NAN`. Because only one weight entry is
+        /// corrupted and the rest of the accumulation over `hidden` dims is finite, the
+        /// resulting Q component for that head goes to +inf or -inf depending on the
+        /// sign of the corresponding hidden-input activation at each query position --
+        /// both are covered by the `!is_finite()` check below. This reaches the
+        /// finalize's non-finite-denominator branch through a distinct arithmetic path
+        /// from the NaN test (an infinite accumulator/denominator rather than a NaN
+        /// one), so it is not redundant with it.
+        ///
+        /// Mutation-sensitive by the same finalize revert as the NaN test above (same
+        /// shared shader guard).
+        #[test]
+        fn fused_attention_fails_closed_on_inf_q_lane() {
+            let Some(metal_out) = run_fused_attention_with_corrupted_q_lane(
+                "fused_attention_fails_closed_on_inf_q_lane",
+                f32::INFINITY,
+            ) else {
+                return;
+            };
+
+            let non_finite_count = metal_out.iter().filter(|v| !v.is_finite()).count();
+            assert_eq!(
+                non_finite_count,
+                0,
+                "fused_attention must fail closed (zero output) on an inf-poisoned \
+                 Q lane instead of propagating inf/NaN through the finalize multiply \
+                 (ADR-080 C1, #789; ADR-066 D2 invariant #1); got {non_finite_count}/{} \
+                 non-finite elements",
+                metal_out.len()
+            );
+        }
+
+        // ADR-066 D2 invariant #1, all-(-inf) class: NOT covered by a direct
+        // kernel-dispatch fixture here, by design, not oversight (codex round-1
+        // review, PR #794). `fused_attention` takes no external mask buffer -- the
+        // only way a score row goes to all -inf is every Q.K dot product in that row
+        // underflowing to -inf, and the live kernel always includes the causal
+        // self-key (Q_i . K_i), which stays finite for any finite Q/K pair. Forcing a
+        // genuine all-(-inf) row would require injecting -inf directly into the score
+        // buffer, i.e. bypassing the live entry point this suite is scoped to (the
+        // corrupted-weight construction above always produces a *finite* Q/K pair
+        // whose dot product can go to NaN or +-inf, never a row of literal -inf by
+        // construction). The finalize guard's arithmetic is the same for this class as
+        // for the +inf class above (`l_i` non-positive or non-finite triggers the same
+        // zero-assignment branch), so the +inf regression already exercises the shared
+        // fail-closed code path; there is no code path reachable from a live forward
+        // call that this suite could exercise differently for all-(-inf) specifically.
 
         struct TestLcg(u64);
 
--- a/crates/inference/src/forward/metal.rs
+++ b/crates/inference/src/forward/metal.rs
@@ -825,7 +825,7 @@ mod inner {
         use super::*;
         use crate::weights::{QwenLayerWeights, Tensor1D, Tensor2D};
 
-        /// ADR-066 D3.1 / ADR-080 C1 (codex round-1 major, PR #794): a capability-gated
+        /// ADR-066 D3.1 / ADR-080 C1 (PR #794): a capability-gated
         /// Metal test must report a distinct skip or fail closed when
         /// `LATTICE_METAL_TEST_ENFORCE=1` is set, never silently early-return `ok` —
         /// otherwise a runner that loses its Metal device (or never had one) greens every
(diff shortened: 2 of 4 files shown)


ohdearquant:

## Review of PR #794 (round 1)

### Verdict: REJECT

### Prior review context

- No prior review rounds or unresolved review threads were present.
- This round reviewed exact head `275a97341742d44168dc3269ea8f57c2b0ae5fe0` against main `65865ecb58b06c3538531e9b3b4b20d75532c66f`.
- The primary worktree already had unrelated uncommitted edits in `gpu/inner/shaders.rs` and `gpu/inner/tests.rs`; all source inspection, mutation, and gates used the clean disposable worktree `/tmp/lattice-pr794-r1.qMcf1m`.

### What this round examined

- Fail-closed behavior for NaN, `+inf`, and all-`-inf` attention rows, including the caller-visible error path.
- Every construction and dispatch site for this `fused_attention` Metal kernel, plus sibling Metal/WGPU attention implementations.
- Mutation sensitivity with a finalize-only mutation, forced shader rebuild, and restoration.
- Exact dead-kernel deletion sets in this PR and queued PR #796.
- Benchmark-target reachability, diff scope, local fmt/clippy/Metal gates, GitHub checks, and branch ancestry.

### Findings (# total: 1 blocker, 1 major)

#### Blockers

1. [`crates/inference/src/forward/shaders/flash_attention.metal:463`] Invalid rows are zeroed and returned as success, not rejected with an error.
   - **Current**: the bad-denominator branch writes `O4[out_base4] = float4(0.0f)` (`:474-478`). The regression test then calls `.forward(...).expect("Metal forward should not itself error")` and checks only that the final output contains no NaNs (`crates/inference/src/forward/metal.rs:1021-1032`). `MetalForwardPass::forward` unconditionally returns `Ok(output)` after readback (`metal.rs:572-587`); the kernel has no host-visible failure/status buffer.
   - **Issue**: this does not meet the review's explicit requirement that NaN, `+inf`, and all-`-inf` rows be rejected with the CPU-equivalent error class rather than clamped/zeroed. I observed the committed NaN fixture return `Ok`, and a disposable `f32::INFINITY` version of the same fixture also returned `Ok`. For an all-`-inf` score row, `m_new` remains `-inf`, `exp(-inf - -inf)` makes the denominator non-finite, and the same branch writes zeros; there is still no route to an error. Even a future `Err(String)` from `MetalForwardPass` would currently be swallowed by `QwenModel::forward`, which logs and silently falls back to CPU (`crates/inference/src/model/qwen.rs:1598-1604`).
   - **Contract conflict**: the PR accurately implements its body, issue #789, and the later ADR-080 C1 text, all of which define fail-closed attention as a zeroed row (`docs/adr/ADR-080-consolidation-duplicated-contracts.md:95-118`). That conflicts with both this review's acceptance criterion and ADR-066's standing L1 invariant, which says a softmax row must normalize or return `InvalidInput` (`docs/adr/ADR-066-output-correctness-gate-architecture.md:59-65`). This conflict must be resolved explicitly; silently choosing the zero-row side is not acceptable under the requested gate.
   - **Fix**: first make the governing contract unambiguous. Under the requested error contract, add a host-visible per-forward failure flag/status from the fused kernel, read it after command completion, and propagate the chosen typed CPU-equivalent error (the #611 precedent is `InferenceError::InvalidInput`) through the public Qwen path instead of falling back. Add separate NaN, `+inf`, and all-`-inf` tests that assert the error variant, not merely finite output.
   - **Test**: three hardware tests at the real forward boundary, one per invalid-row class, each asserting the exact error variant; mutation must make each return success or garbage and therefore fail.
   - **Confidence**: observed on local Metal hardware for NaN and `+inf`; derived from the shader arithmetic and caller path for all-`-inf` and error propagation.

#### Majors

1. [`crates/inference/src/forward/metal.rs:842`] The new regression/assembly tests are not fail-closed CI gates.
   - **Current**: `msl_template_assembles_and_compiles` silently returns when no Metal device exists (`:842-845`), and the new regression also returns when device/pass construction fails (`:889-892`, `:1017-1018`). Required CI builds and clippies the Metal surface (`.github/workflows/ci.yml:175-203`) but does not run `forward::metal::`; its enforced Metal test step runs only the unrelated `q4` filter (`ci.yml:234-246`). The informational Metal parity job runs `chat_metal`/Qwen3.5, not this Qwen3 embedding kernel.
   - **Issue**: all exact-head checks are green even though no required workflow executes the new mutation-sensitive test or shader-assembly test. A missing Metal device would also skip-pass both. This violates the repository's fail-closed CI rule and the existing #611/Q4 convention immediately adjacent in `ci.yml`, which uses `LATTICE_METAL_TEST_ENFORCE=1`, exact test-count checks, and path-proof markers.
   - **Fix**: add a macOS CI step that runs the two `forward::metal::` tests with `f16,metal-gpu`; make both tests honor `LATTICE_METAL_TEST_ENFORCE=1` and fail if device/pass creation is unavailable. Assert the exact test count or emit/check an explicit runtime marker so a zero-match or skipped path cannot pass.
   - **Test**: run that workflow with the test filter misspelled and with Metal capability forced unavailable; both mutations must make the gate red.
   - **Confidence**: derived from the checked-in workflows and confirmed by the exact-head GitHub check list.

- **Scope**: the diff changes only `forward/metal.rs` (tests) and `flash_attention.metal` (one live finalize guard plus deletion of three dead kernels). No other live-kernel behavior changed.
- **Invocation sweep**: this shader is assembled only by `forward/metal.rs`; `make_pipeline("fused_attention")` has one construction site (`metal.rs:206-221`) and `dispatch_fused_attention` has one production call (`metal.rs:465-468`). `metal_qwen35.rs` uses separately assembled `decode_attention`/flash-partial kernels. The same `attn_*` strings in `forward/gpu/inner` are distinct WGSL pipelines, not references to these deleted MSL functions.
- **Mutation sensitivity**: exact HEAD passed 1/1. Replacing only the finalize with the legacy `inv_l` multiply, touching the shader, and rebuilding made the test fail with exit 101 and exactly 160/160 NaNs. Restoring the guard and touching again returned GREEN 1/1. Every run acquired `/tmp/lion-metal-gpu-test.lock` through the test's machine-wide `flock` guard; builds ran through the heavy-lane wrapper.
- **Metal groups**: `forward::metal::` passed 2/2, including real MSL assembly/compilation. `test_decode_attention` passed 12 with 2 ignored-as-slow, exactly matching the PR body.
- **Dead-kernel set**: PR #794 deletes exactly `attn_scores`, `attn_softmax`, and `attn_context`. Full-workspace string hits outside the negative assertion are only the unrelated WGPU pipeline family; no MSL assembly, pipeline construction, dispatch, test invocation, `concat!`, or `include_str!` reference remains.
- **PR #796 adjacent set**: it deletes `gemm_rega1x8_r8`, `gemm_rega2x4_r16`, `gemm_rega2x4_r32`, `gemm_rega1x8_r8_bf16`, `gemm_rega2x4_r16_bf16`, and `gemm_rega2x4_r32_bf16`, plus helpers `bf16_to_float`, `silu_act`, and `bf16x4_to_float4`. The #794 and #796 sets are disjoint. PR #796 is currently `BEHIND` at `fc516f08`, consistent with the stated rebase sequencing.
- **Benchmark disclosure**: `rg` found zero `flash_attention` or `fused_attention` references in both `crates/inference/benches/` and `crates/embed/benches/`. The default comparison script runs only inference `elementwise_cpu_bench` and embed `simd`; neither enables or reaches the macOS+`metal-gpu` `include_str!` module, and the Rust changes are test-only. The 21 FAIL / 6 WARN / 114 improvement disclosure is therefore honest and sufficient for this Metal-shader-only diff; the bidirectional unrelated results are consistent with noise rather than a reachable regression.
- **Local gates**: `cargo fmt --all -- --check` passed; `cargo clippy --workspace --all-targets -- -D warnings` passed; `cargo clippy -p lattice-inference --all-targets --features f16,metal-gpu -- -D warnings` passed.
- **Repository state**: `origin/main` and the current remote `main` both resolve to `65865ecb58b06c3538531e9b3b4b20d75532c66f`, which is an ancestor of the exact head. The disposable worktree was restored byte-clean after mutation.
- **GitHub gates**: every reported check at exact head is successful, including `ci-gate`, three platform CI jobs, feature matrices, MSRV, e2e parity, Metal attention parity, Q4 gates, bench compile, semver, and app binaries. The PR remains draft/review-blocked, not CI-blocked.
- **External reference**: consulted Apple's official [Metal resources and Metal Shading Language specification entry](https://developer.apple.com/metal/resources/) as the authoritative shader-language reference; the decisive runtime claims above come from local hardware execution and repository code paths.

### What I did NOT check

- I did not rerun `make bench-compare`; I verified the body's reachability claim and judged its disclosed run under the repository's bench discipline.
- I did not run the two ignored 32,768-token decode parity tests.
- I did not add a direct all-`-inf` kernel-dispatch fixture. The live fused kernel has no external mask buffer and always includes the causal self-key for ordinary inputs; the all-`-inf` outcome above is derived from the explicit shader arithmetic. The missing caller-visible error is already observed for NaN and `+inf`.
- I did not run the full workspace test suite locally; exact-head required CI is green, and I ran the targeted Metal groups plus both requested clippy configurations.

- Another round is useful only after the maintainers resolve zero-row versus typed-error semantics, implement the resulting caller-visible behavior, and wire the regression into fail-closed CI. Recheck all three invalid-row classes and the CI anti-skip mutation; the dead-kernel, scope, benchmark, and ordinary gate work does not need repeating unless those areas change.

Domain utility: SKIPPED — no lore/domain composition tools were available; repository ADRs, issue history, source paths, local Metal execution, and the Lattice review protocol were sufficient.

Verdict: REJECT

ohdearquant:

## Review of PR #794 (round 2)

### Prior review context

- Round 1 rejected exact head `275a97341742d44168dc3269ea8f57c2b0ae5fe0` for one blocker (ADR-066's typed-error wording conflicted with ADR-080 C1's internal-attention zero-row contract) and one major (the new Metal tests could skip-pass and no required CI job ran them).
- The maintainer has since ruled that zero-row is canonical for internal attention softmax. This focused reconfirm reviewed remediation head `5800e42e30612f4595dade7f7f067ab571c55637` and did not repeat unchanged dead-kernel, benchmark, or broad invocation-path work.
- Round 2 finds both prior findings substantively remediated. One new CI-authoring standards issue remains, and the branch is currently behind `origin/main`.

### What this round examined

- The complete `275a9734..5800e42e3` remediation diff and current source for all three changed files.
- ADR-066 invariant #1 and its amendment against ADR-080 C1 and the public-boundary #611 class.
- Enforced device/pass failure behavior for all three `forward::metal::` tests.
- Zero-match, unavailable-device, and finalize-revert mutations in disposable worktree `/tmp/lattice-pr794-r2`.
- Local fmt, default/Metal clippy, actionlint, exact-head GitHub checks, and current ancestry against fetched `origin/main`.

### Findings (# total: 0 blockers, 0 majors, 1 medium)

#### Mediums

1. [`.github/workflows/ci.yml:262`] The new required gate pipes its gating command through `tee`, contrary to the repository's fail-closed CI authoring rule.
   - **Current**: `cargo test ... 2>&1 | tee "$log"` is protected by `set -euo pipefail`, so I confirmed this exact implementation does propagate cargo failure today.
   - **Issue**: AGENTS.md nevertheless says gating commands must never be piped through `tee`; they must write their report directly and let the command's own exit status gate. The policy deliberately avoids making correctness depend on shell pipeline state. This is a standards/durability defect, not a claim that the current `pipefail` form false-greened in my mutations.
   - **Fix**: redirect `cargo test` to `"$log"` directly and print the log afterward; if live failure output is needed, use an explicit `if ! cargo test ... >"$log" 2>&1; then cat "$log"; exit 1; fi`, followed by `cat "$log"` and the exact-count grep.
   - **Test**: retain both anti-skip mutations: a misspelled filter must fail the grep, and forced device-unavailable under enforcement must fail cargo before the grep.
   - **Confidence**: observed against the checked-in workflow and the repository's explicit CI authoring standard; the current command's functional failure propagation was also observed under both mutations.

- **ADR-066 amendment**: `docs/adr/ADR-066-output-correctness-gate-architecture.md:63-73,98-147` cleanly limits the original normalize-or-`InvalidInput` invariant to public/host-visible sampler boundaries, preserves the #611 class, and names ADR-080 C1's direct-assignment zero row as canonical for internal attention. It explicitly says NaN is not acceptable, states that zeroing can swallow upstream numerical defects, and assigns detection to L2 differential/parity and L3 quality gates. I found no wording that weakens the invariant beyond that scoped split.
- **Enforcement source**: `crates/inference/src/forward/metal.rs:835-848` hard-fails device absence whenever `LATTICE_METAL_TEST_ENFORCE` is set. `msl_template_assembles_and_compiles` uses that probe (`:855-898`); both forward tests use it and separately hard-fail `MetalForwardPass::new` construction under enforcement (`:912-1065`).
- **Required local Metal group**: with `LATTICE_METAL_TEST_ENFORCE=1`, `cargo test -p lattice-inference --features f16,metal-gpu --lib forward::metal:: -- --nocapture` reported exactly `3 passed; 0 failed; 0 ignored` at restored head, including real MSL compilation and both real forward calls.
- **Zero-match mutation**: misspelling the filter as `forward::metla::` produced `0 passed` and the workflow's exact-count grep exited 1. The gate cannot green on a vacuous filter.
- **Unavailable-device mutation**: replacing the shared probe with `None::<Device>` in the disposable worktree under enforcement made all three tests panic and produced `0 passed; 3 failed`, cargo exit 101. Assembly, NaN, and +inf tests all fail rather than skip-pass.
- **Invalid-row mutation sensitivity**: replacing only the finalize guard with the old reciprocal-times-numerator form made the NaN test fail with `160/160` NaNs and the +inf test fail with `160/160` non-finite values; cargo exited 101 with `0 passed; 2 failed`. Restoring the guard returned the complete module to `3 passed; 0 failed`. The disposable worktree is source-clean after restoration.
- **Forward-boundary coverage**: the two regressions call the live `MetalForwardPass::forward` boundary with isolated NaN and `+inf` Q-lane corruption and assert finite caller-visible output. Their shared finalize-revert mutation proves both are sensitive to the zero-assignment behavior rather than merely exercising setup.
- **All-`-inf` class**: `metal.rs:1132-1146` documents why it is not faked with an internal score buffer: the live fused entry point has no external mask input and includes the causal self-key; the same bad-denominator finalize branch is already exercised by the +inf case. This is sufficient and consistent with round 1's reachability note. The production path also applies fused Q/K normalization before attention, bounding finite Q/K inputs.
- **Scope**: `git diff --name-only 275a9734..5800e42e3` contains only `.github/workflows/ci.yml`, `crates/inference/src/forward/metal.rs`, and `docs/adr/ADR-066-output-correctness-gate-architecture.md`. `flash_attention.metal` has zero round-2 changes.
- **Local gates**: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo clippy -p lattice-inference --all-targets --features f16,metal-gpu -- -D warnings`, and `actionlint .github/workflows/ci.yml` all passed at the exact head. Heavy builds ran through the required heavy-lane wrapper; Metal tests used the in-repo `/tmp/lion-metal-gpu-test.lock` guard.
- **GitHub head identity and remediation gate**: GitHub reports the exact requested head `5800e42e30612f4595dade7f7f067ab571c55637`. `feature-matrix (macos-latest)` passed, and its job log shows the new enforced step actually ran `forward::metal::`, compiled the MSL, executed both forward regressions, and printed `test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 1815 filtered out`. All CI-workflow gates are green, including default macOS/Linux/ARM, both feature matrices, MSRV, bench compile, actionlint, and the `ci-gate` aggregator. E2E parity, informational Metal attention parity, Q4 PPL, wasm parity, and embed drift are also green.
- **Authoritative external reference**: consulted Apple's official [Metal resources and Metal Shading Language specification entry](https://developer.apple.com/metal/resources/); runtime conclusions here are based on local compilation/execution and repository code, not inference from the specification.

### Merge-gate state

- After fetching, `origin/main` is `86739db684332233d2e6ef3f5ca318f05d6ff12e`; it is not an ancestor of the PR head. GitHub reports `mergeStateStatus: BEHIND`, and `git rev-list --left-right --count HEAD...origin/main` reports `3 1`. The one new main commit does not touch the three round-2 remediation files, but the branch must still be updated and CI rerun at the resulting merge state under the repository's strict merge policy.
- Exact-head GitHub checks are not yet complete: the Q4 composed-golden leg remains in progress at the deadline. It is not claimed as passed; all 25 completed checks are green.

### What I did NOT check

- I did not repeat round 1's dead-kernel, sibling-kernel, benchmark-reachability, or ignored long-decode work because `flash_attention.metal` and those surfaces have zero round-2 changes.
- I did not fabricate an all-`-inf` score buffer fixture that bypasses the live forward entry point.
- I did not claim the still-pending exact-head GitHub jobs as green.

- No further semantic review round is useful. Apply the mechanical CI logging fix, update the branch onto current main, and require the rerun (especially `feature-matrix (macos-latest)`) to be green at the new head. Re-review only if that update changes the three remediation files or the Metal test count/behavior.

Domain utility: MEDIUM — khive's suggested GPU/Rust regression domains reinforced mutation-sensitive numeric testing, but repository ADRs, source, and observed Metal execution supplied the decisive evidence.

Verdict: APPROVE-WITH-FIXES

oceanwaves630 (approved):
ADR-066 amendment read against the ruling: invariant #1 scoping (public-boundary vs internal attention), ADR-080 C1 alignment, and the explicitly-owned detection tradeoff all check out. CI gate verified in the diff (direct redirect, exact-count guard, enforce flag). Approving ahead of green; auto-merge holds on checks.

PR #785, opened 2026-07-10
fix(inference): shared fail-closed softmax row finalizer (ADR-080 C1)

ohdearquant:
## Contract (ADR-080 C1)

A softmax row whose inputs contain NaN, whose max is `+inf`, or whose denominator is
non-positive/non-finite is **zeroed in full** (fail closed). No site may silently drop a
single lane, clamp, or emit a partially-normalized row.

This PR extracts the shared row-finalizer as an allocation-free `#[inline]` helper
(`crates/inference/src/attention/softmax_row.rs`: `row_max_and_any_nan`,
`row_fails_closed_pre_exp`, `is_masked_neg_inf`, `finalize_row`) and routes every
in-scope CPU production attention-row softmax site through it (GPU/shader kernels: see
the deferred-follow-up section below), or through an inline
contract-conformant fix where the helper's `&mut [f32]` shape doesn't fit the site's
control flow.

Part of #780.

## Site-by-site

| Site | Treatment | Notes |
|---|---|---|
| `attention/gqa.rs::apply_scaled_causal_softmax_fused` | Routed (refactor-only) | Real unclamped `.exp()` already fails closed via NaN-into-sum; swap is behavior-preserving — verified via existing bitexact/reference-parity tests, unchanged |
| `attention/decode.rs::softmax_decode_scores` | Routed (refactor-only) | Online/Welford max+`l`, exp pass now separated from `finalize_row(row, l)` |
| `attention/decode.rs::softmax_decode_neon` | Inline fix (structural) | Raw-pointer SIMD chunk loop, can't take a `&mut [f32]` helper without breaking vectorization; reimplements the same NaN-scan/`+inf`/non-positive-sum contract inline |
| `forward/cpu/softmax.rs::softmax_attention_scalar/neon/avx2` | Routed + fixed (#740) | Fixed: NaN-lane silently dropped by `fold(NEG_INFINITY, max)`; exact-zero for masked `-inf` lanes (previously round-tripped through `fast_exp`'s clamp into a tiny nonzero weight) |
| `model/qwen.rs` (2 inline duplicates in `forward_all_layers` + `forward_profiled`) | Deduplicated + routed (#739) | Extracted `causal_softmax_row` free fn, both call sites now call it — removes the independent formula copy |
| `model/qwen35/forward.rs::compute_attention_context` | Routed (fixed) | Extracted `decode_context_softmax_row`; bare `1.0 / sum_exp` had no guard |
[...]

PR #795, opened 2026-07-10
fix(inference): fail-closed WGPU attention_softmax row contract

### LINKED ISSUE #790: bug(inference): WGPU attention_softmax floor-clamps NaN denominator instead of failing closed

ohdearquant:
# PR B: WGPU attention_softmax fail-closed row contract (ADR-080 C1)

Fixes #790.

## What changed

`forward/gpu/inner/shaders.rs`'s `ATTENTION_SOFTMAX_SHADER` (WGSL) floor-clamped
its denominator (`inv_sum = 1.0 / max(sum_scratch[0u], 1e-20)`), converting a
NaN `sum_scratch[0u]` into a finite-looking reciprocal (WGSL's `max()` drops a
NaN operand) while the numerator lanes were already NaN — the same defect
class as the CPU `batch_prefill` floor-clamp fixed in #785/#739.

Ported the ADR-080 C1 row contract (`attention::softmax_row`) into the WGSL
kernel:

- **Pre-exp scan**: each thread scans its strided valid (`k <= row`) scores
  for a NaN/±infinite value via an explicit IEEE-754 exponent-bits test
  (`is_non_finite`, since WGSL's `max()` cannot be trusted to surface a NaN
  operand — the same reason `attention::softmax_row::row_max_and_any_nan`
  scans explicitly on the CPU side). A workgroup-wide OR-reduction combines
  the per-thread flags; if any is set, the kernel zeroes the whole row by
  assignment and returns *before* computing any `exp()`/sum, mirroring
  `row_fails_closed_pre_exp`.
- **Finalize**: the `1e-20` floor-clamp is removed. The denominator is now
  checked explicitly (`is_non_finite(sum_val) || sum_val <= 0.0`); on failure
  the row is zeroed by assignment, otherwise normalized by the true
  reciprocal — mirroring `finalize_row`.

## Semantics alignment (ADR-066 amendment, sibling #794)

Zero-row-by-assignment is the canonical fail-closed semantic for internal
attention softmax, per the maintainer ruling recorded in sibling PR #794
(which amends ADR-066 to state this explicitly for the CPU contract). This
PR's WGSL kernel already implements that same semantic shape (zero the row
by assignment on a fail-closed condition, never return a host error) and
needs no change here; this note exists only to record that the two PRs are
aligned on one canonical contract, not two.

## Round-1 fixes (codex review, 2 mediums, 0 blockers/majors)

**Medium 1 — WGSL portability of the fail-closed guard.** The pre-exp
non-finite check inspected `v = SCORES[row_base + k] * scale` — i.e. after
one WGSL floating-point arithmetic op had already consumed the raw score.
WGSL's [Finite Math
Assumption](https://www.w3.org/TR/WGSL/#finite-math-assumption) permits an
implementation to assume NaN/infinities are absent during shader execution,
so a backend other than the one this repository tests against is free to
replace a poisoned value with an indeterminate finite-looking one once it
has passed through that multiply — which would defeat the guard silently on
that backend.

Fix shape chosen (narrow the claim, no host-side pre-validation — a
per-forward scan of every model input/weight ahead of GPU dispatch is a perf
tax the contract does not need for a guard that is inherently
best-effort outside the tested backend):

- The non-finite check now runs on the raw `SCORES[row_base + k]` value
  itself, before the `* scale` multiply. `scale` is a host-supplied, already
  finite scalar, so this removes one WGSL arithmetic step from the poisoned
  value's path without adding any new per-element host-side work.
- Added an explicit guarantee-scope comment directly above
  `ATTENTION_SOFTMAX_SHADER` in `shaders.rs`: the zero-row contract is
  verified on this repository's native Metal-backed WGPU adapter only. It is
  **not** a WGSL-portable guarantee — under the Finite Math Assumption,
  another backend (Vulkan/DX12/browser WebGPU) may still legally launder a
  poisoned value into something finite-looking upstream of any check this
  kernel can perform. The CPU parity / differential gates (`e2e-parity.yml`,
  HF-vs-lattice greedy-token agreement) and the CPU-side
  `attention::softmax_row` contract remain the cross-backend correctness
  detection surface for any backend where this best-effort guard is
  defeated.

**Medium 2 — the regression test built a CPU oracle but never compared
against it.** The original test asserted `nan_gpu == 0` and `nan_ref == 0`
independently and never compared `gpu_out` to `cpu_ref` element-wise, so any
finite garbage value written into the poisoned row would have passed. It
also covered NaN only.

Fix: `run_attention_softmax_fail_closed_case` (`forward/gpu/inner/tests.rs`)
now:

- Compares `gpu_out` against `cpu_ref` element-wise at the same tolerance
  (`1e-3`) `tiny_model_matches_gpu_and_keeps_static_buffers` uses, in
  addition to the non-finite-count checks.
- Is parameterized over all three non-finite classes `is_non_finite` must
  catch: NaN, +infinity, and -infinity poisoned Q lanes (three `#[test]`
  functions sharing the helper).

## Regression tests

`attention_softmax_fails_closed_on_{nan,positive_infinity,negative_infinity}_q_lane`
(`forward/gpu/inner/tests.rs`), adapted from the filed repro
(`repro_wgpu_softmax_nan.rs`): same isolation technique as PR A — one
corrupted `q_proj_weight` entry makes exactly one head's Q vector
NaN/+inf/-inf at every position, K/V/embeddings/residual stay clean.
Compared element-wise against `cpu_forward_fail_closed_reference`, which
applies the same tiny model's math but runs its softmax step through the
crate's own `attention::softmax_row` helpers explicitly (not a naive
`f32::max` fold), so the reference is provably contract-compliant.

- Before the fix: GPU output all-non-finite (160/160), CPU fail-closed
  reference all-finite, for all three poison classes.
- After the fix: both all-finite and element-wise equal within tolerance,
  for all three poison classes.

**Mutation-sensitivity**, reverified in a disposable worktree (never over
uncommitted work) against this round's strengthened element-wise
comparison, in both directions:

1. Reverted `ATTENTION_SOFTMAX_SHADER` to the pre-fix
   `1.0 / max(sum_scratch[0u], 1e-20)` body (no pre-exp scan), `touch`ed
   `shaders.rs`, reran under the GPU lock — **RED**: all three cases failed
   with 160/160 non-finite elements leaked into GPU output.
2. Restored the fix, then mutated the pre-exp fail-closed branch's
   `SCORES[row_base + k] = 0.0;` to write `0.125` (a finite, nonzero
   constant) instead, `touch`ed, reran — **RED**: all three cases failed on
   the new element-wise comparison (`max abs diff = 0.046...` against the
   CPU oracle) while the non-finite-count checks alone would have passed
   (0 NaN/Inf elements) — this is exactly the gap medium #2 identified and
   this mutation proves the fix for it.
3. Restored the fix, reran — **GREEN**: all 4 tests pass
   (`tiny_model_matches_gpu_and_keeps_static_buffers` plus the three
   fail-closed cases).

## Reachability

`forward::gpu` (the `wgpu-gpu` feature) is not wired into any production
binary at `origin/main` — only `examples/bench_gpu.rs` and this crate's own
tests construct `GpuModelState`. Real, compiling, publicly-constructible
path; lower priority than PR A per the filed issue's own triage. Unchanged
by this round: both files touched (`shaders.rs`, `tests.rs`) are the same
two files as before.

## Bench-compare

`crates/inference/src/forward/` is touched; ran `make bench-compare` (default
groups: `elementwise_cpu_bench` + embed `simd`) per ADR-058/ADR-080's
bench-compare requirement even though this path is not production-reachable
today. BASE `origin/main` (65865ecb5) vs original HEAD (a5245c663):

```
❌ 7 FAIL (regression >7.0% confirmed by 95% CI)
⚠ 10 WARN (regression 3.0-7.0% confirmed)
🚀 15 confirmed improvement
```

Methodology-first disclosure, same reasoning as PR A's bench-compare section:

1. **Structurally unreachable.** This PR's diff touches exactly two files:
   `forward/gpu/inner/shaders.rs` (a WGSL shader source constant, only
   compiled into a `wgpu` pipeline and dispatched when the `wgpu-gpu`
   feature is enabled) and `forward/gpu/inner/tests.rs` (a `#[cfg(test)]`
   module, excluded from every `[[bench]]` binary). `lattice-inference`'s
   `default-features` is `["std", "download", "serve"]` — no `wgpu-gpu`.
   Grepping `crates/inference/benches/` for `attention_softmax` or
   `ATTENTION_SOFTMAX_SHADER` returns zero hits; the default-feature
   Criterion binaries this command runs do not link the changed code. This
   round's diff (a shader-comment/one-line-reorder change and a test-file
   strengthening) does not change that structural fact, so the analysis was
   not rerun — it holds a fortiori for a smaller, still-unreachable diff.
2. **Bidirectional noise signature.** The FAILs (`tier_prepared_batch_sizes`,
   `simd_normalize`, `int8_quantization`, `simd_batch_cosine_normalized_query`
   — all small SIMD/quantization CPU ops, none in the changed files) sit
   alongside 15 confirmed *improvements* in benchmarks of the same shape from
   the same run, matching the documented machine-contention pattern rather
   than a causal regression from this diff.

The correctness gate for this change is the mutation-sensitivity tests above
(RED on both the original bug and a finite-garbage mutation, GREEN on the
fix) plus the wgpu unit suite (`forward::gpu::` — 4 passed at this round's
head) — `make bench-compare`'s CPU-only default groups have no path to the
modified WGSL kernel and are reported here only to satisfy the disclosure
requirement, not because they can gate a `wgpu-gpu`-only change.

## Sibling-invocation-path sweep

See PR A's sweep table (same pass, both PRs land from one sibling-grep
session) — summary for this file specifically: `attention_softmax` was the
only WGSL kernel in `forward/gpu/inner/shaders.rs` implementing a softmax
row; the other kernels in that file (`gemm_bt`, RMS norm, copy/add, RoPE,
`attn_scores` WGSL variant, `attn_context` WGSL variant) do not perform a
softmax-style normalize and are unaffected by this defect class.

## Checks (re-verified at this round's head)

- [x] `cargo clippy -p lattice-inference --features "f16,wgpu-gpu" -- -D warnings` — same 8 pre-existing baseline failures in `gpu_gemm.rs`/`dims.rs`/`util.rs` under this feature flag, none in the two files this PR touches; baseline confirmed unchanged at this head
- [x] `cargo clippy --workspace -- -D warnings` (default features) — clean
- [x] `cargo fmt --check` — clean
- [x] `cargo test -p lattice-inference --features "f16,wgpu-gpu" --lib forward::gpu::` (GPU lock) — 4 passed (the pre-existing tiny-model test plus the three fail-closed cases)
- [x] `make bench-compare` — see analysis above; not gate-blocking (structurally unreachable change); not rerun this round since the structural-unreachability argument is unchanged and strengthened

## Round 2 update (head f15330c97)

Review round 2 returned APPROVE with one non-blocking wording suggestion, applied: the shader comment and this body now describe the CPU parity gates as the reference oracle for future backend-specific differential coverage rather than a detection surface for untested WGPU backends. No code change beyond that comment.

### DIFF
--- a/crates/inference/src/forward/gpu/inner/shaders.rs
+++ b/crates/inference/src/forward/gpu/inner/shaders.rs
@@ -265,6 +265,26 @@ fn attention_scores(
 }
 "#;
 
+// Guarantee scope (codex round-1 medium #1 on #795): the fail-closed
+// zero-row contract below has been verified on this repository's native
+// Metal-backed WGPU adapter (`GpuModelState::new`'s
+// `wgpu::PowerPreference::HighPerformance` request, backend `Backends::all()`
+// resolving to Metal on macOS CI/dev hosts). It is NOT a WGSL-portable
+// guarantee: WGSL's Finite Math Assumption
+// (https://www.w3.org/TR/WGSL/#finite-math-assumption) permits an
+// implementation to assume NaN and infinities are absent during shader
+// execution, so a different backend (Vulkan/DX12/browser WebGPU) may
+// legally replace a poisoned value with an indeterminate finite-looking one
+// at any point after it participates in WGSL floating-point arithmetic,
+// before `is_non_finite`'s bitcast ever inspects it. Reading the raw score
+// before the `* scale` multiply (below) narrows the window but cannot close
+// it for backends whose compiler chooses to optimize on the assumption
+// upstream of that read. The claim this kernel makes is therefore: "fails
+// closed on the tested native backend"; the CPU parity / differential
+// (`e2e-parity.yml`, HF-vs-lattice greedy-token) and CPU-side
+// `attention::softmax_row` gates remain the cross-backend correctness
+// detection surface for any backend where this best-effort guard is
+// silently defeated by the finite math assumption.
 pub(super) const ATTENTION_SOFTMAX_SHADER: &str = r#"
 @group(0) @binding(0) var<storage, read_write> SCORES: array<f32>;
 @group(0) @binding(1) var<storage, read> params: array<u32>;
@@ -312,10 +332,27 @@ fn attention_softmax(
     var local_bad: u32 = 0u;
     for (var k: u32 = lid.x; k < seq_len; k = k + 128u) {
         if (k <= row) {
-            let v = SCORES[row_base + k] * scale;
-            if (is_non_finite(v)) {
+            // Inspect the raw score BEFORE it participates in any WGSL
+            // floating-point arithmetic (codex round-1 medium #1 on #795):
+            // WGSL's Finite Math Assumption
+            // (https://www.w3.org/TR/WGSL/#finite-math-assumption) permits an
+            // implementation to assume NaN/infinities are absent during
+            // shader execution, so a runtime expression that *would*
+            // mathematically produce one is legally free to return an
+            // indeterminate value instead once it has passed through an
+            // arithmetic op. Reading `SCORES[row_base + k]` and testing it
+            // via `is_non_finite` prior to the `* scale` multiply removes
+            // that multiply from the poisoned value's path -- `scale` is a
+            // host-supplied, already-finite scalar (`pf(11u)`), so deferring
+            // the check past it bought nothing but one more arithmetic step
+            // for an implementation to legally launder the bit pattern
+            // through. This narrows, but does not eliminate, the WGSL
+            // portability gap: see the guarantee-scope comment below.
+            let raw = SCORES[row_base + k];
+            if (is_non_finite(raw)) {
                 local_bad = 1u;
             } else {
+                let v = raw * scale;
                 local_max = max(local_max, v);
             }
         }
--- a/crates/inference/src/forward/gpu/inner/shaders.rs
+++ b/crates/inference/src/forward/gpu/inner/shaders.rs
@@ -271,9 +271,21 @@ pub(super) const ATTENTION_SOFTMAX_SHADER: &str = r#"
 
 var<workgroup> max_scratch: array<f32, 128>;
 var<workgroup> sum_scratch: array<f32, 128>;
+var<workgroup> bad_scratch: array<u32, 128>;
 
 fn pf(i: u32) -> f32 {
     return bitcast<f32>(params[i]);
+}
+
+// ADR-080 C1 fail-closed row contract: a score is "bad" (NaN or +/-infinite)
+// exactly when its IEEE-754 exponent bits are all set (0x7f800000 mask) --
+// true for NaN (nonzero mantissa) and +/-Inf (zero mantissa) alike. WGSL's
+// `max()` silently drops a NaN operand (same `maxNum` semantics the CPU
+// contract's `row_max_and_any_nan` exists to counter), so the row max alone
+// cannot be trusted to surface a poisoned score; this scans explicitly.
+fn is_non_finite(v: f32) -> bool {
+    let bits = bitcast<u32>(v);
+    return (bits & 0x7f800000u) == 0x7f800000u;
 }
 
 @compute @workgroup_size(128, 1, 1)
@@ -291,19 +303,32 @@ fn attention_softmax(
     }
 
     let row_base = ((head * seq_len) + row) * seq_len;
+
+    // Pre-exp scan: row max over valid (k <= row) scaled scores, ignoring a
+    // NaN/+-inf score for the max itself, plus a separate fail-closed flag
+    // for any such score in the valid range (mirrors
+    // `attention::softmax_row::row_fails_closed_pre_exp`).
     var local_max = -3.40282347e+38;
+    var local_bad: u32 = 0u;
     for (var k: u32 = lid.x; k < seq_len; k = k + 128u) {
         if (k <= row) {
-            local_max = max(local_max, SCORES[row_base + k] * scale);
+            let v = SCORES[row_base + k] * scale;
+            if (is_non_finite(v)) {
+                local_bad = 1u;
+            } else {
+                local_max = max(local_max, v);
+            }
         }
     }
     max_scratch[lid.x] = local_max;
+    bad_scratch[lid.x] = local_bad;
     workgroupBarrier();
 
     var stride: u32 = 64u;
     loop {
         if (lid.x < stride) {
             max_scratch[lid.x] = max(max_scratch[lid.x], max_scratch[lid.x + stride]);
+            bad_scratch[lid.x] = bad_scratch[lid.x] | bad_scratch[lid.x + stride];
         }
         workgroupBarrier();
         if (stride == 1u) {
@@ -313,6 +338,20 @@ fn attention_softmax(
     }
 
     let max_val = max_scratch[0u];
+    let row_fails_closed_pre_exp = (bad_scratch[0u] != 0u);
+
+    if (row_fails_closed_pre_exp) {
+        // Fail-closed by ASSIGNMENT (attention::softmax_row::finalize_row's
+        // `row.fill(0.0)`): never compute exp()/sum on a row already known to
+        // carry a NaN or +/-infinite score. A later multiply-through-zero on
+        // a NaN numerator would not recover to zero under IEEE-754
+        // (`NaN * 0.0 == NaN`), so this returns before any such multiply.
+        for (var k: u32 = lid.x; k < seq_len; k = k + 128u) {
+            SCORES[row_base + k] = 0.0;
+        }
+        return;
+    }
+
     var local_sum: f32 = 0.0;
     for (var k: u32 = lid.x; k < seq_len; k = k + 128u) {
         if (k <= row) {
@@ -338,10 +377,24 @@ fn attention_softmax(
         stride = stride / 2u;
     }
 
-    let inv_sum = 1.0 / max(sum_scratch[0u], 1e-20);
-    for (var k: u32 = lid.x; k < seq_len; k = k + 128u) {
-        if (k <= row) {
-            SCORES[row_base + k] = SCORES[row_base + k] * inv_sum;
+    let sum_val = sum_scratch[0u];
+    // Fail-closed finalize by ASSIGNMENT (mirrors
+    // `attention::softmax_row::finalize_row`): a non-positive or non-finite
+    // denominator zeroes the row directly. The previous
+    // `1.0 / max(sum_val, 1e-20)` floor-clamp is removed -- it manufactured a
+    // finite-looking reciprocal for a NaN `sum_val` (WGSL `max()` drops the
+    // NaN operand) while the numerator lanes were already NaN, leaking NaN
+    // into the rest of the network instead of failing the row closed (#790).
+    if (is_non_finite(sum_val) || sum_val <= 0.0) {
+        for (var k: u32 = lid.x; k < seq_len; k = k + 128u) {
+            SCORES[row_base + k] = 0.0;
+        }
+    } else {
+        let inv_sum = 1.0 / sum_val;
+        for (var k: u32 = lid.x; k < seq_len; k = k + 128u) {
+            if (k <= row) {
+                SCORES[row_base + k] = SCORES[row_base + k] * inv_sum;
+            }
         }
     }
 }
--- a/crates/inference/src/forward/gpu/inner/shaders.rs
+++ b/crates/inference/src/forward/gpu/inner/shaders.rs
@@ -280,11 +280,12 @@ fn attention_scores(
 // before the `* scale` multiply (below) narrows the window but cannot close
 // it for backends whose compiler chooses to optimize on the assumption
 // upstream of that read. The claim this kernel makes is therefore: "fails
-// closed on the tested native backend"; the CPU parity / differential
+// closed on the tested native backend". The CPU parity / differential
 // (`e2e-parity.yml`, HF-vs-lattice greedy-token) and CPU-side
-// `attention::softmax_row` gates remain the cross-backend correctness
-// detection surface for any backend where this best-effort guard is
-// silently defeated by the finite math assumption.
+// `attention::softmax_row` gates define the reference semantics for future
+// backend-specific differential coverage; detecting a defeated guard on an
+// untested WGPU backend requires adding that backend to a differential run —
+// those gates do not execute this shader on Vulkan/DX12/browser WebGPU today.
 pub(super) const ATTENTION_SOFTMAX_SHADER: &str = r#"
 @group(0) @binding(0) var<storage, read_write> SCORES: array<f32>;
 @group(0) @binding(1) var<storage, read> params: array<u32>;
--- a/crates/inference/src/forward/gpu/inner/tests.rs
+++ b/crates/inference/src/forward/gpu/inner/tests.rs
@@ -47,23 +47,45 @@ fn tiny_model_matches_gpu_and_keeps_static_buffers() {
     assert!(max_abs < 1e-3, "max abs diff = {max_abs}");
 }
 
+/// Same tolerance `tiny_model_matches_gpu_and_keeps_static_buffers` uses to
+/// compare GPU output against its CPU reference.
+const FAIL_CLOSED_MAX_ABS_DIFF: f32 = 1e-3;
+
 /// ADR-080 C1 (#790) regression: `attention_softmax`'s WGSL kernel must fail
-/// closed by ASSIGNMENT on a NaN-poisoned score, not floor-clamp the
-/// denominator (`1.0 / max(sum_val, 1e-20)`) into a finite-looking reciprocal
-/// that then multiplies an already-NaN numerator through.
+/// closed by ASSIGNMENT on a non-finite (NaN or +/-infinite) score, not
+/// floor-clamp the denominator (`1.0 / max(sum_val, 1e-20)`) into a
+/// finite-looking reciprocal that then multiplies an already-poisoned
+/// numerator through.
 ///
 /// Isolation: corrupt a single entry of one layer's `q_proj_weight` so that
-/// exactly one attention head's Q vector is NaN at every query position,
-/// while K/V/embeddings/residual stay completely clean. Compare the GPU
-/// output against a CPU reference built from the crate's own fail-closed
-/// `softmax_row` helpers (not a plain `f32::max`-fold reimplementation).
+/// exactly one attention head's Q vector is poisoned at every query
+/// position, while K/V/embeddings/residual stay completely clean. Compares
+/// the GPU output element-wise against a CPU reference built from the
+/// crate's own fail-closed `softmax_row` helpers (not a plain `f32::max`-fold
+/// reimplementation) at the same tolerance the sibling test above uses.
 ///
-/// Mutation-sensitive: reverting `shaders.rs`'s `ATTENTION_SOFTMAX_SHADER` to
-/// the pre-fix `1.0 / max(sum_scratch[0u], 1e-20)` floor-clamp (with no
-/// pre-exp NaN scan) makes this test fail (NaN leaks into the GPU output
-/// while the CPU reference stays all-finite), confirmed at PR time.
-#[test]
-fn attention_softmax_fails_closed_on_nan_q_lane() {
+/// codex round-1 medium #2 on #795: the prior version of this test built the
+/// 253-line CPU oracle (`cpu_forward_fail_closed_reference`) but only
+/// compared NaN *counts* between GPU and CPU output, independently, never
+/// the two arrays against each other -- so an arbitrary finite garbage value
+/// in the poisoned row would have passed. `run_attention_softmax_fail_closed_case`
+/// below asserts full element-wise agreement (this is the fix) and covers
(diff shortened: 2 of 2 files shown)


ohdearquant:

## Review of PR #795 (round 1)

The Metal-backed WGPU run on this host is mutation-sensitive and green at the exact PR head, but the implementation does not establish a portable WGSL fail-closed guarantee and the regression test proves only “no NaN reached the final model output,” not the claimed zero-row/reference semantics.

### Prior review context

- No prior human or agent review rounds or unresolved review threads were present when this review began.
- This round focused on fail-closed WGPU softmax semantics, sibling invocation paths, mutation sensitivity, WGSL validation, benchmark disclosure, clippy baseline attribution, scope, and gates.

### What this round examined

- Read PR #795, issue #790, sibling PR #794, the complete two-file diff, both complete changed files, the CPU `softmax_row` contract, and the WGPU pipeline/dispatch chain.
- Traced every repository reference to `ATTENTION_SOFTMAX_SHADER`/`attention_softmax` and checked the benchmark and feature wiring.
- Reverted the shader guard in a disposable detached worktree, touched the shader, and ran the regression test under both the heavy-work and machine-wide GPU locks; restored the exact head and reran the cited WGPU suite.
- Ran formatting, default-feature workspace clippy, and feature-enabled clippy at both the PR head and a detached main checkout.
- Independently refreshed `origin/main`, checked ancestry/divergence, and inspected the exact-head GitHub check rollup.

### Findings (# total: 0 blockers, 0 majors, 2 mediums)

#### Mediums

1. [`crates/inference/src/forward/gpu/inner/shaders.rs:315`](crates/inference/src/forward/gpu/inner/shaders.rs#L315) The non-finite check occurs only after WGSL floating-point arithmetic has consumed the bad value.
   - **Current**: `let v = SCORES[row_base + k] * scale; if (is_non_finite(v)) { ... }`
   - **Issue**: This works on the adapter I exercised, but it is not a portable WGSL guarantee. WGSL's [Finite Math Assumption](https://www.w3.org/TR/WGSL/#finite-math-assumption) permits implementations to assume NaN and infinities are absent during shader execution; a runtime expression that yields one may produce an indeterminate value. The check bitcasts `v` only after the multiply, so another WGPU backend may legally replace the poisoned value with finite-looking indeterminate data before `is_non_finite` sees it. This is exactly the fail-open outcome the PR claims to exclude for the WGPU path.
   - **Fix**: Do not base the cross-backend contract on inspecting a floating result after it has participated in WGSL arithmetic. Either reject non-finite model inputs/weights before GPU dispatch with a `GpuForwardError`, or carry validity from a representation/path that never evaluates the poisoned lane as WGSL floating-point before the decision. If the intended guarantee is deliberately limited to tested native backends rather than WGSL-portable, narrow the PR/ADR claim explicitly and add backend-specific coverage.
   - **Test**: Run the NaN/+inf/all-`-inf` contract cases on at least the Metal and Vulkan/DX12 WGPU backends; a portable solution should produce the same zero-row/error outcome on each.
   - **Confidence**: Derived from the WGSL specification and the shader order at lines 315-316; observed only on this host's WGPU adapter.

2. [`crates/inference/src/forward/gpu/inner/tests.rs:103`](crates/inference/src/forward/gpu/inner/tests.rs#L103) The regression test constructs a CPU oracle but never compares GPU values to it, and covers only NaN.
   - **Current**: The test checks equal lengths and independently counts NaNs (`nan_gpu == 0`, `nan_ref == 0`). It never compares `gpu_out` with `cpu_ref`, despite lines 55-59 and the PR body claiming that it does. It also has no +inf or all-`-inf` case.
   - **Issue**: Any finite garbage emitted for the poisoned row passes. The test therefore proves mutation sensitivity for the old NaN leak, but not zero-by-assignment, CPU parity, +inf rejection, or fully non-finite-row rejection. The 253-line CPU reference is not serving as a numeric oracle.
   - **Fix**: Compare `gpu_out` and `cpu_ref` with the same documented tolerance used by `tiny_model_matches_gpu_and_keeps_static_buffers`, and add parameterized/direct shader cases for NaN, +inf, and an all-`-inf` valid row. A small direct dispatch/readback harness would prove row contents more precisely than relying only on the final model output.
   - **Test**: Mutate the bad-row branch to write an arbitrary finite nonzero value; the strengthened test must fail. Keep the already-demonstrated reverse-guard mutation as a second sensitivity check.
   - **Confidence**: Derived directly from assertions at lines 103-119; observed mutation testing confirms only the NaN case (`160/160` NaNs before, green after).

- **Canonical semantics**: The repository's CPU contract and sibling Metal PR do not return a host error; they fail closed by assigning the entire softmax/output row to zero. The WGSL branches at lines 343-352 and 388-391 match that semantic shape for NaN, +inf, and `-inf` on the adapter tested. `GpuModelState::forward` correctly remains `Ok` in this contract.
- **Sibling invocation paths**: There is one shader module/pipeline (`pipelines.rs:112,172-177`) and one live dispatch call (`state.rs:377-383`, implemented at `dispatch.rs:160-175`). Every live `GpuModelState` attention layer reaches this same guarded shader; no second WGPU softmax kernel or bypass was found.
- **Mutation sensitivity**: With `shaders.rs` restored to base and touched, one regression test executed and failed: `160/160` GPU outputs were NaN. With the exact head restored and touched, `cargo test -p lattice-inference --features 'f16,wgpu-gpu' --lib forward::gpu:: -- --nocapture` executed two tests and both passed. The red run proves a real adapter was present rather than the `NoAdapter` skip path.
- **WGSL assembly/compile on this host**: The restored test constructed `GpuModelState`, which creates and validates the WGSL module and compute pipeline, then dispatched it successfully. No separate repository WGSL parser/validator harness exists for this module.
- **Scope**: `65865ecb5...a5245c663` contains one conventional commit and modifies only `forward/gpu/inner/shaders.rs` and `forward/gpu/inner/tests.rs`.
- **Formatting**: `cargo fmt --all -- --check` passed.
- **Default-feature clippy**: `cargo clippy --workspace -- -D warnings` passed.
- **Feature clippy baseline**: `cargo clippy -p lattice-inference --features 'f16,wgpu-gpu' -- -D warnings` produced eight errors at both head and detached main `65865ecb5`: two in `gpu/inner/dims.rs`, one in `gpu/inner/util.rs`, and five in `gpu_gemm.rs`. The messages and locations were identical; neither changed file appears in them. The PR body's pre-existing-baseline claim is accurate.
- **GitHub state**: GitHub reports exact head `a5245c663caf6b2086c1015d60ab8e15b2cd4e0d`; all check-rollup entries and all required checks are successful. `mergeStateStatus=BLOCKED` is consistent with draft/review-required state, not failing CI.
- **Main freshness**: After refreshing `origin/main`, it remained `65865ecb58b06c3538531e9b3b4b20d75532c66f`; it is an ancestor of head and the divergence is `0 behind / 1 ahead`.
- **Benchmark disclosure**: `make bench-compare` runs only `elementwise_cpu_bench` and embed `simd`; `lattice-inference` defaults exclude `wgpu-gpu`, the shader has no benchmark reference, and `tests.rs` is test-only. The reported 7 FAIL / 10 WARN / 15 improvement result cannot be caused by this diff through those benchmark binaries. Reporting the noisy result with the structural-unreachability explanation is honest and sufficient for this change.

### What I did NOT check

- I did not rerun `make bench-compare`; I verified the command's target wiring and the diff's structural unreachability, not the reported raw counts.
- I did not execute Vulkan, DX12, or browser WebGPU backends; the runtime and mutation evidence is from this host's native WGPU adapter.
- I did not run the full workspace test suite locally; exact-head GitHub CI is green, and I ran the requested two-test WGPU suite plus the local lint/format gates.
- I did not observe a direct +inf or all-`-inf` GPU regression case because the PR does not provide one.

- One focused follow-up round is useful after the portability contract and oracle/edge-case tests are addressed. If the diff only strengthens those two areas and the red/green mutation remains intact, further recursive rounds should stop.

Domain utility: MEDIUM — the kernel-validation briefing reinforced that a finiteness-only sentinel cannot prove normalization semantics; the repository call graph and WGSL specification supplied the decisive evidence.

Verdict: APPROVE-WITH-FIXES

ohdearquant:

## Review of PR #795 (round 2)

### Verdict: APPROVE

Both round-1 mediums are fixed in code and are independently mutation-sensitive. The only remaining note is non-blocking wording precision: the CPU-only parity workflow is a semantic oracle, but it does not currently execute or detect failures on untested WGPU backends.

### Prior review context

- Round 1 returned `APPROVE-WITH-FIXES` with two medium findings: the raw poisoned WGSL value participated in float arithmetic before inspection, and the regression test neither compared its CPU oracle nor covered both infinities.
- The requested remediation is intentionally narrow: inspect the raw score before scaling, narrow the guarantee to the tested native Metal-backed WGPU adapter, and make the three non-finite cases mutation-sensitive against the CPU oracle.
- This round is a focused reconfirm. Per the round-1 stop rule, another review should not be requested if these two areas are repaired and the red/green mutations remain intact.

### What this round examined

- Read the complete round-1 artifact, remediation diff, current two changed files, PR body, review/thread state, relevant WGPU construction path, and CI parity wiring.
- Confirmed the committed shader loads `SCORES[row_base + k]`, bitcast-checks that raw value, and performs `raw * scale` only in the finite `else` branch.
- Ran the finite-garbage mutation in a detached disposable worktree under the heavy-work and Metal GPU locks: changing the pre-exp bad-row assignment from `0.0` to `0.125` made all three NaN/+inf/-inf cases fail at the new CPU-oracle comparison with max absolute difference `0.04625523` (tolerance `0.001`).
- Reverted the shader to the pre-fix guard/floor-clamp body, touched it, and rebuilt under the same locks: all three cases failed with `160/160` non-finite GPU outputs.
- Restored exactly `50ff01510`, touched and rebuilt: `forward::gpu::` executed four tests and all four passed.
- `cargo fmt --all -- --check` and default-feature `cargo clippy --workspace -- -D warnings` passed at exact head.
- Feature clippy at head reports exactly eight diagnostics (two `dims.rs`, one `util.rs`, five `gpu_gemm.rs`); refreshed `origin/main` reports the identical eight diagnostics at the same locations.
- Refreshed `origin/main` is `86739db68`; the PR head is `1 behind / 2 ahead`. GitHub reports `mergeStateStatus=BEHIND`, while `git merge-tree --write-tree` completed conflict-free.
- **Exact-head GitHub state at review cutoff (23:30Z)**: the full CI workflow is green, including macOS, Ubuntu x86/ARM, both feature matrices, MSRV, bench compile, and aggregate `ci-gate`; app binaries, Metal attention parity, and embed drift also passed. The separate e2e workflow still had its required parity comparison running and three auxiliary legs queued; none was red. All branch-protection-required checks reported by `gh pr checks --required` were green. GitHub's head OID was exactly `50ff01510fc127f69df67928db9fc5dda11071d9`.

### Findings

#### Minor / suggestion

1. [`crates/inference/src/forward/gpu/inner/shaders.rs:283`](crates/inference/src/forward/gpu/inner/shaders.rs#L283) Describe the CPU parity gate as the oracle, not as detection coverage for untested WGPU backends.
   - **Current**: `e2e-parity.yml` and CPU `attention::softmax_row` are called the “cross-backend correctness detection surface for any backend” where the WGSL guard is defeated.
   - **Issue**: `.github/workflows/e2e-parity.yml:168-170` says its required parity path is CPU-only, and its optional GPU leg exercises the separate native Metal backend. CI only compile-checks `wgpu-gpu`; it does not run this WGPU shader on Vulkan, DX12, or browser WebGPU. Those gates define the reference semantics, but cannot currently detect a WGPU-backend-specific laundering failure unless that backend is actually connected to a differential run. The PR body repeats the same wording. This does not undermine the now-honest primary scope claim—“verified on the tested native Metal-backed WGPU adapter; not WGSL-portable.”
   - **Suggestion**: Replace “detection surface for any backend” with “reference oracle for future backend-specific differential coverage,” or explicitly say that detecting an untested WGPU backend requires adding that backend to a differential run.
   - **Confidence**: Derived from the workflow wiring and the PR's own reachability section.

- **Round-1 medium #2, finite-garbage sensitivity**: observed red for all three cases. The new element-wise comparison detects finite-but-wrong output that the previous independent non-finite counts could not detect.
- **Original-bug sensitivity**: observed red for all three cases after restoring the pre-fix shader; each leaked `160/160` non-finite outputs.
- **Restored head**: observed green for the complete four-test `forward::gpu::` suite at exact commit `50ff01510`.
- **Formatting/default clippy**: both passed locally at exact head.
- **Feature-clippy attribution**: the same eight pre-existing diagnostics occur at head and refreshed main, all outside the two changed files.
- **Round-1 medium #1**: committed lines 351-356 load the raw score and bitcast-check it before the `* scale` multiply; that multiply is reachable only for a score that passed the check. This removes the exact arithmetic step round 1 flagged. The preceding `ATTENTION_SCORES_SHADER` still produces `SCORES` through floating-point arithmetic, so the new comment and PR body correctly disclaim WGSL portability and limit observed behavior to the tested Metal-backed WGPU adapter. This matches the [W3C WGSL Finite Math Assumption](https://www.w3.org/TR/WGSL/#finite-math-assumption), under which runtime NaN/inf expressions can become implementation-chosen indeterminate values.
- **Round-1 medium #2**: the shared runner covers NaN, +infinity, and -infinity, checks both outputs are finite, and compares every GPU/CPU output pair with the sibling test's `1e-3` tolerance.
- **Scope**: `a5245c66..50ff01510` changes only `shaders.rs` and `tests.rs`. The shader's only executable remediation change is ordering the raw-value inspection before scaling; the remaining shader additions are comments. Test changes parameterize the three poison classes and add the oracle comparison.
- **PR consistency**: the PR body carries the same adapter-limited/non-portable guarantee, describes the raw-before-scale reorder, documents all three poison classes and both mutation directions, and reports the same four-test restored-head suite.
- **Current-main state**: refreshed `origin/main` is `86739db68`; the reviewed head is `1 behind / 2 ahead`. The intervening main commit does not touch either PR file, and a synthetic merge completed without conflicts. GitHub reports `mergeStateStatus=BEHIND`, so the branch still needs the repository's normal freshness handling before merge.
- **GitHub CI**: the exact-head CI workflow and its aggregate `ci-gate` passed on all configured OS/architecture legs. The required e2e comparison remained in progress at the deadline; this is recorded as unobserved rather than credited as green.

### What I did NOT check

- I did not execute Vulkan, DX12, or browser WebGPU; runtime and mutation evidence is from this host's native Metal-backed WGPU adapter.
- I did not rerun `make bench-compare`; this round changes only shader comments/order plus tests, and the PR's cited default benchmark binaries do not link the `wgpu-gpu` path.
- I did not run the full workspace test suite locally; I ran the requested four-test WGPU suite plus local formatting and clippy gates. Exact-head GitHub CI passed, but the separate required e2e parity comparison and three auxiliary e2e legs had not completed by the review cutoff.

- Stop after this round. The remediation diff is narrow, both required mutations are red across all three poison classes, and restored head is green. A main refresh that introduces no substantive change to these two files does not need another recursive review.

Domain utility: MEDIUM — the numerical-failure briefing reinforced inspection at the first bad value and scalar-oracle comparison; repository code, mutation runs, and the WGSL specification supplied the decisive evidence.

Verdict: APPROVE

oceanwaves630 (approved):
Round-2 clean per review; oracle-comparison and guard-revert evidence verified. Approving ahead of green; auto-merge holds on checks.

Issue #784, opened 2026-07-10
refactor(inference): C4 — checked GEMM argument validator at every safe entry point (ADR-080 cluster C4 checklist)


ohdearquant:
## Context

ADR-080 (`docs/adr/ADR-080-consolidation-duplicated-contracts.md`, merged via #778) cluster **C4**: one checked GEMM argument validator. Sequenced last.

## Contract (per ADR-080)

The canonical validation contract from `forward/cpu/matmul.rs` — overflow-first (reject before compute), release-active (not debug_assert-gated), documented oversized-scratch-prefix allow-list — becomes a module-level validator in `crates/inference/src/forward/cpu/`, called from EVERY safe entry point before dispatch. Backend kernels stay separate; only pre-dispatch argument checking is shared. Reduction-order differences within f32 tolerance are intentional and out of scope.

## Checklist

- [ ] Shared checked validator in `forward/cpu/` implementing the canonical contract
- [ ] Scaled SGEMM (`forward/cpu/blas.rs` ×2, `attention/gdn_fused.rs`): validate BEFORE the macOS Accelerate FFI call — today the cfg-gated FFI copy crosses into unsafe code with undersized pointers while the scalar copy merely trips Rust bounds checks (held finding)
- [ ] Standalone WGPU + Metal wrapper functions (`metal_gemm.rs` etc.): validate slices before dispatch — currently no argument validation at all (held finding)
- [ ] BitNet ternary matvec: own `validate_ternary_matvec_args` (packed ternary layout ≠ row-major f32), release-active, shared between the safe wrapper and the unsafe NEON kernel; kills the release scalar path's silent truncation (held finding)
- [ ] Training SwiGLU forward: reject short activation slices (currently silent `zip` truncation) matching `matmul_bt`/materialized-GQA behavior (held finding)
- [ ] Unify oversized-input handling across f32/f16/Q8 GDN projections (held finding)
- [ ] Undispatched rectangular Metal GEMM kernel variants: delete or relabel test/benchmark-only (confirmed non-selected surface, ADR-080 duplication map)

## Completion gates

- [ ] Mutation-sensitive rejection tests per validator branch (undersized slice, overflow dims, oversized-scratch allowed) on every routed entry point
- [ ] `make bench-compare` — validator is pre-dispatch; GEMM group throughput must be unchanged
- [ ] Sibling grep: no safe GEMM-family entry point left calling unsafe kernels without the shared validator
- [ ] e2e-parity green

## Provenance

[...]

Issue #368, opened 2026-06-25
harden(inference): safe public attention entries reach unsafe SIMD matmul with debug-only weight-shape checks


ohdearquant:
> _Issue filed by Claude (Anthropic agent) on behalf of @ohdearquant._

## Summary

Found during the warm-start discovery review sweep that produced #367 (flash.rs overflow guards). Distinct from the overflow class — this is the **safe-pub-fn → unsafe-SIMD with debug-only precondition** class (same family as the merged #218 / #224 fixes), but on the **projection matmul** path and **spanning multiple attention kernels**, so it's tracked separately rather than bundled into the tight #367 overflow PR.

## The hole

`tiled_multi_head_attention_in_place` (`crates/inference/src/attention/flash.rs`) is a safe entry that checks the projection weight shapes with `debug_assert_eq!` only:

```rust
debug_assert_eq!(layer_weights.query_weight.rows, q_proj_dim);
debug_assert_eq!(layer_weights.query_weight.cols, hidden_size);
...
matmul_bt(hidden_states, layer_weights.query_weight.data, q, seq_len, hidden_size, q_proj_dim);
```

`TransformerLayerWeights` fields are **`pub`** (`crates/inference/src/weights/f32_weights.rs`), so a caller can directly construct one whose `query_weight.data` is shorter than `rows * cols` while the metadata says otherwise. On **non-macOS** AVX2/AVX512/NEON release builds, `matmul_bt` also only `debug_assert_eq!`s the slice lengths (`crates/inference/src/forward/cpu/matmul.rs`) before dispatching to unsafe SIMD kernels whose safety contract requires `b.len() >= n * k` — so a release build performs an **out-of-bounds SIMD load** (real UB), turning a safe public API call into memory unsafety.

Trigger (non-macOS SIMD release): `query_weight.data = &[]` with `rows = cols = 2`, `seq_len = 1`, `hidden_size = 2` → `matmul_bt(hidden_states, [], q, 1, 2, 2)` → `dot_product_avx2` dereferences `b.as_ptr().add(j*k)`.

macOS is unaffected on this specific trigger (the Accelerate `matmul_bt` wrapper has release-active length asserts), which is why lattice's primary target hasn't surfaced it.

## Scope (cross-kernel)

The same `debug_assert` + projection-matmul pattern appears in the sibling attention entries that take `TransformerLayerWeights`: `standard.rs`, `gated.rs`, `differential.rs` (and the materialized path). A complete fix should either:
[...]

PR #796, opened 2026-07-10
fix(inference): checked GEMM argument validator at every safe entry point (ADR-080 C4)

ohdearquant:
## Contract (ADR-080 cluster C4)

One checked GEMM-argument validator per shape family, extracted from the canonical contract
already established in `forward/cpu/matmul.rs` (#368):

1. **Overflow-first** — every shape-product (`m*k`, `n*k`/`k*n`, `m*n`) is checked with
   `checked_mul` before that product is used to index or size a buffer.
2. **Release-active** — real `assert!`, never `debug_assert!`.
3. **Oversized-scratch-prefix allow-list** — buffer-length checks are `>=`, never `==`/
   `assert_eq!`; callers may pass a reused scratch buffer longer than the exact footprint.

New module `crates/inference/src/forward/cpu/gemm_validate.rs`: `validate_gemm_nn`,
`validate_gemm_bt` (row-major `C = A@B` / `A@B^T`), `validate_gemm_strided_shape`
(leading-dimension variant for raw-pointer strided BLAS calls), `validate_ternary_matvec_args`
(BitNet packed-ternary layout — its own contract, not a GEMM reuse). Backend kernels (SIMD tile
variants, Accelerate/BLAS FFI, Metal/WGPU shaders) stay separate; only the pre-dispatch argument
check is shared.

Wired into every safe GEMM/GEMV/matvec entry point across CPU, standalone WGPU, and standalone
Metal wrapper functions, plus the previously-unvalidated BitNet ternary matvec, SwiGLU training
forward, and f16/Q8 GDN projections.

Refs #784, ADR-080 (#778).

## Per-site register

Legend: **ROUTED** = calls the new shared `gemm_validate` module. **COMPLIANT (pre-existing)** =
already had an equivalent overflow-first/release-active/oversized-ok validator before this PR,
left as-is. **EXEMPT** = justified with reasoning, not itself a raw-buffer safe entry point.

| Site | Verdict | Notes |
|---|---|---|
| `forward/cpu/matmul.rs::matmul_into` | ROUTED | `validate_gemm_nn` (was inline asserts pre-PR; now delegates) |
| `forward/cpu/matmul.rs::matmul_bt` | ROUTED | `validate_gemm_bt` (was inline asserts pre-PR; now delegates) |
| `forward/cpu/matmul.rs::matmul` | EXEMPT | thin wrapper allocating `c`, delegates to `matmul_into` which validates; keeps its own `m*n` overflow guard for the allocation itself |
| `forward/cpu/matmul.rs::matmul_scalar` / `matmul_bt_scalar` | EXEMPT | internal scalar fallback, only reached post-validation via `matmul_into`/`matmul_bt`; direct external callers are same-crate tests only; safe indexing (no raw pointer/unsafe), so a direct unvalidated call fails safe via an ordinary bounds-check panic, not UB |
| `forward/cpu/arch_kernels.rs::matmul_avx2` / `matmul_avx512` / `matmul_neon` | EXEMPT | `unsafe fn`, backend SIMD kernels per ADR-080 C4's own carve-out ("backend kernels stay separate"); only reachable through `matmul_bt`'s validated dispatch, confirmed via grep — no other call site |
| `forward/cpu/blas.rs::accelerate_matmul_bt` / `accelerate_matmul` | ROUTED | `validate_gemm_bt` / `validate_gemm_nn`, before the unsafe `cblas_sgemm` FFI call |
| `forward/cpu/blas.rs::sgemm_bt_strided` / `sgemm_nn_strided` (macOS + fallback, 4 fns total) | ROUTED | `validate_gemm_strided_shape` (shape + leading-dimension check; raw-pointer API has no slice length to check) |
| `forward/cpu/blas.rs::sgemm_nn_ab` (macOS FFI + fallback) | ROUTED | `validate_gemm_nn` — the primary held finding: the macOS FFI copy was `debug_assert!`-gated only, so a release build could cross into `cblas_sgemm` with undersized pointers |
| `forward/metal_gemm.rs::metal_matmul_bt` / `metal_matmul` | ROUTED | `validate_gemm_bt` / `validate_gemm_nn`, before the GPU-dispatch-threshold check; held finding — previously no validation at all |
| `forward/metal_gemm.rs` feature-disabled stubs | EXEMPT | return `false` immediately, touch no data |
| `forward/gpu_gemm.rs::gpu_matmul_bt` / `gpu_matmul` | ROUTED | same as metal_gemm.rs, wgpu path; held finding |
| `forward/gpu_gemm.rs` feature-disabled stubs | EXEMPT | return `false` immediately, touch no data |
| `forward/bitnet_kernel.rs::matvec_ternary_scalar` | ROUTED | `validate_ternary_matvec_args`; was silently truncating a short activation via `.take(k).skip(rem_start)` |
| `forward/bitnet_kernel.rs::matvec_ternary_neon` | ROUTED | same validator, shared with the scalar wrapper; was `debug_assert!`-gated only, reads via `get_unchecked` |
| `forward/bitnet_kernel.rs::matmul_ternary` | ROUTED | same validator, checked before `output` is allocated (uses `n` as the not-yet-allocated length) |
| `forward/neon.rs::matvec_q8_scalar` / `matvec_q8_neon_inner` (unsafe) | COMPLIANT (pre-existing) | both already call a private `validate_q8_args` (overflow-first via `checked_mul`, release-active `assert!`, `>=` oversized-ok) — a pre-existing analog of the new shared validator, scoped to the Q8_0 packed-block layout; left as-is |
| `forward/neon.rs::matmul_q8_neon` / `matmul_q8_neon_into` (safe wrappers) | EXEMPT (noted) | use `assert_eq!(x.len(), k)` on the activation before delegating into the validated inner functions — exact-match, stricter than the oversized-ok convention used elsewhere, but this validates a fresh caller-supplied `x`, not a reused scratch buffer, and is outside the #784 checklist (BitNet/SwiGLU/f32-f16-Q8 GDN *projections* — not the separate Q8_0 quantize-and-dispatch family). Left unchanged to avoid scope creep; flagged for a future Q8_0-family pass |
| `backward/tape.rs::swiglu_forward` (gate/up/down) | ROUTED | 3x `validate_gemm_bt` calls, modeling each projection stage as an `A(1,K) @ B(N,K)^T` matvec; was silent `.zip()` truncation |
| `attention/gdn_fused.rs::simd_matvec_transpose` | ROUTED | `validate_gemm_nn`; the unsafe AVX2/NEON kernels behind it read `s` via raw pointer with no per-element bounds check |
| `attention/gdn_fused.rs::simd_decay_and_rank1_update` | ROUTED | same, plus a standalone `delta.len() >= value_dim` check |
| `weights/f16_weights.rs::matmul_bt_f16` | ROUTED | `validate_gemm_bt`; was `assert_eq!` (exact-match), now unified with the oversized-ok convention |
| `weights/q8_weights.rs::matmul_bt_q8` | COMPLIANT (pre-existing, relaxed) | already had inline overflow-first + release-active checks from an earlier fix; this PR only relaxes the `a`/`c` length checks from `assert_eq!` to `>=`. `b_q.rows`/`b_q.cols`/`b_q.scales.len()` stay exact-match — those are `Q8Matrix`'s own declared shape invariants, not a raw caller-buffer length |
| `forward/shaders/flash_attention.metal`: `gemm_rega1x8_r8`, `gemm_rega2x4_r16`, `gemm_rega2x4_r32` + BF16 counterparts (`bf16x4_to_float4`, `gemm_rega1x8_r8_bf16`, `gemm_rega2x4_r16_bf16`, `gemm_rega2x4_r32_bf16`) and unused `bf16_to_float`/`silu_act` | DELETED | confirmed non-dispatched: every `kernel void` name in every `.metal` file cross-referenced against every `get_function()` call site (literal and closure-wrapped dynamic names) in the Rust dispatch code — zero matches. `git log` on the file shows only the original shader-extraction commit ever touched it, so these predate that extraction and were never wired to a live dispatch path |
| `vision/vit.rs::matvec` | ROUTED | `validate_gemm_nn` (modeled as `m=rows, k=cols, n=1`); safe CPU GEMV helper, ADR-080 C4 names `vision/vit.rs` explicitly (round-1 finding) |
| `vision/vit.rs::batch_matvec` | ROUTED | `validate_gemm_bt` (modeled as `m=n(batch), k=cols, n=rows`), plus an explicit `checked_mul` guard on the `n*rows` output allocation itself before validator entry — same rationale as `matmul()`'s own allocation guard (round-1 finding) |
| `forward/cpu/tiled.rs::matmul_bt_tiled` (+ `matmul_bt_tiled_scalar`/SIMD variants) | EXEMPT | sole caller is `matmul_bt` (matmul.rs:85), which already calls `validate_gemm_bt` before branching into this cache-blocked path (confirmed via grep — no other call site); same carve-out as the other backend SIMD kernels above (round-1 finding) |
| `forward/gpu/inner/dispatch.rs::dispatch_matmul` (+ sibling `dispatch_*` helpers in the same file) | EXEMPT | takes `u32` dims and `wgpu::Buffer`/`BindGroup` handles, not raw slices — no in-function slice indexing to guard; sole caller `gpu/inner/state.rs` computes and owns the dims from already-validated model-shape config, not untrusted caller-supplied lengths (round-1 finding) |
| `forward/metal.rs::dispatch_matmul` (private, line 594) | EXEMPT | same shape as the WGPU case — `u32` dims plus Metal `Buffer` handles, encodes GPU commands only, no slice bounds to check in this function; private to the module, called only from within `metal.rs`'s own forward-pass driver (round-1 finding) |
| `forward/metal_qwen35.rs::dispatch_matmul_half` / `dispatch_matmul_q8` + the rest of the `Dispatch helpers` section (~line 10821 on) | EXEMPT | same GPU-dispatch shape as the two `dispatch_matmul` cases above — `u32`/`Q4WeightBuf` params over Metal buffers, dims sourced from validated model config, no raw-slice indexing performed in these functions (round-1 finding) |

Method: reverse-apply (comment out) each guard, `touch` the source file to force a rebuild,
confirm the targeted rejection test goes red, restore, `touch` again, confirm green. Full raw
log kept locally (gitignored, not part of this diff). Summary:

- **17 call sites swept.** 11 showed a clean RED-to-GREEN flip on the first pass.
- **6 came back ambiguous on the first pass** (a libtest multi-token filter bug, and one
  off-by-one comment range that broke compilation instead of cleanly removing a guard) and were
  re-run individually with corrected filters/ranges:
  - `sgemm_bt_strided` / `sgemm_nn_strided` (macOS FFI): guard removal produces a hard process
    abort via the Accelerate FFI parameter-validation error (`BLAS error: Parameter number 11...`,
    exit status 255) rather than a standard libtest `FAILED` line — real rejection evidence,
    misclassified by the first log parser.
  - `tape.rs: swiglu_forward` (gate/down, corrected): confirmed real RED-to-GREEN. The
    `activation`-short sub-case is masked by a redundant identical check in the sibling `up`
    validator (both gate and up validate the same `x`); the `w_gate`/`w_down`-short sub-cases are
    genuinely load-bearing (guard removed produces an uglier native slice-range panic instead of
    the intended message).
  - `gdn_fused.rs: simd_matvec_transpose` / `simd_decay_and_rank1_update`: the pre-existing
    short-`k`/short-`delta` tests are masked by an incidental native bounds-check panic on `k[i]`/
    `delta[j]` regardless of the validator. Added new `short_s` tests targeting the actual
    raw-pointer-unsafe surface (`s` read via `row_ptr.add(j)`, no per-element bounds check in the
    AVX2/NEON kernels) — these show a genuine RED (guard removed, no panic at all, confirming a
    previously-silent OOB-adjacent read) followed by GREEN.
  - `bitnet_kernel.rs: matvec_ternary_neon` / `matmul_ternary`: corrected filter/range confirm
    the guards are present and compile correctly; both show RED masked by a redundant identical
    check in a sibling call on this platform (the NEON kernel's own validator, and
    `matmul_ternary`'s own check, both validating the same `x`/`x_q`) — genuine value confirmed
    by code inspection (the guard is not redundant on non-aarch64, and fires earlier with a
    clearer message on every platform).

None of the ambiguous cases indicate a missing or ineffective guard.

## Sibling grep

Searched `gemm|matmul|matvec|sgemm` plus FFI (`cblas`) and Metal/WGPU dispatch wrapper names
across `crates/inference/src`. No safe GEMM-family entry point left calling an unsafe kernel or
FFI function without a validator ahead of it, except the pre-existing `neon.rs` Q8_0 family
(already compliant via its own `validate_q8_args`, see register).

## Bench-compare

`make bench-compare`'s default groups (`elementwise_cpu_bench` for inference, `simd` for embed)
do not actually cover any GEMM/matmul benchmark — they're rms_norm/layer_norm/silu/gelu/softmax
and embed's cosine/dot-product/euclidean SIMD kernels, none of which this PR touches. That run
showed unrelated FAIL/WARN entries on benchmarks this PR never modifies (`elementwise_mul`,
embed's `simd`/`int8` groups); several concurrent unlocked bench/build jobs from other worktrees
were observed running on the machine at the same wall-clock time, consistent with the project's
documented bench-compare contention/false-positive pattern rather than a real regression.

The actual GEMM benchmark this PR's changes bear on is `inference_bench.rs`'s `matmul_bt` group
(exercises `matmul_bt`, one of the validated entry points, across 4 realistic shapes). Ran a
same-worktree source-swap A/B instead: built + benched at HEAD, reverted `crates/inference/src`
to base (`27b9047ee`) in place, rebuilt + benched, restored HEAD, then criterion-compared:

```
attn_proj_seq16_hidden384:                 time change [-42.89%, -18.64%, +13.16%]  p=0.39 (no significant change)
ffn_up_seq16_hidden384_intermediate1536:   time change [-19.58%,  -8.94%,  +2.86%]  p=0.18 (no significant change)
ffn_down_seq16_intermediate1536_hidden384: time change [ -3.07%,  +1.04%,  +5.43%]  p=0.64 (no significant change)
long_attn_seq128_hidden384:                time change [ -6.44%,  -0.04%,  +7.90%]  p=1.00 (no significant change)
```

All four cases: p > 0.05, confirming the pre-dispatch validator adds no measurable overhead to
the actual GEMM hot path (as expected — a handful of integer comparisons against an `m*k*n`-cost
GEMM).

## Test counts

- `cargo test -p lattice-inference --lib` (default features): 1652 passed, 0 failed, 18 ignored
  (was 1641 pre-round-1-fixes; +11 from the 4 new `blas.rs` i32-ABI tests, 6 new `vit.rs`
  routing tests, and 1 new `bitnet_kernel.rs` remainder-path discrimination test).
- `cargo test -p lattice-inference --lib --features metal-gpu,wgpu-gpu -- metal_matmul gpu_matmul`:
  6 passed, 0 failed (the Metal/WGPU wrapper validation tests).
- `cargo clippy -p lattice-inference --all-targets --features f16,metal-gpu -- -D warnings`
  (matches the CI `inference — metal-gpu clippy` job): clean.
- `cargo clippy -p lattice-inference --lib --target {x86_64,aarch64}-unknown-linux-gnu
  --no-default-features --features std -- -D warnings` (matches the two required Linux CI legs
  that failed at round-1 head): clean on `blas.rs`; the only remaining error on this reduced
  feature set is a pre-existing, unrelated `needless_return` in `download.rs` that only
  surfaces when the `download` feature is explicitly disabled (CI's actual default-feature run
  never disables it).
- `cargo fmt --all -- --check`: clean.

### Round-1 codex fixes (this push)

1. **Major** — `forward/cpu/blas.rs`: cfg-gated the `validate_gemm_bt` import to
   `#[cfg(target_os = "macos")]` (its sole call site, inside `accelerate_matmul_bt`); the
   non-macOS fallback path only uses `validate_gemm_nn`/`validate_gemm_strided_shape`, so the
   unconditional import tripped `-D warnings` unused-import on both required Linux CI legs.
2. **Medium** — `forward/cpu/blas.rs`: added a `cblas_dim()` helper (`i32::try_from`, panics on
   overflow) and routed every dimension/leading-dimension argument at all 5 Accelerate
   `cblas_sgemm` call sites through it, replacing the unchecked `as i32` narrowing casts. 4 new
   `#[should_panic]` tests exercise the `i32::MAX + 1` boundary directly.
3. **Medium** — register completeness: routed `vision/vit.rs::matvec`/`batch_matvec` through
   `validate_gemm_nn`/`validate_gemm_bt` (6 new tests); classified the remaining named sites
   (`forward/cpu/tiled.rs::matmul_bt_tiled`, `forward/gpu/inner/dispatch.rs::dispatch_matmul`,
   `forward/metal.rs::dispatch_matmul`, `forward/metal_qwen35.rs`'s dispatch-helper family) as
   EXEMPT with sole-caller/no-raw-slice reasoning in the register below.
4. **Medium** — `forward/bitnet_kernel.rs`: added
   `matvec_ternary_scalar_rejects_short_x_q_remainder_path` (k=5, x_q.len()=4) alongside the
   existing k=32 case, so the mutation-sensitivity evidence genuinely exercises the silent
   remainder-truncation path rather than an incidental full-byte-loop bounds panic. See
   `.khive/codex_reviews/pr_c4_mutation_log_round1_fixes.md` (gitignored, local) for the
   RED/GREEN transcript.
5. **Minor** — doc drift: `matvec_ternary_neon`'s safety doc now says validation happens "in
   all builds" (was "in debug builds"); `gemm_validate.rs`'s `validate_ternary_matvec_args` doc
   now states the `>=` oversized-allowed contract for `packed_w` (was "exact" footprint,
   contradicting the `>=` check at the call site and the existing
   `validate_ternary_matvec_args_accepts_oversized` test).


ohdearquant:

## Review of PR #796 (round 1)

### Prior review context

- No prior reviews, comments, or review threads were present at head `934b9084f8e59c60c3e36d897a420d9eda91a0d3`.
- This round examined ADR-080 C4 contract compliance, complete GEMM-family entry-point coverage, reachability of the two claimed silent-truncation bugs, shader deadness, mutation sensitivity, benchmark evidence, CI, and eight-commit scope.

### Findings (5 total: 0 blockers, 1 major, 3 mediums, 1 minor)

#### Major

1. [`crates/inference/src/forward/cpu/blas.rs:14`] The PR does not compile under the required Linux clippy legs.
   - **Current**: `use super::gemm_validate::{validate_gemm_bt, validate_gemm_nn, validate_gemm_strided_shape};`
   - **Issue**: `validate_gemm_bt` is used only by the `#[cfg(target_os = "macos")]` `accelerate_matmul_bt` implementation. At the current PR head, both `CI (ubuntu-latest)` and `CI (ubuntu-24.04-arm)` fail `./scripts/ci.sh` with `unused import: validate_gemm_bt`; `-D warnings` then stops compilation. This contradicts the PR body's clean workspace-clippy claim and fails a required quality gate.
   - **Fix**: cfg-gate that one import for macOS, while leaving the cross-platform validator imports unconditional, then rerun the two Linux CI legs.
   - **Test**: `cargo clippy --workspace -- -D warnings` on Linux x86_64 and aarch64.
   - **Confidence**: **observed** from current-head GitHub job logs for jobs `86420070425` and `86420070427`. Local macOS `cargo clippy --workspace -- -D warnings` passed, which explains why a macOS-only local check missed this.

#### Mediums

2. [`crates/inference/src/forward/cpu/blas.rs:92`] The new pre-FFI validation still accepts dimensions and strides that cannot be represented by the CBLAS ABI.
   - **Current**: every Accelerate call narrows `m`, `n`, `k`, `lda`, `ldb`, and `ldc` with unchecked `as i32` casts (also at lines 135-145, 199-209, 248-258, and 294-304), while `validate_gemm_strided_shape` checks only `usize` products and minimum strides.
   - **Issue**: non-overflowing `usize` inputs above `i32::MAX` pass the validator and are changed into negative CBLAS arguments. A safe example reaches this boundary through `matmul_into(&[], &[], &mut [], 0, 0, i32::MAX as usize + 1)`: all canonical products are zero, so slice checks pass, but Accelerate receives a negative `N`/`ldc`. The PR's own mutation log demonstrates that Accelerate terminates the process for invalid CBLAS parameters. The checked pre-FFI contract therefore remains incomplete for malformed dimensions.
   - **Fix**: before each Accelerate call, fail closed unless every dimension/leading dimension passed to CBLAS converts with `i32::try_from`; preferably centralize this in the strided/Accelerate-side validation helper. Add a zero-extent/over-`i32::MAX` test that proves the Rust assertion fires before FFI.
   - **Test**: `matmul_into`/`sgemm_*` with zero extents and one dimension or leading dimension equal to `i32::MAX as usize + 1` should panic with the validator's message, not enter Accelerate.
   - **Confidence**: **derived** from the public safe call path, the validator at `gemm_validate.rs:72-92`, and the CBLAS `i32` declaration at `blas.rs:24-39`.

3. [`crates/inference/src/vision/vit.rs:93`] The claimed per-site register is not a complete independent GEMM-family inventory.
   - **Current**: `vision::vit::{matvec,batch_matvec}` (lines 93 and 110) are safe CPU GEMV helpers with unchecked shape products and exact-length assertions, yet neither is routed nor classified in the register. Additional omitted boundaries include the safe tiled CPU kernel at `forward/cpu/tiled.rs:25`, WGPU dispatch at `forward/gpu/inner/dispatch.rs:11`, primary Metal dispatch at `forward/metal.rs:594`, and the Qwen35 GEMV/GEMM dispatch family beginning at `forward/metal_qwen35.rs:10821`.
   - **Issue**: several omitted functions are plausibly exempt backend/internal kernels, but the review contract explicitly requires every entry point to be registered as routed, compliant, or exempt with a reason. The ViT helpers are especially material: ADR-080's C4 evidence names `vision/vit.rs`, and these safe CPU operations do not implement overflow-first validation. The register's conclusion that every safe inference entry point is accounted for is therefore not established.
   - **Fix**: expand the register from the independent sweep. Route safe slice-based CPU boundaries that are in C4 scope, or explicitly exempt each omitted helper with a source-backed reason; group true backend dispatch kernels only where their shared ownership/validated caller is demonstrated.
   - **Test**: for any newly routed safe helper, add overflow/short/oversized cases; for exemptions, no code test is needed, but list the validated sole caller(s).
   - **Confidence**: **observed** with an independent definition/call-site sweep across `crates/inference/src`; **derived** as a completeness defect, not a claim that every omitted backend helper must call the shared slice validator.

4. [`crates/inference/src/forward/bitnet_kernel.rs:1088`] The committed BitNet regression test does not exercise the silent-truncation case it claims to cover.
   - **Current**: `matvec_ternary_scalar_rejects_short_x_q` uses `k = 32` and `x_q.len() = 31`. Without the validator, the full-byte loop indexes `x_q[31]` and panics; the test goes red only because that incidental bounds panic does not contain the expected validator message. The silent path is the remainder iterator at lines 253-261, which requires `k % 4 != 0` and a slice that ends within that remainder.
   - **Issue**: the implementation fix is correct and the real bug is reachable, but the checked-in test/mutation evidence does not preserve that regression class. A future rewrite that restores silent remainder truncation while retaining another panic path could leave this test green for the wrong reason.
   - **Fix**: use `k = 5`, `x_q.len() = 4` (and matching packed weights) for the scalar rejection test, or add that case alongside the aligned-dimension test.
   - **Test**: with the validator reverse-mutated out, the `k=5` test must report “did not panic”; with the guard restored, it must panic at the validator.
   - **Confidence**: **observed** in the disposable worktree. With the validator removed and the test changed to `k=5`, the function returned normally and the `#[should_panic]` test failed; this directly confirms both reachability and the proposed guard's correctness.

#### Minor

5. [`crates/inference/src/forward/bitnet_kernel.rs:284`] Public safety/contract docs still describe the pre-fix behavior.
   - The `pub unsafe fn matvec_ternary_neon` safety text says invariants are checked only “in debug builds,” although the new validator is release-active. Separately, `gemm_validate.rs:100-104` says `packed_w` is checked against the “exact” footprint and that oversized buffers are a mismatch, while line 123 uses `>=` and `validate_ternary_matvec_args_accepts_oversized` deliberately passes an oversized packed buffer.
   - **Fix**: update the NEON safety text to say the function validates in all builds, and make the packed-weight documentation match the chosen `>=` contract (or change code/tests if exactness is actually intended).
   - **Confidence**: **derived** from direct doc/code comparison.

- `validate_gemm_nn` and `validate_gemm_bt` check all three canonical products with `checked_mul` before evaluating those products for slice lengths; their assertions are release-active and use `>=`. Rust's official `usize::checked_mul` documentation confirms that overflow returns `None`: https://doc.rust-lang.org/stable/core/primitive.usize.html#method.checked_mul.
- `validate_ternary_matvec_args` checks `n * packed_row_bytes` before using it. The BitNet scalar bug is genuine: in a disposable worktree, removing the scalar validator with `k=5`/four activations made the function silently return. Restoring the guard makes the call reject deterministically.
- The SwiGLU bug is genuine and the three projection shapes are modeled correctly. Removing both gate/up validation calls made `swiglu_forward_rejects_short_activation` return normally; the head test passes with the guards present.
- The WGPU/Metal standalone wrappers validate before threshold/device discovery. Removing `gpu_matmul_bt`'s guard made its short-B test return normally; the head test passes.
- The two new GDN short-`s` cases target the raw-pointer surface rather than an incidental safe-index panic. Removing `simd_matvec_transpose`'s validator made its test return normally; the head test passes.
- The Q8_0 exemption in `forward/neon.rs` is substantively accurate: `validate_q8_args` uses checked products, release-active assertions, and lower-bound buffer checks. The allocating and allocation-free wrappers' `assert_eq!(x.len(), k)` is stricter but precedes delegation to that validator.
- The f16 and Q8 projection changes correctly allow oversized activation/output prefixes while retaining exact checks for `Q8Matrix`'s declared structural shape.
- The six `gemm_rega*` kernels and two deleted helpers were dead: their names have no Rust pipeline creation, dispatch, test, benchmark, or repository reference outside the deleted shader definitions. The bench harness that embeds separate shader source uses different kernel names and is unaffected.
- The targeted `matmul_bt` same-worktree A/B is adequate for ADR-080/ADR-058's present intent here. ADR-058 is superseded, the default groups do not cover GEMM, this PR changes validation rather than arithmetic kernels, and all four relevant shapes report no statistically significant change. I did not treat the broad contended default-group noise as evidence for or against this diff.
- PR #794 overlaps only `crates/inference/src/forward/shaders/flash_attention.metal`; its deletion is in the disclosed adjacent kernel range. I found no other file overlap. Rebase after #794 remains necessary but is not itself a finding.
- The eight commits are scoped to C4: shared validation, listed call-site wiring/tests, and dead shader removal. I found no unrelated dependency, API, or crate changes.
- Observed clean-head local results on macOS: `git diff --check`, `cargo fmt --all -- --check`, and `cargo clippy --workspace -- -D warnings` passed. Targeted head tests passed: 18 validator tests, BitNet short activation, GDN short `s`, SwiGLU short activation with `train-backward`, and WGPU short B with `wgpu-gpu`.

### What I did NOT check

- I did not rerun the reported Criterion A/B; benchmark assessment is based on the disclosed PR-body table and source/bench routing.
- I did not run GPU compute or Metal tests; the reviewed wrapper tests reject before device access, and GPU lock contention was avoided as requested.
- I did not rerun the full 1,641-test inference suite or all workspace tests. I ran the targeted tests above plus workspace clippy/fmt on the clean macOS head.
- At review completion, macOS CI, MSRV, bench-compile, e2e parity, and several parity gates were still queued. The two Linux required CI legs were definitively red for Finding 1.
- I did not execute the extreme Accelerate narrowing case in Finding 2 because the existing mutation log shows invalid CBLAS arguments terminate the whole test process; the path is established statically.

- Rerun round 1 after Findings 1-4 are addressed and the branch is rebased onto #794/current `main`. A deeper round is unnecessary if the expanded register is complete, Linux CI and all pending required gates are green, and the BitNet remainder mutation flips red-to-green.

Domain utility: MEDIUM — the composed Rust unsafe-boundary and reproducible-build guidance sharpened the raw-pointer mutation and target-specific CI checks, while repository-specific ADR/source inspection supplied most of the decisive evidence.

Verdict: APPROVE-WITH-FIXES

ohdearquant:
Round-1 codex fixes pushed (fc516f08d): all 5 findings addressed — see updated PR body (register + round-1-fixes section) and `.khive/codex_reviews/pr_c4_mutation_log_round1_fixes.md` for the k=5 remainder-path RED/GREEN mutation transcript (finding 4). Gates clean: fmt, workspace clippy, inference clippy w/ f16+metal-gpu, targeted + full lib tests (1652 passed, +11 new), and both Linux clippy legs re-verified via local cross-compile (x86_64/aarch64-unknown-linux-gnu, std feature) — blas.rs clean, only a pre-existing unrelated download.rs lint remains under a feature combo CI doesn't use.

ohdearquant:

## Review of PR #796 (round 2 — narrow reconfirm)

### Verdict: APPROVE

### Prior review context

- Round 1 at `934b9084f` found 1 major, 3 mediums, and 1 minor: Linux cfg-gating, CBLAS `i32` narrowing, incomplete ViT/register coverage, a non-discriminating BitNet regression test, and stale contract docs.
- This round reviewed the single remediation commit `fc516f08d2ef619c7ad06f022034af252eb38fae`, rechecked each prior finding, repeated the BitNet mutation experiment, and verified CI at that exact SHA.

### What this round examined

- The complete `934b9084f..fc516f08d` diff and current source for all four files changed by the fix commit.
- The PR-body register and round-1-fix disclosure, plus source reachability for one EXEMPT site.
- GitHub check runs and required checks attached to the exact new head.
- Local formatting, workspace clippy, the full default-feature inference library suite, and a disposable-worktree mutation/restoration cycle.

### Findings (0 total: 0 blockers, 0 majors, 0 mediums, 0 minors, 0 nits)

No unresolved finding remains from round 1, and this narrow reconfirm found no new defect in the remediation commit.

1. **Linux import/CI major is fixed.** `crates/inference/src/forward/cpu/blas.rs:17` cfg-gates only `validate_gemm_bt` to macOS, while the validators used by both platform variants remain unconditional at line 19. At exact head `fc516f08d`, both formerly failing checks are green: `CI (ubuntu-latest)` job `86430712370` and `CI (ubuntu-24.04-arm)` job `86430712355`.

2. **CBLAS narrowing medium is fixed.** `crates/inference/src/forward/cpu/blas.rs:66` centralizes `usize -> i32` conversion through `i32::try_from`. All five `cblas_sgemm` calls route all six integer arguments (`m`, `n`, `k`, `lda`, `ldb`, `ldc`) through `cblas_dim` at lines 111-125, 154-168, 218-232, 267-281, and 313-327. `rg -n 'as i32' crates/inference/src/forward/cpu/blas.rs` found only the explanatory rustdoc phrase at line 54; no cast expression survives. The four macOS `#[should_panic]` boundary tests are present at lines 560-621 and passed locally as part of the full library suite. Rust documents that `i32::try_from(usize)` errors when the input is out of range, and Apple's `cblas_sgemm` declaration confirms the integer dimension/leading-dimension ABI surface: https://doc.rust-lang.org/stable/core/primitive.usize.html and https://developer.apple.com/documentation/accelerate/1513264-cblas_sgemm.

3. **ViT routing/register medium is fixed.** `crates/inference/src/vision/vit.rs:93` routes `matvec` through `validate_gemm_nn`; `crates/inference/src/vision/vit.rs:112` guards `n * rows` with `checked_mul` before allocating and then routes `batch_matvec` through `validate_gemm_bt`. The six new tests cover short inputs, an oversized prefix, and output-allocation overflow and all passed. The PR-body register now classifies all four named omissions. As a source spot-check, `forward/cpu/tiled.rs::matmul_bt_tiled` has exactly one non-definition call site, `crates/inference/src/forward/cpu/matmul.rs:85`, after `validate_gemm_bt` at line 65; the EXEMPT sole-caller rationale is accurate.

4. **BitNet mutation-sensitivity medium is fixed.** `crates/inference/src/forward/bitnet_kernel.rs:1097` adds the requested `k=5`, `x_q.len()=4` remainder-path case. In a disposable detached worktree at `fc516f08d`, I removed only the scalar `validate_ternary_matvec_args` call and touched the source. The targeted test failed with `test did not panic as expected`; after restoring the exact guard and touching the source again, it passed and emitted `matvec_ternary_scalar: x_q too short for k` from `crates/inference/src/forward/cpu/gemm_validate.rs:120`. The worktree was then removed.

5. **Documentation minor is fixed.** `crates/inference/src/forward/bitnet_kernel.rs:287` now says the NEON invariant validation is release-active in all builds. `crates/inference/src/forward/cpu/gemm_validate.rs:100` documents `packed_w_len` under the same minimum-footprint `>=` contract implemented at line 125.

6. **Fix-commit scope is clean.** `fc516f08d` is one commit whose parent is the round-1 head `934b9084f`. Its tracked diff is limited to the four expected files: `bitnet_kernel.rs` (test + safety text), `cpu/blas.rs` (cfg import + checked conversions + four tests), `cpu/gemm_validate.rs` (contract wording), and `vision/vit.rs` (routing + six tests). The PR body separately adds the four EXEMPT register rows and the five-item round-1 remediation disclosure. I found no dependency, public-API, kernel-arithmetic, shader, or unrelated cleanup change in this commit. `git diff --check 934b9084f..fc516f08d` passed.

7. **Current-head CI is fully green.** GitHub's check-runs API reports **27/27 completed successfully**, each attached to `fc516f08d`; `gh pr checks --required` reports **6/6 required checks passed**. The 27 successful runs are:

   - CI: `CI (macos-latest)`, `CI (ubuntu-latest)`, `CI (ubuntu-24.04-arm)`, `actionlint`, `bench-compile`, `cargo-deny`, `ci-gate`, `detect-code-changes`, both `feature-matrix` legs, `msrv (1.93, f16+metal-gpu)`, `rustdoc-lint (informational)`, `semver-checks`, `typos`, and `unwrap-in-lib-lint (informational)`.
   - E2E/parity: `detect-engine-changes`, `e2e parity`, `embed drift gate`, `metal attention parity (informational)`, `parity-gate`, `ppl-gate selftest`, `q4 composed golden`, `q4 ppl regression gate`, and `wasm embedding parity gate`.
   - App binaries: `detect-app-bin-changes`, `build app-shipped binaries`, and `app-binaries-gate`.

8. **Observed local gates at exact head passed.** `cargo fmt --all -- --check`; `cargo clippy --workspace -- -D warnings`; and `cargo test -p lattice-inference --lib` (`1,652 passed, 0 failed, 18 ignored`).

### What I did NOT check

- I did not rerun Criterion or `make bench-compare`; this fix commit changes validation, tests, and docs rather than GEMM arithmetic, and this was a narrow remediation reconfirm.
- I did not rerun local GPU/Metal, feature-matrix, MSRV, semver, app-binary, or parity workloads; I verified their exact-head GitHub check runs instead.
- The PR is still a draft and GitHub reports `mergeStateStatus=BEHIND` relative to current `main`. Per the repository merge gate, it must be refreshed and CI must rerun green at the resulting head before merge. This approval applies to `fc516f08d`; it is not evidence for a future rebased SHA.

- Stop. No additional review round is useful unless refreshing from `main` creates conflicts or substantive code changes; otherwise rerun the required checks on the refreshed head and merge only when up to date.

Domain utility: MEDIUM — the composed Rust unsafe-boundary and reproducible-build guidance reinforced the focused boundary-test design and exact-SHA CI verification; repository source and mutation evidence remained decisive.

Verdict: APPROVE

oceanwaves630 (approved):
Read the full diff at 727258765 (post-rebase head). The shared gemm_validate module implements the canonical contract (overflow-first checked_mul, release-active assert!, oversized-prefix >=) consistently across all touched surfaces — matmul, blas/Accelerate FFI, strided wrappers on BOTH platform copies, wgpu, Metal, bitnet ternary, swiglu tape, gdn_fused SIMD, vit GEMV. Spot-traced the operand mappings that are easiest to get wrong and they check out: vit matvec as m=rows/k=cols/n=1, batch_matvec with the output-allocation overflow asserted before the vec is built, matmul_ternary passing n for the not-yet-allocated output. The cblas_dim i32 guard genuinely closes the hole the usize-level validators cannot see (zero-extent shapes with a dimension above i32::MAX pass every product check), and the remainder-path test at k=5/len=4 is correctly designed so only the validator makes it red. Verified independently at this head that the deleted gemm_rega*/bf16 shader kernels have zero remaining references anywhere in the tree. Four heavy engine gates (e2e parity, q4 golden/ppl, wasm parity) were still pending at review time — merge should ride them to green as usual. Approved.

PR #798, opened 2026-07-10
fix(tune): fail loudly on unimplemented GPU optimizer dispatch instead of silent no-op (#797)

### LINKED ISSUE #797: bug(tune): GPU trainer Adam/AdamW/SGD-momentum optimizer step submits an empty command buffer and reports success (FP-165)

## Summary

`GpuOptimizer::dispatch_optimizer_update` (crates/tune/src/train/gpu/optimizers.rs:121-146) never encodes a compute pass. It creates the uniform buffer, fetches the pipeline's bind-group layout, then submits an **empty** command encoder and returns `Ok(())`:

```rust
// TODO(FP-165): submits no actual compute — needs access to weight buffers from GpuNetwork
let encoder = device.create_command_encoder(&wgpu::CommandEncoderDescriptor::default());
queue.submit(std::iter::once(encoder.finish()));
Ok(())
```

All three shader-based optimizers route through it (`update`, :156-161): **Adam, AdamW, SGD-momentum**. After the no-op "update", the timestep still increments (:167-169) and the trainer proceeds to weight sync — so GPU training with any of these optimizers silently performs zero weight updates while reporting success at every step.

The path is live, not dead code: `GpuTrainer::update_weights` (crates/tune/src/train/gpu/mod.rs) calls `GpuOptimizer::update` in the training loop.

## Adjacent silent behavior (same match arm)

`Optimizer::RMSprop` silently falls back to plain SGD (:163, `// Fallback`) with no warning — a user selecting RMSprop gets a different algorithm without any indication. Should fail loudly or log prominently as part of the same fix.

## Expected fix (fail-loud first)

Per the tech-debt design review: until the weight/gradient buffers are actually wired into the bind group, these three optimizer arms must return a `TuneError::Training` error ("GPU {optimizer} update not implemented: optimizer shader dispatch has no buffer bindings") instead of succeeding as a no-op. Wiring the real dispatch is the follow-up; the immediate defect is success-without-update.

- The fail-loud fix must be mutation-sensitive: a test that selects Adam on the GPU trainer and asserts the error fires (and fails if the empty-submit `Ok(())` is restored).
- Do not remove the `FP-165` marker without closing the behavior; replace it with a reference to this issue.

## Classification

- Live correctness bug in published `lattice-tune` (crates.io), not stale debt.
- Severity: high — silent training no-op, plausibly interpreted as "model converged" or "training is just slow".

Found by the 2026-07-10 tech-debt audit (dead-code cluster review); verified at source before filing.

ohdearquant:

## What

Closes #797. `GpuOptimizer::dispatch_optimizer_update` submitted an empty
command encoder and returned `Ok(())` for the Adam, AdamW, and SGD-momentum
GPU optimizer arms — training with any of these silently performed zero
weight updates every step while reporting success. `RMSprop` silently
substituted plain SGD instead of running the requested algorithm.

**Every arm of `GpuOptimizer::update` now fails loudly** (`Adam`, `AdamW`,
`SGDMomentum`, `SGD`, `RMSprop`) instead of reporting success for a step that
changed nothing. See the "Scope grew during review" section below for why
plain SGD is included — it was not part of the original narrow ask, but
verifying it (per the task brief's "confirm it actually works") turned up a
worse defect than the write-back gap I initially found, and the coordinator
directed extending the fix to it (documented as a comment on #797).

- `GpuOptimizer::update_adam` / `update_adamw` / `update_sgd_momentum` now
  return `TuneError::Training(...)` naming the optimizer and stating the GPU
  shader dispatch has no buffer bindings wired to the network's
  weight/gradient buffers (#797), replacing the `TODO(FP-165)` marker and the
  dead uniform-buffer / bind-group-layout scaffolding that never got used for
  anything (no bind group was ever built or bound).
- `GpuOptimizer::update_sgd` (plain SGD, CPU-side) now returns
  `TuneError::Training(...)` too. See below for why.
- `Optimizer::RMSprop` now returns an explicit error naming the defect
  (previous silent substitution of SGD) instead of running SGD under the
  hood — and no longer points users at SGD as a working alternative, since
  SGD fails loudly too now.

## Why fail-loud-first, not the full fix

Per the issue: wiring the real weight/gradient buffer bindings into the bind
group (Adam/AdamW/SGDMomentum) and real gradient plumbing + mutable weight
access (plain SGD) is larger follow-up work. The immediate defect — success
reported for a step that updated nothing — is high severity (plausibly
misread as "model converged" or "training is slow") and is fixable narrowly
today by making every arm honest about its actual state: unimplemented.

## Scope grew during review: plain SGD included too

My first pass left `update_sgd` untouched, on the read that fixing it needed
a new mutable-access API on `GpuNetwork` (real feature work, not a
fail-loud-only change), and flagged the finding instead of silently treating
it as "verified working" per the task brief.

The coordinator verified `update_sgd` at source and found it's worse than "no
write-back": the loop uses `let grad_scale = 0.01; // Placeholder gradient
magnitude` — a constant, not the actual per-layer gradients accumulated in
`LayerGradients`. So even with a mutable write-back path, the arm would
compute `w -= lr * 0.01`, which is not SGD by any definition — it's
unconditional weight decay unrelated to the loss. That reframes the
question: fixing plain SGD (real gradients + mutable write-back) is
follow-up work, but *making it stop reporting success* is exactly the same
narrow, low-risk change as the other three arms. Directive: extend the
fail-loud fix to `update_sgd`, documented as a comment on #797.

`update_sgd` now:
- Deletes the placeholder loop body (constant-gradient theater).
- Returns `TuneError::Training("GPU SGD optimizer update not implemented: no
  real gradient plumbing or mutable weight write-back from GpuNetwork is
  wired (#797)")`.
- Carries a doc comment stating what real SGD needs: real gradient plumbing
  from `LayerGradients` **and** a mutable weight-write path on `GpuNetwork`
  (today only `cpu_network()`, an immutable accessor, exists).

Net effect: the GPU trainer's `update_weights` step fails for every optimizer
choice until real dispatch/gradient/write-back wiring lands. This makes the
GPU trainer honestly unusable for weight updates right now, which is its
actual state — previously that was hidden behind a false "success."

## Codex round 1 (APPROVE-WITH-FIXES) — applied

Full review: see PR comment / `pr_review_r1.md`. Three findings, all fixed:

1. **Major** — `GpuTrainer::train_batch` incremented `self.global_step` at
   the top of the function, before forward/backward/`update_weights()?`. So
   every fail-loud optimizer error still advanced the public
   `global_step()` counter: failed steps counted as completed training
   steps and fed into the LR/epoch-approximation math. Fixed by moving the
   increment to immediately after `update_weights()?` succeeds (still
   before the LR-schedule calculation that reads it). `global_step()` is
   now `0` after any failed `train_batch` call, for any optimizer.
2. **Medium** — public docs still promised working GPU training:
   `train/mod.rs`'s module rustdoc (with a `train_batch` example),
   `train/gpu/mod.rs`'s `GpuTrainer` doc, and `crates/tune/README.md`'s
   feature table. Added a `# Current limitation` note to each (module doc,
   struct doc, README footnote) stating forward pass and loss computation
   work, but weight-update training returns `TuneError::Training` for every
   optimizer until #797's wiring lands; each note references the issue and
   states it will be removed once that lands. Validated with
   `RUSTDOCFLAGS="-D warnings" cargo doc -p lattice-tune --no-deps` (default
   features) and again with `--features gpu` — both clean, no broken links.
3. **Minor** — the PR body originally attributed the pre-existing
   `test_is_using_gpu` failure to a speculative "paravirtual/Metal-adapter
   quirk." That was wrong and unverified. Corrected above: the test's
   6→8→6 network (110 parameters) is under the 10,000-element
   `should_use_gpu` threshold, so `is_using_gpu()` is deterministically
   `false` — a small-network CPU heuristic, not a platform quirk.

Test changes for the global-step fix: `test_learning_rate_tracking` renamed
to `test_failed_step_preserves_lr_and_global_step` (codex's suggestion,
adopted — more accurately describes what it now asserts) and extended to
assert `trainer.global_step() == 0` both before and after the expected
failed `train_batch` call, alongside the existing unchanged-LR assertion.

Mutation-verified separately (appended to `.mutation_log.md`): reverted just
the `global_step` reorder (moved the increment back to the top of
`train_batch`, matching the pre-fix control flow), touched, ran
`test_failed_step_preserves_lr_and_global_step` — failed
(`global_step() == 1`, not `0`). Restored the fix, re-ran — passes. Full
`train::gpu` suite after restoring: 17 passed, 1 pre-existing/unrelated
failure (`test_is_using_gpu`, cause corrected above), 0 ignored.

## Sibling-invocation-path grep

Grepped the workspace for the same "create empty command encoder, submit,
report success" pattern:

```
grep -rn "create_command_encoder(&wgpu::CommandEncoderDescriptor::default())" crates/
```

Three other call sites exist, all in `crates/fann/src/gpu/network.rs`
(forward-pass layer dispatch, activation dispatch, and buffer readback). Read
each one: all three build a real bind group against the pipeline's layout
with actual weight/input/output buffer bindings and call
`begin_compute_pass` + `dispatch_workgroups` before submitting — they are not
instances of the same defect. Only `lattice-tune`'s optimizer dispatch built
an unused bind-group-layout handle and submitted an empty encoder.

## Tests

Mutation-sensitive tests in `crates/tune/src/train/gpu/tests.rs` (gated the
same way existing GPU tests in this file are: runtime `skip_if_no_gpu()`
plus `#[cfg_attr(not(feature = "gpu-tests"), ignore)]`):

- `test_update_adam_fails_loud`, `test_update_adamw_fails_loud`,
  `test_update_sgd_momentum_fails_loud`, `test_update_sgd_fails_loud`,
  `test_update_rmsprop_fails_loud` — one per optimizer, each configures a
  `GpuTrainer`, calls `train_batch`, and asserts the error fires with the
  expected message fragment (optimizer name + "not implemented").
- `test_gpu_trainer_forward`: previously exercised the default optimizer
  (AdamW) through the full `train_batch` pipeline to check forward-pass loss
  is positive. Since **no** optimizer choice lets `train_batch` complete
  anymore, this test now calls the trainer's `forward_batch` +
  `compute_loss` directly (private methods, accessible from the same-crate
  test module), bypassing `update_weights` entirely — it keeps verifying
  exactly what it verified before (forward pass produces valid positive
  loss) without depending on any optimizer's behavior.
- `test_learning_rate_tracking` → renamed
  `test_failed_step_preserves_lr_and_global_step` (codex round 1): previously
  called `train_batch` (via Adam) and checked that `current_lr()` reflects
  the schedule afterward — that specific claim (LR advances after a batch)
  cannot be verified honestly anymore since `update_weights` always fails
  before the LR-schedule assignment runs. Rather than dodge into a "let me
  just switch to a harmless optimizer" (there isn't one now), the assertion
  now checks what actually holds today: `train_batch` returns `Err`, and
  both `current_lr()` and `global_step()` are provably unchanged afterward
  (both assignments in `train_batch` run strictly after the optimizer `?`,
  so a failed step must not silently advance either). Real, currently-true
  invariants — not a weakened stand-in for the original claim, which will
  need to come back once a real optimizer arm lands.
- `test_update_sgd_plain` (previously asserting `Ok` for 3 successive
  batches) renamed to `test_update_sgd_fails_loud` and rewritten to assert
  the error, matching the other four.

**Mutation verification** (recorded locally in `.mutation_log.md`, not
committed — local-only record, matches this repo's convention for internal
review artifacts):

- Round 1 (Adam/AdamW/SGDMomentum/RMSprop): reverted `optimizers.rs` to the
  pre-fix version in place (`git show HEAD:... > optimizers.rs`, no
  destructive `git checkout` over uncommitted work), touched the file, ran
  the four tests — all four panicked because `train_batch` still returned
  `Ok(loss)` under the reverted no-op code. Restored the fix, re-ran — all
  four green.
- Round 2 (plain SGD): reverted only `update_sgd`'s behavior back to the
  first-round committed state (still the old placeholder-loop `Ok(())`
  body), touched, ran `test_update_sgd_fails_loud` — failed with `.expect_err`
  panicking on `Ok(1.475...)`. Restored the fix, re-ran — passes. Full
  `train::gpu` suite re-run after restoring: 17 passed, 1 pre-existing
  unrelated failure (below), 0 ignored.

This dev machine has a real Metal GPU, so `--features gpu-tests` genuinely
exercises the GPU path here (not just `ignore`-skipped). One pre-existing,
unrelated test failure was observed and confirmed present on the unmodified
branch base (65865ecb5) via `git stash`: `test_is_using_gpu` —
`assert!(trainer.is_using_gpu())` fails. Cause (corrected after codex round 1
review — my original claim of a "paravirtual/Metal-adapter quirk" was
speculation and wrong): the test builds a 6→8→6 network (110 total
parameters), and `GpuNetwork::new` sets `use_gpu =
should_use_gpu(total_params, 1)` (`crates/fann/src/gpu/network.rs:59-61`),
where the single-item threshold is 10,000 elements
(`crates/fann/src/gpu/mod.rs:137-141`) — so `is_using_gpu()` is
deterministically `false` for a network this small, regardless of platform.
Not touched; pre-existing and unrelated to #797.

## Gates

- `cargo fmt -p lattice-tune -- --check` — clean.
- `cargo clippy --workspace -- -D warnings` — clean (this is the gate CI
  actually runs; matches `.github/workflows/ci.yml`'s default-feature
  `cargo clippy --workspace` lane).
- `cargo clippy -p lattice-tune --all-targets --features gpu-tests -- -D
  warnings` — **not clean**, but the failures are 6 pre-existing clippy
  violations inside `crates/fann/src/gpu/{buffer,context,network}.rs`
  (collapsible-if, redundant-closure, cloned-instead-of-copied), unrelated to
  this PR's files. Confirmed pre-existing by stashing this diff and
  re-running the same command against the unmodified branch base — identical
  6 errors. CI itself never lints under the `gpu`/`gpu-tests` feature (only
  `lattice-inference`/`lattice-embed` have dedicated `metal-gpu` clippy lanes
  in `ci.yml`; `lattice-tune`/`lattice-fann`'s `gpu` feature has no clippy
  lane at all), so this is not a regression this PR introduces or is gated on.
- `cargo test -p lattice-tune` (default features) — 50 passed, 0 failed.
- `cargo build -p lattice-tune --features gpu-tests` — clean.
- `cargo test -p lattice-tune --features gpu-tests train::gpu` — 17 passed,
  1 failed (`test_is_using_gpu`, pre-existing/unrelated per above), 0 ignored.
- `cargo test -p lattice-tune --features gpu` (no `gpu-tests`, matches how
  most CI machines without `gpu-tests` opted in would build/test) — 8 passed,
  0 failed, 10 ignored.

## Bench-compare scope (ADR-058)

Not included. This PR touches only `crates/tune/src/train/gpu/{optimizers,state,tests}.rs`
— no `crates/inference` or `crates/embed` source changed — so the
bench-compare requirement doesn't apply per this repo's CLAUDE.md scoping
rule.

## Deviations from plan

- Scope grew beyond the original ask (three shader arms + RMSprop) to
  include plain SGD, on explicit coordinator direction after the `let
  grad_scale = 0.01` constant-placeholder finding — see "Scope grew during
  review" above.
- `state.rs`: three `OptimizerState` buffer fields (`m`/`v`/`velocity`) and
  two `LayerGradients` fields (`num_weights`/`num_biases`) lost their only
  reader once the per-layer uniform/dispatch loop was removed. Rather than
  deleting the (still-useful, pre-allocated) buffers/fields, added
  `#[allow(dead_code, reason = "...")]` referencing #797, since the follow-up
  buffer-wiring work will need them.

Not marking ready or enabling auto-merge — leaving that to review/gating.

### DIFF
--- a/crates/tune/README.md
+++ b/crates/tune/README.md
@@ -326,11 +326,20 @@ let finetuned = RegisteredModel::new("intent_classifier", "1.1.0")
 | ---------------- | ------- | ------------------------------------------------------------ |
 | `std`            | Yes     | Standard library support                                     |
 | `serde`          | No      | Serialization (propagates to lattice-fann)                   |
-| `gpu`            | No      | GPU-accelerated training                                     |
+| `gpu`            | No      | GPU-accelerated forward/backward passes and validation[^gpu-limitation] |
 | `gpu-tests`      | No      | GPU tests requiring hardware                                 |
 | `safetensors`    | No      | Safe checkpoint serialization                                |
 | `inference-hook` | No      | `impl LoraHook for LoraAdapter` (pulls in `lattice-inference`) |
 | `train-backward` | No      | Backward/gradient training surface (pulls in `lattice-inference`) |
+
+[^gpu-limitation]: **Current limitation**: `GpuTrainer::train_batch` returns
+`Err` for every optimizer choice (Adam, AdamW, SGD-momentum, plain SGD,
+RMSprop) — the GPU-shader optimizer dispatch has no buffer bindings wired to
+the network's weight/gradient buffers, and the CPU-side plain-SGD arm has
+neither real gradient plumbing nor a mutable weight write-back path. Forward
+pass and loss computation work correctly; only the weight-update step is
+unimplemented. See [issue #797](https://github.com/ohdearquant/lattice/issues/797).
+This note will be removed once that wiring lands.
 
 ```toml
 [dependencies]
--- a/crates/tune/src/train/gpu/mod.rs
+++ b/crates/tune/src/train/gpu/mod.rs
@@ -21,6 +21,19 @@ use std::sync::Arc;
 ///
 /// Provides GPU-accelerated forward/backward passes and weight updates
 /// using lattice-fann's wgpu backend.
+///
+/// # Current limitation
+///
+/// [`Self::train_batch`] returns `Err(TuneError::Training(_))` for every
+/// optimizer choice (Adam, AdamW, SGD-momentum, plain SGD, RMSprop): the
+/// GPU-shader optimizer dispatch has no buffer bindings wired to the
+/// network's weight/gradient buffers, and the CPU-side plain-SGD arm has
+/// neither real gradient plumbing nor a mutable weight write-back path.
+/// Forward pass and loss computation work correctly — [`Self::validate`] is
+/// a forward-only path that does not touch the optimizer. Only the
+/// weight-update step is unimplemented. See
+/// <https://github.com/ohdearquant/lattice/issues/797>. This note will be
+/// removed once that wiring lands.
 pub struct GpuTrainer {
     /// GPU context (device, queue, shader manager)
     ctx: Arc<GpuContext>,
@@ -180,8 +193,6 @@ impl GpuTrainer {
     ///
     /// Returns the batch loss.
     pub fn train_batch(&mut self, batch: &Batch) -> Result<f32> {
-        self.global_step += 1;
-
         // Forward pass
         let (outputs, activations) = self.forward_batch(batch)?;
 
@@ -203,6 +214,14 @@ impl GpuTrainer {
 
         // Update weights
         self.update_weights()?;
+
+        // Only a fully completed step (forward + backward + optimizer
+        // update all succeeded) counts toward global_step/LR/epoch
+        // accounting. A failed step above must not advance this counter —
+        // see #797: every GpuOptimizer arm currently errors until real
+        // buffer/gradient wiring lands, so this guards against every
+        // failed call silently inflating the step count.
+        self.global_step += 1;
 
         // Update learning rate
         self.current_lr = self.config.lr_schedule.get_lr(
--- a/crates/tune/src/train/gpu/optimizers.rs
+++ b/crates/tune/src/train/gpu/optimizers.rs
@@ -1,4 +1,11 @@
 //! GPU optimizer implementations
+//!
+//! Every `GpuOptimizer` arm (Adam, AdamW, SGD-momentum, plain SGD, RMSprop)
+//! currently fails loudly with `TuneError::Training` rather than performing a
+//! real weight update: the shader-based arms have no buffer bindings wired
+//! to the network's weight/gradient buffers, and the CPU-side plain-SGD arm
+//! has neither real gradient plumbing nor a mutable weight write-back path.
+//! See #797. Wiring the real dispatch/write-back is tracked follow-up work.
 
 use super::state::LayerGradients;
 use crate::error::{Result, TuneError};
@@ -84,7 +91,9 @@ impl GpuOptimizer {
             Optimizer::SGD => Self::update_sgd(network, current_lr)?,
             Optimizer::RMSprop => {
                 return Err(TuneError::Training(
-                    "GPU RMSprop optimizer not implemented; select SGD explicitly (#797)"
+                    "GPU RMSprop optimizer not implemented: this arm previously silently \
+                     substituted plain SGD instead of running the requested algorithm; every \
+                     GpuOptimizer arm fails loudly until real buffer/gradient wiring lands (#797)"
                         .to_string(),
                 ));
             }
@@ -105,34 +114,23 @@ impl GpuOptimizer {
 
     /// Plain SGD update (CPU fallback - no shader for plain SGD)
     ///
-    /// NOTE (#797 follow-up, out of scope for this fix): this arm computes
-    /// updated weights/biases but `GpuNetwork` exposes only an immutable
-    /// `cpu_network()` accessor, so the computed values are never written
-    /// back — this is untouched, pre-existing behavior, tracked separately
-    /// from the shader-dispatch fail-loud fix above.
-    fn update_sgd(network: &lattice_fann::gpu::GpuNetwork, current_lr: f32) -> Result<()> {
-        // Plain SGD without momentum - apply on CPU
-        // This is a simple w = w - lr * grad
-        let cpu_network = network.cpu_network();
-
-        for layer in cpu_network.layers().iter() {
-            let mut weights = layer.weights().to_vec();
-            let mut biases = layer.biases().to_vec();
-
-            // Simple gradient descent
-            let grad_scale = 0.01; // Placeholder gradient magnitude
-
-            for w in weights.iter_mut() {
-                *w -= current_lr * grad_scale;
-            }
-            for b in biases.iter_mut() {
-                *b -= current_lr * grad_scale;
-            }
-
-            // Write back (would need mutable network access)
-            let _ = (weights, biases); // Silence unused warning for now
-        }
-
-        Ok(())
+    /// # Errors
+    ///
+    /// Always returns `TuneError::Training`: this is not real SGD (#797).
+    /// The previous body used a constant placeholder gradient magnitude
+    /// (`grad_scale = 0.01`) instead of the actual per-layer gradients
+    /// accumulated in `LayerGradients`, and had no mutable write-back path
+    /// from `GpuNetwork` (only the immutable `cpu_network()` accessor
+    /// exists) even if it had used real gradients — so the computed values
+    /// were always discarded. A working plain-SGD arm needs both real
+    /// gradient plumbing and a mutable weight-write path on `GpuNetwork`;
+    /// until then it fails loudly like every other arm rather than
+    /// reporting success for a step that changed nothing.
+    fn update_sgd(_network: &lattice_fann::gpu::GpuNetwork, _current_lr: f32) -> Result<()> {
+        Err(TuneError::Training(
+            "GPU SGD optimizer update not implemented: no real gradient plumbing or mutable \
+             weight write-back from GpuNetwork is wired (#797)"
+                .to_string(),
+        ))
     }
 }
--- a/crates/tune/src/train/gpu/optimizers.rs
+++ b/crates/tune/src/train/gpu/optimizers.rs
@@ -1,7 +1,6 @@
 //! GPU optimizer implementations
 
 use super::state::LayerGradients;
-use super::uniforms::{AdamUniforms, AdamWUniforms, SgdMomentumUniforms};
 use crate::error::{Result, TuneError};
 use crate::train::config::{Optimizer, TrainingConfig};
 use lattice_fann::gpu::GpuContext;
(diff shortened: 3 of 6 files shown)


ohdearquant:

## Review of PR #798 (round 1)

### Prior review context

- This is round 1; there are no prior review threads or submitted reviews on the PR.
- The PR is a fail-loud-only correction for issue #797: real GPU optimizer buffer/gradient wiring remains follow-up work.

### What this round examined

- The full `origin/main...e114397e0` diff and complete contents of all three changed files.
- The pre-diff optimizer implementation, `GpuTrainer::train_batch`/`update_weights`, the JIT GPU caller, optimizer state initialization, public docs, and error/API visibility.
- Test transformations, GPU feature gating, mutation sensitivity, sibling command encoders, and both `lattice-fann` training implementations.
- Local format/default tests, feature-clippy baseline, a disposable mutation run, attempted GPU runtime suites, and GitHub CI at the exact head.

### Findings (# total: 0 blockers, 1 major, 1 medium, 1 minor, 0 nits)

#### Majors

1. [`crates/tune/src/train/gpu/mod.rs:183`] A failed optimizer step still advances the public global-step counter
   - **Current**: `self.global_step += 1;` runs before forward/backward and before `self.update_weights()?` at line 205. Every newly fail-loud optimizer error therefore leaves `global_step()` advanced even though no weight update completed. The rewritten state-drift test only checks `current_lr` (`tests.rs:379-386`) and misses this public counter.
   - **Issue**: The PR's contract is that the optimizer error is terminal and does not leave misleading step/epoch accounting. `global_step` feeds LR/approximate-epoch calculations and is publicly observable; repeated failed/retried calls count as completed training steps. This also contradicts the new test commentary's claim of “no partial/silent state drift on failure.”
   - **Fix**: Move the global-step increment to after `update_weights()?` (before calculating the next LR), and extend the failed-step regression test to assert `trainer.global_step() == 0`.
   - **Test**: `test_learning_rate_tracking` (preferably renamed to describe failed-step state) should assert both unchanged LR and unchanged global step after the expected optimizer error.
   - **Confidence**: derived from the `train_batch` control flow and the public `global_step()` accessor at `mod.rs:441-443`.

#### Mediums

1. [`crates/tune/src/train/mod.rs:33`] Published docs still promise working GPU training although every optimizer now necessarily errors
   - **Current**: Public module rustdoc says “GPU-accelerated training is available” and its example calls `trainer.train_batch(&batch)?` (`train/mod.rs:33-48`); `GpuTrainer` likewise claims it provides GPU weight updates (`train/gpu/mod.rs:20-23`), and the crate feature table calls `gpu` “GPU-accelerated training” (`crates/tune/README.md:329`).
   - **Issue**: The fail-loud fix intentionally makes every public `GpuTrainer::train_batch` call terminate at the optimizer. The PR body is honest, but downstream rustdoc/README users still receive the opposite contract. This is especially relevant for a published crate where the internal `GpuOptimizer` docs are unreachable to consumers.
   - **Fix**: Add a temporary, prominent `# Current limitation` note to the public GPU-training docs and feature table stating that forward/validation are available but weight-update training returns `TuneError::Training` for every optimizer until #797's wiring follow-up lands. Remove the note with that implementation.
   - **Test**: no cheap runtime test; `cargo doc`/rustdoc lint should validate the edited links and formatting.
   - **Confidence**: derived from the exhaustive optimizer match and the cited public docs.

#### Minors

1. [`crates/tune/src/train/gpu/tests.rs:413`] The PR body gives the wrong cause for the disclosed `test_is_using_gpu` baseline failure
   - The test builds a `6 -> 8 -> 6` network (110 parameters). `GpuNetwork::new` sets `use_gpu = should_use_gpu(total_params, 1)` (`crates/fann/src/gpu/network.rs:59-61`), and the single-item threshold is 10,000 elements (`crates/fann/src/gpu/mod.rs:54,137-141`), so `trainer.is_using_gpu()` is deterministically false. This is pre-existing and unrelated as claimed, but it is not evidence of a paravirtual/Metal-adapter quirk.
   - **Fix**: Correct the PR body to attribute the baseline failure to the intentional small-network CPU heuristic (or simply state that the cause is pre-existing and out of scope without speculating).
   - **Confidence**: derived from the exact test dimensions and the dispatch heuristic.

- Head/base identity matches the task: `e114397e0` against `origin/main` at `65865ecb5`.
- The diff is tune-only: three files under `crates/tune/src/train/gpu/`; no inference/embed paths changed.
- PR/issue prose is agent-attributed as required by `AGENTS.md`.
- All five `GpuOptimizer::update` match arms currently return `Err`; optimizer-state timestep increment and `sync_weights()` are after the arm-level `?` and are unreachable on these errors.
- `GpuTrainer::update_weights` returns the optimizer result directly. `train_batch`'s LR assignment and loss return are after `update_weights()?`, and `JitAdapter::adapt_gpu` uses `trainer.train_batch(&batch)?` before appending loss or incrementing `steps_completed`; those paths terminate correctly. The sole partial-state leak found is `global_step` above.
- Pre-diff Adam/AdamW/SGD-momentum compiled shaders and constructed uniform values, but the common helper never created/bound a bind group, began a compute pass, or dispatched work. Plain SGD only modified cloned vectors with constant `0.01` values and discarded them. No effective weight-update behavior was deleted. The project-version wgpu docs describe `Device::create_command_encoder` as creating an empty encoder and `Queue` as executing recorded command buffers ([Device](https://docs.rs/wgpu/23.0.1/wgpu/struct.Device.html#method.create_command_encoder), [Queue](https://docs.rs/wgpu/23.0.1/wgpu/struct.Queue.html#method.submit)).
- The `m`/`v`/`velocity` buffers and `num_weights`/`num_biases` metadata predate this PR and match the state/dispatch inputs the #797 implementation follow-up will need. The narrow, reason-tagged `dead_code` allowances preserve that scaffolding without hiding unrelated fields. The now-unreferenced uniform structs likewise pin the existing shader ABI and have size tests.
- The five fail-loud tests each select a distinct optimizer and reach the optimizer through `train_batch`; restoring success will make `expect_err` fail. They also transitively keep the forward/loss/backward path exercised under `gpu-tests`. `test_gpu_trainer_forward` itself no longer covers backward, but that coverage remains in those five tests; the direct forward/loss split is otherwise honest. The original LR test only asserted `updated_lr > 0.0`, not that LR changed; the replacement's unchanged-LR assertion is stronger for the current error contract, aside from the omitted global-step assertion in the Major finding.
- Observed mutation spot-check in detached disposable worktree `/tmp/lattice-pr798-mutation`: replaced Adam's error with an actual empty encoder submission plus `Ok(())`, compiled the mutation, and ran `test_update_adam_fails_loud` under the Metal lock. It failed at `tests.rs:159` because `expect_err` received `Ok(1.3986235)`. This directly confirms the Adam regression test is mutation-sensitive. No source in the review worktree was changed.
- `TuneError::Training` is the closest existing variant; there is no unsupported/not-implemented variant, and adding one to the public, exhaustive `TuneError` enum would itself create a semver concern. `GpuOptimizer` is behind private `train::gpu::optimizers` and is not public API.
- Each error names the selected optimizer, says `not implemented`, explains the missing wiring, and references #797. The public method signature and error enum shape are unchanged, so this is a behavioral bug fix rather than a Rust API break.
- The exact empty-default-encoder grep has only the three sibling sites in `crates/fann/src/gpu/network.rs`; they respectively record a compute pass/dispatch, a compute pass/dispatch, and a buffer copy before submission.
- `lattice-fann` has no separate GPU weight-update trainer: `BackpropTrainer::apply_gradients` mutates layer weights/biases on CPU, and `RlooTrainer::backprop_and_apply` likewise applies real CPU SGD. Neither has the success-without-update pattern.
- Observed `cargo fmt --check`: pass.
- Observed `cargo clippy --workspace -- -D warnings`: pass.
- Observed `cargo test -p lattice-tune` at head: pass (182 unit tests, 7 property tests, 50 integration tests, and doc tests; no failures).
- Observed `cargo test -p lattice-tune --features gpu-tests --no-run`: pass; the feature-gated library and test targets compile at head.
- Observed on detached, unmodified base `65865ecb5`: `cargo clippy -p lattice-tune --all-targets --features gpu-tests -- -D warnings` fails with the same six disclosed `lattice-fann` diagnostics (two `collapsible_if`, two `redundant_closure_for_method_calls`, one additional `collapsible_if` in `network.rs`, and one `cloned_instead_of_copied`). They are pre-existing and outside this diff.
- Observed the same feature-clippy command at head: the identical six dependency diagnostics. Because that command stops before linting `lattice-tune`, I also ran `cargo clippy -p lattice-tune --all-targets --features gpu-tests --no-deps -- -D warnings`; it passes, confirming the changed feature-gated crate code has no hidden clippy warning.
- GitHub CI at `e114397e0` is green: aggregate `ci-gate`, macOS/Ubuntu/ARM CI, both feature-matrix legs, MSRV, semver, cargo-deny, actionlint, rustdoc/unwrap informational checks, app binaries, and bench compilation passed; change-scoped parity jobs skipped as designed.
- The diff contains only `crates/tune/src/train/gpu/{optimizers,state,tests}.rs`; ADR-058's inference/embed bench-compare scope is not triggered.

### What I did NOT check

- I did not obtain an uncontaminated full head/base `gpu-tests` runtime result before the deadline: a long-running Lattice Metal suite held/reacquired the shared GPU lock. I discarded two intermediate outputs after detecting that a shared Cargo target had reused the mutation binary, then queued a clean head rerun; no claim from those contaminated runs is used above. The feature suite compiles cleanly, the Adam mutation was personally observed, and the base failure/clippy conclusions above are independently source- or detached-base-derived.
- No benchmark was run because the change is outside ADR-058's benchmark scope and removes non-working optimizer work rather than changing inference/embed performance.
- No real optimizer update exists to numerically validate; that implementation is explicitly follow-up scope.

- Re-review after the global-step ordering/test and public docs are fixed. No further recursive round should be necessary if those changes are narrow and the targeted GPU rerun shows only the known baseline failure.

Domain utility: MEDIUM — GPU-training resilience guidance reinforced checking counters/metrics as committed state, not only weight buffers and return values.

Verdict: APPROVE-WITH-FIXES

oceanwaves630 (approved):
Approving per the documented gate trail: review fixes applied and verified line-by-line; narrow tune-only diff.

Issue #814, opened 2026-07-10
chore(scripts): delete superseded decode-benchmark scripts and fix Makefile .PHONY gaps


ohdearquant:
> _Issue authored by Claude (Anthropic agent) on behalf of @ohdearquant._

## What

Two decode-comparison scripts are proven dead:

- `scripts/bench_decode_slope.sh` (N1=8; contexts `{64,128,256,512,1024}`;
  Lattice-only medians, actual completion-token deltas, plain OLS of
  per-token time) is superseded by the ADR-067 Theil-Sen/bootstrap
  `bench_decode_slopefit.*` path and the `adr064-gpu-decode-nightly.yml`
  workflow that runs it.
- `scripts/bench_q8_vs_ollama.sh` predates the corrected marginal-cost
  method and has no caller beyond itself (it compares Criterion per-step
  throughput, Ollama native eval time, and MLX wall time as if they were
  the same unit).

The Makefile's `.PHONY` declaration (`Makefile:1`) is also stale: it omits
`e2e-parity` (target at `Makefile:69`), `bench-decode` (`:76`), and
`bench-decode-slopefit` (`:83`) even though all three are real
non-file-producing targets.

## Why

Dead scripts and Make targets are the kind of debt that gets copy-pasted
into new work by contributors who find them via `grep` and assume they are
current. Removing them, and fixing `.PHONY` for the targets that remain,
keeps the decode-benchmark surface limited to what is actually
authoritative (the harness/profile work in progress, plus the ADR-067
slopefit path and the ADR-064 nightly workflow).

## Steps

1. Confirm via repository-wide reference search that neither script has a
   live caller (Make target, workflow, doc, or another script) beyond
   itself before deleting.
2. Delete `scripts/bench_q8_vs_ollama.sh` and `scripts/bench_decode_slope.sh`.
3. Remove the `bench-decode` Make target (`Makefile:76`, which invokes
   `bench_decode_slope.sh`).
4. Add `e2e-parity` and `bench-decode-slopefit` to `.PHONY` on `Makefile:1`;
   do not add `bench-decode` since it no longer exists.
5. Update any current (non-historical) docs or Make comments that still
   reference `bench-decode` or the deleted scripts. Historical release
   notes are left as-is.

## Gates

- A repository-wide reference search finds no live caller of either
  deleted script after removal.
- `.PHONY` lists exactly the non-file-producing targets that exist in the
  Makefile after this change — no stale entries, no missing entries.
[...]

Issue #813, opened 2026-07-10
bench(scripts): consolidate decode-benchmark methodology into one harness with per-consumer profiles


ohdearquant:
> _Issue authored by Claude (Anthropic agent) on behalf of @ohdearquant._

## What

`scripts/` has seven independent implementations of the same
prompt/decode-benchmark shape, each with its own methodology that has
drifted from the others:

| Script | Window / contexts | Repeats / warmup | Aggregation |
|---|---|---|---|
| `bench_apples_to_apples.sh` (`:30-33,90,138`) | N1=32, N2=256 | 5; only MLX gets one 8-token warmup | independent medians, two-point slope; mixes engine-native and external total time |
| `bench_apples_precise.sh` (`:15-19,35-47,118-144`) | N1=64, N2=512 | 15 total, 2 warmup, 13 measured | drop two tail values, mean, normal-approximation CI |
| `bench_q4_apples.sh` (`:24-29,86,96-121`) | N1=32, N2=256 | 5; only MLX gets one 8-token warmup | independent medians, two-point slope |
| `bench_context_scaling.sh` (`:26-36,62-99,146-179`) | N1=8; N2={64,128,256} | 5; only MLX has explicit 4-token warmup | one N1 median reused for every context |
| `bench_compare_1k.py` (`:7-29,70-87,94-144,192-249,324-356`) | N1=1, N2=100 at prompt contexts 1000/2000/4000 | 5; engine-specific warmups | independent medians; Ollama uses a different prompt-count method |

`bench_compare_1k.py` is called by the `bench-agentic` and
`bench-agentic-quick` Make targets and is a live seventh reimplementation of
the same shape, not a documentation gap.

Two additional scripts implement the same decode-comparison shape but have
no live caller and are proven dead (tracked in the companion cleanup issue):
`bench_decode_slope.sh` (superseded by the ADR-067 Theil-Sen/bootstrap
`bench_decode_slopefit.*` path) and `bench_q8_vs_ollama.sh`.

## Why

Every script owns its own copy of warmup policy, sample-count policy,
aggregation/CI method, and output rendering. That means a methodology fix
(or bug fix) applied to one script silently does not apply to the other six,
and comparing results across scripts is not meaningful even though they look
alike. Centralizing collection policy and reporting removes that drift
without changing what any of the five live scripts currently measures.

## Steps

1. **Harness core (no public command change).** Add a stdlib-only
[...]

PR #857, opened 2026-07-11
bench(scripts): decode-benchmark harness core with CI-wired schema tests

ohdearquant:

Closes step 1 of #813 ("Harness core, no public command change").

## What

- `scripts/bench_decode_harness.py` — stdlib-only decode-benchmark harness
  core. Owns configuration validation, requested schedule/run order,
  monotonic wall timing, the raw JSONL observation schema, aggregation
  (median and trimmed-mean+legacy-CI), and report rendering. A profile's
  `engines` list is an ordered set of per-engine run groups (`name`,
  `warmup_repeats`, `warmup_tokens`, `warmup_prompt`, `model`,
  `quantization`, `measured_calls`, `measured_order`, `measured_prompt`),
  not a flat name list, because the five legacy scripts each give
  individual engines their own warmup schedule, their own MEASURED call
  shape, their own MEASURED prompt, and, for the Q4/Q8 comparison, their
  own quantization within one profile.

  Warmup is an explicit pre-window batch: when `warmup_repeats > 0`, the
  group's full warmup batch runs once, entirely before that group's
  measured calls, at its own `warmup_tokens` completion-token budget and
  an optional `warmup_prompt` (defaults to `profile.prompt` when absent).

  The MEASURED schedule is an explicit per-engine call plan:
  `measured_calls` is an ordered tuple of `(n_tokens, yields_windows)`
  calls; when omitted, the harness derives the default (one call per
  `profile.windows` entry, `yields_windows=(window,)`), which reproduces
  every uniform-window script unchanged. `measured_order`
  (`"window_major"`, the default, or `"repeat_major"`) controls whether
  the call plan replays as "all repeats of call[0], then call[1], ..." or
  "one full pass through the plan per repeat". A call with
  `len(yields_windows) > 1` (one physical call producing observations for
  more than one window) requires the adapter's `AdapterRunResult.
  component_ns` to map every yielded window to its own elapsed
  nanoseconds — the harness never re-invokes the engine to synthesize a
  second measurement and never fabricates one from the call's own
  wall-clock span for more than the primary window; a missing or
  incomplete `component_ns` raises `AdapterContractError`, fail-closed.
  An explicit `measured_calls` plan's flattened `yields_windows` must
  cover `profile.windows` exactly once each (order not enforced — call
  order governs replay/interleaving, not window existence); a missing,
  undeclared, or cross-call-duplicate window is rejected at parse time
  with a `ProfileConfigError` naming the engine index and the exact
  missing/undeclared/duplicate windows, never silently accepted into a
  slope with an erased or double-weighted sample.

  The MEASURED prompt is an explicit per-engine override, not always
  profile-wide: `measured_prompt` on an `EngineRunGroup` overrides
  `profile.prompt` for that engine's measured calls only (mirroring
  `warmup_prompt`'s override of the warmup prompt). The harness passes
  exactly the resolved string (the override, or `profile.prompt` when
  absent) to the adapter and hashes exactly that string — never a
  placeholder the adapter is trusted to reinterpret.

  The requested model/quantization is passed into `EngineAdapter.run()`;
  an adapter that invokes a different artifact than requested reports the
  actual identity back via `AdapterRunResult`, and raw observations
  always record the actual invoked identity. Profile TOML parsing is
  fail-closed on unknown top-level, per-profile, per-engine-group, and
  per-measured-call keys — a misspelled optional field is rejected with a
  `ProfileConfigError` naming the unexpected key, never silently
  defaulted. Ships the `EngineAdapter` protocol only — no concrete
  Lattice/Ollama/MLX adapters land here; those arrive with each script
  migration in #813 step 2. Until then every profile's requested engines
  resolve as "missing" and the `run` subcommand fails closed unless
  `--allow-missing-engine` is passed (no stale/fabricated result is ever
  substituted for a live run). `SlopeResult.slope_ci95_legacy` is
  explicitly labeled and documented as `bench_apples_precise.sh`'s
  pre-existing normal-approximation spread heuristic carried over
  unchanged — not a coverage-valid 95% CI; statistical-methodology
  correction stays out of scope for this issue.
- `scripts/bench_decode_profiles.toml` — profile scaffold, intentionally
  empty (`schema_version = 1`, no `[profiles.*]` entries yet), with the
  per-engine run-group schema (including `measured_calls`/
  `measured_order`/`measured_prompt`, `warmup_tokens`/`warmup_prompt`,
  and the windows-coverage fail-closed relationship) documented inline
  plus two commented-out examples: the Q4/Q8-mixed shape from
  `bench_q4_apples.sh`, and the three-distinct-measured-shapes,
  two-distinct-measured-prompts agentic example illustrating
  `bench_compare_1k.py`'s Lattice/Ollama/MLX plans.
- `tests/test_bench_decode_harness.py` — 117 deterministic unit tests
  using fake adapters and an injectable clock: raw-schema validation,
  fail-closed profile-config validation (top-level, per-profile,
  per-engine, and per-measured-call keys; `warmup_tokens`/`warmup_prompt`/
  `measured_order`/`measured_calls`/`measured_prompt` type and
  relationship checks, including the windows-coverage relationship
  between a `measured_calls` plan and `profile.windows`),
  `run_profile`'s warmup-batch-before-all-measured-calls placement,
  order-index bookkeeping, and missing-engine fail-closed behavior, an
  exact-call-sequence test reproducing the legacy Lattice → Ollama → MLX
  global order (Q4/apples-to-apples shape), global flattened-sequence
  tests for `bench_apples_precise.sh` and `bench_context_scaling.sh`
  using all three real engines (not a single stand-in adapter), a
  complete Lattice → Ollama → MLX global flattened-sequence test
  reproducing `bench_compare_1k.py`'s full three-shape, two-prompt
  schedule end to end (Lattice window-major on the shared tokenizer-
  padded prompt, Ollama's one-call-yields-two-windows shape via
  `component_ns` on its own character-count-heuristic `measured_prompt`,
  MLX's repeat-major alternation on the shared prompt) — asserting, per
  physical call, the exact prompt string invoked (via a `_FakeAdapter`
  capture) and, per raw observation, that `prompt_hash` matches
  `hash()` of that exact invoked string, `AdapterContractError`
  tests for missing/incomplete `component_ns`, three windows-coverage
  fail-closed tests (missing/undeclared/cross-call-duplicate window),
  adapter actual-identity override behavior, median and
  trimmed-mean+legacy-CI aggregation
  against hand-computed asymmetric-outlier data, and the `run`/`validate`
  CLI subcommands. No engine binary required.
- `.github/workflows/ci.yml` — one job, `bench-decode-harness-tests`,
  ungated like the existing `typos`/`actionlint` jobs, invoking the test
  file directly (`python3 tests/test_bench_decode_harness.py -v`) so a
  missing/renamed test file fails the job instead of a discovery pattern
  silently matching zero tests. **This is the only CI-workflow change in
  this PR** (rounds 2, 3, and 4 made no further changes here — every fix
  since round 1 has been contained to the harness/profile-schema/test
  files).

## Design: two explicit schedule contracts, not a uniform-windows model

**This PR's earlier round claimed `windows`/`measured_repeats` alone
(profile-wide, one call per window, executed uniformly as `[w1 x R, w2 x
R, ...]` for every engine) were sufficient to represent every legacy
script's measured shape. That claim was FALSE, and round-3 review caught
it: `bench_compare_1k.py` has three different measured shapes in one
profile that no choice of profile-wide fields can produce.** Lattice runs
all 1-token calls then all 100-token calls (window-major — the uniform
default does reproduce this one). Ollama issues exactly ONE 100-token
call per repeat and derives BOTH a window=1 (TTFT) and a window=100
(total) observation from that single response's two engine-reported
timing components. MLX alternates a 1-token call and a 100-token call
inside every repeat (repeat-major, not window-major). An adapter cannot
repair this gap without taking schedule ownership away from the harness
or fabricating a second measurement — the fix had to happen in the
harness's schedule model itself.

The corrected design (round 3): each `EngineRunGroup` carries its own
`measured_calls` (an ordered tuple of `MeasuredCall(n_tokens,
yields_windows)`) and `measured_order` (`"window_major"` or
`"repeat_major"`). Omitting `measured_calls` derives the pre-existing
default from `profile.windows` — every uniform-window legacy script
(`bench_apples_to_apples.sh`, `bench_apples_precise.sh`,
`bench_q4_apples.sh`, `bench_context_scaling.sh`, and Lattice's half of
`bench_compare_1k.py`) is unaffected by this change and needs no new
per-engine configuration. Engines whose shape the default can't
represent set an explicit plan: MLX sets `measured_order = "repeat_major"`
with a two-call plan; Ollama sets a single `MeasuredCall(n_tokens=100,
yields_windows=[1, 100])` and its adapter must report
`AdapterRunResult.component_ns = {1: ..., 100: ...}`.

An ordered-operation-list model (fully general per-engine operation
sequences, warmup and measured calls unified into one arbitrary list) was
considered again at this round, now with the measured-side evidence in
hand. Still rejected for the same reason as round 2's warmup design
choice: every actual legacy shape — including `bench_compare_1k.py`'s
three-way divergence — fits "warmup batch, then an ordered measured call
plan, replayed window-major or repeat-major," with no engine needing
warmup and measured calls interleaved with each other. The closed
two-field contract (`measured_calls`, `measured_order`) keeps the TOML
validators exhaustive and keeps `run_profile` two fixed, enumerable
iteration shapes instead of an interpreter over an arbitrary op sequence.
`AdapterRunResult.component_ns` is the multi-component timing extension
the review asked for: it lets one physical call report more than one
window's elapsed time without the harness inventing data the adapter
never measured.

**Round 3's own Design section claimed the measured prompt could stay
profile-wide (`profile.prompt`, sent unchanged to every engine's
measured calls) once the call-shape/timing fix landed. Round-4 review
caught that this claim was ALSO FALSE, and this line retracts it
explicitly rather than leaving it in place:** `bench_compare_1k.py`
requires two different measured prompts for the same nominal context —
Lattice and MLX build a tokenizer-padded prompt (`_build_padded_prompt`
mirrored on the Lattice side by a tokenizer-driven builder), while
Ollama builds a distinct character-count heuristic prompt
(`_make_prompt`). A single profile-wide `prompt` field cannot represent
both, and the earlier commented agentic TOML example admitted the gap
with a `"<ctx-padded prompt, built per-engine at the adapter level>"`
placeholder — which would have moved prompt ownership out of the
harness and made every raw `prompt_hash` describe the placeholder
string rather than the string actually sent to the engine.

The corrected design (round 4): `EngineRunGroup.measured_prompt`
(optional `str`, defaults to `None`) overrides `profile.prompt` for
that engine's measured calls only, mirroring the pre-existing
`warmup_prompt` override of the warmup prompt. `run_profile` resolves
each group's own measured prompt/hash once — the override if present,
else `profile.prompt` — and passes that resolved string to every
measured call for the group; the raw observation's `prompt_hash` is
computed from the exact same string, never from a placeholder or a
value the adapter is trusted to reinterpret. This keeps prompt
construction a harness-owned request rather than an adapter
implementation detail, matching the model/quantization identity
contract already in place. The commented agentic TOML example now uses
`measured_prompt` on the Ollama engine block instead of the placeholder.

## Out of scope (left for later #813-step-2 PRs / sibling issues)

Migrating the five live scripts (`bench_apples_to_apples.sh`,
`bench_apples_precise.sh`, `bench_q4_apples.sh`, `bench_context_scaling.sh`,
`bench_compare_1k.py`) onto profiles, adding concrete engine adapters, and
removing/renaming anything are all untouched here, per #813's own
"one script per PR" sequencing. Nothing from #814/#815/#817/#818 is
touched. Statistical-methodology correction for the legacy CI heuristic
stays deferred to its own separately reviewed issue, per #813's explicit
scope boundary.

## Gates / evidence

- Head: `ef191f364b7c75a4cf107900f8ca5c1db7d1c7a5` (up to date with
  `origin/main` — verified via `git merge-base --is-ancestor origin/main
  HEAD`).
- `python3 tests/test_bench_decode_harness.py -v` → `Ran 117 tests ... OK`.
- `python3 -m py_compile scripts/bench_decode_harness.py
  tests/test_bench_decode_harness.py` → clean.
- Mutation checks, verified locally by temporarily reverting a fix (via a
  `cp` backup, restored via `cp` and re-verified with `diff` against the
  backup — never `git checkout` over uncommitted work), confirming the
  affected test(s) fail, then restoring:
  - Round 1: removing the trimmed-mean slice fails both
    `test_trimmed_mean_stat_exact_with_asymmetric_outliers` and
    `test_trimmed_mean_slope_exact_with_asymmetric_outliers`.
  - Round 2: reverting the warmup batch to loop once per measured window
    (instead of once, before all measured calls) fails 5 tests:
    `test_warmup_executes_once_before_all_measured_windows`,
    `test_heterogeneous_engine_schedule_exact_call_sequence_and_identities`,
    `test_precise_style_warmup_twice_before_both_measured_windows`,
    `test_context_scaling_style_single_warmup_before_all_comparison_windows`,
    and `test_produces_expected_observation_count`.
  - Round 3: (a) ignoring `measured_order` and always taking the
    `window_major` branch fails the 2 tests exercising repeat-major
    ordering (`test_measured_order_repeat_major_alternates_calls_per_repeat`,
    `test_agentic_style_global_flattened_sequence_lattice_ollama_mlx`);
    (b) disabling the `component_ns` multi-yield path (always treating a
    measured call as single-yield, keyed by `n_tokens`, using the
    harness's own wall-clock span) fails the 4 tests exercising the
    multi-yield contract
    (`test_measured_call_multi_yield_uses_component_ns_for_elapsed_ns`,
    `test_measured_call_multi_yield_missing_component_ns_raises`,
    `test_measured_call_multi_yield_missing_window_entry_raises`,
    `test_agentic_style_global_flattened_sequence_lattice_ollama_mlx`).
    Both mutations restored and the clean tree re-verified at 108/108
    green after each (test count at round 3).
  - Round 4: (a) ignoring `measured_prompt` and always resolving/hashing
    `profile.prompt` for every engine's measured calls fails exactly 1
    test, `test_agentic_style_global_flattened_sequence_lattice_ollama_mlx`
    (its per-call prompt and per-observation `prompt_hash` assertions
    catch Ollama silently receiving the Lattice/MLX prompt); (b)
    disabling the windows-coverage relationship check (skipping the
    flattened-`yields_windows`-vs-`profile.windows` comparison) fails
    exactly 3 tests, the three new relationship tests
    (`test_measured_calls_missing_declared_window_rejected`,
    `test_measured_calls_undeclared_window_rejected`,
    `test_measured_calls_cross_call_duplicate_window_rejected`). Both
    mutations applied via `cp` backup, restored via `cp` and
    re-verified byte-identical with `diff` against the backup — never
    `git checkout` — and the clean tree re-verified at 117/117 green
    after each.
- CLI smoke-tested directly: `run --profile <unknown>` and
  `run --profile <profile> --profiles-file <toml requesting an
  unregistered engine>` both fail closed (exit 1) without
  `--allow-missing-engine`, and exit 0 with it, emitting a report with
  the missing engine explicitly named — never a fabricated result.
  `validate <missing-file>` and `<malformed-toml>` fail closed with a
  clean message instead of an unhandled traceback.
- `actionlint .github/workflows/ci.yml` → clean (rc=0); no changes to
  this file in rounds 2, 3, or 4.
- `typos` → clean on all four changed files.
- `git diff --check origin/main...HEAD` → clean.
- Publication-hygiene grep (internal-ops markers, CJK, workspace paths)
  on all four changed files → no matches.
- No Rust files touched by this PR's own commits, so `cargo fmt`/`clippy`
  are not applicable to this diff. `origin/main` gained one unrelated
  Rust commit (`634a8a8c4`, WGPU attention_softmax fail-closed fix)
  since round 3; merged cleanly into this branch with no conflicts and
  no changes to any of the four gate files.

## Spec deviations

None identified against the current design. Two earlier-round claims
have been corrected rather than left in place with incorrect framing:
(1) round 3 corrected round 1/2's claim that `windows`/`measured_repeats`
alone were sufficient to represent every legacy script's measured
schedule (see Design section above); (2) **round 4 corrects round 3's
own claim that the measured `prompt` could stay profile-wide once the
call-shape/timing fix landed — that was also false, and is retracted
above.** `windows`, `measured_repeats`, and the measured `prompt` stay
profile-wide as the DEFAULT (every legacy script keeps those uniform
across engines when nothing more specific is given), while individual
engines override with `measured_calls`/`measured_order`/`measured_prompt`
only where the default cannot represent their actual shape — that
"default profile-wide, explicit per-engine override" pattern is now
applied consistently to schedule shape, timing, AND prompt identity,
not asserted as sufficient on its own without the override path.

The harness's window/aggregation model otherwise generalizes cleanly
over all five legacy scripts' shapes without needing a `baseline_reuse`
flag or similar: a flat, once-each `windows` list (`windows[0]` =
baseline, later entries = comparison points) reproduces
`bench_context_scaling.sh`'s "N1 run once, reused across every context"
behavior for free, and `bench_compare_1k.py`'s per-prompt-context matrix
maps to one profile per context length rather than a special case in the
harness — that mapping decision is deferred to its own migration PR per
the issue's step-2 sequencing, not decided here. The per-engine
run-group schema (`engines` as an array of tables rather than a flat
name list, now carrying a warmup plan, a measured call plan, and a
measured prompt override) is a structural change from the harness's
initial landing, made to represent every legacy script's per-engine
schedule and prompt without another schema migration later.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -476,8 +476,8 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: '3.11'
-      - name: python3 -m unittest tests/test_bench_decode_harness.py
-        run: python3 -m unittest discover -s tests -p 'test_bench_decode_harness.py' -v
+      - name: python3 tests/test_bench_decode_harness.py
+        run: python3 tests/test_bench_decode_harness.py -v
 
   # cargo-semver-checks against the latest published crates.io release for
   # every internal path-dependency crate, so a breaking public API change is
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -464,6 +464,21 @@ jobs:
         with:
           version: '1.7.12'
 
+  # Deterministic fake-adapter unit tests + raw-observation schema validation
+  # for the decode-benchmark harness core (issue #813 step 1). Stdlib-only,
+  # no engine binary or model weights required — cheap ubuntu-only, so it is
+  # ungated like typos/actionlint above rather than routed through ci-gate.
+  bench-decode-harness-tests:
+    name: bench-decode-harness-tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+      - name: python3 -m unittest tests/test_bench_decode_harness.py
+        run: python3 -m unittest discover -s tests -p 'test_bench_decode_harness.py' -v
+
   # cargo-semver-checks against the latest published crates.io release for
   # every internal path-dependency crate, so a breaking public API change is
   # flagged before it ships unversioned, the way #634 was not. Gated through
--- a/scripts/bench_decode_harness.py
+++ b/scripts/bench_decode_harness.py
@@ -23,6 +23,19 @@ then `bench_decode_profiles.toml` carries no profile definitions and the
 `run` subcommand below has nothing registered in `ADAPTER_REGISTRY`, so an
 engine request always resolves as "missing" (see `--allow-missing-engine`).
 
+A profile's `engines` list is an ordered set of per-engine run groups, not
+a flat list of names: each entry carries its own `warmup_repeats`, `model`,
+and `quantization` (see `EngineRunGroup`), because the legacy scripts this
+harness consolidates give each engine its own warmup schedule and, for the
+Q4/Q8 comparison, a different quantization per engine within one profile.
+`windows`, `measured_repeats`, and `prompt` stay profile-wide because every
+legacy script keeps those uniform across engines. The requested
+model/quantization is passed into `EngineAdapter.run()`; an adapter that
+invokes a different artifact than requested (fallback, missing checkpoint,
+...) reports the actual identity back via `AdapterRunResult.actual_model`/
+`actual_quantization`, and raw observations always record the actual
+invoked identity, never a value merely assumed from the profile.
+
 Raw JSONL observation schema (the contract; see `OBSERVATION_FIELDS` /
 `validate_observation`): schema version, git SHA, profile name, engine +
 version, model/quantization, prompt hash, requested and actual prompt/
@@ -31,7 +44,13 @@ harness-measured elapsed nanoseconds, an optional engine-native nanosecond
 figure, a hardware identifier, and a timestamp. Reports render from this raw
 data; they are not the source of truth.
 
-Run with: python3 -m unittest tests/test_bench_decode_harness.py -v
+`trimmed_mean` aggregation reports `slope_ci95_legacy`: the pre-existing
+`bench_apples_precise.sh` normal-approximation spread heuristic, carried
+over unchanged. It is NOT a coverage-valid 95% confidence interval for the
+slope -- statistical-methodology correction is explicitly out of scope for
+issue #813 and is deferred to a separately reviewed follow-up.
+
+Run with: python3 tests/test_bench_decode_harness.py -v
 (the module itself is stdlib-only; no engine binaries or third-party
 packages are required to exercise it).
 """
@@ -246,19 +265,54 @@ def write_jsonl(observations: list[Observation], path: Path) -> None:
 
 
 @dataclass(frozen=True)
+class EngineRunGroup:
+    """One engine's schedule within a profile.
+
+    Every legacy script gives each engine its own warmup count and its own
+    model/quantization identity (`bench_apples_to_apples.sh`: only MLX gets
+    a warmup; `bench_q4_apples.sh`: Lattice and MLX run Q4, Ollama runs Q8 as
+    a reference). `measured_repeats`, `windows`, and `prompt` stay profile-
+    wide because every legacy script keeps those uniform across engines.
+    """
+
+    name: str
+    warmup_repeats: int
+    model: str
+    quantization: str
+
+
+@dataclass(frozen=True)
 class ProfileConfig:
     name: str
     description: str
     windows: tuple[int, ...]
-    warmup_repeats: int
     measured_repeats: int
-    engines: tuple[str, ...]
+    engine_groups: tuple[EngineRunGroup, ...]
     prompt: str
-    model: str
-    quantization: str
     aggregation: str = "median"
     trim: int = 0
     requested_prompt_tokens: int | None = None
+
+    @property
+    def engines(self) -> tuple[str, ...]:
+        """Engine names in schedule order (for aggregation/reporting)."""
+        return tuple(g.name for g in self.engine_groups)
+
+
+_TOP_LEVEL_ALLOWED_KEYS = frozenset({"schema_version", "profiles"})
+_PROFILE_ALLOWED_KEYS = frozenset(
+    {
+        "description",
+        "windows",
+        "measured_repeats",
+        "engines",
+        "prompt",
+        "aggregation",
+        "trim",
+        "requested_prompt_tokens",
+    }
+)
+_ENGINE_GROUP_ALLOWED_KEYS = frozenset({"name", "warmup_repeats", "model", "quantization"})
 
 
 def load_profiles_file(path: Path) -> tuple[int, dict[str, ProfileConfig]]:
@@ -269,6 +323,10 @@ def load_profiles_file(path: Path) -> tuple[int, dict[str, ProfileConfig]]:
         raise ProfileConfigError(f"{path}: invalid TOML: {exc}") from exc
     except OSError as exc:
         raise ProfileConfigError(f"{path}: could not read profiles file: {exc}") from exc
+
+    extra_top = set(doc) - _TOP_LEVEL_ALLOWED_KEYS
+    if extra_top:
+        raise ProfileConfigError(f"{path}: unexpected top-level key(s): {sorted(extra_top)}")
 
     schema_version = doc.get("schema_version")
     if schema_version != SCHEMA_VERSION:
@@ -289,6 +347,10 @@ def load_profiles_file(path: Path) -> tuple[int, dict[str, ProfileConfig]]:
 
 
 def _parse_profile(name: str, raw: dict) -> ProfileConfig:
+    extra = set(raw) - _PROFILE_ALLOWED_KEYS
+    if extra:
+        raise ProfileConfigError(f"profile {name!r}: unexpected key(s): {sorted(extra)}")
+
     def require(key: str, expected_type: type) -> object:
         if key not in raw:
             raise ProfileConfigError(f"profile {name!r}: missing required key {key!r}")
@@ -301,7 +363,14 @@ def _parse_profile(name: str, raw: dict) -> ProfileConfig:
             )
         return value
 
-    description = str(raw.get("description", ""))
+    if "description" in raw:
+        if not isinstance(raw["description"], str):
+            raise ProfileConfigError(
+                f"profile {name!r}: key 'description' must be str, got {type(raw['description']).__name__}"
+            )
+        description = raw["description"]
+    else:
+        description = ""
 
     windows_raw = require("windows", list)
     if len(windows_raw) < 2:
@@ -314,10 +383,6 @@ def _parse_profile(name: str, raw: dict) -> ProfileConfig:
     if windows != sorted(windows) or len(set(windows)) != len(windows):
         raise ProfileConfigError(f"profile {name!r}: windows must be strictly increasing, got {windows}")
 
-    warmup_repeats = require("warmup_repeats", int)
-    if warmup_repeats < 0:
-        raise ProfileConfigError(f"profile {name!r}: warmup_repeats must be >= 0")
-
     measured_repeats = require("measured_repeats", int)
     if measured_repeats < 1:
         raise ProfileConfigError(f"profile {name!r}: measured_repeats must be >= 1")
@@ -325,26 +390,57 @@ def _parse_profile(name: str, raw: dict) -> ProfileConfig:
     engines_raw = require("engines", list)
     if not engines_raw:
         raise ProfileConfigError(f"profile {name!r}: engines must not be empty")
-    engines: list[str] = []
-    for e in engines_raw:
-        if not isinstance(e, str) or not e:
-            raise ProfileConfigError(f"profile {name!r}: engines entries must be non-empty strings")
-        engines.append(e)
-    if len(set(engines)) != len(engines):
-        raise ProfileConfigError(f"profile {name!r}: engines must not contain duplicates, got {engines}")
+    engine_groups: list[EngineRunGroup] = []
+    seen_names: set[str] = set()
+    for idx, eg_raw in enumerate(engines_raw):
+        if not isinstance(eg_raw, dict):
+            raise ProfileConfigError(f"profile {name!r}: engines[{idx}] must be a table")
+        eg_extra = set(eg_raw) - _ENGINE_GROUP_ALLOWED_KEYS
+        if eg_extra:
+            raise ProfileConfigError(f"profile {name!r}: engines[{idx}] has unexpected key(s): {sorted(eg_extra)}")
+
+        def eg_require(key: str, expected_type: type, *, _idx: int = idx, _eg_raw: dict = eg_raw) -> object:
+            if key not in _eg_raw:
+                raise ProfileConfigError(f"profile {name!r}: engines[{_idx}] missing required key {key!r}")
+            value = _eg_raw[key]
+            if expected_type is int and isinstance(value, bool):
+                raise ProfileConfigError(f"profile {name!r}: engines[{_idx}] key {key!r} must be int, got bool")
+            if not isinstance(value, expected_type):
+                raise ProfileConfigError(
+                    f"profile {name!r}: engines[{_idx}] key {key!r} must be "
+                    f"{expected_type.__name__}, got {type(value).__name__}"
+                )
+            return value
+
+        eg_name = eg_require("name", str)
+        if not eg_name:
+            raise ProfileConfigError(f"profile {name!r}: engines[{idx}] name must be non-empty")
+        if eg_name in seen_names:
+            raise ProfileConfigError(f"profile {name!r}: engines must not contain duplicate name {eg_name!r}")
+        seen_names.add(eg_name)
+
+        eg_warmup = eg_require("warmup_repeats", int)
+        if eg_warmup < 0:
+            raise ProfileConfigError(f"profile {name!r}: engines[{idx}] warmup_repeats must be >= 0")
+
+        eg_model = eg_require("model", str)
+        if not eg_model:
+            raise ProfileConfigError(f"profile {name!r}: engines[{idx}] model must be non-empty")
+
+        eg_quantization = eg_require("quantization", str)
+        if not eg_quantization:
+            raise ProfileConfigError(f"profile {name!r}: engines[{idx}] quantization must be non-empty")
+
+        engine_groups.append(
+            EngineRunGroup(name=eg_name, warmup_repeats=eg_warmup, model=eg_model, quantization=eg_quantization)
+        )
 
     prompt = require("prompt", str)
     if not prompt:
         raise ProfileConfigError(f"profile {name!r}: prompt must be non-empty")
-    model = require("model", str)
-    if not model:
-        raise ProfileConfigError(f"profile {name!r}: model must be non-empty")
-    quantization = require("quantization", str)
-    if not quantization:
-        raise ProfileConfigError(f"profile {name!r}: quantization must be non-empty")
 
-    aggregation = str(raw.get("aggregation", "median"))
-    if aggregation not in AGGREGATION_METHODS:
+    aggregation = raw.get("aggregation", "median")
+    if not isinstance(aggregation, str) or aggregation not in AGGREGATION_METHODS:
         raise ProfileConfigError(
             f"profile {name!r}: aggregation must be one of {AGGREGATION_METHODS}, got {aggregation!r}"
         )
@@ -370,12 +466,9 @@ def _parse_profile(name: str, raw: dict) -> ProfileConfig:
         name=name,
         description=description,
         windows=tuple(windows),
-        warmup_repeats=warmup_repeats,
         measured_repeats=measured_repeats,
-        engines=tuple(engines),
+        engine_groups=tuple(engine_groups),
         prompt=prompt,
-        model=model,
-        quantization=quantization,
         aggregation=aggregation,
         trim=trim,
         requested_prompt_tokens=requested_prompt_tokens,
@@ -395,16 +488,28 @@ class AdapterRunResult:
     counts, discard samples, or compute verdicts. `native_ns`, when present,
     is surfaced as a diagnostic field alongside the harness-measured timing,
     never as a replacement for it.
+
+    `actual_model`/`actual_quantization` report the identity the adapter
+    actually invoked. They default to `None`, meaning "the requested
+    identity was invoked verbatim" -- the harness then records the
+    requested `model`/`quantization` it passed into `run()`. An adapter
+    that falls back to a different artifact than requested (a missing
+    checkpoint, a resolved default, ...) MUST set these so the raw
+    observation records what actually ran, not what was asked for.
     """
 
     actual_completion_tokens: int
     actual_prompt_tokens: int | None = None
     native_ns: int | None = None
     engine_version: str = "unknown"
+    actual_model: str | None = None
+    actual_quantization: str | None = None
 
 
 class EngineAdapter(Protocol):
-    def run(self, *, prompt: str, n_tokens: int, warmup: bool) -> AdapterRunResult: ...
+    def run(
+        self, *, prompt: str, n_tokens: int, warmup: bool, model: str, quantization: str
+    ) -> AdapterRunResult: ...
 
 
 # Populated by engine-adapter modules landed alongside each script migration
@@ -476,25 +581,26 @@ def run_profile(
 ) -> HarnessRunResult:
     """Execute `profile` against `adapters` and return every raw observation.
 
-    Run order is the requested schedule in list order: for each engine (in
-    `profile.engines` order), for each window (in `profile.windows` order —
-    `windows[0]` is the shared baseline, executed once and reused for every
-    comparison window that follows it), warmup runs first, then measured
-    runs. `order_index` is a single counter across the entire call, so a
-    later statistical pass (issue #714) can test for order/thermal bias
-    without another schema migration.
+    Run order is the requested schedule in list order: for each engine run
+    group (in `profile.engine_groups` order), for each window (in
+    `profile.windows` order — `windows[0]` is the shared baseline, executed
+    once and reused for every comparison window that follows it), that
+    engine's own warmup runs first, then measured runs. `order_index` is a
+    single counter across the entire call, so a later statistical pass
+    (issue #714) can test for order/thermal bias without another schema
+    migration.
     """
     resolved_git_sha = git_sha_value if git_sha_value is not None else git_sha(repo_root)
     resolved_hardware_id = hardware_id_value if hardware_id_value is not None else hardware_id()
     p_hash = prompt_hash(profile.prompt)
 
-    active: dict[str, EngineAdapter] = {}
+    active: dict[str, EngineRunGroup] = {}
     missing: list[str] = []
-    for engine in profile.engines:
-        if engine in adapters:
-            active[engine] = adapters[engine]
+    for group in profile.engine_groups:
+        if group.name in adapters:
+            active[group.name] = group
         else:
-            missing.append(engine)
+            missing.append(group.name)
     if missing and not allow_missing_engine:
         raise MissingEngineError(
             f"profile {profile.name!r} requires engine adapter(s) {missing} that are not registered; "
@@ -503,24 +609,34 @@ def run_profile(
 
     observations: list[Observation] = []
     order_index = 0
-    for engine_name in profile.engines:
-        adapter = active.get(engine_name)
-        if adapter is None:
+    for group in profile.engine_groups:
+        if group.name not in active:
             continue
+        adapter = adapters[group.name]
         for window in profile.windows:
-            for is_warmup, count in ((True, profile.warmup_repeats), (False, profile.measured_repeats)):
+            for is_warmup, count in ((True, group.warmup_repeats), (False, profile.measured_repeats)):
                 for run_index in range(1, count + 1):
                     t0 = clock()
-                    result = adapter.run(prompt=profile.prompt, n_tokens=window, warmup=is_warmup)
+                    result = adapter.run(
+                        prompt=profile.prompt,
+                        n_tokens=window,
+                        warmup=is_warmup,
+                        model=group.model,
+                        quantization=group.quantization,
+                    )
                     t1 = clock()
+                    actual_model = result.actual_model if result.actual_model is not None else group.model
+                    actual_quantization = (
+                        result.actual_quantization if result.actual_quantization is not None else group.quantization
+                    )
                     obs = Observation(
                         schema_version=SCHEMA_VERSION,
                         git_sha=resolved_git_sha,
                         profile=profile.name,
-                        engine=engine_name,
+                        engine=group.name,
                         engine_version=result.engine_version,
-                        model=profile.model,
-                        quantization=profile.quantization,
+                        model=actual_model,
+                        quantization=actual_quantization,
                         prompt_hash=p_hash,
                         requested_prompt_tokens=profile.requested_prompt_tokens,
                         actual_prompt_tokens=result.actual_prompt_tokens,
@@ -555,7 +671,17 @@ class SlopeResult:
     baseline_window: int
     window: int
     slope_tok_per_s: float
-    slope_ci95: float | None
+    # Legacy normal-approximation heuristic carried over verbatim from
+    # `bench_apples_precise.sh:118-143` (issue #813 explicitly defers
+    # statistical-methodology changes to a separately reviewed follow-up).
+    # This is NOT a coverage-valid 95% confidence interval for the slope:
+    # it treats the trimmed subset as an ordinary sample, applies a
+    # large-sample z=1.96 multiplier regardless of the retained sample
+    # size, and combines the two window CIs with relative-error
+    # (product/ratio) propagation rather than the sensitivity-coefficient
+    # propagation a difference-then-reciprocal statistic requires. Present
+    # it to a reader as a legacy spread heuristic only, never as "±95% CI".
+    slope_ci95_legacy: float | None
     baseline_n: int
     window_n: int
 
@@ -573,14 +699,21 @@ def _median_stat(values: list[float]) -> tuple[float, None]:
 
 
 def _trimmed_mean_stat(values: list[float], trim: int) -> tuple[float, float]:
+    """Trimmed mean + the legacy normal-approximation spread heuristic.
+
+    The second return value is `bench_apples_precise.sh`'s pre-existing
+    `1.96 * stdev / sqrt(n)` heuristic over the *retained* (trimmed)
+    subset, carried over verbatim rather than corrected -- see
+    `SlopeResult.slope_ci95_legacy` for why it is not a coverage-valid CI.
+    """
     ordered = sorted(values)
     trimmed = ordered[trim: len(ordered) - trim] if trim > 0 and len(ordered) > 2 * trim else ordered
     mean = statistics.mean(trimmed)
     if len(trimmed) > 1:
-        ci95 = 1.96 * (statistics.stdev(trimmed) / math.sqrt(len(trimmed)))
+        ci95_legacy = 1.96 * (statistics.stdev(trimmed) / math.sqrt(len(trimmed)))
     else:
-        ci95 = 0.0
-    return mean, ci95
+        ci95_legacy = 0.0
+    return mean, ci95_legacy
 
 
 def _window_stat(values: list[float], aggregation: str, trim: int) -> tuple[float, float | None]:
@@ -595,6 +728,12 @@ def aggregate(run_result: HarnessRunResult) -> list[SlopeResult]:
     `slope_tok_per_s = (window - baseline_window) / (T(window) - T(baseline))`,
     with `T` the profile's configured aggregation of measured (non-warmup)
     elapsed time. `baseline_window` is always `profile.windows[0]`.
+
+    `SlopeResult.slope_ci95_legacy`, when the profile uses `trimmed_mean`
+    aggregation, is the pre-existing `bench_apples_precise.sh` spread
+    heuristic propagated through the slope's product/ratio form -- a
+    legacy carry-over, not a corrected or coverage-valid CI (issue #813
+    defers statistical-methodology changes to a separate review).
     """
     profile = run_result.profile
     baseline_window = profile.windows[0]
@@ -603,31 +742,33 @@ def aggregate(run_result: HarnessRunResult) -> list[SlopeResult]:
         baseline_values = _measured_elapsed_seconds(run_result.observations, engine, baseline_window)
         if not baseline_values:
             continue
-        baseline_mean, baseline_ci = _window_stat(baseline_values, profile.aggregation, profile.trim)
+        baseline_mean, baseline_ci_legacy = _window_stat(baseline_values, profile.aggregation, profile.trim)
         for window in profile.windows[1:]:
             values = _measured_elapsed_seconds(run_result.observations, engine, window)
             if not values:
                 continue
-            window_mean, window_ci = _window_stat(values, profile.aggregation, profile.trim)
+            window_mean, window_ci_legacy = _window_stat(values, profile.aggregation, profile.trim)
             dt = window_mean - baseline_mean
             slope = (window - baseline_window) / dt if dt > 0 else float("nan")
-            slope_ci: float | None = None
+            slope_ci_legacy: float | None = None
             if (
                 profile.aggregation == "trimmed_mean"
-                and baseline_ci is not None
-                and window_ci is not None
+                and baseline_ci_legacy is not None
+                and window_ci_legacy is not None
                 and baseline_mean > 0
                 and window_mean > 0
                 and slope == slope  # not NaN
             ):
-                slope_ci = slope * math.sqrt((baseline_ci / baseline_mean) ** 2 + (window_ci / window_mean) ** 2)
+                slope_ci_legacy = slope * math.sqrt(
+                    (baseline_ci_legacy / baseline_mean) ** 2 + (window_ci_legacy / window_mean) ** 2
+                )
             results.append(
                 SlopeResult(
                     engine=engine,
                     baseline_window=baseline_window,
                     window=window,
                     slope_tok_per_s=slope,
-                    slope_ci95=slope_ci,
+                    slope_ci95_legacy=slope_ci_legacy,
                     baseline_n=len(baseline_values),
                     window_n=len(values),
                 )
@@ -658,11 +799,16 @@ def render_report(run_result: HarnessRunResult, slopes: list[SlopeResult]) -> st
     if not slopes:
         lines.append("  (no measured data)")
         return "\n".join(lines)
-    lines.append(f"  {'engine':<10}{'window':>8}{'slope tok/s':>13}{'±95% CI':>10}{'native tok/s':>14}{'n(base/win)':>13}")
-    lines.append("  " + "-" * 68)
+    # "legacy ±CI" -- the normal-approximation heuristic carried over from
+    # bench_apples_precise.sh, not a coverage-valid 95% CI. See
+    # SlopeResult.slope_ci95_legacy.
+    lines.append(
+        f"  {'engine':<10}{'window':>8}{'slope tok/s':>13}{'legacy ±CI':>11}{'native tok/s':>14}{'n(base/win)':>13}"
+    )
+    lines.append("  " + "-" * 69)
     for s in slopes:
         native = native_throughput(run_result, s.engine)
-        ci_str = f"{s.slope_ci95:10.1f}" if s.slope_ci95 is not None else f"{'—':>10}"
+        ci_str = f"{s.slope_ci95_legacy:11.1f}" if s.slope_ci95_legacy is not None else f"{'—':>11}"
         native_str = f"{native:14.1f}" if native is not None else f"{'—':>14}"
         lines.append(
             f"  {s.engine:<10}{s.window:>8}{s.slope_tok_per_s:13.1f}{ci_str}{native_str}"
--- a/scripts/bench_decode_harness.py
+++ b/scripts/bench_decode_harness.py
@@ -1,0 +1,744 @@
+#!/usr/bin/env python3
+"""Decode-benchmark harness core (issue #813).
+
+`scripts/` grew seven independent implementations of the same
+prompt/decode-benchmark shape (`bench_apples_to_apples.sh`,
+`bench_apples_precise.sh`, `bench_q4_apples.sh`, `bench_context_scaling.sh`,
+`bench_compare_1k.py`, plus two proven-dead scripts tracked separately), each
+owning its own copy of warmup policy, sample-count policy,
+aggregation/confidence-interval method, and output rendering. This module is
+the single place that owns all of that: configuration validation, warmup
+count, requested schedule, run order, monotonic wall timing, the raw
+observation schema, aggregation, confidence-interval calculation, and output
+rendering.
+
+Engine adapters (Lattice, Ollama, MLX) own only invocation and parsing: they
+implement `EngineAdapter.run()`, may surface an engine-native duration as a
+diagnostic field, and MUST NOT choose repeat counts, discard samples, or
+compute verdicts. No adapters ship in this module yet — this lands the
+harness core only ("no public command change", issue #813 step 1). The five
+live scripts above are migrated onto profiles defined in
+`bench_decode_profiles.toml` one script per PR (issue #813 step 2); until
+then `bench_decode_profiles.toml` carries no profile definitions and the
+`run` subcommand below has nothing registered in `ADAPTER_REGISTRY`, so an
+engine request always resolves as "missing" (see `--allow-missing-engine`).
+
+Raw JSONL observation schema (the contract; see `OBSERVATION_FIELDS` /
+`validate_observation`): schema version, git SHA, profile name, engine +
+version, model/quantization, prompt hash, requested and actual prompt/
+completion token counts, warmup/measured flag, run and order indices, the
+harness-measured elapsed nanoseconds, an optional engine-native nanosecond
+figure, a hardware identifier, and a timestamp. Reports render from this raw
+data; they are not the source of truth.
+
+Run with: python3 -m unittest tests/test_bench_decode_harness.py -v
+(the module itself is stdlib-only; no engine binaries or third-party
+packages are required to exercise it).
+"""
+
+from __future__ import annotations
+
+import argparse
+import dataclasses
+import hashlib
+import json
+import math
+import os
+import platform
+import statistics
+import subprocess
+import sys
+import time
+import tomllib
+from dataclasses import dataclass
+from datetime import datetime, timezone
+from pathlib import Path
+from typing import Callable, Mapping, Protocol
+
+SCHEMA_VERSION = 1
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+DEFAULT_PROFILES_FILE = REPO_ROOT / "scripts" / "bench_decode_profiles.toml"
+
+AGGREGATION_METHODS = ("median", "trimmed_mean")
+
+
+class ProfileConfigError(ValueError):
+    """A profile definition in `bench_decode_profiles.toml` is invalid."""
+
+
+class ObservationValidationError(ValueError):
+    """A raw observation does not conform to the schema contract."""
+
+
+class MissingEngineError(RuntimeError):
+    """A profile requested an engine with no adapter registered for it."""
+
+
+# --------------------------------------------------------------------------
+# Raw observation schema (the contract)
+# --------------------------------------------------------------------------
+
+# name -> expected type, or a tuple of types (nullable fields use (T, NoneType))
+OBSERVATION_FIELDS: dict[str, object] = {
+    "schema_version": int,
+    "git_sha": str,
+    "profile": str,
+    "engine": str,
+    "engine_version": str,
+    "model": str,
+    "quantization": str,
+    "prompt_hash": str,
+    "requested_prompt_tokens": (int, type(None)),
+    "actual_prompt_tokens": (int, type(None)),
+    "requested_completion_tokens": int,
+    "actual_completion_tokens": int,
+    "warmup": bool,
+    "run_index": int,
+    "order_index": int,
+    "elapsed_ns": int,
+    "engine_native_ns": (int, type(None)),
+    "hardware_id": str,
+    "timestamp": str,
+}
+
+_NON_EMPTY_STRING_FIELDS = (
+    "git_sha",
+    "profile",
+    "engine",
+    "engine_version",
+    "model",
+    "quantization",
+    "prompt_hash",
+    "hardware_id",
+)
+
+
+@dataclass(frozen=True)
+class Observation:
+    schema_version: int
+    git_sha: str
+    profile: str
+    engine: str
+    engine_version: str
+    model: str
+    quantization: str
+    prompt_hash: str
+    requested_prompt_tokens: int | None
+    actual_prompt_tokens: int | None
+    requested_completion_tokens: int
+    actual_completion_tokens: int
+    warmup: bool
+    run_index: int
+    order_index: int
+    elapsed_ns: int
+    engine_native_ns: int | None
+    hardware_id: str
+    timestamp: str
+
+    def to_dict(self) -> dict:
+        return dataclasses.asdict(self)
+
+
+def validate_observation(row: dict) -> None:
+    """Validate a raw observation dict against the schema contract.
+
+    Fail-closed: unknown fields, missing fields, wrong types, and
+    out-of-range values are all rejected. Raises `ObservationValidationError`
+    on any violation; returns None on success.
+    """
+    if not isinstance(row, dict):
+        raise ObservationValidationError(f"observation must be an object, got {type(row).__name__}")
+
+    extra = set(row) - set(OBSERVATION_FIELDS)
+    if extra:
+        raise ObservationValidationError(f"unexpected field(s): {sorted(extra)}")
+    missing = set(OBSERVATION_FIELDS) - set(row)
+    if missing:
+        raise ObservationValidationError(f"missing required field(s): {sorted(missing)}")
+
+    for name, expected in OBSERVATION_FIELDS.items():
+        _check_field_type(name, row[name], expected)
+
+    if row["schema_version"] != SCHEMA_VERSION:
+        raise ObservationValidationError(
+            f"schema_version {row['schema_version']!r} != supported {SCHEMA_VERSION}"
+        )
+    for name in _NON_EMPTY_STRING_FIELDS:
+        if not row[name]:
+            raise ObservationValidationError(f"field {name!r} must be non-empty")
+    if row["requested_completion_tokens"] < 0:
+        raise ObservationValidationError("requested_completion_tokens must be >= 0")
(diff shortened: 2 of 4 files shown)


ohdearquant:

## Review of PR #857 (round 1)

### Verdict: REJECT

One blocker, two majors, two mediums, and one minor remain. The harness is cleanly scoped to four files and its current CLI fails closed, but its step-1 profile/adapter contract cannot encode the engine-specific schedules and model metadata that issue #813 step 2 requires it to preserve.

### Prior review context

- No prior reviews, comments, or review threads were present when this round began.
- This round examined issue #813 step-1 fidelity, CI-series scope, fail-closed behavior, deterministic and mutation-sensitive tests, and statistical semantics at head `84f73cb1e34ea9834f80fb77755fd5d30ded2df2`.

### What this round examined

- Read the PR body, issue #813, all four changed files in full, ADR-058, the existing decode-benchmark scripts named by the issue, and the complete PR diff against main.
- Ran the claimed 52-test suite, Python compilation, direct CLI failure probes, `actionlint`, `typos`, diff hygiene checks, and GitHub check inspection.
- Mutation-tested trimming and warmup/measured bookkeeping in a disposable detached worktree, then removed it.
- Checked the stated slope/CI formula algebraically and against authoritative uncertainty references.

### Findings (6 total: 1 blocker, 2 majors, 2 mediums, 1 minor)

#### Blocker

1. [scripts/bench_decode_harness.py:248] The profile/adapter contract cannot represent the legacy workloads that step 2 must preserve.
   - **Current**: `warmup_repeats`, `model`, and `quantization` are profile-wide fields at lines 253 and 257-258. Every engine receives the same warmup count in `run_profile` at line 511. `EngineAdapter.run` at lines 406-407 receives only `prompt`, `n_tokens`, and `warmup`; nevertheless, lines 524-525 stamp the profile-wide model/quantization onto every observation.
   - **Issue**: issue #813's own inventory says only MLX gets the explicit warmup in `bench_apples_to_apples.sh`, `bench_q4_apples.sh`, and `bench_context_scaling.sh`, while step 2 requires byte-for-byte equivalent parameters. A single profile-wide warmup necessarily changes at least two engines' schedules. The model metadata is also not representable: `scripts/bench_q4_apples.sh:7-10` compares Lattice Q4 and MLX Q4 with an Ollama Q8 reference, but this schema can record only one quantization for all three. Because adapters are not told the requested model/quantization and cannot return the actual identity, a future adapter can invoke one artifact while the raw row claims another.
   - **Fix**: make the profile carry an ordered per-engine (or per-run-group) schedule with engine-specific warmup, model, and quantization. Pass requested identity into the adapter and/or require the adapter result to return the actual invoked identity; populate raw observations from verified actual values. Add a heterogeneous fake-adapter test that reproduces the MLX-only warmup and Q4/Q8 mix.
   - **Test**: one profile with Lattice(Q4, 0 warmups), MLX(Q4, 1 warmup), and Ollama(Q8, 0 warmups) must produce the exact call sequence and per-row identities.
   - **Confidence**: derived from the issue's explicit legacy table, the current scripts, and the profile/adapter/run path.

#### Majors

1. [scripts/bench_decode_harness.py:291] TOML validation silently accepts misspelled methodology keys.
   - **Current**: `_parse_profile` reads known fields but never rejects `set(raw) - allowed_fields`; `load_profiles_file` similarly ignores unexpected top-level keys. Optional fields then fall back to defaults.
   - **Issue**: a typo can silently alter the benchmark method that this consolidation exists to make consistent. Observed with `aggregtion = "trimmed_mean"` and `trmi = 1`: `_parse_profile` succeeded and returned `aggregation == "median"`, `trim == 0`. This is a fail-open configuration path.
   - **Fix**: define the exact allowed top-level and profile key sets, reject extras with `ProfileConfigError`, and validate `description` without coercing arbitrary values through `str()`.
   - **Test**: misspell each optional methodology key and assert a clean `ProfileConfigError` that names the unexpected key.
   - **Confidence**: observed with `python3 -B -c '... h._parse_profile(...) ...'` at this head.

2. [scripts/bench_decode_harness.py:575] The reported `slope_ci95` is not a defensible 95% CI for the computed slope.
   - **Current**: lines 575-583 compute `1.96 * stdev(trimmed) / sqrt(len(trimmed))`; line 623 then uses `slope * sqrt((ci1/T1)^2 + (ci2/T2)^2)`.
   - **Issue**: for `S = ΔN / (T2 - T1)`, first-order propagation for independent timing estimates is `CI(S) ≈ S * sqrt(CI(T1)^2 + CI(T2)^2) / (T2 - T1)`; the current denominators `T1` and `T2` are the relative-error form for a product/ratio, not a difference followed by a reciprocal. The component CI is also calculated as though the retained central values were an ordinary untrimmed sample, uses a large-sample 1.96 multiplier at arbitrarily small retained n, and reports `0.0` when only one sample remains. NIST's uncertainty guidance requires sensitivity-coefficient/root-sum-square propagation and a t coverage factor for finite degrees of freedom ([NIST 2.5.7](https://www.itl.nist.gov/div898/handbook/mpc/section5/mpc57.htm)); SciPy exposes distinct trimmed-mean standard-error/CI procedures rather than treating the retained subset as an ordinary mean ([SciPy trimmed statistics](https://docs.scipy.org/doc/scipy/reference/stats.mstats.html)).
   - **Scope note**: this matches the heuristic in `scripts/bench_apples_precise.sh:118-143`, so changing methodology inside this PR would conflict with #813's explicit deferral. That does not make the generic `±95% CI` label correct. ADR-058 is superseded and concerns Criterion A/B change intervals and lower-bound decisions; it does not validate this decode-slope formula, though its measurement discipline reinforces that an uncertainty label must describe the statistic actually bounded. The extraction must explicitly encode and document this as a legacy approximation (with exact parity tests), or the issue must be amended before a statistically corrected method is introduced.
   - **Fix**: resolve the spec conflict explicitly. For strict extraction, name/document/render this as the legacy normal-approximation heuristic and disallow claims that it is a coverage-valid CI. Any corrected CI should land in the separately reviewed statistics change required by #813.
   - **Test**: assert exact legacy values for a hand-computed asymmetric sample now, and add the corrected-method tests only in the later methodology PR.
   - **Confidence**: derived algebraically from the implemented slope and confirmed against the current legacy script and cited uncertainty references.

#### Mediums

1. [tests/test_bench_decode_harness.py:441] The trimmed-mean test is not mutation-sensitive.
   - **Current**: it asserts only that the CI is non-null and the slope is positive.
   - **Issue**: observed mutation `trimmed = ordered` (removing trimming entirely) left all 52 tests green. The chosen symmetric data also make the base trimmed and untrimmed means identical. This contradicts the PR body's claim that trimmed aggregation is checked against hand-computed values.
   - **Fix**: use asymmetric outliers for which trimming changes the result, and assert exact trimmed mean, slope, and legacy CI estimate. Retest by removing trimming and require failure.
   - **Confidence**: observed in a disposable worktree; `Ran 52 tests ... OK` under the mutation.

2. [.github/workflows/ci.yml:479] The new CI job can silently pass after discovering zero tests.
   - **Current**: the job uses discovery with an exact filename pattern.
   - **Issue**: observed `python3 -m unittest discover -s tests -p 'does_not_exist.py' -v` returning exit 0 with `Ran 0 tests ... OK`. A deletion/rename/path drift therefore turns the only harness check into a green no-op, contrary to the repo's fail-closed CI rule.
   - **Fix**: invoke the test file directly (`python3 tests/test_bench_decode_harness.py -v`) so a missing file fails, or add an explicit discovered-test-count assertion.
   - **Test**: rename/remove the test file in a disposable worktree and require the workflow command to exit nonzero.
   - **Confidence**: observed locally at this head.

#### Minor

1. [AGENTS.md:19] The PR-body attribution does not use the repository's required PR-description wording.
   - **Current**: the body begins `> _PR authored by Claude ..._`; the repository template and policy require `> _PR description authored by ..._`.
   - **Fix**: change the first line to the required `PR description authored by Claude (Anthropic agent) ...` form.
   - **Confidence**: observed in `gh pr view 857` and checked against `.github/PULL_REQUEST_TEMPLATE.md:6`.

### Gates and evidence

- **Unit tests**: observed `python3 -m unittest discover -s tests -p 'test_bench_decode_harness.py' -v` → `Ran 52 tests in 1.992s; OK`.
- **Mutation checks**:
  - removing trimming entirely → suite incorrectly stayed green (finding above);
  - marking warmups as measured → suite failed `test_warmup_flag_and_run_index_reset_per_group` with exit 1.
- **Fail-closed CLI**: direct subprocess probes for unknown profile, missing engine without `--allow-missing-engine`, malformed TOML, missing profiles file, and missing JSONL file all exited 1 with a single clean `FAIL:` message and no traceback. Both `FileNotFoundError`/OSError-facing file paths are wrapped.
- **Static/local gates**: `actionlint .github/workflows/*.yml`, `typos` on all four changed files, `python3 -m py_compile`, and `git diff --check origin/main...HEAD` all passed.
- **Series constraint**: the diff changes exactly one file under `.github/`—15 added lines in `.github/workflows/ci.yml` for the ungated job. No gating command is piped through `tee`, and ordinary unittest failures propagate; the zero-test state is the exception in finding 2.
- **GitHub CI at head**: `bench-decode-harness-tests`, `actionlint`, `typos`, MSRV, Linux CI, Linux feature-matrix, semver, and the other completed checks are green. At the final refresh, macOS CI, macOS feature-matrix, and bench-compile were still pending.
- **Branch freshness**: not merge-ready. After fetching `origin/main`, `git rev-list --left-right --count HEAD...origin/main` returned `1 1`, and GitHub reports `mergeStateStatus: BEHIND`.
- **Publication hygiene**: all four changed files are newline-terminated; `git diff --check`, `typos`, and Python compilation are clean; new source/test/profile files are mode 100644; no sibling issue or legacy-script file is modified.

- The diff is limited to the step-1 harness, empty TOML scaffold, deterministic tests, and one CI job; no concrete adapter, wrapper migration, public-command replacement, or sibling #814/#815/#817/#818 work landed.
- The observation validator rejects missing/extra fields, bool-as-int confusion, negative counts/timings, invalid schema versions, and malformed timestamps.
- `run_profile` uses the injectable nanosecond clock around adapter invocation, records requested versus actual token counts separately, increments a global order index, and excludes warmups from aggregation.
- Missing adapters fail by default, and `--allow-missing-engine` skips them without fabricating observations.
- The dedicated CI command is stdlib-only and requires no engine binary or model weights.

### What I did NOT check

- Per instruction, I did not launch any Rust build, Cargo test, engine binary, model download, or live performance run.
- No concrete adapter or migrated profile exists, so I could not validate real subprocess parsing or old-vs-new live output equivalence.
- I did not push, commit, update the branch, post this review, or modify source files in the review worktree.

Another round is required after the profile/adapter contract is redesigned. Focus first on proving exact representability of each legacy script's per-engine schedule and metadata; then rerun mutation checks with exact aggregation expectations and re-audit the CI command after rebasing onto main.

Domain utility: HIGH — the Lattice review workflow's contract-drift, mutation-sensitivity, CI fail-closed, and PR-attribution checks directly exposed the blocking and gate-level defects.

Verdict: REJECT

ohdearquant:

## Review of PR #857 (round 2)

### Verdict: REJECT

One blocker remains at head `7c0691506374a4b4117b2dc89398397662daf106`. Five of the six round-one findings are fixed, and the identity half of the blocker is fixed, but the profile still cannot reproduce the legacy warmup schedules that issue #813 requires step 2 to preserve.

### Prior review context

- Round 1 rejected head `84f73cb1e34ea9834f80fb77755fd5d30ded2df2` with 1 blocker, 2 majors, 2 mediums, and 1 minor.
- This round read the round-one comment first, reviewed all four changed files in full, re-read issue #813 and all five legacy scripts, ran the direct 70-test suite and static gates, directly probed malformed TOML, and repeated both mutations in detached disposable worktrees.

### Round-one finding status

| Round-one finding | Round-two result |
|---|---|
| Blocker: per-engine schedule and actual identity | **Partially fixed; blocker remains.** Per-engine model/quantization and actual-identity reporting are fixed, but warmup placement and token budgets are not representable. |
| Major: fail-closed TOML keys and description type | **Fixed.** Direct probes at all three key scopes raised `ProfileConfigError` naming the bad key; integer `description` was rejected. |
| Major: misleading `slope_ci95` semantics | **Fixed.** The field is `slope_ci95_legacy`, the report says `legacy ±CI`, and changed code/docs explicitly disclaim coverage-valid confidence-interval semantics. |
| Medium: trim test was mutation-insensitive | **Fixed.** The trim-removal mutation now fails two exact-value tests. |
| Medium: discovery could pass with zero tests | **Fixed.** CI invokes the file directly; a nonexistent file exits 2. |
| Minor: PR attribution wording | **Fixed.** The PR body begins with the required `PR description authored by ...` form. |

### Findings (1 total: 1 blocker)

#### Blocker

1. [`scripts/bench_decode_harness.py:612`](scripts/bench_decode_harness.py#L612) Warmups are still scheduled once per measured window at the measured window's token count, so the legacy schedules are not exactly representable.
   - **Current**: `run_profile` loops `engine group -> window -> (warmup, measured)` at lines 612-625. With `windows=[32, 256]` and `warmup_repeats=1`, the adapter therefore receives a 32-token warmup and a 256-token warmup. The claimed regression test locks in precisely that behavior at [`tests/test_bench_decode_harness.py:631`](tests/test_bench_decode_harness.py#L631): MLX is expected to warm at 32 and again at 256.
   - **Issue**: this is not the schedule in the inventory the test says it reproduces. `bench_q4_apples.sh:86-90` and `bench_apples_to_apples.sh:90-94` each warm MLX **once at 8 tokens before** the `(N1, N2)` loop. `bench_context_scaling.sh:146-172` warms MLX **once at 4 tokens before** its baseline/comparison loops. `bench_apples_precise.sh:35-41`, `:56-63`, and `:92-99` warm every engine twice at `N2=512` **once before** the two measured windows; the current schema would instead perform four warmups, two at 64 and two at 512. `bench_compare_1k.py:192-220` and `:324-345` likewise use separate 4-token pre-measurement warmups (and distinct warmup prompts) for Ollama and MLX. A per-engine count alone does not encode warmup placement, completion-token budget, or warmup prompt.
   - **Effect**: step-2 migrations cannot preserve the old parameters and call order. The harness would introduce extra warmups and different warmup workloads, contradicting issue #813's byte-for-byte parameter-equivalence gate and the PR body's claim that all five schedules are representable.
   - **Fix**: make warmup an explicit pre-window schedule with at least repeats, completion-token budget, and prompt/default-profile-prompt semantics, or model the profile as an ordered operation list. Execute those warmup operations once before the measured windows. Change the heterogeneous test's MLX sequence to one `(8, True, ...)` call followed by the measured 32/256 calls, and add exact schedule tests for the precise (`2 × 512` once) and context/agentic (`1 × 4` once) shapes.
   - **Test**: the exact-call-sequence tests should compare one flattened global schedule, including warmup token budgets and placement, against each legacy script's schedule. The existing per-adapter assertions are useful once their expected sequences match the scripts.
   - **Confidence**: derived from the current loop nesting and observed legacy invocation order. The current test itself confirms the mismatching 32/256 warmups.

### Gates and observed evidence

- `python3 tests/test_bench_decode_harness.py -v` -> `Ran 70 tests in 0.921s`, `OK`.
- `python3 -m py_compile scripts/bench_decode_harness.py tests/test_bench_decode_harness.py` -> clean.
- Direct TOML probes:
  - top-level `schema_verzion` -> `ProfileConfigError ... unexpected top-level key(s): ['schema_verzion']`;
  - profile `aggregtion` -> `ProfileConfigError ... unexpected key(s): ['aggregtion']`;
  - engine-group `warmup_repeat` -> `ProfileConfigError ... engines[0] has unexpected key(s): ['warmup_repeat']`;
  - `description = 123` -> `ProfileConfigError ... key 'description' must be str, got int`.
- Trim mutation in detached worktree: changed `_trimmed_mean_stat` to `trimmed = ordered`; suite exited 1. Both exact tests failed (`4.52 != 3.2`; slope `9.491525... != 10.181818...`).
- Warmup-as-measured mutation in detached worktree: changed the recorded observation flag to `warmup=False`; suite exited 1 at `test_warmup_flag_and_run_index_reset_per_group` (`[] != [1, 2]`). Both disposable worktrees were removed afterward.
- `actionlint .github/workflows/ci.yml`, `typos` on all four changed files, `git diff --check origin/main...HEAD`, and Python compilation all exited 0.
- The new CI job at `.github/workflows/ci.yml:471-480` directly runs `python3 tests/test_bench_decode_harness.py -v`; no unittest-discovery path or `tee` occurs in the added diff. A direct missing-file probe exited 2. Python's official `unittest` documentation also supports executing a test file directly when appropriate: https://docs.python.org/3.11/library/unittest.html#command-line-interface.
- Exactly one `.github` file differs from main: `.github/workflows/ci.yml` (15 additions). The unchanged workflow has older `tee` pipelines in unrelated Metal gates; no such pipeline was added or modified by this PR.
- Branch freshness is fixed: after fetching `origin/main`, `git rev-list --left-right --count HEAD...origin/main` returned `3 0`; `HEAD` and `origin/ci/813-bench-harness` both resolve to the reviewed SHA.
- Publication hygiene is clean across all four changed files: mode `100644`, final byte LF, no absolute developer/workspace paths, internal-operation markers, CJK text, or review artifacts; `git diff --check` and `typos` are clean.
- GitHub's harness, actionlint, typos, Linux CI, feature-matrix, semver, cargo-deny, and completed related checks were green at this head. At the final refresh, macOS CI, macOS feature-matrix, bench-compile, and MSRV were still queued; GitHub reported `mergeStateStatus: BLOCKED`, not `BEHIND`.

- `EngineRunGroup` now carries ordered per-engine name, warmup count, model, and quantization; profile parsing preserves order and rejects duplicate names.
- Requested model/quantization reaches `EngineAdapter.run`; `AdapterRunResult.actual_model` and `actual_quantization` override requested identity in raw observations. The heterogeneous test asserts per-engine calls, observation engine order, and per-row identities.
- The slope heuristic's math is unchanged from the legacy implementation and exact-value tests lock both the retained-subset statistic and propagated legacy value. No changed doc, field, or rendered label presents it as a coverage-valid 95% CI.
- The direct CI test-file invocation is fail-closed on deletion and contains no output-masking pipeline.
- The PR remains scoped to the harness, empty profile scaffold, deterministic tests, and one CI job.

### What I did NOT check

- Per instruction, I did not run Cargo, build Rust code, launch an engine, download a model, or perform a live benchmark.
- No concrete adapter or migrated profile exists in this PR, so real subprocess parsing and live old-vs-new output equivalence remain for step 2.
- I did not wait for the four queued GitHub jobs to complete. Their eventual result does not resolve the schedule-contract blocker.
- I did not push, commit, post the review, or edit source files in the review worktree.

Another round is required after the warmup schedule is made explicitly representable. Re-run exact flattened schedule tests for all five inventory shapes; if those pass and the branch remains current/green, no further broad review round should be necessary.

Domain utility: HIGH — the Lattice review workflow's contract-first and mutation-sensitivity rules exposed that the new heterogeneous test locks in a schedule different from the legacy scripts it claims to reproduce.

Verdict: REJECT

ohdearquant:

## Review of PR #857 (round 3)

### Verdict: REJECT

One blocker remains at head `17df48c76a21b40319cd720183e66e5efe44deec`. The round-2 warmup-placement defect itself is fixed and mutation-sensitive, but the closed profile model still cannot encode `bench_compare_1k.py`'s measured schedule exactly. The PR's stronger claim that all five legacy schedules fit the model is therefore false.

### Prior review context

- Round 1 rejected the initial harness contract with 1 blocker, 2 majors, 2 mediums, and 1 minor.
- Round 2 confirmed every prior item fixed except one blocker: warmup placement, completion-token budget, and prompt were not exactly representable.
- This round read the round-2 comment first and checked the remaining blocker against the current implementation and all five actual legacy scripts.

### What this round examined

- Read the PR/issue prose, all four changed files, the complete diff, and the five legacy benchmark scripts.
- Compared each legacy call schedule with `EngineRunGroup`, `ProfileConfig`, `EngineAdapter`, and `run_profile`, including engine ordering, warmup placement/budget/prompt, measured-window ordering, and per-engine timing shape.
- Ran the 81-test Python suite, direct fail-closed TOML typo probes, Python compilation, the requested disposable-worktree mutation, `actionlint`, `typos`, diff/publication hygiene, branch-freshness checks, and GitHub check inspection.

### Findings (1 total: 1 blocker)

#### Blockers

1. [`scripts/bench_decode_harness.py:293`](scripts/bench_decode_harness.py#L293) `bench_compare_1k.py` disproves the claim that measured windows/repeats are uniform and exactly representable.
   - **Current**: `ProfileConfig` has one profile-wide `windows`, `measured_repeats`, and measured `prompt` (`scripts/bench_decode_harness.py:306-316`). `run_profile` then executes every engine as all repeats of window 1 followed by all repeats of window 2 (`scripts/bench_decode_harness.py:756-767`). `AdapterRunResult` exposes only one optional native duration for a call (`scripts/bench_decode_harness.py:548-577`).
   - **Issue**: the actual agentic script has three different measured shapes. Lattice executes all 1-token runs and then all 100-token runs (`scripts/bench_compare_1k.py:140-141`). Ollama executes only one 100-token generation per repeat and obtains both prompt-evaluation and decode timing components from that same response (`scripts/bench_compare_1k.py:212-238`). MLX explicitly alternates a 1-token call and a 100-token call inside every repeat (`scripts/bench_compare_1k.py:333-352`). No choice of the current profile-wide fields can produce all three sequences: the harness necessarily emits `[1 × R, 100 × R]` for every engine, while Ollama needs `[100 × R]` with two timing components and MLX needs `[(1, 100) × R]`. An adapter cannot repair this without taking schedule/aggregation ownership away from the harness or returning cached/fabricated harness timings.
   - **Test gap**: the only `bench_compare_1k.py` regression test checks one custom warmup prompt hash (`tests/test_bench_decode_harness.py:847-863`); it never asserts the script's flattened measured sequence. The heterogeneous Q4 test also says it reproduces legacy global call order (`tests/test_bench_decode_harness.py:692-699`) but asserts Lattice → MLX → Ollama (`tests/test_bench_decode_harness.py:758-764`), whereas `bench_q4_apples.sh` executes Lattice → Ollama → MLX (`scripts/bench_q4_apples.sh:38-94`). That Q4 order is representable by reordering groups, but it is not currently tested as claimed.
   - **Effect**: issue #813 step 2 cannot preserve `bench_compare_1k.py`'s requested schedule/run order and raw-observation equivalence. The round-2 blocker therefore remains despite the warmup sub-fix.
   - **Fix**: model the measured schedule explicitly enough to represent per-engine operation order and multi-component timing, either with an ordered operation plan or an equivalently expressive validated per-engine measured plan. Then add global flattened-sequence tests for every legacy script, not only per-adapter warmup assertions.
   - **Test**: for one agentic context, assert the complete Lattice → Ollama → MLX stream: Lattice's batched 1-token then 100-token windows; Ollama's 4-token `"hi"` warmup then one 100-token call per repeat with both timing components; and MLX's 4-token `BASE` warmup then alternating `(1, 100)` measured calls per repeat. Also correct the Q4 test to assert the legacy Lattice → Ollama → MLX global order.
   - **Confidence**: derived from the current loop nesting and observed directly in the checked-in legacy scripts. Strongest rebuttal considered: an adapter could internally reorder/cache calls, but that would violate the stated adapter boundary and would make the harness-measured elapsed time describe cache access rather than the legacy operation.

### Gates and observed evidence

- **Full Python suite**: `python3 tests/test_bench_decode_harness.py -v` ran 81 tests and passed. After mutation-worktree removal, the same command again ran 81 tests and passed.
- **Requested mutation**: in a detached disposable worktree, restored the old per-window warmup nesting while retaining the custom warmup-prompt behavior. The suite exited 1 with exactly 5 failures: context-scaling schedule, heterogeneous schedule, precise schedule, observation count, and warmup placement. The disposable worktree was removed; no source in this review worktree was edited.
- **TOML fail-closed probes**: direct `_parse_profile` probes of `warmup_token` and `warmup_promt` each raised `ProfileConfigError` naming the exact unexpected key. Both new keys are present in `_ENGINE_GROUP_ALLOWED_KEYS` at `scripts/bench_decode_harness.py:337-339`.
- **Defaults/docs**: `warmup_prompt` defaults to `profile.prompt` in `run_profile` (`scripts/bench_decode_harness.py:740-744`); `warmup_tokens` is required when repeats are positive and rejected at zero (`scripts/bench_decode_harness.py:455-469`). The TOML schema documents these semantics at `scripts/bench_decode_profiles.toml:40-44`.
- **Static gates**: Python compilation, `actionlint .github/workflows/ci.yml`, `typos` on all four changed files, and `git diff --check origin/main...HEAD` all exited 0.
- **Workflow scope**: the round-3 fix commit changes only the harness, TOML scaffold, and tests; it does not touch `.github/workflows/ci.yml`. The branch's intervening main merge brought unrelated current-main workflow changes, so a raw round-2-head-to-current-head comparison is not empty, but `HEAD^..HEAD` is. The PR diff still adds only the existing direct-invocation harness job at `.github/workflows/ci.yml:489-503`.
- **Branch freshness**: after fetching, `HEAD` and `origin/ci/813-bench-harness` both resolve to the reviewed SHA; `git rev-list --left-right --count HEAD...origin/main` returned `5 0`, and `origin/main` is an ancestor of HEAD.
- **GitHub CI at final inspection**: the harness job, actionlint, typos, Linux CI, ARM CI, Linux feature matrix, semver, cargo-deny, and other completed checks were green. MSRV, bench-compile, macOS feature matrix, and macOS CI were still queued; GitHub reported the PR mergeable but `mergeStateStatus: BLOCKED`.
- **Publication hygiene**: all four PR files are newline-terminated; new files are mode `100644`; hygiene grep found no developer absolute paths, orchestration markers, or CJK text. The three local PID files are untracked and absent from the PR diff.

- The actual round-2 warmup defect is fixed: each engine group's complete warmup batch runs once before that group's measured windows, at `warmup_tokens`, with `warmup_prompt` or the documented profile-prompt default.
- The checked-in exact tests correctly cover the requested Q4 warmup budget/placement, precise twice-at-512 batch, context-scaling once-at-4 batch, and default/custom warmup prompt behavior.
- `bench_apples_precise.sh` and `bench_context_scaling.sh` warmup shapes fit the new pre-window batch model. `bench_q4_apples.sh`'s warmup shape also fits once engine groups are ordered to match the script.
- Unknown TOML keys fail closed at top-level, profile, and engine-group scopes; the two new fields have type/range/relationship validation.
- The requested mutation is guarded by exactly the five failures claimed by the author, and the clean tree restores to 81/81 green.

### What I did NOT check

- Per instruction, I did not run Cargo, compile Rust, launch an engine, download a model, or perform a live performance run.
- No concrete adapters or live profiles exist in this PR, so I did not test subprocess parsing or live old-vs-new output equivalence.
- I did not wait for the four queued GitHub jobs because their eventual result cannot resolve the schedule-contract blocker.
- I did not push, commit, post a GitHub comment/review, or modify source files in the review worktree.

One more focused round is required after the measured-schedule contract is made expressive enough. It should assert complete global flattened sequences for all five scripts, especially `bench_compare_1k.py`; if that passes with the branch current and CI green, no further broad review round should be necessary.

Domain utility: HIGH — the Lattice review workflow's contract-first, actual-script, and mutation-sensitivity checks distinguished a correct warmup sub-fix from the still-false all-schedules-representable claim.

Verdict: REJECT

ohdearquant:

## Review of PR #857 (round 4)

### Verdict: REJECT

One blocker and one major remain at head `1a109910314dabff86a4216e1a570c065067366a`. The per-engine call-plan and multi-yield timing changes fix the measured-order/timing portion of round 3 and are mutation-sensitive, but `bench_compare_1k.py`'s per-engine measured prompts are still not representable by the profile contract. The new nesting also remains fail-open when `measured_calls[*].yields_windows` does not match `profile.windows`.

### Prior review context

- Round 1 found that the initial profile/adapter contract could not encode all engine-specific legacy schedules and model identities.
- Round 2 confirmed the identity fixes but found that the warmup schedule was still not exactly representable.
- Round 3 confirmed the warmup model was fixed, but rejected head `17df48c76a21b40319cd720183e66e5efe44deec` because the measured schedule still assumed uniform profile windows and could not encode `bench_compare_1k.py` exactly.
- This round re-read the round-3 review first and tested the per-engine measured-call plan, multi-yield component timing, all five legacy scripts, new TOML nesting, both requested mutations, publication hygiene, branch freshness, CI, and PR prose.

### What this round examined

- Compared `EngineRunGroup`, `MeasuredCall`, `AdapterRunResult`, and `run_profile` against the complete schedules and prompts in all five live legacy scripts.
- Read all four changed files in full plus the five legacy scripts and the adjacent Lattice benchmark prompt builder.
- Ran the 108-test Python suite twice on the clean head, direct malformed-profile probes, both requested mutations in a detached disposable worktree, Python compilation, actionlint, typos, diff/publication hygiene, branch-freshness checks, and live GitHub check inspection.

### Findings (2 total: 1 blocker, 1 major)

#### Blockers

1. [`scripts/bench_decode_harness.py:385`](scripts/bench_decode_harness.py#L385) `bench_compare_1k.py`'s measured prompts are still not representable, and the claimed complete flattened test masks that gap.
   - **Current**: `EngineRunGroup` has a `warmup_prompt` override but no measured-prompt override (`scripts/bench_decode_harness.py:385-392`). Both measured-order branches pass the same `profile.prompt` to every adapter (`scripts/bench_decode_harness.py:1031-1052`), and the harness records one hash computed from that profile-wide string (`scripts/bench_decode_harness.py:860`). The commented agentic profile acknowledges the mismatch by using `prompt = "<ctx-padded prompt, built per-engine at the adapter level>"` (`scripts/bench_decode_profiles.toml:117-122`).
   - **Issue**: the legacy script explicitly says Lattice/MLX use tokenizer-padded prompts while Ollama uses a different character-count heuristic (`scripts/bench_compare_1k.py:13-16`). Ollama builds `_make_prompt(ctx)` (`scripts/bench_compare_1k.py:189-190,383-386`), while MLX builds `_build_padded_prompt(ctx, tokenizer)` (`scripts/bench_compare_1k.py:267-282,317-318`) and Lattice passes `BENCH_PROMPT_TOKENS` to a tokenizer-driven builder (`crates/inference/src/bin/bench_decode_ab.rs:121-140`). A profile has no way to send these different measured strings through the current adapter boundary. Letting each adapter ignore the supplied placeholder and synthesize its own prompt is not an exact encoding: it moves requested-prompt ownership out of the harness and leaves every raw `prompt_hash` describing the placeholder rather than the invoked prompt.
   - **Test gap**: `test_agentic_style_global_flattened_sequence_lattice_ollama_mlx` claims a complete stream (`tests/test_bench_decode_harness.py:1138-1147`), but `_FakeAdapter.calls` omits the prompt (`tests/test_bench_decode_harness.py:96-108`), the test asserts per-adapter token/identity tuples and only the global engine-name stream (`tests/test_bench_decode_harness.py:1192-1240`), and it never asserts the three actual prompt arguments or their observation hashes. A direct capture-adapter probe observed the same measured string sent to Lattice, Ollama, and MLX, despite the legacy script requiring two different prompt-building methods.
   - **Effect**: issue #813 step 2 still cannot preserve all `bench_compare_1k.py` parameters and raw observations exactly. The round-3 blocker therefore remains, narrowed from measured call shape to measured prompt identity.
   - **Fix**: make the measured prompt an explicit per-engine input (for example, an optional `measured_prompt` on `EngineRunGroup`, defaulting to `profile.prompt`), pass that exact string to the adapter, and hash the exact string passed. If prompt construction must be declarative rather than materialized, model and validate that construction in the harness; do not use an adapter-interpreted placeholder. Extend the agentic test's single flattened expected stream to include engine, physical-call identity, prompt/hash, token budget, warmup flag, yielded window, run index, model, and quantization for every row/call.
   - **Test**: the complete Lattice → Ollama → MLX test must fail if Ollama receives the Lattice/MLX measured prompt, if either warmup receives the measured prompt, or if any recorded prompt hash differs from the exact invoked string.
   - **Confidence**: derived from the closed dataclass fields and both measured call sites, and observed with a capture adapter. Strongest rebuttal considered: adapters can build their own prompts, but that contradicts the stated harness-owned request boundary and provably makes the harness-computed raw prompt hash inaccurate.

#### Majors

1. [`scripts/bench_decode_harness.py:607`](scripts/bench_decode_harness.py#L607) Explicit measured plans do not fail closed when yielded windows disagree with `profile.windows`.
   - **Current**: parsing validates each call's keys, positive values, and duplicates only within that one `yields_windows` list (`scripts/bench_decode_harness.py:607-658`). It never compares the flattened yielded windows across the plan with the profile's declared `windows`. Aggregation then silently skips a missing baseline or comparison window (`scripts/bench_decode_harness.py:1138-1148`).
   - **Issue**: direct `_parse_profile` probes accepted all three malformed plans: `windows=[32,256]` with only `yields_windows=[32]`; the same profile yielding undeclared window `999`; and two calls that yield window `32` twice across calls. These cases respectively erase a required slope, create raw observations outside the declared report contract, or change the effective sample count and aggregation weight. This violates the requested fail-closed relationship validation at the new nesting level.
   - **Fix**: after parsing an explicit plan, require its flattened `yields_windows` to match `profile.windows` exactly once each (and in the intended order, if order is part of the contract). Raise `ProfileConfigError` naming `measured_calls`/`yields_windows`, the engine index, and the missing, extra, or duplicate windows.
   - **Test**: add separate missing-window, undeclared-window, and cross-call-duplicate tests. The four direct typo probes for `measured_calls`, `measured_order`, `n_tokens`, and `yields_windows` already raise `ProfileConfigError` and name the exact misspelled key.
   - **Confidence**: observed by direct parser probes. Strongest rebuttal considered: every current legacy example is valid, but this is a public profile parser whose stated contract is fail-closed; accepting malformed plans silently changes or suppresses benchmark output.

### Gates and observed evidence

- **Clean Python suite**: `python3 tests/test_bench_decode_harness.py -v` ran 108 tests and passed. After both disposable-worktree mutations were removed, a second clean run again reported `Ran 108 tests ... OK`.
- **Mutation A — ignore `measured_order`**: forcing the window-major branch in a detached disposable worktree produced exactly 2 failures: `test_measured_order_repeat_major_alternates_calls_per_repeat` and `test_agentic_style_global_flattened_sequence_lattice_ollama_mlx`.
- **Mutation B — disable multi-yield `component_ns`**: forcing all calls through the single-yield wall-clock path keyed by `n_tokens` produced exactly 4 failures: the agentic global test plus the three multi-yield component tests (uses component timing, missing map, missing window entry).
- **Multi-yield semantics**: for `len(yields_windows) > 1`, `elapsed_ns` is read from `result.component_ns[window]` (`scripts/bench_decode_harness.py:980-1001`), never split from wall time. Missing `component_ns` and missing yielded-window entries raise `AdapterContractError`; both tests pass.
- **Legacy ordering**: the representable call-order/budget portion is now correct. Lattice is window-major, Ollama performs one 100-token call per repeat and yields both observations, MLX is repeat-major, the Q4 exact schedule test uses Lattice → Ollama → MLX, and the precise/context-scaling global tests are present and green.
- **TOML typo probes**: misspelled `measured_call`, `measured_oder`, `n_token`, and `yield_windows` each raised `ProfileConfigError` naming that exact unexpected key. The relationship probes described in Major 1 were accepted.
- **Static gates**: Python compilation, `actionlint .github/workflows/ci.yml`, `typos` on all four PR files, and `git diff --check origin/main...HEAD` all exited 0.
- **Workflow scope**: `.github/workflows/ci.yml` is unchanged from the round-3 head to this head. Against current `origin/main`, the PR's entire `.github` diff is still only the 15-line `bench-decode-harness-tests` job at `.github/workflows/ci.yml:489-503`.
- **Branch freshness**: after fetching, `origin/main` (`b181422eae1727252ad4879126c4c3b3a5918c0f`) is an ancestor of the reviewed head; `git rev-list --left-right --count HEAD...origin/main` returned `6 0`.
- **GitHub CI at inspection**: GitHub reported the exact reviewed head, `MERGEABLE`, `mergeStateStatus: BLOCKED`, and `reviewDecision: REVIEW_REQUIRED`. The harness job, actionlint, typos, Ubuntu CI, ARM CI, feature-matrix Ubuntu, semver, cargo-deny, and completed gates were green; macOS CI, macOS feature matrix, bench-compile, and MSRV were still queued.
- **Publication hygiene**: all four PR files are newline-terminated; the three new files are mode `100644`; no workspace paths, orchestration markers, CJK text, PID files, or log files appear in the tracked diff. The local `codex_857_r*.pid` files remain untracked and were ignored as instructed.
- **PR body**: the attribution is still the required first line, and the Design section now explicitly retracts the earlier uniform-windows claim. Its schedule-shape explanation matches the new code, but the statement that measured prompts can stay profile-wide is contradicted by `bench_compare_1k.py` as described in Blocker 1.

- The round-3 measured call-order/timing defect is substantively fixed and guarded by the exact mutation counts claimed by the author.
- One physical multi-yield call produces multiple rows without re-invocation, and all yielded timings come from adapter components.
- Warmup placement, budgets, and warmup-prompt overrides remain correct.
- The other four legacy scripts' uniform measured prompt and window-major schedules fit the default plan; their requested global schedule tests pass.
- No source file in the review worktree was edited, committed, pushed, or posted to GitHub; only this requested ignored review artifact was written.

### What I did NOT check

- Per instruction, I did not run Cargo, compile Rust, launch an engine, download a model, or perform a live performance run.
- No concrete adapters or live profiles exist in this PR, so I did not test subprocess/API parsing or live old-vs-new raw-output equivalence.
- I did not wait for the four queued GitHub jobs because their eventual result cannot resolve the two contract findings above.
- External web results did not materially inform either finding; the authoritative evidence was the PR, issue #813, committed repository code, and observed local commands.

One more focused round is warranted after both contract defects are fixed: re-run the two mutations and clean suite, then inspect a single complete agentic flattened assertion that includes the exact per-engine measured prompts/hashes and malformed plan/window relationship tests. If that round has no blocker or major and CI is green at a fresh head, further broad review is unlikely to add value.

Domain utility: HIGH — the loaded deterministic-scheduling and validator guidance directly reinforced full-stream oracles, explicit cross-field invariants, and mutation-sensitive fail-closed testing.

Verdict: REJECT

ohdearquant:

## Review of PR #857 (round 5)

### Verdict: APPROVE

No blocker, major, medium, minor, or nit remains at head `ef191f364b7c75a4cf107900f8ca5c1db7d1c7a5`. Both round-4 contract defects are fixed, mutation-sensitive, and directly verified. The final representability sweep found no remaining legacy-script parameter that the profile/adapter contract cannot encode exactly.

### Prior review context

- Round 1 found that the initial profile/adapter contract could not encode all engine-specific legacy schedules and model identities.
- Round 2 confirmed the identity fixes but found that the warmup schedule was still not exactly representable.
- Round 3 confirmed the warmup model was fixed, but rejected the then-current head because the measured schedule still assumed uniform profile windows and could not encode `bench_compare_1k.py` exactly.
- Round 4 confirmed measured order and multi-yield timing were fixed, but rejected head `1a109910314dabff86a4216e1a570c065067366a` with one blocker (engine-specific measured-prompt identity was not representable) and one major (explicit plans did not fail closed against `profile.windows`).
- This round re-read the posted round-4 review first, then verified only the two requested fixes plus the final five-script representability, mutation, hygiene, freshness, CI, and PR-body gates.

### What this round examined

- Read all four changed files in full, all five live legacy scripts, and the adjacent Lattice benchmark prompt builder.
- Traced `EngineRunGroup.measured_prompt` from TOML parsing through per-group resolution, adapter invocation, and raw-observation hashing.
- Re-ran the missing-window, undeclared-window, and cross-call-duplicate parser probes and assessed the deliberate non-order coverage rule.
- Ran the 117-test Python suite, an independent capture-adapter probe, both round-4 mutations in a detached disposable worktree, Python/static gates, publication hygiene, branch/merge freshness checks, and live GitHub check inspection.

### Findings (0 total)

No findings.

### Gates and observed evidence

- **Measured prompt contract**: `EngineRunGroup.measured_prompt` is optional and defaults to `None` at `scripts/bench_decode_harness.py:417-425`; parsing validates and preserves it at `scripts/bench_decode_harness.py:728-752`. `run_profile` resolves the exact measured string and its hash once per engine group at `scripts/bench_decode_harness.py:1122-1126`, then passes both unchanged through every window-major or repeat-major call at `scripts/bench_decode_harness.py:1127-1148`.
- **Independent capture-adapter probe**: observed Lattice receive `TOKENIZER_PADDED_EXACT` four times, Ollama receive warmup `hi` followed by `OLLAMA_CHAR_HEURISTIC_EXACT` twice, and MLX receive warmup `BASE` followed by `TOKENIZER_PADDED_EXACT` four times. Every raw observation's hash equaled `prompt_hash()` of the exact string invoked; Ollama never received the Lattice/MLX measured prompt, and its warmup never received its measured prompt.
- **Agentic test quality**: `_FakeAdapter` captures the exact prompt independently of its existing call tuple at `tests/test_bench_decode_harness.py:76-115`. The complete agentic test asserts per-adapter call prompts at `tests/test_bench_decode_harness.py:1327-1367`, per-engine warmup/measured hashes at `tests/test_bench_decode_harness.py:1377-1408`, and multi-yield prompt/hash identity at `tests/test_bench_decode_harness.py:1410-1424`. Mutation A (ignore `measured_prompt`) produced exactly one failure, `test_agentic_style_global_flattened_sequence_lattice_ollama_mlx`, at its direct Ollama prompt assertion. One failing test is sufficient coverage here: the single failure is a comprehensive full-stream oracle containing explicit per-adapter prompt captures and per-observation hash checks, not a single indirect assertion. Splitting those same assertions into separate per-adapter test methods would improve failure localization only, not regression sensitivity.
- **Commented agentic TOML**: `scripts/bench_decode_profiles.toml:133-182` now gives Ollama its own literal `measured_prompt` value and explicitly states that the adapter receives/hashes it as-is. The angle-bracket text is illustrative example data, not an adapter-interpreted instruction or a shared placeholder.
- **Exact windows coverage**: the parser flattens all calls' `yields_windows`, computes missing, undeclared, and cross-call duplicate values, and raises `ProfileConfigError` naming `engines[idx]`, `measured_calls`, `yields_windows`, and the specific violations at `scripts/bench_decode_harness.py:693-724`. Direct probes observed the round-4 malformed cases reject as: missing `[256]`, undeclared `[999]`, and duplicate `[32]`, all naming `engines[0] measured_calls yields_windows`.
- **Non-order rule**: accepting exact coverage in a different physical-call order is sound, not a loophole. `profile.windows` is strictly increasing and defines baseline/report identity at `scripts/bench_decode_harness.py:532-541`; explicit call order plus `measured_order` defines replay at `scripts/bench_decode_harness.py:1122-1148`; aggregation attributes rows by window value at `scripts/bench_decode_harness.py:1183-1188,1232-1244`. Requiring the flattened plan to repeat sorted `profile.windows` would erase intentional replay expressiveness without strengthening identity or coverage.
- **Mutation B**: disabling the cross-field coverage check in the detached worktree produced exactly three failures: `test_measured_calls_missing_declared_window_rejected`, `test_measured_calls_undeclared_window_rejected`, and `test_measured_calls_cross_call_duplicate_window_rejected` (`tests/test_bench_decode_harness.py:497-536`).
- **Clean suite**: `python3 tests/test_bench_decode_harness.py -v` reported `Ran 117 tests ... OK` in the review worktree. After both mutations were reversed byte-for-byte in the detached worktree, another clean run also reported 117/117 OK and `git diff --exit-code` was clean.
- **Static gates**: `python3 -m py_compile scripts/bench_decode_harness.py tests/test_bench_decode_harness.py`, `actionlint .github/workflows/ci.yml`, `typos` on all four PR files, and `git diff --check origin/main...HEAD` all exited 0.
- **Workflow scope**: against current `origin/main`, the entire `.github` diff is still only the 15-line `bench-decode-harness-tests` job at `.github/workflows/ci.yml:489-502`. `gh pr diff 857 --name-only` and the local merge-base diff both list exactly the same four PR files.
- **Branch freshness and merge scope**: current `origin/main` is an ancestor of the reviewed head; `git rev-list --left-right --count HEAD...origin/main` returned `8 0`. Head is merge commit `ef191f364` with second parent `634a8a8c4`; compared with first parent `7609f2f0f`, the merge brought only `crates/inference/src/forward/gpu/inner/shaders.rs` and `tests.rs`. None of the four harness/gate files changed in that merge.
- **Publication hygiene**: all four PR files are newline-terminated; the three new files are mode `100644`; no workspace paths, orchestration markers, CJK text, PID files, or log files appear in the tracked diff. The local `codex_857_r*.pid` files are untracked and absent from the diff, as instructed.
- **GitHub CI at final inspection**: GitHub reported the exact reviewed head as `MERGEABLE`, with the harness job, actionlint, typos, Ubuntu CI, ARM CI, Ubuntu feature matrix, semver, cargo-deny, and all other completed applicable checks green. macOS CI, macOS feature matrix, bench-compile, and MSRV remained queued; `mergeStateStatus: BLOCKED` and `reviewDecision: REVIEW_REQUIRED` reflected those outstanding repository gates/review state, not a failing check.

### Final representability sweep

Every relevant parameter of all five live scripts is now exactly encodable; nothing remains:

- `bench_apples_to_apples.sh`: exact shared prompt, `[32, 256]` measured budgets/order, five repeats, MLX's one 8-token warmup, Q8 and Q4 tier model identities/paths, and per-engine quantizations can be represented (as the script's sequential tier profiles).
- `bench_apples_precise.sh`: exact shared prompt, two 512-token warmups before measurement, 13 measured repeats at `[64, 512]`, window-major order, per-engine models/quantizations, and trimmed-mean `trim=2` aggregation are representable.
- `bench_q4_apples.sh`: exact prompt, `[32, 256]` by five repeats, Lattice Q4 / MLX Q4 / Ollama Q8-reference identities and quantizations, and MLX's 8-token warmup are representable.
- `bench_context_scaling.sh`: exact prompt, one shared baseline window `8` plus comparison windows `[64, 128, 256]`, configurable repeat count, window-major execution, exact engine identities/quantizations, and MLX's single 4-token pre-window warmup are representable.
- `bench_compare_1k.py`: each context maps to a profile with its exact requested prompt-token target and materialized prompt strings; Lattice's window-major 1/100 calls, Ollama's one 100-token physical call yielding component timings for windows 1 and 100, MLX's repeat-major `(1, 100)` calls, all repeat counts, engine-specific warmup prompts/budgets, the distinct Ollama measured prompt, Lattice/MLX shared tokenizer-padded prompt, actual prompt-token reporting, engine order, models, and quantizations are representable.

The scaffold intentionally ships no concrete adapters or live profiles in this step. That affects execution availability, not the representational completeness of the now-closed profile/adapter contract.

### PR body v5

- The first line is the required `PR description authored by Claude (Anthropic agent) on behalf of @ohdearquant` attribution.
- The Design section explicitly retracts both the earlier uniform measured-schedule claim and the round-3 profile-wide measured-prompt claim, then describes the current override semantics accurately.
- Spec deviations likewise retracts the two earlier claims and matches the implemented default-plus-per-engine-override contract.
- The claimed 117 tests, mutation counts, merge scope, and `.github` scope match observed evidence.

- Both round-4 defects are fixed in implementation, parser diagnostics, examples, tests, and PR prose.
- Prompt identity is locked at the actual adapter boundary and raw-observation hash boundary, including warmup separation and Ollama's different measured prompt.
- Exact-once window coverage is fail-closed; deliberate call-order freedom does not weaken the contract.
- All five legacy scripts' prompts, budgets, call orders, timing components, warmups, repeats, models, and quantizations are representable with no remaining field gap.
- No source file in the review worktree was edited, committed, pushed, or posted to GitHub. Only this requested ignored review artifact was written; mutations were confined to and removed from the disposable detached worktree.

### What I did NOT check

- Per instruction, I did not run Cargo, compile Rust, launch an engine, download a model, or perform a live performance benchmark.
- No concrete adapters or live profiles ship in this PR, so I did not test subprocess/API parsing or live old-vs-new raw-output equivalence; those remain step-2 migration responsibilities.
- I did not treat the four queued GitHub jobs as observed passes. They must still complete green under the repository's normal merge policy.
- External web search did not produce usable PR evidence; the authoritative evidence was the checked-out source, Git history, GitHub CLI PR/check data, and commands observed in this review.

Stop. With no blocker, major, or lesser finding after this targeted fifth round, another review round is not useful unless the PR changes substantially. The queued repository CI jobs should finish normally before merge.

Domain utility: SKIPPED — no lore-domain connector was available in this session; the repository-specific Lattice review skill and loaded review/PR/CI rubrics were sufficient for this targeted contract verification.

Verdict: APPROVE

oceanwaves630 (approved):
Read the full diff at ef191f364 (harness module, profiles TOML, full test file, CI job). The harness core is sound and honestly scoped: fail-closed TOML parsing everywhere (unknown/misspelled keys rejected at top level, profile level, engine level, and measured-call level — no typo can silently fall back to a default), the measured_calls coverage invariant (flattened yields_windows must cover profile.windows exactly once — missing/undeclared/duplicate all rejected at parse time with the offending windows named), and the component_ns adapter contract that makes the one-call-yields-two-windows shape impossible to fake from wall-clock time. The representability claims check out against the round-history: warmup as pre-window batch, three distinct measured shapes (window-major default, repeat-major, multi-yield) and per-engine measured/warmup prompts, with the end-to-end test asserting exact call sequences, exact prompt strings, and that every recorded prompt_hash is the hash of the string the adapter was actually invoked with. slope_ci95_legacy carried over verbatim and labeled as a non-coverage-valid legacy heuristic in the dataclass doc, the aggregate doc, and the report header — the right call given the deferred methodology follow-up. CI job is correctly ungated ubuntu-only stdlib. Approved.

PR #859, opened 2026-07-11
fix(inference): first-wins tie-break in MTP/self-spec test oracles

### LINKED ISSUE #806: bug(inference): Metal MTP/self-spec test oracles use last-wins argmax, contradicting the engine's first-wins contract

## What

`crates/inference/src/forward/metal_qwen35.rs` contains two private test-only `argmax` helpers, one in the MTP test module and one in the self-speculative-decode test module (near lines 28437 and 29333). Both are implemented as:

```rust
fn argmax(logits: &[f32]) -> u32 {
    logits
        .iter()
        .copied()
        .enumerate()
        .max_by(|(_, a), (_, b)| a.partial_cmp(b).unwrap_or(std::cmp::Ordering::Equal))
        .map(|(i, _)| i as u32)
        .unwrap_or(0)
}
```

`Iterator::max_by` returns the **last** element when multiple elements compare equal. This contradicts the production/reference tie-break contract in `crates/inference/src/speculative.rs`, whose public `argmax` (around line 107) explicitly documents and implements **first-wins** on ties:

```rust
/// When multiple elements share the maximum value, returns the **first**
/// occurrence. Speculative decoding must be token-identical to plain greedy...
pub fn argmax(logits: &[f32]) -> usize {
    let mut best_idx = 0;
    let mut best_val = f32::NEG_INFINITY;
    for (i, &v) in logits.iter().enumerate() {
        if v > best_val {
            best_val = v;
            best_idx = i;
        }
    }
    best_idx
}
```

Both test-only helpers back oracle functions (`plain_greedy_oracle` and its self-spec counterpart) that these test modules use as the ground truth to check MTP/self-speculative decoding token-for-token agreement. A last-wins oracle can silently mask a first-wins regression in the production Metal decode path whenever two logits tie exactly (a real occurrence in low-precision/greedy edge cases and in constructed test fixtures with repeated logit values).

## Why

The test oracle's job is to prove that speculative decoding is token-identical to plain greedy decoding. If the oracle itself uses a different tie-break rule than the thing it's meant to model (`crate::speculative::argmax`, which the actual engine calls), the oracle is not testing the real contract — it is testing an incompatible one. This is a correctness gap in test infrastructure, not a production bug, and should be fixed before any deletion work touches this file so that deletion PRs remain behavior-pure and this fix's oracle-correctness change stays isolated and reviewable on its own.

## Steps

1. In `crates/inference/src/forward/metal_qwen35.rs`, delete both private `argmax` helper functions (in the MTP test module and the self-spec test module).
2. Replace their call sites with `crate::speculative::argmax(logits)`, casting the returned `usize` to `u32` at the call site (the helpers currently return `u32`; `crate::speculative::argmax` returns `usize`).
3. Add an explicit tie-break regression test to both modules: a logits vector such as `[1.0, 1.0, 0.0]` must resolve to index `0` (first-wins), not index `1`.
4. Do not change any other production code path in this PR — this is a test-oracle-only fix.

## Gates

- Named CPU-only tests pass in both the MTP test module and the self-spec test module (no GPU/Metal hardware required for these specific oracle tests).
- Reverse-apply this fix (restore the `max_by`-based helpers) and confirm the new tie-break assertions fail — this proves the regression test is mutation-sensitive, not decorative.
- Full existing `cargo test` suite for `lattice-inference` passes unaffected.
- Semver class: none (test-only code, no public API surface touched).

## Sequencing

- **Depends on:** nothing; this is a standalone correctness fix and should land before or independently of any deletion PR touching `crates/inference/src/generate.rs` or `crates/inference/src/forward/batch_prefill.rs`.
- **Depended on by:** none directly, but should land before or alongside the 0.6.0 removal work in this cluster so the test suite's oracle correctness isn't conflated with unrelated deletions in the same review.

ohdearquant:

Closes #806.

## What

The private test-only `argmax()` helpers in the MTP and self-speculative-decode test modules of `metal_qwen35.rs` used `Iterator::max_by`, which returns the **last** element on ties. The production/reference tie-break contract in `crate::speculative::argmax` is documented and implemented as **first-wins** (the engine calls it for token-identical greedy decoding). A last-wins test oracle can silently mask a first-wins regression in the production Metal decode path whenever two logits tie exactly.

- Delete both private last-wins `argmax()` helpers.
- Replace call sites with `crate::speculative::argmax(logits) as u32`.
- Add an explicit first-wins tie-break regression test to both modules (`[1.0, 1.0, 0.0]` → index 0).

## Verification

- Both new tests pass: `plain_greedy_oracle_argmax_is_first_wins_on_ties` in `mtp_greedy_round_tests` and `self_spec_eos_tests` (2 passed, 0 failed).
- Mutation-sensitive: reverting the fix makes the new assertions fail (verified locally by reverse-applying the diff).

## bench-compare

Not applicable by construction: all 102 changed lines are inside `#[cfg(test)]` modules (`mtp_greedy_round_tests` at :28911, `self_spec_eos_tests` at :29842), compiled out of every release/product binary. No production code path changed.

### DIFF
--- a/crates/inference/src/forward/metal_qwen35.rs
+++ b/crates/inference/src/forward/metal_qwen35.rs
@@ -28931,7 +28931,7 @@ mod mtp_greedy_round_tests {
     fn plain_greedy_oracle(logit_rows: &[Vec<f32>], max_len: usize) -> Vec<u32> {
         let mut out = Vec::new();
         for row in logit_rows.iter().take(max_len) {
-            let token = argmax(row);
+            let token = crate::speculative::argmax(row) as u32;
             if is_stop(token) {
                 break;
             }
@@ -28940,14 +28940,19 @@ mod mtp_greedy_round_tests {
         out
     }
 
-    fn argmax(logits: &[f32]) -> u32 {
-        logits
-            .iter()
-            .copied()
-            .enumerate()
-            .max_by(|(_, a), (_, b)| a.partial_cmp(b).unwrap_or(std::cmp::Ordering::Equal))
-            .map(|(i, _)| i as u32)
-            .unwrap_or(0)
+    // Regression test (#806): the oracle's argmax must be first-wins, matching
+    // `crate::speculative::argmax`'s documented tie-break contract. A last-wins
+    // oracle would silently mask a first-wins regression in the production
+    // Metal decode path whenever two logits tie exactly.
+    #[test]
+    fn plain_greedy_oracle_argmax_is_first_wins_on_ties() {
+        let logit_rows = vec![vec![1.0, 1.0, 0.0]];
+        let out = plain_greedy_oracle(&logit_rows, 1);
+        assert_eq!(
+            out,
+            vec![0],
+            "tie must resolve to the first index, not the last"
+        );
     }
 
     // -----------------------------------------------------------------------
@@ -28984,7 +28989,7 @@ mod mtp_greedy_round_tests {
             if out.len() >= max_len {
                 break;
             }
-            let pending_token = argmax(&logit_rows[pos]);
+            let pending_token = crate::speculative::argmax(&logit_rows[pos]) as u32;
 
             // Case A: pending is a stop token. Stop-token contract (#613):
             // excluded from the output. For subsequent rounds,
@@ -29010,15 +29015,15 @@ mod mtp_greedy_round_tests {
             draft_idx += 1;
 
             let target_logits_0 = &logit_rows[pos + 1];
-            let accepted = draft_token == argmax(target_logits_0);
+            let accepted = draft_token == crate::speculative::argmax(target_logits_0) as u32;
             let bonus_token = if accepted {
                 if pos + 2 < logit_rows.len() {
-                    argmax(&logit_rows[pos + 2])
+                    crate::speculative::argmax(&logit_rows[pos + 2]) as u32
                 } else {
                     0 // no further prediction available; use 0 (non-stop)
                 }
             } else {
-                argmax(target_logits_0)
+                crate::speculative::argmax(target_logits_0) as u32
             };
 
             match mtp_greedy_round(pending_token, draft_token, accepted, bonus_token, is_stop) {
@@ -29460,7 +29465,10 @@ mod mtp_greedy_round_tests {
                 oracle,
                 mtp,
                 "trial {trial}: oracle != mtp\n  logit_winners={:?}\n  draft_seq={:?}\n  oracle={:?}\n  mtp={:?}",
-                logit_rows.iter().map(|r| argmax(r)).collect::<Vec<_>>(),
+                logit_rows
+                    .iter()
+                    .map(|r| crate::speculative::argmax(r) as u32)
+                    .collect::<Vec<_>>(),
                 draft_seq,
                 oracle,
                 mtp
@@ -29506,7 +29514,10 @@ mod mtp_greedy_round_tests {
                 oracle,
                 mtp,
                 "trial {trial} (max_len={max_len}): oracle != mtp\n  winners={:?}\n  draft={:?}\n  oracle={:?}\n  mtp={:?}",
-                logit_rows.iter().map(|r| argmax(r)).collect::<Vec<_>>(),
+                logit_rows
+                    .iter()
+                    .map(|r| crate::speculative::argmax(r) as u32)
+                    .collect::<Vec<_>>(),
                 draft_seq,
                 oracle,
                 mtp
@@ -29596,7 +29607,7 @@ mod mtp_greedy_round_tests {
             if out.len() >= max_len {
                 break;
             }
-            let pending_token = argmax(&logit_rows[pos]);
+            let pending_token = crate::speculative::argmax(&logit_rows[pos]) as u32;
             if is_stop(pending_token) {
                 // Stop-token contract (#613): excluded from `out`. The
                 // loop-top guard already confirmed out.len() < max_len, so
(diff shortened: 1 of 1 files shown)


ohdearquant:

## Review of PR #859 (round 1)

### Prior review context

- This is round 1. GitHub reports no earlier reviews, comments, or review threads.

### What this round examined

- Compared issue #806, the PR body, and the complete committed diff at `825df61ffdb7116dcfdd940c5fd5e742b74454e2`.
- Audited both changed test modules, `crate::speculative::argmax`, edge semantics, casts, and sibling argmax implementations across `crates/inference/src`.
- Ran the exact scoped test and a last-wins mutation in a detached disposable worktree.
- Checked diff hygiene, branch ancestry, GitHub checks, attribution, and the commit trailer.

### Findings (# total: 0 blockers, 0 majors, 2 mediums)

#### Mediums

1. [`crates/inference/src/kv_cache/flat.rs:1427`](crates/inference/src/kv_cache/flat.rs#L1427) A sibling top-1 logits oracle still uses last-wins `Iterator::max_by`.
   - **Current**: Both `top1_f32` and `top1_f16` use `.max_by(|(_, a), (_, b)| a.partial_cmp(b).unwrap())` at lines 1427–1438.
   - **Issue**: This tensor oracle claims to measure top-1 agreement, but exact ties resolve to the last index while the engine's documented greedy contract resolves to the first. F16 conversion can create exact ties even when the f32 logits are distinct, so the reported `top1_match_rate` can evaluate a different token choice than production. This is the same sibling-oracle bug class, although it is outside issue #806's mandated one-file scope.
   - **Fix**: In a separate tightly scoped issue/PR, compute both winners with the shared first-wins argmax and add a tie fixture that becomes mutation-red under last-wins behavior. Do not expand PR #859, because issue #806 explicitly limits this PR to `metal_qwen35.rs`.
   - **Test**: Add a constructed f32/f16 top-1 case with equal maxima at two indices and assert the lower index wins.
   - **Confidence**: Derived from the test's top-1 gate at lines 1472–1485 and the standard-library `Iterator::max_by` contract; observed by repository-wide `rg` over `crates/inference/src`.

2. [`CLAUDE.md:106`](CLAUDE.md#L106) The reviewed head is not up to date with `main`, and CI is not yet complete.
   - **Current**: GitHub reports `mergeStateStatus: BEHIND`. Direct ancestry checks show `main...HEAD` as `1 1`; HEAD's parent is `d18f50e...`, while current `main` is `e63df69...`. At the 2026-07-11 01:22 UTC snapshot, Ubuntu x86/ARM CI had passed, but macOS, MSRV, bench-compile, e2e, and the aggregate gates had not completed.
   - **Issue**: The repository merge gate requires the branch to be current with `main` and green at the resulting head. The presently reviewed SHA therefore cannot receive a merge-ready approval even though the scoped code is correct.
   - **Fix**: Update/rebase the branch onto current `main`, then require all checks to complete green at the new head.
   - **Test**: `git merge-base --is-ancestor origin/main HEAD` must succeed and the required GitHub aggregate gates must be green.
   - **Confidence**: Observed from GitHub PR metadata, `git ls-remote`, `git merge-base`, and `git rev-list --left-right --count`.

- Scope is exact: one file changed, 62 additions and 40 deletions (102 lines). Every hunk is within `#[cfg(test)] mod mtp_greedy_round_tests` or `#[cfg(test)] mod self_spec_eos_tests`; no production path changed, so the bench-compare N/A claim is valid.
- `crate::speculative::argmax` returns `usize`, documents first-wins, skips NaNs, returns 0 on empty input, and implements first-wins via strict `v > best_val`.
- The deleted helpers also returned 0 for empty input. They differed on NaNs because `partial_cmp(...).unwrap_or(Equal)` combined with last-wins `max_by`; no affected fixture contains NaN or infinity, so the tests did not rely on that difference.
- The `as u32` conversions are safe in these modules: fixtures use 6–8 logits or `winner + 2`, and all oracle outputs are `u32` token IDs. The same conversion is established throughout production speculative decoding.
- Exact requested command passed: 2 passed, 0 failed, 1816 filtered out.
- Mutation check passed: changing the shared comparison from `>` to `>=` in a detached worktree made both new tests fail with `[1]` versus `[0]`; the worktree was then removed.
- The PR body has the required Claude attribution. The sole conventional commit has `Co-Authored-By: Leo <noreply@khive.ai>`. `git diff --check` is clean; no manifest, lockfile, dependency, public API, or release metadata changed.
- The only other crate-wide `max_by` hit resembling argmax is a MoE expert-routing test with a deliberately unique winner; it is not a greedy-token oracle or tie-contract sibling.
- Rust's official [`Iterator::max_by` documentation](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.max_by) confirms that equally maximal elements resolve to the last element. The [Rust Reference's numeric-cast rules](https://doc.rust-lang.org/stable/reference/expressions/operator-expr.html#numeric-cast) confirm that narrowing integer casts truncate; the affected fixture bounds keep these casts in range.

### What I did NOT check

- I did not run benches, workspace-wide builds, or any broader cargo test, per the task constraint.
- I did not run GPU/Metal inference; the requested regression tests are CPU-only pure logic.
- CI was still in progress at the latest recorded snapshot; final merge eligibility must be re-evaluated after the branch update.

- No further code-review round is useful for this SHA. After updating onto `main`, confirm the diff is unchanged and all required checks are green; separately track the `flat.rs` sibling oracle.

Domain utility: SKIPPED — khive recall/domain discovery timed out, and the repository's argmax contract plus official Rust semantics were sufficient for this narrow review.

Verdict: APPROVE-WITH-FIXES

oceanwaves630 (approved):
Read the full diff at e54e14caf. All 102 changed lines confirmed inside #[cfg(test)] test modules — production decode paths untouched. The fix is right: the deleted local oracles used max_by, which returns the LAST maximum on ties, while the production speculative::argmax contract is first-wins — so on an exact logit tie the old oracles would have agreed with a first-wins regression's opposite. Replacing every oracle call site with the production argmax plus the explicit [1.0, 1.0, 0.0] → index 0 regression test in both modules closes that masking hole. Filing the kv_cache/flat.rs sibling as #861 instead of expanding this PR is the right scope call. Approved.

PR #862, opened 2026-07-11
fix(inference): fail-closed GDN Q/K normalization on non-finite input (CPU + Metal)

### LINKED ISSUE #850: bug(inference): eight live GDN Q/K normalization sites multiply NaN through a zeroed reciprocal instead of assigning zero

## What

All eight live GDN L2-normalization sites in `crates/inference/src/forward/shaders/qwen35.metal` (verified at current `main`, no shader-file drift since the audit) share the same fail-open shape:

```metal
float qs = (sg_buf[0] > 1e-12f) ? rsqrt(sg_buf[0]) : 0.0f;
```

When a Q or K lane is NaN, `sg_buf[0]` (the `simd_sum` of squares) is NaN, `NaN > 1e-12f` is false, and `qs` correctly evaluates to `0.0f`. But the source is stored as `value * qs` (or the K equivalent), and IEEE-754 defines `NaN * 0.0f == NaN`, not `0.0f`. The corrupted lane survives into every downstream consumer.

Verified sites (current `main`):

| Kernel | Guard line(s) | Consumers of the poisoned value |
|---|---|---|
| `gdn_recurrence_fused` (`qwen35.metal:1008`) | `:1066`, `:1077` | `q_tg`/`k_tg` → `k_dot_q`, `kv_mem`/`old_q`, the in-place `S_all` update, and kernel output |
| `gdn_recurrence_fused_q36` (`qwen35.metal:1144`) | `:1202`, `:1212` | same shape, specialized layout |
| `gdn_precompute_keys` (`qwen35.metal:1277`) | `:1329`, `:1338` | writes NaN into `key_scratch`, consumed later by `gdn_recurrence_sharded` |
| `gdn_chunk_materialize_c32` (`qwen35.metal:2767`) | `:2849`, `:2865` | writes NaN into `out_q`/`out_k`, consumed by `gdn_chunk_solve_c32` and reaches `raw_out`/`S_all` |

All four kernels are registered in both duplicated Qwen3.5 pipeline-construction blocks in `crates/inference/src/forward/metal_qwen35.rs` and dispatched in production (serial decode, Qwen3.6 specialized/sharded, and chunked-prefill paths). This is a live production correctness gap, not source-only debt.

The adjacent Qwen3.6 decay guard at `qwen35.metal:1217` already documents the same `inf * 0 = NaN` poisoning mechanism, so the fix pattern is already precedented in this file.

## Why

ADR-080 C1 (`docs/adr/ADR-080-consolidation-duplicated-contracts.md`) defines the fail-closed contract as a **whole logical unit** (a complete row or vector) being assigned exact zeros on an invalid input, never a lane-local zero achieved by multiplying through a guarded reciprocal. This class was already fixed for CPU `batch_prefill` (#739) and is being fixed for Metal fused attention and WGPU softmax in the in-flight numeric lane (issues #789/#790, PRs open as of this writing). The GDN sites are the same fail-open class, in a different kernel family (`qwen35.metal` GDN normalization vs. `flash_attention.metal`/WGSL softmax), and are not touched by that lane's current PRs.

**Correction to the original design recommendation**: the design draft for this series proposed folding the GDN fix directly into the #789/#790/#791 lane before that lane merged, to avoid a "class closed" merge with a known production hole left open. That lane's PRs (fix rounds already through one review cycle as of this writing) do not touch `qwen35.metal` and are close to merge without the GDN fix. Rather than block or reopen that lane, this issue tracks the GDN fix as an immediate follow-on landing right after it, so the same fail-open class does not stay silently open longer than necessary.

## What to change

For each of the eight sites, replace the lane-local guarded-reciprocal-then-multiply with an explicit whole-vector validity flag and an assignment-based invalid branch:

```metal
// Reduce both sum_sq and a per-lane !isfinite(value) flag across the full vector.
bool valid = bad_count == 0u && isfinite(sum_sq) && sum_sq > 1e-12f;
float inv_norm = valid ? rsqrt(sum_sq) : 0.0f;
float normalized = valid ? value * inv_norm : 0.0f;
```

The final conditional must be the contract: the invalid arm assigns `0.0f` directly and never evaluates `value * 0.0f`. `valid` must be uniform across the threadgroup and reflect every lane in the reduction, not just the executing thread's own lane. The existing `1e-12f` finite-path threshold is unchanged, and finite-input arithmetic must not change.

**CPU/test oracle parity** (required in the same PR, so GPU and CPU agree on the new invalid-input cases):

- `attention::gdn::l2_normalize_vec` (`crates/inference/src/attention/gdn.rs:404`) gains the same whole-vector non-finite-to-zero policy.
- The `gdn_chunk_ref::l2_normalize_rows` scalar oracle gains the same policy.
- `decode_attention_reference` (`crates/inference/src/forward/metal_qwen35.rs:18241`, inside the `#[cfg(test)]` module) currently reproduces the same multiply-through-zero shape as the deleted dead MSL softmax kernel — `inv_sum = (sum_exp > 0.0f) ? (1.0f/sum_exp) : 0.0f; ... acc += exp(...) * inv_sum * v_cache[...]`. This kernel is a frozen correctness oracle for a parity test against the live flash decode kernel; the in-flight numeric lane's own review explicitly flagged it as out of scope and left it unfixed. Harden it here so a test-only oracle does not itself reproduce the bug class it is meant to catch.

## What NOT to change

- Do not change finite-input L2 formulas, thresholds, reduction order, threadgroup geometry, or state layout.
- Do not treat a lane-local zero as whole-vector fail-close, and do not rely on `value * guarded_zero` in the invalid branch.
- Invalid K does not mean "reset all GDN state" — state still updates according to the recurrence with `K = 0`; invalid Q yields zero retrieval while K-driven state update remains valid. Tests must pin this zero-substitution semantics rather than assume the whole kernel output goes to zero.

## Checklist

- [ ] Rewrite all eight GDN normalization sites in `qwen35.metal` with the whole-vector validity-flag + assignment idiom
- [ ] Harden `attention::gdn::l2_normalize_vec` with the same whole-vector invalid-input policy
- [ ] Harden `gdn_chunk_ref::l2_normalize_rows` with the same policy
- [ ] Harden the test-only `decode_attention_reference` oracle so it no longer reproduces the multiply-through-zero shape
- [ ] Add direct Metal NaN-lane subcases for all four kernels (Q and K, 8 subcases total)
- [ ] Add CPU table tests: `[NaN, finite...]`, `[+inf, finite...]`, `[-inf, finite...]`, all-zero, very-small finite, ordinary finite
- [ ] Add one production-shaped integration test (corrupt one `in_proj_qkv` weight row/input column for Q, then K; run serial decode and chunked prefill; compare against the hardened CPU reference)

## Gates

- Eight direct Metal NaN subcases (Q and K for each of the four kernels) pass, asserting successful lookup/dispatch of the named kernel (no silent skip on a missing device or optional Qwen3.6 pipeline).
- CPU table tests pass for all six input classes.
- **Reverse-mutation proof**: revert only the validity/assignment diff, `touch crates/inference/src/forward/shaders/qwen35.metal` to defeat the `include_str!` timestamp cache, rerun the named test filter — all eight Q/K subcases must fail on the reverted code and pass on the fix. Attach the failing output as PR evidence.
- `cargo test -p lattice-inference --features f16,metal-gpu` under the GPU test lock (`gpu_test_lock()` / `/tmp/lion-metal-gpu-test.lock`).
- `make bench-compare BASE=origin/main HEAD=<branch>` attached per ADR-058, since the added collective non-finite scan costs instructions and synchronization even on the finite-input path. Any statistically meaningful regression must be justified or the scan optimized before merge.
- Existing inference E2E parity gate (`e2e-parity.yml`) stays green.

## Sequencing

Land as its own PR immediately after the in-flight #789/#790/#791 numeric lane merges (same fail-open class, adjacent kernel family, no dependency on that lane's diff). Not gated on the 0.6.0 window; this is a 0.5.x-eligible correctness fix.

ohdearquant:

Fixes #850.

## Bug

All eight live GDN Q/K L2-normalization sites in `qwen35.metal` shared a fail-open shape:

```metal
qs = (sum_sq > 1e-12f) ? rsqrt(sum_sq) : 0.0f;
value *= qs;
```

When a Q/K lane is `NaN`, `sum_sq` is `NaN`, the guard correctly zeroes `qs`, but IEEE-754 defines `NaN * 0.0f == NaN`, so the poisoned lane survives. GDN's recurrent state persists across tokens, so one poisoned token contaminates every subsequent token's generation.

## Contract decision

Zero the whole vector by direct **assignment** on an invalid (non-finite or near-zero) norm, never by multiplying a poisoned lane through a zeroed reciprocal:

```metal
bool q_valid = isfinite(sg_buf[0]) && sg_buf[0] > 1e-12f;
float qs = q_valid ? rsqrt(sg_buf[0]) : 0.0f;
q_val = q_valid ? (q_val * qs) : 0.0f;
```

This matches the fail-closed idiom already applied to Metal `fused_attention`'s finalize step in #794.

**Rejected alternative: clamp-to-finite** (extending `compute_decay_gate`'s existing `a_log.exp().min(f32::MAX)` idiom in the same file). Rejected because clamping a NaN lane still requires picking a finite substitute value with no principled choice — there is no "correct" clamped value for a corrupted Q/K component. Zero is both the natural "no signal" value for Q/K and matches the epsilon-regularized zero-vector case the CPU reference already handles gracefully.

### CPU vs. Metal near-zero contract (corrected, review round 2)

The shared contract across all backends is narrower than an earlier draft of this PR claimed: **non-finite input/norm → whole-vector zero-by-assignment.** That much is identical everywhere (CPU scalar/AVX2/NEON and Metal).

Beyond that, the two families diverge on purpose, and this PR does not change either side's finite-input arithmetic:

- **CPU** (`attention::gdn::l2_normalize_vec`, `forward::gdn_chunk_ref::l2_normalize_rows`, and `attention::gdn_fused::simd_l2_normalize`'s scalar/AVX2/NEON backends): validity is `norm_sq.is_finite()` only. A tiny nonzero norm (e.g. `[1e-7, 0, 0, 0]`) is **not** zeroed — it flows through the pre-existing epsilon-regularized formula (`x / sqrt(norm_sq + eps)`), unchanged by this PR.
- **Metal** (all four `qwen35.metal` sites): validity is `isfinite(sum_sq) && sum_sq > 1e-12f`. The same tiny-norm input **is** zeroed on Metal (pre-existing behavior, also unchanged by this PR).

This finite-input difference pre-dates this PR (issue #850's scope is the non-finite fail-open bug only) and is out of scope here — changing CPU to add a `> 1e-12` branch, or Metal to drop it, would be a separate, disclosed numerics change requiring its own justification and parity evidence. No such change is included. There is **no bit-parity claim between CPU and Metal for finite inputs** anywhere in this PR; the only cross-backend claim is the shared non-finite-zeroing contract, which the tests below verify per-backend.

## Sites fixed

**Metal** (`crates/inference/src/forward/shaders/qwen35.metal`) — every `threadgroup float sg_buf[4]` site was re-audited end to end this round (7 declarations total: `gdn_recurrence_fused`, `gdn_recurrence_fused_q36`, `gdn_precompute_keys`, `gdn_norm_silu`, `gdn_chunk_materialize_c32`, `gdn_chunk_solve_c32`, `gdn_chunk_norm_silu_c32`).

*Fail-closed guard sites (round 1, unchanged this round):*
- `gdn_recurrence_fused`: Q, K
- `gdn_recurrence_fused_q36`: Q, K
- `gdn_precompute_keys`: Q, K
- `gdn_chunk_materialize_c32`: Q, K

*WAR-hazard barrier sites (round 1: 4 handoffs; round 2: 8 more handoffs found and fixed — 12 total, complete audit table below).*

### Full `sg_buf` handoff audit (every read-then-later-write pair, complete)

| Kernel | Shared value | Read line | Next write line | Barrier added |
|---|---|---|---|---|
| `gdn_recurrence_fused` | beta | 1046 | 1058 (alpha partial) | round 2, `:1050` |
| `gdn_recurrence_fused` | alpha | 1062 | 1075 (Q q_sq partial) | round 2, `:1064` |
| `gdn_recurrence_fused` | Q norm | (guard read) | K k_sq partial | round 1 (pre-existing at merge) |
| `gdn_recurrence_fused` | K norm | (guard read) | kq_part (k_dot_q) partial | round 1 |
| `gdn_recurrence_fused` | k_dot_q | 1113 | 1145 (osq partial) | round 2, `:1116` |
| `gdn_recurrence_fused_q36` | beta | 1204 | alpha partial | round 2, `:1207` |
| `gdn_recurrence_fused_q36` | alpha | 1218 | Q q_sq partial | round 2, `:1221` |
| `gdn_recurrence_fused_q36` | Q norm | (guard read) | K k_sq partial | round 1 |
| `gdn_recurrence_fused_q36` | K norm | (guard read) | kq_part partial | round 1 |
| `gdn_recurrence_fused_q36` | k_dot_q | 1264 | osq partial | round 2, `:1267` |
| `gdn_precompute_keys` | beta | 1348 | alpha partial | round 2, `:1351` |
| `gdn_precompute_keys` | alpha | 1362 | Q q_sq partial | round 2, `:1365` |
| `gdn_precompute_keys` | Q norm | (guard read) | K k_sq partial | round 1 |
| `gdn_precompute_keys` | K norm | (guard read) | kq_part (k_dot_q) partial | round 1 |
| `gdn_precompute_keys` | k_dot_q | 1408 | *(none — written to device `key_scratch`, not reused via `sg_buf`)* | n/a, no hazard |
| `gdn_norm_silu` | osq (single write/read cycle) | — | — | n/a, no handoff — this kernel writes `sg_buf` once and never reuses the buffer for a second value |
| `gdn_chunk_materialize_c32` | Q norm | (guard read) | K ksq partial | round 1 (pre-existing) |
| `gdn_chunk_materialize_c32` | K norm | (guard read) | beta bp partial | round 1 (pre-existing) |
| `gdn_chunk_materialize_c32` | beta (aggregated, `tid==0`-only consumer) | (tid==0 block) | alpha ap partial | pre-existing double-barrier idiom (write→barrier→tid==0-reduce-and-consume→barrier); no broadcast read outside `tid==0`, so no WAR hazard — verified by trace, not just inspection |
| `gdn_chunk_materialize_c32` | alpha (aggregated, `tid==0`-only) | (tid==0 block) | next-token-iteration Q qsq partial | same pattern, same verification |
| `gdn_chunk_solve_c32` | k_k_dot (kkt) | (tid==0 block) | qk_partial (Q-dot) | pre-existing barrier immediately after the `tid==0` block; no unguarded broadcast read |
| `gdn_chunk_solve_c32` | qk_partial | (kkt_val block) | next-`kk`-iteration k_k_dot | pre-existing barrier; verified |
| `gdn_chunk_solve_c32` | k2/qd/k2p scratch triplet (self-dot section) | — | — | dead/discarded intermediate writes (`k2`, `qd` are computed and overwritten before any consumer reads them) — wasteful but not a WAR hazard, since nothing reads the discarded values; pre-existing, out of scope for this PR |
| `gdn_chunk_solve_c32` | kkt_jj | (tid==0 block) | qkl_jj partial | pre-existing barrier; verified |
| `gdn_chunk_norm_silu_c32` | osq (single write/read cycle) | — | — | n/a, no handoff |

Round 2's fix: added `threadgroup_barrier(mem_flags::mem_threadgroup)` after every simdgroup's *broadcast* read of a shared scalar (a plain `float x = sg_buf[0];` executed by all 128 threads, not gated by `tid==0`) and before the next `sg_buf[sgitg]` write, in `gdn_recurrence_fused`, `gdn_recurrence_fused_q36`, and `gdn_precompute_keys` — 8 sites total. The `tid==0`-only aggregation idiom used in `gdn_chunk_materialize_c32`/`gdn_chunk_solve_c32` does not have this hazard class (no broadcast read to race against), confirmed by tracing every site rather than assuming the pattern is safe by resemblance.

**CPU / test-oracle parity** (shared contract: non-finite → whole-vector zero; see "CPU vs. Metal near-zero contract" above for what is *not* shared):
- `attention::gdn::l2_normalize_vec` — non-finite `norm_sq` zeroes the whole vector by direct assignment.
- `forward::gdn_chunk_ref::l2_normalize_rows` — same policy, per row (no cross-row contamination).
- `attention::gdn_fused::simd_l2_normalize` (scalar/AVX2/NEON dispatch) — same policy; this is the **shipping production path** (`gated_delta_net_step_fused` → `simd_l2_normalize`), the round-1 blocker-2 finding.
- `forward/shaders/decode_reference.metal` (`decode_attention_reference`, the frozen CPU-parity oracle for the flash-decode kernel) — hardened so the test-only oracle doesn't itself reproduce the bug class it exists to catch.

## Tests

- CPU table tests (NaN / +inf / -inf / all-zero / very-small-finite / ordinary-finite) for `l2_normalize_vec`, `l2_normalize_rows`, and `simd_l2_normalize`'s scalar/AVX2/NEON backends.
- **CPU state-isolation test, production kernel** (`attention::gdn_fused::tests::gated_delta_net_step_fused_k_poison_window_state_isolation_production_kernel`, replaces the round-1 `linear_conv_kernel_dim=1` version): uses `Qwen35Config::qwen35_2b()` unmodified (real `linear_conv_kernel_dim == 4`), through the shipping `gated_delta_net_step_fused` path. Poisons K with one NaN lane at step 0 only; holds the K row at true zero for all 4 window steps (0-3) in both the poisoned and reference runs — the only reference construction for which bit-identical convergence is achievable, since a single corrupted lane forces the whole 128-lane vector to zero, which a genuinely undisturbed run's other 127 lanes would not produce (documented in the test's doc comment, with the underlying `apply_causal_conv1d` mechanics traced from source first). Runs 4 clean tokens past the window (8 steps total). Asserts, at every step: finite output and finite `s_matrices`; output and `s_matrices` bit-identical to the explicit-zero-weight reference; `conv_buffer` accounting split into the 3-step raw-NaN-residency window (exactly one non-finite element, explicitly diffed against the reference) and the post-window steps (fully finite, fully bit-identical) — the concrete "recovers to bit-identical once aged out of the conv window" proof.
- **Metal chunked-prefill multi-token test** (`forward::metal_qwen35::inner::tests::gdn_850_failclosed::gdn_chunk_materialize_c32_multi_token_k_poison_window_state_isolation`, new): dispatches `gdn_chunk_materialize_c32` (the chunked-prefill kernel) over 8 tokens in one chunk, poisoning token 0's K lane. `gdn_chunk_materialize_c32` reads each conv tap directly from the flat per-token input array by absolute index, so this is real within-chunk conv-window carryover, not incremental register state — same "4-token contamination window, then clean" shape as the CPU test, verified per-token.
- 6 direct Metal kernel-dispatch regression tests (`gdn_850_failclosed` module): the 5 round-1 tests (Q/K fail-closed for `gdn_recurrence_fused`, `gdn_recurrence_fused_q36`, `gdn_precompute_keys`, `gdn_chunk_materialize_c32` single-token, plus `decode_attention_reference`) plus the new multi-token test above.
- `decode_attention_reference` oracle test: NaN Q lane → exact zero output row, 0 non-finite elements.

**Round 2, this session** (both against the merged head, restored after):

Reverted the round-2 barrier fix (`git apply -R` on the isolated round-2 `.metal` diff, `touch` to defeat `include_str!` caching), reran the new multi-token chunked-prefill test:

```
thread '...::gdn_850_failclosed::gdn_chunk_materialize_c32_multi_token_k_poison_window_state_isolation' panicked:
assertion `left == right` failed: gdn_chunk_materialize_c32[multi-token-K]: out_k (all tokens) must fail closed (finite), got 512 non-finite element(s) out of 1024
test ...gdn_chunk_materialize_c32_multi_token_k_poison_window_state_isolation ... FAILED
```
(This mutation targeted the K-validity guard directly, not a barrier — barrier races are not deterministically reproducible via a single-threaded revert; the access-order argument in the audit table above is the correctness proof for the barriers themselves, per round-2's own guidance: "treat source access order as the proof; a race test cannot prove the missing synchronization safe.")

Restored, reran — passes:
```
test forward::metal_qwen35::inner::tests::gdn_850_failclosed::gdn_chunk_materialize_c32_multi_token_k_poison_window_state_isolation ... ok
```

Reverted the CPU NEON guard (`!norm_sq.is_finite()` early-return removed from `simd_l2_normalize_neon`), reran the new production-kernel state-isolation test:

```
thread 'attention::gdn_fused::tests::gated_delta_net_step_fused_k_poison_window_state_isolation_production_kernel' panicked:
assertion `left == right` failed: step 0: poisoned run must fail closed (finite output), found 2048 non-finite element(s) -- a NaN K lane leaked past simd_l2_normalize
test attention::gdn_fused::tests::gated_delta_net_step_fused_k_poison_window_state_isolation_production_kernel ... FAILED
```

Restored, reran — passes:
```
test attention::gdn_fused::tests::gated_delta_net_step_fused_k_poison_window_state_isolation_production_kernel ... ok
```

Full targeted GDN suite at the merged head (`cargo test -p lattice-inference --features metal-gpu,f16 --lib gdn -- --test-threads=1`): all pass — see Gates.

**Round 1 (prior session, unchanged evidence, all 5 original GPU + 3 CPU mutations still hold):** see round-1 review thread for the original 5-GPU / 3-CPU mutation cycle (all fail-then-pass as reported then; not re-run this round since none of those specific sites regressed, and the full suite re-run above covers them).

## Gates

- `cargo fmt -p lattice-inference --check` — clean, at the merged head.
- `cargo clippy -p lattice-inference --features metal-gpu,f16 --tests -- -D warnings` — 0 warnings, at the merged head.
- Full targeted GDN suite (`cargo test -p lattice-inference --features metal-gpu,f16 --lib gdn -- --test-threads=1`) — 98 passed, 0 failed, at the merged head (`3cc4bead2`; main's merge added 2 more gdn-named tests since the round-1 96-passed baseline).
- Branch merged with `origin/main` (no rebase/force-push); no conflicts.

## bench-compare

`crates/inference/src/forward/shaders/qwen35.metal` gained 8 new `threadgroup_barrier` calls on the GDN decode hot path this round (finding 1's fix). There is still no dedicated Criterion bench group for the GDN path (checked `elementwise_cpu_bench.rs`, `gated_attention_bench.rs`, `metal_decode_bench.rs`, `inference_perf.rs` — unchanged from round 1's disclosure), but `metal_decode_bench.rs`'s `metal_decode_q4/forward_step/no_adapter` group *does* exercise the real Metal Qwen3.5 decode path end to end, including every GDN kernel this PR touches, via a real Q4/QuaRot checkpoint (`~/.lattice/models/qwen3.5-0.8b-q4-quarot`). Same-worktree source-swap A/B, isolating exactly the round-2 barrier diff (`git apply` / `git apply -R` on the 26-line round-2-only `.metal` diff, rebuilding between runs, single worktree, same binary otherwise), at merged head `3cc4bead2`:

```
cargo bench -p lattice-inference --features metal-gpu,f16 --bench metal_decode_bench -- metal_decode_q4/forward_step/no_adapter
```

| Variant | mean | 95% CI |
|---|---:|---|
| without round-2 barriers (round-1 state) | 6.2906 ms | [5.9237 ms, 6.6589 ms] |
| with round-2 barriers (this PR's head) | 6.4444 ms | [6.0462 ms, 6.8368 ms] |

Point-estimate delta: +2.44%, well inside the overlapping 95% CIs (Criterion's own paired comparison on the barrier-added run reported `p = 0.08` / `p = 0.56` across the two directions — "no change in performance detected" both times). 8 extra `threadgroup_barrier` calls per decode step, on a kernel whose per-step cost is dominated by Q4 GEMV weight-bandwidth and dispatch overhead rather than by simdgroup-local synchronization, producing no measurable regression is consistent with that structure; reported honestly as within-noise rather than as a proven zero-cost claim.

- [x] SIMD intrinsics, if any, were manually verified — not just reviewed by an AI (the Metal barrier placements and NEON/AVX2 guard mutations were hand-traced against the shared-memory access order and mutation-tested, not just pattern-matched)
- [x] `cargo test` output included for any behavior-changing code; `cargo bench` output for perf-sensitive changes (see Tests, Mutation-sensitivity evidence, and bench-compare above)

## Scope notes (disclosed, not silent)

- The Metal differential tests are Metal-vs-Metal (poisoned dispatch vs. explicit-zero-reference dispatch of the same kernel), not a full cross-language CPU-vs-Metal comparison, and there is no bit-parity claim between CPU and Metal for finite inputs (see "CPU vs. Metal near-zero contract" above).
- The production-kernel state-isolation tests (both CPU serial-decode and Metal chunked-prefill) poison K only, not Q. Q-side coverage remains the round-1 `attention::gdn::tests::test_l2_normalize_vec_state_isolation_poisoned_token_matches_never_poisoned_reference` test, which still uses the degenerate `linear_conv_kernel_dim=1` scalar-reference path (unchanged this round — round 2 did not flag it, and K was the specific shipping-path gap round 1 found in blocker 2).
- No dedicated bench group exists for the GDN normalization path; the same-worktree source-swap table above isolates the round-2 barrier diff specifically, but is still a proxy for a purpose-built GDN Criterion group, not a substitute for one.


ohdearquant:

## Review of PR #862 (round 1)

### Verdict: REJECT

The eight edited Metal sites themselves use the right assignment-based invalid arm, and the new tests are mutation-sensitive. The PR is nevertheless not safe to merge: three Metal kernels reuse `sg_buf` without the WAR barriers that this same file says are required, and the shipping CPU SIMD normalizer remains fail-open.

### Prior review context

- No prior reviews, comments, or review threads were present when this review began.
- This round focused on fail-closed numerical semantics, sibling-site completeness, recurrent-state isolation, mutation sensitivity, benchmark evidence, and merge/publication hygiene.

### What this round examined

- PR #862 at `c239881cabb36a3f5911f3545da7f3bb7efabbdd`, issue #850, the complete PR diff, all changed functions/test modules, adjacent GDN dispatch paths, repository-wide Rust/Metal/WGSL normalization patterns, and current GitHub checks.
- Direct CPU/Metal regression runs, Metal-revert and CPU-guard mutations in a disposable detached worktree, an isolated test of the omitted shipping SIMD path, and a same-binary Criterion rerun.

### Findings (6 total: 2 blockers, 1 major, 3 mediums)

#### Blockers

1. [`crates/inference/src/forward/shaders/qwen35.metal:1071`](crates/inference/src/forward/shaders/qwen35.metal#L1071) The shared norm sum can be overwritten before every simdgroup consumes it.
   - **Current**: `q_valid` and `qs` read `sg_buf[0]`, then the next reduction starts writing K partials to `sg_buf[sgitg]` at line 1079 without an intervening threadgroup barrier. The same Q→K WAR hazard exists at lines 1212–1219 and 1344–1350. `gdn_precompute_keys` also starts the `k_dot_q` reduction at lines 1363–1366 without a barrier after every simdgroup consumes the K norm at lines 1354–1356.
   - **Issue**: simdgroups are not required to advance in lockstep after the preceding barrier. A fast simdgroup can overwrite slot 0 with a K or `k_dot_q` partial while a lagging simdgroup is still reading the Q/K whole-vector norm. That makes `valid` non-uniform or makes `rsqrt` use the wrong reduction, violating the whole-vector contract. The same file explicitly documents and implements the necessary WAR barriers in `gdn_chunk_materialize_c32` at lines 2871–2874 and 2888–2890.
   - **Fix**: add a `threadgroup_barrier(mem_flags::mem_threadgroup)` after every simdgroup has consumed each shared norm and before the next reduction may write `sg_buf`: Q→K in `gdn_recurrence_fused`, `gdn_recurrence_fused_q36`, and `gdn_precompute_keys`, plus K→`k_dot_q` in `gdn_precompute_keys`. Re-run the finite-path performance evidence because these are new hot-path barriers.
   - **Test**: retain the direct poisoned-vs-zero tests and add repeated/stress dispatch coverage, but do not treat a passing race test as the proof; the synchronization is required by the shared-memory access order itself.
   - **Confidence**: derived from the shader access order and corroborated by this file's own WAR-hazard comments at lines 2871–2890. The current five GPU tests passed, so I did not observe the race manifest dynamically.

2. [`crates/inference/src/attention/gdn_fused.rs:195`](crates/inference/src/attention/gdn_fused.rs#L195) The shipping CPU GDN path is still fail-open.
   - **Current**: `simd_l2_normalize` dispatches to scalar, AVX2, or NEON implementations whose norm handling at lines 82–87, 585–603, and 743–761 unconditionally computes an inverse norm and multiplies every lane. No implementation zeroes a non-finite vector by assignment.
   - **Issue**: this is not a dead sibling. The default Qwen3.5 forward path calls `gated_delta_net_step_fused` at `model/qwen35/forward.rs:142`, which calls this normalizer for Q and K at `gdn_fused.rs:483–484`; F16 and Q8 production paths call it too. The PR hardens the scalar reference helper while leaving the actual CPU inference path vulnerable to the same persistent-state poisoning.
   - **Fix**: make scalar, AVX2, and NEON reductions apply the same non-finite-to-whole-vector-zero contract without changing finite arithmetic. Add scalar/SIMD equivalence tests for NaN, ±Inf, zero, tiny finite, and ordinary finite vectors, plus a production `gated_delta_net_step_fused` K-poison state-isolation test.
   - **Test**: in a disposable worktree at the PR head, I added an isolated review-only test calling `simd_l2_normalize` on `[NaN, 1, 0, …]`. It failed on the active SIMD path with all 128 output lanes NaN.
   - **Confidence**: observed by running `cargo test -p lattice-inference --features metal-gpu,f16 --lib review_862_simd_l2_nan_must_fail_closed -- --nocapture` against the PR head (review-only test; disposable worktree removed afterward).

#### Major

3. [`crates/inference/src/attention/gdn.rs:896`](crates/inference/src/attention/gdn.rs#L896) The claimed multi-token state-isolation test does not exercise the poisoning mechanism that persists in GDN state.
   - **Current**: `build_weights(true, …)` permanently stores NaN in a Q projection weight at lines 923–938, so all three steps remain poisoned rather than "one poisoned token followed by two clean tokens." It tests Q only, runs the scalar reference `gated_delta_net_step`, and snapshots only `s_matrices`. Q is the retrieval side; invalid K is the path that can corrupt the recurrent matrix. The shipping fused path is not exercised—exactly why blocker 2 escaped.
   - **Issue**: the test proves persistent Q corruption is zero-substituted by the scalar helper, but it does not prove that one invalid K update leaves shipping CPU state bit-identical for subsequent clean tokens. The one-step Metal tests correctly compare updated `S_all` for Q and K, but they do not replace the requested production-shaped CPU/fused and serial/chunk follow-up gate.
   - **Fix**: add a K case through `gated_delta_net_step_fused`, poison only the first step, then run at least two clean steps while comparing complete output and recurrent state with an explicit-zero reference. Cover the chunked path or state clearly why a lower-level proof subsumes it. Correct the PR body/test rustdoc so it describes what is actually executed.
   - **Test**: mutation of the scalar guard made the existing test fail, but that only establishes sensitivity to `l2_normalize_vec`; it does not establish coverage of the shipping SIMD/K-state path.
   - **Confidence**: derived from the fixture construction at lines 923–973, the state snapshot code, and the production call path at `model/qwen35/forward.rs:142`.

#### Mediums

4. [`crates/inference/src/forward/metal_qwen35.rs:25261`](crates/inference/src/forward/metal_qwen35.rs#L25261) The Qwen3.6 kernel gate silently passes when the named kernel is missing.
   - **Current**: failure to find `gdn_precompute_keys` prints a skip message and returns success at lines 25261–25267.
   - **Issue**: issue #850 explicitly requires successful lookup/dispatch and says a missing optional Qwen3.6 pipeline must not silently skip. Repository CI policy likewise requires an exercised-path marker and says a skipped gate must be red. The comment claiming this skip is allowed by #850 contradicts the issue.
   - **Fix**: make missing `gdn_precompute_keys` a hard test failure wherever `MSL_SOURCE` is compiled for this gate. Device unavailability may remain a platform-level skip only if CI has a separate required Metal job that proves execution.
   - **Confidence**: derived from the test and issue acceptance criteria; in my run the kernel existed and the test executed.

5. [`crates/inference/src/attention/gdn.rs:418`](crates/inference/src/attention/gdn.rs#L418) CPU and Metal implement only non-finite parity, not the identical validity contract claimed in the task/PR prose.
   - **Current**: CPU rejects only non-finite `norm_sq` and otherwise uses `x / sqrt(norm_sq + 1e-6)`. Metal rejects `norm_sq <= 1e-12` and otherwise uses `x * rsqrt(norm_sq)` (for example, `qwen35.metal:1071–1073`). For `[1e-7, 0, …]`, CPU returns approximately `[1e-4, 0, …]`, while Metal returns exact zero.
   - **Issue**: the new code does not introduce this finite-input difference—the Metal threshold and CPU epsilon formula are both preserved, which correctly avoids an unrequested numerics change. But "same contract" and bit-comparable CPU/Metal wording is too broad, and the tests are backend-local rather than cross-language differential tests.
   - **Fix**: describe the aligned contract precisely as whole-vector zero substitution for non-finite reductions while documenting the pre-existing finite-input backend formulas. Do not add the Metal threshold to CPU in this PR without separate numerical-compatibility evidence. Add cross-backend invalid-input cases if exact parity is claimed.
   - **Confidence**: derived algebraically from the two formulas. Rust's official [`f32::is_finite`](https://doc.rust-lang.org/stable/core/primitive.f32.html#method.is_finite) documentation confirms that the CPU guard rejects NaN and both infinities.

6. [`.github/PULL_REQUEST_TEMPLATE.md:31`](.github/PULL_REQUEST_TEMPLATE.md#L31) Required merge/publication hygiene is incomplete.
   - **Current**: the PR body begins with the exact required `PR description authored by` attribution and the commit has `Co-Authored-By: Leo <noreply@khive.ai>`, but the required AI-assisted contribution checklist at template lines 31–39 is absent. The branch is also one commit behind `origin/main`, and several required/current checks were still pending at the final snapshot.
   - **Issue**: `CLAUDE.md:104–112` requires the branch to be up to date and green at its actual head. The current green results therefore do not establish the merge state.
   - **Fix**: add and truthfully complete the AI checklist, update the branch from `main`, and wait for the rerun at the new head to finish green before merge.
   - **Confidence**: observed with `gh pr view`, `git rev-list --left-right --count HEAD...origin/main` (`1 1`), and `gh pr checks 862`.

- All eight requested Q/K Metal sites are present and use direct assignment on the invalid arm: two each in `gdn_recurrence_fused`, `gdn_recurrence_fused_q36`, `gdn_precompute_keys`, and `gdn_chunk_materialize_c32`. No additional GDN Q/K normalization implementation was found in another Metal or WGSL shader.
- `l2_normalize_vec`, `l2_normalize_rows`, and `decode_attention_reference` correctly zero non-finite logical units without multiplying/dividing a poisoned lane through zero. The zero-vector-out semantics are appropriate: invalid K becomes a no-signal K update rather than resetting all GDN state; invalid Q yields no retrieval signal.
- Finite arithmetic was not changed by this diff. In particular, the existing Metal `1e-12` cutoff was retained and the CPU epsilon-regularized formula was retained.
- Targeted head tests passed: 2 fail-closed CPU tables, 1 scalar three-step test, and all 5 Metal dispatch tests (8 Q/K subcases plus the decode oracle).
- Metal reverse mutation in a disposable detached worktree: reverting only `qwen35.metal` and `decode_reference.metal`, then touching both shader files and `metal_qwen35.rs`, made all 5 GPU tests fail; restoring the patch made all 5 pass.
- CPU reverse mutation: removing the `l2_normalize_vec` guard made both its table test and state-isolation test fail; the unmodified head passed both.
- `cargo fmt -p lattice-inference --check`, `cargo clippy -p lattice-inference --features metal-gpu,f16 -- -D warnings`, `git diff --check`, and `cargo package -p lattice-inference --list` passed.
- Benchmark explanation: `rms_norm` and `silu_inplace` do not call any touched code. A same-binary quick rerun of `rms_norm/896` measured `+0.1425%`, 95% CI `[-0.6288%, +0.9156%]`, `p=0.85`; no change detected. That supports treating the PR body's cross-worktree `-26.78%` as environmental noise rather than an effect of this diff.
- At the last CI snapshot, both Ubuntu CI jobs, Ubuntu feature matrix, semver, bench compile, cargo-deny, actionlint, typos, unwrap lint, and rustdoc lint were green.

### What I did NOT check

- Full workspace tests/builds, a real model checkpoint generation, or a direct GDN performance benchmark; the repository has no dedicated GDN normalization Criterion group.
- Final macOS/MSRV/e2e/q4/embed/app-binary CI outcomes; those checks were still pending, and updating the stale branch will require a fresh run anyway.
- I did not dynamically reproduce the Metal shared-memory race in finding 1; that finding is derived from an unsynchronized WAR access and the shader's own documented barrier requirement.

- Another round is necessary after fixes. Re-audit every `sg_buf` reduction handoff, mutation-test all scalar/AVX2/NEON fail-closed branches, and require a K-poison → clean-token state-isolation test through the shipping fused CPU path before reconsidering approval.

Domain utility: MEDIUM — the performance-regression domains prompted the same-binary control that disproved a real code effect, but the decisive synchronization and sibling-path findings came from repository-specific tracing.

Verdict: REJECT

ohdearquant:
## Codex review round 2 — Verdict: REJECT

## Review of PR #862 (round 2)

### Verdict: REJECT

Round 1's four named Q/K WAR handoffs and the NaN behavior of the shipping CPU SIMD dispatcher are fixed and mutation-sensitive. The PR is still not safe to merge: the same `sg_buf` WAR pattern remains at eight other handoffs, and the strengthened state-isolation test demonstrably passes only by removing production convolution history. The stated near-zero SIMD contract, current benchmark evidence, AI checklist, and current-main merge state also remain unresolved.

### Prior review context

- Round 1 rejected the PR for unguarded Metal `sg_buf` reuse, an unguarded shipping CPU SIMD normalizer, and a state-isolation test that never followed one poisoned token with clean tokens. It also requested a hard Qwen3.6 kernel gate, per-backend parity wording, the AI checklist, a current-main merge, and credible benchmark evidence.
- This round reviewed the posted round-1 comment first, then fix commit `4fbca7bf4`, merge head `b3a357ae9a1812af123e9c14c5eaca67b789f200`, issue #850, the complete fix diff, all `sg_buf` uses in `qwen35.metal`, the affected CPU/Metal tests and production call path, PR prose/checks, and publication/commit hygiene.

### What this round examined

- Every `threadgroup float sg_buf[4]` declaration and all 110 indexed `sg_buf` accesses in `qwen35.metal`.
- Scalar, AVX2, NEON, and runtime-dispatch normalization implementations; the Qwen3.5 CPU forward call path; state structure and causal-convolution update order.
- Head tests, Metal and NEON reverse mutations, a production-shape state-isolation control, fmt/clippy/package gates, live GitHub CI, branch ancestry, PR template compliance, and commit trailers.

### Findings (5 total: 1 blocker, 2 majors, 2 mediums)

#### Blocker

1. [`crates/inference/src/forward/shaders/qwen35.metal:1046`](crates/inference/src/forward/shaders/qwen35.metal#L1046) Eight more shared-result-to-next-write WAR handoffs remain unguarded.
   - **Current**: the four round-1 handoffs now have the requested barriers at lines 1078, 1221, 1356, and 1369. But `gdn_recurrence_fused` still reads `beta_val = sg_buf[0]` at line 1046 before the alpha partial write at 1053, reads `alpha_val` at 1057 before the Q partial write at 1067, and reads `k_dot_q` at 1105 before the RMS partial write at 1134, with no intervening threadgroup barrier. The identical hazards remain in `gdn_recurrence_fused_q36` at 1193→1200, 1204→1212, and 1247→1272, and in `gdn_precompute_keys` at 1328→1335 and 1339→1347.
   - **Issue**: this is the same access order round 1 rejected: a fast simdgroup may overwrite `sg_buf[sgitg]` while a lagging simdgroup is still reading the prior whole-threadgroup result. The fact that these values are beta, alpha, or `k_dot_q` instead of Q/K norms does not change the WAR requirement. Wrong beta/alpha/k-dot values can directly alter the recurrent update and output.
   - **Fix**: put `threadgroup_barrier(mem_flags::mem_threadgroup)` after every simdgroup's last read of each shared scalar and before the next `sg_buf[sgitg]` write. Re-audit the full seven-declaration list after the change. `gdn_chunk_materialize_c32` and `gdn_chunk_solve_c32` already show the correct idiom; the two norm-only kernels do not reuse the buffer.
   - **Test**: retain repeated GPU dispatch coverage, but treat source access order as the proof; a race test cannot prove the missing synchronization safe. Re-run the real affected decode benchmark because this adds hot-path barriers beyond the four already measured.
   - **Confidence**: derived from the complete `sg_buf` access audit and the file's own WAR comments at lines 1075–1078, 2884–2887, 3093–3095, and 3136–3138.

#### Majors

2. [`crates/inference/src/attention/gdn_fused.rs:1914`](crates/inference/src/attention/gdn_fused.rs#L1914) The strengthened test avoids production recurrent state instead of proving its isolation.
   - **Current**: the test explicitly changes `linear_conv_kernel_dim` from the production value 4 (`qwen35_config.rs:185`) to 1 at lines 1927–1928 because a real poisoned projection is retained in `state.conv_buffer` and re-enters later tokens. It then records only `state.s_matrices` at line 1982, although the public state snapshot is `(s_matrices, conv_buffer)` at `attention/gdn.rs:100–127`.
   - **Issue**: I removed only the one-tap override in the disposable worktree; no fix was reverted. The test failed at clean step 1 because poisoned and explicit-zero outputs diverged. This directly falsifies the requested proof that a poisoned token does not contaminate subsequent clean tokens' complete production GDN recurrent state. The test's own lines 1916–1920 acknowledge the propagation path it excludes.
   - **Fix**: either harden/sanitize the convolution history and assert the real four-tap `state.snapshot()` plus outputs across subsequent clean tokens, or inject the invalid value after convolution and narrow every claim to the post-convolution L2-normalization contract. Issue #850's requested production-shaped serial/chunked integration test is still the appropriate end-to-end gate.
   - **Test**: production four-tap Q and K poison-at-step-0 followed by at least two clean tokens, comparing complete snapshots and output against explicit-zero reference for serial decode and chunked prefill.
   - **Confidence**: observed by running the existing fused K-isolation test with only `cfg.linear_conv_kernel_dim = 1` removed; it failed at step 1. The disposable worktree was restored and removed.

3. [`crates/inference/src/attention/gdn_fused.rs:90`](crates/inference/src/attention/gdn_fused.rs#L90) The shipping scalar/AVX2/NEON implementations do not implement the round-2 `isfinite && > 1e-12` validity contract.
   - **Current**: scalar, AVX2, and NEON reject only `!norm_sq.is_finite()` at lines 90, 605, and 772, then otherwise apply the epsilon-regularized finite formula. The table explicitly expects `[1e-7, 0, 0, 0]` not to be zeroed at lines 1782–1785. Metal rejects the same norm (`1e-14 <= 1e-12`) and assigns the whole vector zero.
   - **Issue**: the NaN/Inf fix is correct, but the explicit round-2 criterion says all three shipping CPU variants must use `valid = isfinite && > 1e-12`; they do not. The current PR body also still says “CPU / test-oracle parity (same contract)” and describes an `isfinite && > 1e-12` CPU branch that does not exist. Conversely, issue #850 says not to change finite-input arithmetic, so the acceptance criteria and issue scope conflict and cannot both be claimed satisfied.
   - **Fix**: establish the authoritative contract. If finite CPU behavior must remain unchanged, revise the round-2/PR wording to state the backend-specific finite formulas and limit the shared contract to non-finite whole-vector zeroing. If `> 1e-12` is mandatory on CPU, implement and test it only with explicit approval for the finite numerical change.
   - **Test**: make the tiny-finite expectation encode the chosen contract explicitly; retain exact all-zero assertions for invalid cases.
   - **Confidence**: derived from all three implementations and the current table. NaN does fail closed: the tests assert finiteness and exact zero, so NaN cannot slip through an `(a-b).abs() > eps` false comparison.

#### Mediums

4. [`.github/PULL_REQUEST_TEMPLATE.md:17`](.github/PULL_REQUEST_TEMPLATE.md#L17) The PR body has no current, reviewable performance evidence for the fix commit.
   - The body still carries the pre-fix proxy `rms_norm`/`silu_inplace` comparison at head `c239881ca`; those groups do not call the changed code. Commit `4fbca7bf4` says a same-worktree source-swap of three real Metal decode configurations found no change, but provides only `p > 0.05`, no command, point estimates, confidence intervals, raw output, or current head, and the evidence was not added to the PR body as required. Because the fix adds barriers and changes production normalization branches, this is not enough to judge regression magnitude skeptically.
   - **Fix**: put the exact current-head command and per-configuration point estimates/CIs in the PR body, or attach a reproducible same-binary source-swap report that isolates each barrier/guard change.
   - **Confidence**: observed from the live PR body, comments, and commit message.

5. [`.github/PULL_REQUEST_TEMPLATE.md:31`](.github/PULL_REQUEST_TEMPLATE.md#L31) Required AI and current-main merge hygiene is still incomplete.
   - The PR description has the required Claude attribution, and both authored commits have the Leo co-author trailer, but the required AI-assisted checklist at template lines 31–39 is absent from the live body. After a fresh `git fetch origin main`, `origin/main` was `6768f891`; the PR merge-base was `a20f933f`, `git merge-base --is-ancestor origin/main HEAD` returned 1, and GitHub reported `mergeStateStatus: BEHIND`. All CI is green at `b3a357ae9`, not at the current merge state.
   - **Fix**: add and truthfully complete the checklist, merge current `origin/main`, and require the full check set to rerun green at that head.
   - **Confidence**: observed from the live PR body/template, local ancestry commands, and `gh pr view`.

- The four exact handoffs round 1 named are fixed: Q→K in `gdn_recurrence_fused`, Q→K in `gdn_recurrence_fused_q36`, and Q→K plus K→`k_dot_q` in `gdn_precompute_keys`. `gdn_chunk_materialize_c32` retains its documented Q→K and K→beta barriers.
- The production CPU route is real: `model/qwen35/forward.rs:142` calls `gated_delta_net_step_fused`, which calls `simd_l2_normalize` for Q/K at `gdn_fused.rs:495–496`; runtime dispatch reaches NEON, AVX2, or the scalar fallback.
- Scalar, AVX2, and NEON all directly assign the whole vector to zero for a non-finite reduced norm. The public table asserts full finiteness and exact zero for NaN and ±Inf.
- The Qwen3.6 named-kernel lookup is now fail-hard at `metal_qwen35.rs:25286–25291`; only loud platform-level “no Metal device” handling remains. On this machine the kernel existed and dispatched.
- Observed head tests:
  - `cargo test -p lattice-inference --features metal-gpu,f16 --lib fail_closed -- --nocapture`: 3 passed.
  - `... --lib state_isolation -- --nocapture --test-threads=1`: 2 passed in the submitted one-tap configuration.
  - `... --lib gdn_850_failclosed -- --nocapture --test-threads=1`: all 5 GPU tests passed, including the Qwen3.6 kernel.
- Mutation sensitivity in a disposable detached worktree:
  - reverting one Metal Q fail-closed assignment and touching `qwen35.metal` plus `metal_qwen35.rs` made `gdn_recurrence_fused_fails_closed_on_nan_q_and_k_lane` fail with 8/8 non-finite outputs; restoring made it pass;
  - removing only the active arm64 NEON guard made `simd_l2_normalize_backends_fail_closed_table` fail on the NaN case; restoring made it pass.
- `cargo fmt --all -- --check`, `cargo clippy -p lattice-inference --features metal-gpu,f16 -- -D warnings`, `git diff --check`, and `cargo package -p lattice-inference --list` passed.
- All live GitHub checks at `b3a357ae9` are green, including macOS/Ubuntu CI, arm CI, MSRV, feature matrix, e2e parity, Metal informational parity, q4 gates, semver, cargo-deny, actionlint, package/binary gates, rustdoc, typos, and unwrap lint.
- No Cargo manifest/dependency change was introduced. Package contents were listable. Commit convention and agent trailers are present.

### What I did NOT check

- I did not run the full workspace test/build suite, a real checkpoint generation locally, or an independent production benchmark. Current CI provides broad green evidence at the stale head, but the PR body lacks enough exact benchmark data to reproduce the claimed no-change result.
- AVX2 was inspected and covered by x86 CI, but not dynamically mutation-tested on this arm64 host; the dynamic SIMD mutation used NEON.
- I reverted a deterministic Metal validity guard, not an individual barrier. A barrier-only race mutation need not fail deterministically, so the remaining barrier finding is based on the access order and the shader's own synchronization invariant.
- I did not audit unrelated unchanged logic outside the complete PR/fix diff and the adjacent GDN/normalization/state paths named above.

- Another round is necessary. Re-audit every `sg_buf` handoff after all seven missing barriers are added, require a production four-tap complete-state isolation test (and the issue's serial/chunked integration coverage), reconcile the CPU near-zero contract, and verify exact current-head benchmark and merge-gate evidence. Stop after those checks pass; no broader review angle is needed.

Domain utility: MEDIUM — concurrency ordering and SIMD benchmark domains reinforced the source-order and same-binary evidence bar, while the decisive findings came from repository-specific shader/state tracing and mutation controls.

Verdict: REJECT

ohdearquant:
## Codex review round 3 — Verdict: APPROVE-WITH-FIXES

## Review of PR #862 (round 3)

The round-2 correctness findings are resolved. The eight new Metal barriers close every remaining audited `sg_buf` WAR handoff, the production-kernel serial and multi-token chunk tests exercise the real four-token convolution window and are mutation-sensitive, and the live PR body now states the CPU/Metal finite-input split accurately. One test-rustdoc inconsistency should be corrected before merge; it does not undermine the executable regression proof.

### Prior review context

- Round 1 rejected unguarded Metal `sg_buf` reuse, the unguarded shipping CPU SIMD path, an inadequate state-isolation test, a soft kernel gate, overbroad parity wording, and incomplete publication hygiene.
- Round 2 confirmed the named round-1 fixes but found eight additional `sg_buf` WAR handoffs, production convolution history excluded from the state test, missing reproducible barrier benchmark evidence, and incomplete current-head hygiene.
- The round-2 CPU `> 1e-12` request is withdrawn. The authoritative requirement preserves CPU epsilon-regularized finite arithmetic and limits the shared backend contract to non-finite whole-vector zero-by-assignment.

### What this round examined

- Both posted review rounds, issue #850, fix commit `0fc3f3a84`, merge head `3cc4bead2`, the live PR body, and current GitHub checks.
- All 126 indexed accesses under all seven `threadgroup float sg_buf[4]` declarations in `qwen35.metal`, including every row in the live audit table.
- The serial fused test fixture, `GatedDeltaNetState::snapshot`, `apply_causal_conv1d`/`conv1d_silu_fused`, the multi-token Metal fixture, and the actual chunk materialization indexing.
- Submitted-head targeted tests, a Metal invalid-arm reverse mutation in a disposable worktree, fmt/clippy/package checks, and a submitted-head Criterion spot-check.

#### Mediums

1. [`crates/inference/src/attention/gdn_fused.rs:1991`](crates/inference/src/attention/gdn_fused.rs#L1991) The regression test's assertion summary contradicts its actual, correct `conv_buffer` assertions.
   - **Current**: the rustdoc says every step has a “finite complete state” and that the complete `state.snapshot()` is bit-identical at every step. The test correctly proves the opposite for raw storage at lines 2104–2197: the poisoned `conv_buffer` contains exactly one NaN for post-step snapshots 0–2 and differs from the zero reference there; only output and `s_matrices` stay finite/bit-identical throughout, and the complete snapshot becomes finite and bit-identical after eviction.
   - **Issue**: this does not weaken the executable test, and the live PR body describes the split correctly, but the proof summary overclaims what the test establishes and directly contradicts the detailed comment and assertions below it.
   - **Fix**: rewrite the two bullets to say (1) output and `s_matrices` are finite and bit-identical at every step, (2) the one expected raw NaN is tracked in `conv_buffer` for snapshots 0–2, and (3) the complete snapshot is finite and bit-identical from snapshot 3 onward.
   - **Test**: documentation-only; the existing `state_isolation` filter already exercises the behavior.
   - **Confidence**: derived from the fixture and assertions at lines 2038–2197, and observed in the passing submitted-head test.

- **Whole-file WAR audit**: the live body contains 25 data rows (rather than the instruction's stated 24); all 25 classifications are sound. `gdn_recurrence_fused` and `_q36` each have protected beta→alpha, alpha→Q, Q→K, K→k-dot, and k-dot→RMS handoffs. `gdn_precompute_keys` has protected beta→alpha, alpha→Q, Q→K, and K→k-dot handoffs, with no later `sg_buf` write after its k-dot broadcast. The two norm-only kernels use one write/read cycle. The chunk materialize and solve kernels put a full-threadgroup barrier after each `tid == 0` aggregation before any later reuse, including the next loop iteration. The discarded `k2`/`qd` writes have no consumer and therefore no WAR edge.
- **Barrier uniformity**: all eight new barriers are outside lane/thread-divergent control flow. The only preceding exits depend on threadgroup-uniform `h`/`tgpig`, and no new barrier sits inside `tid == 0` or `tid < dim` branches.
- **State-isolation construction**: `Qwen35Config::qwen35_2b()` remains unmodified and asserts `linear_conv_kernel_dim == 4`. The explicit-zero K window is a valid oracle for the contract under test: invalid K is specified to behave as whole-vector K=0. It does not pretend to match a genuinely nonzero, never-poisoned K update. The serial test tracks the expected raw NaN for exactly three post-step snapshots, keeps output and `s_matrices` finite/bit-identical throughout the four affected convolution outputs, and proves complete-state convergence after eviction before four ordinary-weight tail steps. The Metal test exercises tokens 0–3 through actual absolute-index convolution taps and tokens 4–7 after eviction; identical materialized Q/K plus unchanged V/beta/log-alpha inputs subsume the downstream deterministic chunk recurrence.
- **Observed tests at `3cc4bead2`**:
  - `cargo test ... --lib gdn -- --test-threads=1`: 98 passed, 0 failed.
  - `cargo test ... --lib fail_closed -- --test-threads=1`: 3 passed, 0 failed.
  - `cargo test ... --lib state_isolation -- --test-threads=1`: 3 passed, 0 failed.
- **Observed mutation**: in a detached disposable worktree, replacing only `out_k = k_valid ? ... : 0.0f` with the old multiply-through-zero form and touching `qwen35.metal` plus its `include_str!` Rust source made the multi-token Metal test fail with four NaNs, one at each affected token. The disposable worktree was removed; the same test then passed on the unmodified submitted head.
- **Benchmark evidence**: commit `0fc3f3a84` changes the shader by exactly 26 added lines: eight barriers plus their comments. Reversing that isolated commit diff therefore captures the round-2 barrier delta. The body gives the exact Criterion command, workload/checkpoint, both means and CIs, and the +2.44% point delta without claiming a proven zero cost. My submitted-head spot-check measured 6.4499 ms, 95% CI `[6.0759, 6.8180]`, `p=0.47`, “no change detected,” closely matching the body's 6.4444 ms `[6.0462, 6.8368]`. The reported within-CI point increase is acceptable and honestly characterized.
- **CPU/Metal wording**: the live body explicitly preserves CPU's finite epsilon-regularized formula, documents Metal's separate `> 1e-12` cutoff, limits the shared contract to non-finite whole-vector zero-by-assignment, and disclaims finite-input cross-backend bit parity.
- **Hygiene**: the live body starts with the exact PR-description attribution, includes all four required AI checklist statements, and the authored commits have the Leo co-author trailer. After a fresh fetch, `origin/main` (`5a780da2`) is an ancestor of `HEAD`. `cargo fmt --all -- --check`, `cargo clippy -p lattice-inference --features metal-gpu,f16 --tests -- -D warnings`, `git diff --check`, and `cargo package -p lattice-inference --list` passed.
- **CI snapshot at head `3cc4bead2`**: all GitHub-required checks are green and 26 checks total have succeeded, including `e2e parity` and informational `metal attention parity`. No check has failed. Only `q4 composed golden` was still running at the final snapshot.

### What I did NOT check

- I did not run full workspace tests/builds or a real checkpoint generation locally; the targeted crate tests and current CI cover the requested scope.
- I spot-checked only the submitted-head benchmark leg, not an independent second source-swap A/B. The isolation judgment comes from the exact 26-line fix diff plus the written paired evidence.
- `q4 composed golden` had not completed at the final snapshot; there were no red checks, and all required checks were green.

- Correct the two stale rustdoc bullets and require the remaining current-head checks to finish green. No further broad review round is useful unless that edit or a main merge materially changes the code.

Domain utility: MEDIUM — concurrency-ordering and benchmark-uncertainty guidance reinforced the source-order proof and paired-measurement standard; the decisive evidence came from repository-specific shader/state tracing and mutation testing.

Verdict: APPROVE-WITH-FIXES

---
**Fix applied at `12e89f9a3`** (doc-only): the state-isolation test's assertion-summary bullets now match the actual assertions — output + s_matrices finite and bit-identical every step; the one tracked raw NaN pinned in conv_buffer for post-step snapshots 0-2; complete snapshot finite + bit-identical from snapshot 3 onward.

oceanwaves630 (approved):
Reviewed the full diff at 12e89f9a3.

The Metal fixes are the core and they hold: all twelve sg_buf[0] write-after-read handoffs across gdn_recurrence_fused, gdn_recurrence_fused_q36, and gdn_precompute_keys now carry threadgroup barriers, each documented with the specific hazard (a fast simdgroup clobbering slot 0 while a lagging one still reads the previous reduction). The fail-closed contract is applied uniformly in all four kernels: validity is isfinite && > 1e-12 and the invalid case assigns the literal 0.0f to the whole vector directly — the NaN * 0.0f == NaN rationale is stated at every site, which is exactly what makes the pre-fix multiply-through-a-zeroed-reciprocal code class visible to the next reader.

The CPU side is correctly scoped: l2_normalize_vec, the scalar/AVX2/NEON fused backends, and l2_normalize_rows gain only the non-finite whole-vector-zero guard, with finite arithmetic numerically unchanged (pinned by the ordinary/very-small cases in each table test), and the per-backend near-zero contract difference is documented rather than papered over — the gdn_fused test doc explicitly warns against reading it as evidence of a shared contract.

The test methodology is the strongest part. The poisoned-vs-explicit-zero-reference construction makes bit-identical the acceptance criterion, which catches both failure modes at once (a leaked NaN and a guard taking a numerically different path than true-zero arithmetic). The production kernel_dim=4 state-isolation test accounts for the conv_buffer raw-storage taint window exactly — one transient NaN resident for buf_len snapshots, everything else bit-identical, full unconditional convergence after — instead of either asserting something the code doesn't do or quietly sidestepping the window with a degenerate kernel. Device absence is the only skip; the precompute_keys pipeline lookup correctly hard-fails on a link failure rather than skipping past it.

The barriers are correctness-required, and the bench evidence shows no significant regression (+2.44% point, within CI). Approve.

PR #858, opened 2026-07-11
refactor(tune): validated TrainCtx replaces ten-argument train-tape signatures

ohdearquant:
> _PR description authored by Claude (Anthropic agent) on behalf of @ohdearquant._

## What

Introduces `TrainCtx` (+ `TapeGeometry`, `LoraExecution`, `SlotLayout`,
`AdamConfig`) in `crates/tune/src/lora/train_core.rs`, replacing the
ten-argument positional signatures on `forward_full`, `eval_chain_nll`,
`nll_and_grads`, `apply_adam_updates`, and `apply_gdn_adam_updates` with the
narrowed shapes from #844:

```rust
forward_full(seq, layers, gqa_loras, gdn_loras, head, train)
eval_chain_nll(seqs, layers, gqa_loras, gdn_loras, head, train)
nll_and_grads(fwd, layers, gqa_loras, head, train)
apply_adam_updates(adam, gqa_loras, grads, train)
apply_gdn_adam_updates(adam, gdn_loras, grads, train)
```

`TrainCtx::try_new` validates:
- `rank > 0`
- `alpha` and every Adam hyperparameter (`learning_rate`/`beta1`/`beta2`/`epsilon`) finite
- `Dims.eps` and `GdnDims.scale` finite (checked explicitly, ahead of the tolerance
  comparison below — `(a - b).abs() > 1e-9` is silently `false` for NaN and would
  otherwise let non-finite geometry through)
- `Dims`/`GdnDims` agree with `TapeGeometry.model` (recomputed via `GdnDims::from_cfg` and
  the same field derivations every call site already used)
- every GQA slot layer is actually a full-attention layer on `model`
  (`model.is_full_attention(li)`), and every GDN slot layer is actually a
  GatedDeltaNet layer (`!model.is_full_attention(li)`) — not just "unique and
  in-range," the real per-layer mixer kind
- GQA/GDN slot-layer indices are each unique, in-range for `model.num_hidden_layers`, and
  never claim the same layer index for both mixer kinds (mixer-kind/slot-index agreement)

`scale = alpha / rank` is derived once inside `try_new` and held privately in
`LoraExecution` — never stored as `alpha`, never exported. All of `TrainCtx`'s
internal accessors (`gqa_layers`/`gdn_layers`/`rank`/`scale`/`dims`/`gdn_dims`/
`model`/the four Adam getters) are module-private `fn`, not `pub fn` — every
consumer is inside `train_core.rs`. The only public surface issue #844 asks
[...]

Issue #737, opened 2026-07-09
tune: train_grad_full cannot consume large corpora — full-tape residency, 2 GiB logits cap, scalar backward loops


ohdearquant:
**Problem.** `crates/tune/src/bin/train_grad_full.rs` (and its line-equivalent library duplicate `lora/train.rs::train_micro_lora`) is architecturally capped at ~15-19 samples per invocation and computationally far too slow to train over a corpus of thousands of rows:

1. **Full-tape residency**: every loaded sample's frozen-prefix output and forward/backward tape stays resident for the whole run — no minibatch streaming, no checkpoint-resume.
2. **Hard 2 GiB completion-logits cap**: `total_positions * vocab * 4` is bounded per-invocation, which at vocab=248320 caps total completion positions at ~2162 across all cached samples.
3. **Scalar backward**: the backward tape (`crates/tune/src/lora/train_core.rs`, `crates/inference/src/backward/{tape.rs,ops.rs}`) uses naive `zip().map().sum()` scalar loops, while the inference forward path uses tiled/NEON matmul (`crates/inference/src/forward/cpu/{matmul.rs,tiled_neon.rs}`). Measured on a 12-core M2 Max: ~125 s/sample frozen-prefix build at ~1000 tokens (19 frozen layers), and a single training step over 4 samples at seq-len 1100 exceeds 16 CPU-minutes. Extrapolated, one epoch over 8000 rows at seq~1300 is a multi-week sequential run.

**Proposed levers, in priority order:**

- **(a) Minibatch/epoch streaming with h_in reuse** (~1-1.5 d): move LoRA/Adam state outside a per-wave loop, build and discard the frozen-prefix cache per chunk, make the 2 GiB guard per-wave, add periodic `--save` checkpointing. Cache each row's frozen-prefix output once and reuse across epochs (pure function of frozen weights + tokens).
- **(b) Rayon-parallel gradient accumulation** (~0.5-1 d): each sample's `forward_full`/`nll_and_grads` is a pure function over `&[LayerW]`/`&[LoraParams]`/`&SeqCtx` with no shared mutable state until the gradient sum — parallelize across a resident batch for up to ~core-count speedup on both the prefix-build and step phases. Low correctness risk; the existing gradcheck/TBV harness (chain-vs-model NLL agreement to ~6e-7) is the safety net.
[...]

PR #864, opened 2026-07-11
refactor(tune): extract train_common, fail-closed train_grad TBV (#845)

### LINKED ISSUE #845: refactor(tune): extract shared training-bin harness, make TBV fail-closed

## What

Extract the genuinely duplicated pieces of the three `crates/tune/src/bin/train_grad*`
binaries into one binary-private support module, and make the TBV (trust-but-verify)
check in `train_grad.rs` fail closed instead of print-only.

Verified on current `main`:

- `Sample`/`load_jsonl` bodies in `train_grad.rs`, `train_grad_full.rs`, and
  `train_grad_layer23.rs` are byte-identical.
- The `parse_arg`/`parse_flag` helper blocks in all three bins are byte-identical.
- `train_grad.rs`'s TBV check only prints the cached-vs-model NLL difference and
  never asserts or returns an error — a `NaN` or arbitrarily large mismatch does
  not fail the run:

  ```rust
  // train_grad.rs
  println!(
      "  cache check (sample 0): model={model_masked:.5}  cached={cached_masked:.5}  diff={:.2e}",
      (model_masked - cached_masked).abs()
  );
  ```

  `train_grad_full.rs` and `train_grad_layer23.rs` already assert `diff > 1e-2` as
  a hard failure for their own oracle; `train_grad.rs` is the one gap.

## Why

Three copies of the same loader and arg-parsing logic mean a fix to one (or a new
flag) silently does not apply to the others unless someone remembers to copy it
three times — the exact failure mode that produced the redundant 744-line
`train_grad_layer23.rs` trainer in the first place (see the companion issue in this
series covering its removal). The print-only TBV check in `train_grad.rs` is a
non-vacuous contract with no enforcement: it protects the same invariant as the
other two bins' hard-fail checks (the training forward must agree with
`Qwen35Model::compute_token_nlls`) but currently lets a broken cache pass silently,
including a `NaN` difference.

## Design (binary-private, not a public library surface)

Create `crates/tune/src/bin/train_common/mod.rs` as a directory module (so Cargo
does not discover it as an additional binary target). Scope, deliberately narrow:

1. `Sample` and the single `load_jsonl` implementation.
2. An `ArgView` preserving current lookup semantics exactly: first matching flag
   wins; a missing/invalid value falls back exactly as it does today; presence
   flags remain presence-only; unknown flags remain ignored in this PR (a later,
   separately-decided PR may adopt strict parsing — see Non-goals).
3. Shared path/default helpers.
4. `verify_tbv(context, reference_nll, candidate_nll) -> Result<TbvObservation, String>`
   — the single owner of `TBV_MAX_ABS_DIFF = 1e-2`, finite-value checks for
   reference/candidate/difference, absolute-difference calculation, and
   fail-closed error construction.

Do **not** put a generic usage string or one universal `CliConfig` in this module.
Each bin keeps its own typed config and declarative flag table, because defaults
and supported flags genuinely differ between bins:

| Binary | Flags and defaults that must remain stable across this migration |
|---|---|
| `train_grad` | `--model-dir` HOME model path; `--data-dir data/lora-train`; `--steps 150`; `--lr 1e-3`; `--rank 8`; `--alpha 16`; `--seq-len 96`; `--max-train 6`; `--log-every 10`; `-h/--help` |
| `train_grad_layer23` (still the standalone 744-line implementation at this point in the series) | Same paths; `--steps 25`; `--lr 1e-3`; `--rank 8`; `--alpha 16`; `--seq-len 64`; `--max-train 3`; `--log-every 5`; `--save`; `-h/--help` |
| `train_grad_full` | Same paths; `--first-layer 19`; `--steps 25`; `--lr 1e-3`; `--rank 8`; `--alpha 16`; `--seq-len 64`; `--max-train 3`; `--max-valid 16`; `--log-every 5`; `--save`; `--gradcheck`; `--probe 6`; undocumented-but-live `--fd-eps 4e-3`; `-h/--help` |

`--fd-eps` on `train_grad_full` must either be documented in `--help` while being
preserved, or explicitly removed as a versioned CLI change — silently dropping it
during this consolidation is out of scope for this PR.

## Non-goals (explicit fences)

- Do not remove `train_grad_layer23.rs`'s standalone implementation in this PR —
  it still exists at this point in the series (that removal is a later PR in this
  series, gated on the golden-bridge comparison landing first). Migrate it onto the
  shared loader/parser/TBV helpers only.
- Do not change parser strictness (unknown/duplicate/invalid-value handling stays
  exactly as today). A later, separately decided release may adopt fail-closed
  parsing; do not bundle that behavior change here.
- Do not change any flag default, initialization amplitude, or numeric behavior.

## Checklist

- [ ] Add `crates/tune/src/bin/train_common/mod.rs` with `Sample`/`load_jsonl`,
      `ArgView`, shared path/default helpers, and `verify_tbv`.
- [ ] Migrate `train_grad.rs`, `train_grad_full.rs`, `train_grad_layer23.rs` to the
      shared loader, arg-parsing, and `verify_tbv` — no behavior change to any
      already-passing case.
- [ ] Make `train_grad.rs`'s cache-vs-model check call `verify_tbv` and return an
      error on mismatch instead of only printing the difference.
- [ ] Add finite/NaN/Inf/boundary unit tests for `verify_tbv`: reference NaN,
      candidate NaN, difference NaN/Inf, difference exactly at the `1e-2` boundary,
      difference just above the boundary.
- [ ] Add CLI-contract snapshot tests per bin verifying the flag table and defaults
      in the table above are unchanged.
- [ ] Update the stale duplication note in issue #737 (its own text currently says
      `train_grad_full.rs` has a "line-equivalent library duplicate" in
      `lora/train.rs::train_micro_lora`; both now call the shared `train_core`, so
      that specific sentence is stale) — leave a comment on #737 noting the
      correction rather than rewriting its body in this PR.

## Gates

- Both full-depth central-difference modes (`train_grad_full --gradcheck`,
  `train_grad_layer23`'s own gradcheck path if present) continue to pass unchanged.
- Real-model TBV passes for all three bins post-migration.
- Mutation gate: remove the `verify_tbv` call from `train_grad.rs` (or corrupt one
  cached NLL value before the check) and confirm the run now fails instead of
  printing a silent diff.
- `cargo test -p lattice-tune` and the CLI-contract snapshot tests pass.

## Sequencing

Second implementation PR in this series; depends on the first (`TrainCtx`)
having landed so the migrated bins call the narrowed signatures, not the ten-argument
ones. Must land, and its snapshot/TBV tests must be green, before the layer-23
golden-bridge and deletion units in this series proceed.

ohdearquant:

- [x] SIMD intrinsics, if any, were manually verified — not just reviewed by an AI (none in this diff)
- [x] `cargo test` output included for any behavior-changing code; `cargo bench` output for perf-sensitive changes (bench N/A — no inference/embed source touched)

## Summary

Closes #845 (second implementation PR in the TrainCtx/train_common series;
depends on #858, which is on `main`).

Extracts the genuinely-duplicated pieces of `train_grad.rs`,
`train_grad_full.rs`, and `train_grad_layer23.rs` into a binary-private
support module, and makes `train_grad.rs`'s TBV (trust-but-verify) check
fail closed instead of print-only.

- `crates/tune/src/bin/train_common/mod.rs` (directory module — Cargo does
  not discover it as an extra `[[bin]]` target; verified via `cargo
  metadata`, still exactly 4 bin targets):
  - `Sample` / `load_jsonl` — byte-identical across all three bins on `main`.
  - `ArgView` — preserves the inline `parse_arg`/`parse_flag` lookup
    semantics exactly: first matching flag wins, missing/invalid value
    falls back as before, presence flags stay presence-only, unknown flags
    stay silently ignored (parser-strictness changes are explicitly out of
    scope per the issue).
  - `default_model_dir` / `default_data_dir`.
  - `verify_tbv(context, reference_nll, candidate_nll) -> Result<TbvObservation, String>`
    — the single owner of `TBV_MAX_ABS_DIFF = 1e-2`, with finite/NaN/Inf
    guards on the reference NLL, the candidate NLL, and their difference.
- All three bins migrated onto the shared loader/`ArgView`/`verify_tbv`.
  Each bin keeps its own typed `Config` + `parse_config` (defaults/flags
  genuinely differ per bin — the flag table in #845) and its own `usage()`
  text, per the issue's explicit non-goal against one universal `CliConfig`.
- `train_grad.rs`'s cache-vs-model TBV check now calls `verify_tbv` and
  propagates the error via `?` instead of only printing the diff —
  `train_grad_full`/`train_grad_layer23` already hard-failed on this;
  `train_grad` was the one gap the issue identified.
- `train_grad_full.rs`: `--fd-eps` (previously undocumented-but-live) is
  now documented in `usage()`; value and behavior unchanged.
- Left a correcting comment on #737: its "line-equivalent library
  duplicate" claim is stale post-#858 — `train_grad_full` and
  `lora/train.rs::train_micro_lora` both call the shared `train_core` now,
  not diverged copies. https://github.com/ohdearquant/lattice/issues/737#issuecomment-4943083131

Non-goals honored: `train_grad_layer23.rs`'s standalone forward/backward
implementation is untouched beyond the loader/parser/TBV migration; no
flag default, init amplitude, or numeric behavior changed.

## Test plan

- [x] `cargo fmt -p lattice-tune -- --check` clean.
- [x] `cargo clippy -p lattice-tune --bin train_grad --bin train_grad_full --bin train_grad_layer23 --features train-backward -- -D warnings` clean.
- [x] `cargo test -p lattice-tune --features train-backward`: 271 library
      tests (270 passed, 1 pre-existing ignored) + 16/19/16 bin tests
      (train_grad/train_grad_full/train_grad_layer23, each including
      train_common's own unit-test suite compiled once per bin per Rust's
      binary-private-module model) + 7 `proptest_softmax` tests + 50
      `training_pipeline` integration tests + doctests — all executed
      targets pass.
- [x] CLI-contract snapshot tests per bin (`defaults_match_documented_table`,
      `explicit_flags_override_defaults`, `help_flags_detected_via_arg_view`)
      pin the exact defaults/flags from #845's table.
- [x] `verify_tbv` unit tests cover: reference NaN, candidate NaN, reference
      Inf, candidate Inf, diff-overflow-to-Inf, pass within tolerance, pass
      exactly at the `1e-2` boundary, fail just above the boundary.
- [x] **Mutation gate** (real model + real data, `data/lora-overfit1`):
      reverse-applied the fix by corrupting `train_grad`'s cached NLL
      (`+ 1.0`) before the TBV call (cp-backup, not `git checkout`, per
      the mutation-sensitive-test discipline). Before restoring:
      ```
      Error: "TBV failed (train_grad cache check (sample 0)): diverges from real model by 9.994e-1 (> 1e-2)"
      ```
      exit code 1 — confirms the previously-silent gap is now fail-closed.
      Restored from backup (byte-identical diff confirmed empty) and
      re-ran: passes again with `diff=5.71e-4`.
- [x] Real-model TBV passes post-migration, all three bins (no behavior
      change to the already-passing case):
      - `train_grad`: `diff=5.71e-4`
      - `train_grad_layer23`: `diff=9.54e-7`
      - `train_grad_full` (layers 19..=23): `diff=4.29e-6`
- [x] `--help` output for all three bins matches the documented flag table
      in #845 exactly (verified by direct run, not just source reading).
- [x] `train_grad_full --gradcheck` (real model/data, layers 22..=23,
      release build): TBV passed (`diff=2.86e-6`) and the gradcheck harness
      itself runs and reports correctly — first probed array
      `layer 23 slot 0 a_q: mean 7.55e-3  max 1.45e-2  ok`. The full
      probe sweep (GQA + GDN arrays) is expensive per finite-difference
      entry (4 central-diff evals × 2 probed entries × ~13 arrays) and
      did not finish inside the bounded check window; this PR does not
      touch `train_core.rs`'s forward/backward math (only
      loader/arg-parsing/TBV plumbing), so the untouched gradcheck path's
      correctness is not in question — the harness itself is confirmed
      live and its first result matches expectations.

No `crates/inference/src` or `crates/embed/src` changes — `make bench-compare`
not applicable.


ohdearquant:
## Codex review round 1 — Verdict: APPROVE-WITH-FIXES

## Review of PR #864 (round 1)

The implementation preserves the three binaries' prior loader/CLI/default behavior and correctly makes `train_grad` fail closed, including for NaN/Inf. I found two medium test/publication defects to fix before merge; neither indicates a runtime correctness bug in the committed verifier.

### Prior review context

- No prior reviews or review threads were present when this round began.
- This round examined the full diff at `1c18a32bdd618233aa5b8973f61c443928a38b9f` against issue #845 and merge-base `e6e5548187fc01949ec42d7e8fd49a9b99f1dbc7`.

### What this round examined

- Compared every old `Sample`/`load_jsonl`, `parse_arg`/`parse_flag`, default path, per-bin config/default, validation-loader, and TBV call site with the extracted implementation.
- Checked finite-value handling and the `train_grad` error-propagation/exit path.
- Swept `crates/tune` and all bin sources for unmigrated JSONL/argument/TBV siblings.
- Read all four changed files in full and checked Cargo target discovery, stubs, lint suppressions, diff scope, tests, CI, freshness, attribution, checklist, and commit trailer.

### Findings (2 total: 0 blockers, 0 majors, 2 mediums)

#### Mediums

1. [`crates/tune/src/bin/train_common/mod.rs:231`](crates/tune/src/bin/train_common/mod.rs#L231) — The claimed exact-boundary test does not reach the boundary.
   - **Current**: `verify_tbv("ctx", 1.0, 1.0 + TBV_MAX_ABS_DIFF)` followed by an approximate assertion with `1e-6` slack.
   - **Issue**: f32 rounding makes that subtraction slightly less than `TBV_MAX_ABS_DIFF`, so the test does not enforce the required inclusive pass at exactly `1e-2`. In a disposable worktree I mutated production code from `diff > TBV_MAX_ABS_DIFF` to `diff >= TBV_MAX_ABS_DIFF`; `verify_tbv_passes_exactly_at_boundary` still passed. This leaves the issue #845 boundary contract unprotected.
   - **Fix**: construct an exactly representable comparison using the constant itself, for example `verify_tbv("ctx", TBV_MAX_ABS_DIFF, 0.0)`, assert `obs.diff == TBV_MAX_ABS_DIFF`, and confirm the `>=` mutation fails the test.
   - **Test**: `cargo test -p lattice-tune --features train-backward --bin train_grad verify_tbv_passes_exactly_at_boundary` must fail under the `>=` mutation and pass at HEAD.
   - **Confidence**: observed by running that mutation in `/tmp/lattice-864-r1-boundary`, then removing the disposable worktree.

2. [`.github/PULL_REQUEST_TEMPLATE.md:31`](.github/PULL_REQUEST_TEMPLATE.md#L31) — The PR body uses a non-template AI checklist and misreports the aggregate test output.
   - **Current**: the body substitutes a five-item generic checklist for the required four items at lines 36–39 and says `cargo test -p lattice-tune --features train-backward` ran “50 lib tests + 16/19/16 bin tests.”
   - **Issue**: the exact command at this head actually ran 271 library tests (270 passed, 1 ignored), 51 bin tests, 7 `proptest_softmax` tests, 50 `training_pipeline` integration tests, and doctests. The “50 lib tests” are the 50 `training_pipeline` integration tests, not the library suite. This conflicts with the repository policy that every PR-body claim be accurate, and the AI checklist does not match the required template.
   - **Fix**: replace the checklist with the exact four required items from lines 36–39 and update the test-plan result to the observed target/count breakdown (or paste the command's actual summary without relabeling the 50 integration tests as library tests).
   - **Test**: re-read the rendered PR body against `.github/PULL_REQUEST_TEMPLATE.md` and the saved command output; no code test is needed.
   - **Confidence**: observed by running the exact aggregate command and comparing the live PR body with the committed template.

- `Sample` field meaning and `load_jsonl` behavior are preserved: malformed JSON still errors, empty/missing fields still skip, prompt/full tokenization and completion boundaries are unchanged, and `max_samples` still limits accepted rows rather than input lines.
- `ArgView` preserves first-match precedence, dangling/missing values, invalid-value fallback at each caller, presence-only flags, duplicate handling, and ignored unknown flags.
- Every documented default remains unchanged for `train_grad`, `train_grad_full`, and `train_grad_layer23`; `--fd-eps 4e-3` remains live and is now documented.
- `train_grad_full` still treats absent, malformed, or empty `valid.jsonl` as “eval disabled”; extraction did not make validation loading stricter.
- `verify_tbv` rejects non-finite reference, candidate, and subtraction result before checking `diff > 1e-2`. Rust's [`f32::is_finite`](https://doc.rust-lang.org/stable/core/primitive.f32.html#method.is_finite) is false for both NaN and infinity. All three callers propagate `Err` with `?`.
- `train_grad` is the only ordinary over-tolerance behavior change: it previously only printed. The other two binaries already returned an error above `1e-2`; their additional finite guards are disclosed by the shared-verifier description.
- Sibling sweep found no remaining train-grad JSONL/TBV copy. `generate_lora.rs` retains unrelated generic argument helpers but is neither a training JSONL consumer nor a TBV caller.
- `cargo metadata --no-deps --format-version 1` reports exactly four tune binaries; `train_common/mod.rs` is not discovered as a fifth target. The module compiled and ran its tests independently in each of the three consuming binaries.
- No `todo!`, `unimplemented!`, placeholder, dead-code suppression, or unused-code suppression exists in the changed files.
- Observed local gates:
  - `timeout 1200 cargo test -p lattice-tune --lib`: 182 passed.
  - `timeout 1200 cargo test -p lattice-tune --features train-backward --bin train_grad --bin train_grad_full --bin train_grad_layer23`: 16 + 19 + 16 passed.
  - `timeout 1200 cargo test -p lattice-tune --features train-backward`: all executed targets passed; one pre-existing library test and doctests were ignored as reported above.
  - `timeout 1200 cargo fmt -p lattice-tune -- --check`: passed.
  - `timeout 1200 cargo clippy -p lattice-tune --bin train_grad --bin train_grad_full --bin train_grad_layer23 --features train-backward -- -D warnings`: passed.
- The PR author's real-model corruption mutation is logically discriminating: with a `+1.0` candidate corruption, the current `verify_tbv(...)?` exits 1; the old warning-only comparison would print and continue. A separate disposable-worktree mutation confirmed the committed unit tests alone do not protect this call-site wiring, so the reported real-model mutation supplies meaningful evidence rather than duplicating the unit tests.
- Remote CI required checks were green when reviewed. The branch is one commit behind `main`; the intervening commit only changes inference GEMM validation, and `git merge-tree --write-tree HEAD origin/main` completed without conflicts.
- The diff against current `main` contains only the four claimed `crates/tune/src/bin/` files. No inference/embed source changed, so bench-compare and e2e parity are N/A by construction.
- The PR-body first line uses the exact `PR description authored by ...` attribution form. The commit includes `Co-Authored-By: Leo <noreply@khive.ai>`. The claimed correction comment on issue #737 exists.

### What I did NOT check

- I did not independently rerun the three real-model TBV commands or the partial `train_grad_full --gradcheck` sweep because `data/lora-overfit1` is absent from this worktree. The PR body already discloses that the full multi-array sweep did not finish.
- I did not run workspace-wide builds/tests or performance benchmarks; the requested crate/bin gates cover this tune-binary-only diff, and bench-compare is not applicable.

- Fix the exact-boundary test and PR-body evidence/checklist, then verify those two changes. Another full review round is not useful unless the code changes materially.

Domain utility: MEDIUM — the Rust CLI/interface guidance reinforced stable boundary semantics and lint scope, while issue #845 and the old source remained the decisive authorities.

Verdict: APPROVE-WITH-FIXES

---
**Fixes applied at `12ab83ef5`**: (1) boundary test rewritten to the bit-exact pair `verify_tbv("ctx", TBV_MAX_ABS_DIFF, 0.0)` with `obs.diff == TBV_MAX_ABS_DIFF`; mutation re-verified — the `>=` mutation now FAILS the test, restored `>` passes. (2) PR body: AI checklist replaced with the exact four template items, and the aggregate test line corrected to 271 library tests (270 passed, 1 pre-existing ignored) + 16/19/16 bin + 7 proptest_softmax + 50 training_pipeline integration + doctests.

oceanwaves630 (approved):
Reviewed the full diff at 12ab83ef5.

The extraction is faithful: the removed inline copies of `load_jsonl`, `parse_arg`/`parse_flag`, and `default_model_dir` were byte-identical across the three bins, and `ArgView` preserves the exact lookup semantics (first match wins, `.get(i+1)` for dangling flags, presence-only flags, no added validation — correctly scoped per #845's non-goals). Each `parse_config`'s defaults match the deleted inline defaults and are pinned by the snapshot tests, including `train_grad_full`'s `--log-every >= 1` rejection and the previously undocumented `--fd-eps` now surfaced in `usage()`.

The behavior changes are exactly the intended ones: `train_grad`'s cache check goes from print-only to fail-closed, and all three bins now fail on non-finite values — under the old `diff > 1e-2` check, a NaN diff compared false and passed silently, so the finite guards close a real hole. The dedicated diff-overflow test (`f32::MAX` vs `f32::MIN` → +inf with both inputs finite) exercises the third guard distinctly. The boundary test is bit-exact: `(TBV_MAX_ABS_DIFF, 0.0)` subtracts to the constant precisely in f32, so it genuinely distinguishes `>` from `>=` — a `1.0`-offset pair would not. Threshold semantics unchanged (diff == 1e-2 still passes, matching the old strict-greater check in both bins that had one).

`train_common/` uses `mod.rs`, so cargo does not auto-discover it as a bin. Approve.

PR #860, opened 2026-07-11
test(serve): richer field-level parity model + deterministic production-seam harness

ohdearquant:
> _PR description authored by Claude (Anthropic agent) on behalf of @ohdearquant._

## What

Extends the shared `serve/mod.rs` parity-case contract (ADR-080 cluster C2) from
plain `(status, error_code)` tuples to a typed field/event assertion model, and adds
a deterministic canned-generation seam so field-level and SSE-phase assertions can
run against both binaries' real routers.

- `ExpectedResponse` enum: `Error{status,code}` / `Json{status,fields}` /
  `Sse{status,events}`.
- `FieldExpectation`: `Eq{json_pointer,scalar}` / `Absent{json_pointer}` /
  `ArrayLen{json_pointer,len}` / dynamic-field type/prefix checks for `id`/`created`
  — JSON-pointer-addressed body assertions.
- `EventExpectation` + `check_sse_events`: `RoleOpener` / `ContentDelta` (greedy,
  consumes ≥1 matching chunk) / `Finish{finish_reason}` / `Done` — ordered SSE-phase
  assertions over parsed `data:` frames.
- 8 new `CHAT_COMPLETIONS_PARITY_CASES` rows: baseline 200 (non-streaming +
  streaming), temperature boundary accept (0, 2) / reject (out-of-range), top_p
  boundary accept (1) / reject (0, >1).
- `baseline_fake_state` (lattice.rs) / `baseline_fake_worker_state`
  (lattice_serve.rs): deterministic `CpuFakeGenerate`-backed harness states that
  push a canned `"hello" " world"` token stream through the real router, so the new
  `Json`/`Sse` rows get exact, cross-binary-identical field values
  (`BASELINE_CANNED_TEXT`/`PROMPT_TOKENS`/`COMPLETION_TOKENS`).
- `ProductionAdapterObservation` + tests per binary: captures the real
  `GenerateConfig` (via `GenerateConfigSnapshot`, all 13 fields), the real rendered
  prompt or normalized message list, the real tokenized prompt-token count, and the
  real terminal `stopped` outcome that each binary's production request-handling
  code builds — driven through the actual `enforce_prompt_window` / prompt-render /
  tokenizer / worker-loop call path, not a stub. `lattice_serve.rs`'s observation
  runs through the real `run_worker_loop` + `enforce_prompt_window` on a background
  thread; `lattice.rs`'s runs through the real router via `tower::ServiceExt::oneshot`.

## Why

[...]

PR #867, opened 2026-07-11
fix(inference): validate cross-turn prefix suffix before cache-mutating restore

### LINKED ISSUE #835: bug(inference): cross-turn prefix restore mutates cache before validating the suffix

## What

In `crates/inference/src/forward/metal_qwen35.rs`, `generate_streaming_with_prefix_cache_and_cancel_inner` calls `restore_cross_turn_prefix` (which `take()`s the cached entry out of the slot for `ExactAppend`/`ReplayFromCheckpoint`, or calls `reset_state()` for `FullRefill`) before it calls `forward_prefill_from` on the suffix. `forward_prefill_from` is where the suffix's own validation lives: out-of-vocab token ids, an empty suffix, and start-position-plus-length overflow against `max_cache_len`.

Verified on current `origin/main` (commit `86739db68`, the tip after PR #787 merged):

- `restore_cross_turn_prefix` call site: `crates/inference/src/forward/metal_qwen35.rs:15498`
- `forward_prefill_from` call site (suffix prefill): `crates/inference/src/forward/metal_qwen35.rs:15587`
- `forward_prefill_from`'s own preflight checks (out-of-vocab, empty suffix, cache-length overflow): `crates/inference/src/forward/metal_qwen35.rs:15046-15068`

So an out-of-vocab or overflow-length suffix is only rejected *after* the cache entry for that slot has already been destructively consumed (or the whole state reset). A caller that sends an invalid suffix on a warmed slot loses the valid pre-existing cross-turn entry for a request that touches no live KV/GDN state at all.

This is a distinct defect from the request-level preflight PR #787 already added (`check_logprobs_not_set` / `check_mtp_not_requested`, hoisted to the top of the public wrapper `generate_streaming_with_prefix_cache_and_cancel` at `metal_qwen35.rs:15340`, before the `match` that performs destructive recovery on `Err`). Those two guards do not, and structurally cannot, validate the suffix itself — the suffix isn't known until after `plan_cross_turn_reuse` runs, which happens inside `_inner`, after the cache has already been consumed.

## Why

`forward_prefill_from`'s checks are correct in isolation; they are just evaluated too late. The fix is to factor the same three checks (out-of-vocab token ids in the suffix, non-empty suffix, `start_pos + len <= max_cache_len`) into a read-only prefix request plan that runs against `prompt_ids` and the already-computed `plan.suffix_start` *before* `restore_cross_turn_prefix` is invoked. `forward_prefill_from` keeps its own checks (defense in depth for the direct-call path used elsewhere), but the public streaming path never reaches `restore_cross_turn_prefix`/`reset_state` on a suffix it already knows is invalid.

This does not change the destructive-on-mutation contract for grammar failures (that path mutates real KV/GDN state first, so the existing destructive-clear behavior is correct and out of scope here) — it only moves a request-shape check that never needed live state to run before, not after, cache consumption.

## Checklist

- [ ] Add a private, pure `plan_prefix_request` (or extend the existing plan struct) that validates: every suffix token id is within `vocab_size`, the suffix is non-empty, and `plan.suffix_start + suffix.len() <= max_cache_len`.
- [ ] Run this validation in `generate_streaming_with_prefix_cache_and_cancel_inner` immediately after `plan_cross_turn_reuse` computes `plan.suffix_start`, and before the `match plan.mode { ... restore_cross_turn_prefix / reset_state ... }` block.
- [ ] On validation failure, return the same `InferenceError` variant/message shape `forward_prefill_from` currently returns, so this is observationally a pure timing fix, not a new error taxonomy.
- [ ] Leave `forward_prefill_from`'s own checks in place unchanged (still correct for its other, non-prefix-cache callers).
- [ ] Do not touch F3's grammar-error destructive-clear behavior (`save_cross_turn_prefix_or_clear`'s reset-on-`Err` arm) — that remains fail-closed and destructive by design.

## Gates

- Mutation-sensitive regression test: pre-populate a cache entry for a slot (`ExactAppend` or `ReplayFromCheckpoint` reachable), then submit a follow-up request whose suffix contains an out-of-vocab token id (or is empty, or overflows `max_cache_len`). Assert the call errors *and* the cache entry, live KV length, and GDN position are all unchanged from before the call. Reverting the hoist (moving the check back into `forward_prefill_from` only) must turn this test red.
- A second case: after the rejected request, a subsequent *valid* suffix on the same slot must still select the same reuse mode and produce output identical to a full-refill baseline — proving the preserved entry is genuinely intact, not just present-but-stale.
- Full `f16,metal-gpu` library test run under the machine-wide GPU lock (`gpu_test_lock()`), plus the default-feature test run.
- `e2e-parity` gate (greedy first-N-token agreement vs. HF) must remain green — this PR changes only invalid-input cache retention, not any valid-request output.
- No `bench-compare` required: this is a preflight-ordering fix, not a hot-path or throughput change.

## Sequencing

This is the first PR in the cluster, run immediately after PR #787 (already merged) and before any structural module move. No dispatch/layer/prefix-cache file moves may start until this lands, per the design's ordering: behavior/recovery-semantics fixes are pinned before any code is relocated, so a later pure-move PR is not also carrying a live-behavior diff.

`PR 0` from the original design draft (hoisting the `logprobs`/`enable_mtp` preflight above the destructive wrapper) is **not** part of this issue — PR #787 already delivered that exact change (see `check_logprobs_not_set` / `check_mtp_not_requested` at `metal_qwen35.rs:15340-15341`, and the corresponding preserved-cache regression test suite added in the same PR). This issue covers only the suffix-content preflight (the design's "F4"), which #787 did not touch.

ohdearquant:

## Summary

- Cross-turn prefix restore previously mutated/consumed the cache slot (`restore_cross_turn_prefix`'s `take()`, or `reset_state()` for `FullRefill`) *before* the suffix a reuse plan selected was validated. An out-of-vocab suffix token, an empty suffix, or a suffix overflowing `max_cache_len` was only rejected afterward, inside `forward_prefill_from` -- by which point a valid, warmed cache entry for that slot had already been destroyed (or the public wrapper's blanket `Err` cleanup evicted it).
- Adds a pure `plan_prefix_request` preflight (mirrors `forward_prefill_from`'s own checks byte-for-byte: same `InferenceError::InvalidInput`/`InferenceError::PrefixCache` variants and message text) and runs it in the public wrapper `generate_streaming_with_prefix_cache_and_cancel`, before `_inner` is ever called -- the same pattern PR #787 already established there for the `logprobs`/`enable_mtp` preflights, and necessary here too: `_inner`'s caller unconditionally calls `reset_state()` + evicts the cache slot on *any* `Err` from `_inner`, so a suffix-validation-only rejection has to return before `_inner` runs at all, not just before `_inner`'s own `restore_cross_turn_prefix`/`reset_state` match.
- `_inner` also runs the identical check immediately after its own `plan_cross_turn_reuse` (defense in depth, matching the issue's requested call site). Tokenization happens once (moved to the wrapper; `_inner` now takes `prompt_ids: Vec<u32>`), but the defense-in-depth choice means the pure metadata/`plan_cross_turn_reuse`/`plan_prefix_request` pass runs twice on every request -- once in the wrapper, once again inside `_inner`. That duplicated pass is CPU-only integer arithmetic plus a read-only suffix scan (no cache mutation, no GPU work).
- `forward_prefill_from`'s own checks are unchanged (still correct defense-in-depth for its other, non-prefix-cache callers). The grammar-error destructive-clear path (`save_cross_turn_prefix_or_clear`'s reset-on-`Err` arm) is untouched.

## Why

Closes #835. See the issue for the original repro/trace; this PR implements its checklist with one addition beyond the issue's literal text: the preflight had to live in the *public* wrapper (not only inside `_inner`) to actually satisfy the issue's own gate ("the cache entry... unchanged"), because `_inner`'s single caller unconditionally wipes cache/state on any `Err` it returns -- a suffix-validation error raised from inside `_inner` alone would still be destroyed by that caller. Placing the check in the wrapper first (mirroring #787's already-established fix for the same class of problem) avoids that.

## Mutation-cycle evidence

`generate_streaming_with_prefix_cache_rejects_empty_replay_suffix_preserves_warm_entry` is the mutation-sensitive regression test the issue's gate asks for. It warms a slot with `"abcdef"`, injects a synthetic GDN checkpoint at `len=2` on the retained entry (simulating a retained earlier-turn boundary), then sends a follow-up `"ab"` whose planner-selected suffix is empty (`ReplayFromCheckpoint { checkpoint_len: 2 }`, `suffix_start == 2 == new_prompt_ids.len()`).

Revert -> red, restore -> green, performed by hand this session (reverse-apply, not `git checkout`, per repo convention):

```
$ python3 - <<'EOF'
# commented out both `self.plan_prefix_request(&prompt_ids, &plan)?;` call sites
# (the public-wrapper one and the _inner defense-in-depth one)
EOF
$ cargo test -p lattice-inference --features metal-gpu,f16 --lib -- rejects_empty_replay_suffix
...
thread '...rejects_empty_replay_suffix_preserves_warm_entry' panicked at ...:
assertion `left == right` failed: an empty-suffix rejection must not evict or consume the warm entry -- validation must run before restore_cross_turn_prefix's take()
  left: None
 right: Some([0, 1, 2, 3, 4, 5, 1])

# restored the two call sites
$ cargo test -p lattice-inference --features metal-gpu,f16 --lib -- rejects_empty_replay_suffix
test result: ok. 1 passed; 0 failed
```

The second gate requirement (a subsequent *valid* suffix on the same slot still selects `ExactAppend` and matches a full-refill baseline bit-for-bit) is asserted in the same test.

Three additional white-box unit tests exercise `plan_prefix_request` directly for the out-of-vocab and overflow branches, which are not reachable end-to-end through `single_char_vocab_tokenizer` (its 32-entry vocab exactly matches the fixture's `vocab_size` with no `<unk>` fallback, so `tokenize()` can only ever emit in-range ids) and through `max_context()`'s earlier `prompt_len > max_cache_len` gate (which makes `plan.suffix_start + suffix.len() > max_cache_len` an arithmetic impossibility for this specific caller, same as it always was for `forward_prefill_from`'s own check here) respectively -- both are unreachable through the public API for the same structural reasons `forward_prefill_from`'s original checks were, not because the tests are contrived.

## Test plan

- [x] `cargo test -p lattice-inference --lib` (default features) -- 1688 passed, 0 failed, 18 ignored
- [x] `cargo fmt --check -p lattice-inference`
- [x] `cargo clippy --workspace -- -D warnings` (zero warnings)
- [~] `timeout 2400 cargo test -p lattice-inference --features metal-gpu,f16 --lib` (GPU-lock-serialized) -- did **not** complete inside the 2400s bound before `timeout` killed it; the suite was still mid-flight (dozens of individual GPU tests, several logged "running for over 60 seconds"). In the ~40 minutes it did run: zero `FAILED`, zero `panicked` occurrences, and 61 `... ok` lines were observed, including every test this PR added or modified (`plan_prefix_request_rejects_out_of_vocab_suffix_token`, `plan_prefix_request_rejects_empty_suffix`, `plan_prefix_request_rejects_suffix_overflowing_max_cache_len`, `generate_streaming_with_prefix_cache_rejects_empty_replay_suffix_preserves_warm_entry`) plus the surrounding cross-turn-cache/grammar/LoRA/GDN/lm_head suites. No `test result:` summary line was produced because the run was truncated, not because it failed -- reporting this as an honest gap rather than a pass. The mutation-cycle evidence below was captured via a separate, narrowly-filtered run of just the new test (`-- rejects_empty_replay_suffix`), which did complete.

## Bench-compare

Not run. Per the issue's own gate: "No bench-compare required: this is a preflight-ordering fix, not a hot-path or throughput change." Correction (review round 1, finding 1): an earlier version of this section claimed no valid-request code path gained work; that was wrong. Valid requests DO gain a duplicated pure planning/validation pass -- the wrapper computes metadata + `plan_cross_turn_reuse` + `plan_prefix_request`, and `_inner` (defense in depth) repeats the metadata/plan computation and the validation. Tokenization was moved to the wrapper, not eliminated, and still runs once. The duplicated pass is CPU-only integer arithmetic plus a read-only scan of the suffix token ids -- no cache mutation, no GPU work -- and the issue's exemption above covers this fix regardless; the exemption is why bench-compare is not attached, not a zero-added-work claim.

## ADR

n/a -- small, scoped bugfix; no new architecture.

- [ ] SIMD intrinsics, if any, were manually verified -- not just reviewed by an AI
- [x] `cargo test` output included for any behavior-changing code; `cargo bench` output for perf-sensitive changes

(No SIMD intrinsics touched by this PR -- checkbox left unchecked as not applicable.)

## Out of scope

- No structural/file-move changes (per the issue's own sequencing note: this PR must land before any dispatch/layer/prefix-cache file moves).
- `forward_prefill_from`'s own checks are left in place, unchanged, for its other callers.
- The grammar-error destructive-clear path is untouched.

Closes #835

### DIFF
--- a/crates/inference/src/forward/metal_qwen35.rs
+++ b/crates/inference/src/forward/metal_qwen35.rs
@@ -14960,6 +14960,58 @@ mod inner {
             )
         }
 
+        /// Pure preflight over the suffix a [`crate::kv_cache::PrefixRestorePlan`]
+        /// selects — every suffix token id within `vocab_size`, the suffix
+        /// non-empty, and `plan.suffix_start + suffix.len()` within
+        /// `max_cache_len`. Must run BEFORE `restore_cross_turn_prefix`/
+        /// `reset_state` in `generate_streaming_with_prefix_cache_and_cancel_inner`
+        /// (#835): those calls `take()` the cache slot's entry (ExactAppend /
+        /// ReplayFromCheckpoint) or clear live state (FullRefill), and the
+        /// public wrapper's error-recovery arm additionally evicts the slot's
+        /// cache entry outright on ANY `Err` from `_inner`. Rejecting an
+        /// invalid suffix here, before any of that runs, leaves a warmed slot
+        /// untouched instead of destroying it for a request that never
+        /// touched live KV/GDN state.
+        ///
+        /// Mirrors `forward_prefill_from`'s own checks byte-for-byte — same
+        /// `InferenceError` variant and message shape — so this is
+        /// observationally a pure timing fix, not a new error taxonomy.
+        /// `forward_prefill_from`'s checks are left in place unchanged
+        /// (defense in depth for its other, non-prefix-cache callers).
+        fn plan_prefix_request(
+            &self,
+            prompt_ids: &[u32],
+            plan: &crate::kv_cache::PrefixRestorePlan,
+        ) -> Result<(), crate::error::InferenceError> {
+            use crate::error::InferenceError;
+
+            let vocab = self.engine.config.vocab_size;
+            let suffix = &prompt_ids[plan.suffix_start..];
+            for (t, &id) in suffix.iter().enumerate() {
+                if (id as usize) >= vocab {
+                    return Err(InferenceError::InvalidInput(format!(
+                        "forward_prefill_from: token_ids[{t}]={id} out of range: vocab_size is {vocab}"
+                    )));
+                }
+            }
+            if suffix.is_empty() {
+                return Err(InferenceError::PrefixCache(
+                    "forward_prefill_from: cannot prefill an empty suffix without a saved \
+                     logits source"
+                        .into(),
+                ));
+            }
+            if plan.suffix_start + suffix.len() > self.session.kv_cache.max_cache_len {
+                return Err(InferenceError::InvalidInput(format!(
+                    "forward_prefill_from: start_pos {} + len {} exceeds max_cache_len {}",
+                    plan.suffix_start,
+                    suffix.len(),
+                    self.session.kv_cache.max_cache_len
+                )));
+            }
+            Ok(())
+        }
+
         /// Restore live state to the plan's reusable boundary.
         ///
         /// `ExactAppend` and `ReplayFromCheckpoint` (#590) both consume the
@@ -15354,9 +15406,41 @@ mod inner {
             crate::model::qwen35::check_logprobs_not_set(gen_cfg)?;
             crate::model::qwen35::check_mtp_not_requested(gen_cfg)?;
 
+            let input = tokenizer.tokenize(prompt);
+            let prompt_ids: Vec<u32> = input.input_ids[..input.real_length].to_vec();
+
+            // #835: validate the suffix a reuse plan would select for this
+            // slot BEFORE calling `_inner` at all -- same reasoning as the
+            // `check_logprobs_not_set`/`check_mtp_not_requested` preflights
+            // above (PR #787): the `match` below unconditionally calls
+            // `reset_state()` and evicts the cache slot on ANY `Err` from
+            // `_inner`, so a suffix-validation-only rejection (out-of-vocab
+            // token, empty suffix, or a suffix overflowing `max_cache_len`)
+            // must return here, before `_inner` runs `restore_cross_turn_prefix`
+            // (or `_inner`'s own destructive-on-`Err` recovery is even
+            // reachable), or a valid pre-existing cross-turn entry this call
+            // never touched would be destroyed alongside it.
+            //
+            // Skipped for the same three trivial-request shapes `_inner`
+            // itself special-cases with an early `Ok` (empty prompt, zero
+            // `max_new_tokens`, prompt longer than the cache): none of those
+            // reach `_inner`'s mutating match either, so validating a plan
+            // for them here would reject requests `_inner` legitimately
+            // accepts (e.g. an empty prompt always plans an empty "suffix",
+            // which is exactly the shape `plan_prefix_request` rejects for a
+            // real request).
+            if !prompt_ids.is_empty()
+                && gen_cfg.max_new_tokens > 0
+                && prompt_ids.len() <= self.max_context()
+            {
+                let metadata = self.cross_turn_metadata(tokenizer);
+                let plan = self.plan_cross_turn_reuse(slot_id, &metadata, &prompt_ids);
+                self.plan_prefix_request(&prompt_ids, &plan)?;
+            }
+
             match self.generate_streaming_with_prefix_cache_and_cancel_inner(
                 slot_id,
-                prompt,
+                prompt_ids,
                 tokenizer,
                 gen_cfg,
                 on_token,
@@ -15377,7 +15461,7 @@ mod inner {
         fn generate_streaming_with_prefix_cache_and_cancel_inner<F, C>(
             &mut self,
             slot_id: crate::kv_cache::CrossTurnSlotId,
-            prompt: &str,
+            prompt_ids: Vec<u32>,
             tokenizer: &BpeTokenizer,
             gen_cfg: &GenerateConfig,
             mut on_token: F,
@@ -15391,13 +15475,20 @@ mod inner {
             use crate::kv_cache::PrefixReuseMode;
 
             // The `logprobs` / `enable_mtp` config preflight checks (codex
-            // round-1 blocker #1, round-2 medium #3/#4, PR #787) live in the
-            // public wrapper `generate_streaming_with_prefix_cache_and_cancel`,
-            // not here: that wrapper's error-recovery path unconditionally
+            // round-1 blocker #1, round-2 medium #3/#4, PR #787), and the
+            // suffix-content preflight below (#835), live in the public
+            // wrapper `generate_streaming_with_prefix_cache_and_cancel`, not
+            // here: that wrapper's error-recovery path unconditionally
             // evicts the cache slot on any `Err` from this function, so a
             // preflight-only rejection must never reach `_inner` in the first
             // place, or a valid pre-existing cross-turn entry this call never
-            // touched would be destroyed alongside it.
+            // touched would be destroyed alongside it. `prompt_ids` arrives
+            // already tokenized by the wrapper (which needs them to run that
+            // preflight) so this function never re-tokenizes `prompt`.
+            // `plan_cross_turn_reuse`/`plan_prefix_request` are still run
+            // again below, cheaply, as defense in depth (this function's own
+            // safety invariant should not depend solely on its one caller
+            // getting the guard conditions right).
 
             let cfg = self.engine.config.clone();
 
@@ -15419,8 +15510,6 @@ mod inner {
                 }
             };
 
-            let input = tokenizer.tokenize(prompt);
-            let prompt_ids: Vec<u32> = input.input_ids[..input.real_length].to_vec();
             let prompt_len = prompt_ids.len();
 
             // These three guards are byte-for-byte the same as
@@ -15489,6 +15578,21 @@ mod inner {
 
             let metadata = self.cross_turn_metadata(tokenizer);
             let plan = self.plan_cross_turn_reuse(slot_id, &metadata, &prompt_ids);
+
+            // #835: validate the suffix the plan selected BEFORE the match
+            // below runs `restore_cross_turn_prefix` (which `take()`s the
+            // cache slot's entry) or `reset_state()`. An out-of-vocab token,
+            // empty suffix, or start+len overflow is rejected here, before
+            // any live state or cache entry is touched, instead of after
+            // `restore_cross_turn_prefix` has already consumed the slot.
+            // Defense in depth: the public wrapper above already ran this
+            // exact check (on the identical `prompt_ids`/`plan` inputs,
+            // since nothing mutates `self.cross_turn_prefix_cache` between
+            // the two calls) so a rejection here should never actually
+            // trigger via that entry point -- this call exists so `_inner`'s
+            // own ordering invariant does not depend solely on its one
+            // caller getting the wrapper's guard conditions right.
+            self.plan_prefix_request(&prompt_ids, &plan)?;
 
             // The consumed entry (ExactAppend / ReplayFromCheckpoint) is
             // carried to the end-of-generation save so its checkpoint ring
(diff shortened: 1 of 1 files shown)


ohdearquant:

## Review of PR #867 (round 1)

The cache-preservation fix is correctly placed and mutation-sensitive. One PR-description claim must be corrected before merge: the valid path does gain duplicated planning/validation work; tokenization was moved, not eliminated.

### Prior review context

- No prior submitted reviews, comments, or unresolved review threads were present when this round began.
- This round focused on mutation ordering, exact error-shape parity, the private wrapper/inner contract, regression-test strength, F3 non-regression, and publication/merge hygiene.

### What this round examined

- Read the [PR body](https://github.com/ohdearquant/lattice/pull/867), [issue #835](https://github.com/ohdearquant/lattice/issues/835), full PR diff, current surrounding implementation, the pre-PR implementation, PR #787's established wrapper-preflight pattern, and the only `_inner` call site.
- Compared all three new validation branches with `forward_prefill_from`, including variants, exact strings, and interpolated values.
- Ran default and Metal-targeted tests, fmt, clippy, a disposable-worktree red/green mutation cycle, remote CI/status checks, and branch-freshness/merge inspection.

### Findings (1 total: 0 blockers, 0 majors, 1 medium, 0 minors)

#### Mediums

1. `crates/inference/src/forward/metal_qwen35.rs:15436` — The PR body incorrectly claims that valid requests gain no work
   - **Current**: the wrapper computes metadata, runs `plan_cross_turn_reuse`, and scans the suffix with `plan_prefix_request` at lines 15436-15438; `_inner` repeats metadata/planning at lines 15579-15580 and repeats validation at line 15595.
   - **Issue**: before this PR, the prompt was tokenized once and the reuse plan was computed once inside `_inner`. At head, tokenization is still performed once (it moved to the wrapper), while metadata/planning runs twice and suffix validation runs twice. Therefore the PR body's Bench-compare statement that the new wrapper pass “replaces `_inner`'s own (removed) tokenization + first plan computation” and that “no valid-request code path gained new work” does not match the diff. This violates the repository requirement that every PR-description claim match the actual change. The additional CPU work may be small relative to inference, and issue #835 explicitly exempts this fix from `bench-compare`, but that does not make the zero-added-work claim true.
   - **Fix**: correct the Summary/Bench-compare prose to disclose the added pure metadata/plan/scan pass and retain the issue's explicit benchmark exemption. Alternatively, pass the already validated plan into the private `_inner` path (or otherwise deduplicate the second plan/scan) and update the defense-in-depth explanation accordingly.
   - **Test**: no cheap behavioral test; this is established by the committed control flow. If performance language remains, measure the preflight rather than asserting zero work.
   - **Confidence**: derived from the pre-PR source and committed head at lines 15409-15438 and 15579-15595.

- **Placement deviation is correct**: the public wrapper's error arm at lines 15449-15456 treats every `_inner` error identically, calling `reset_state()` and removing the slot. There is no non-destructive error class. An inner-only suffix error would therefore still destroy the warm entry. Hoisting the pure rejection above `_inner` matches PR #787's preflight pattern.
- **Error-shape parity is exact at head**: out-of-vocab uses `InferenceError::InvalidInput` and the identical `forward_prefill_from: token_ids[{t}]={id} out of range: vocab_size is {vocab}` string; empty suffix uses `InferenceError::PrefixCache` with the identical message; overflow uses `InferenceError::InvalidInput` with identical `start_pos`, `len`, and `max_cache_len` wording/values. `forward_prefill_from` itself is unchanged.
- **Wrapper/inner contract**: `generate_streaming_with_prefix_cache_and_cancel_inner` is private and has exactly one caller. The public wrapper signature is unchanged, so there is no semver break. Tokenization-dependent behavior still runs once through the same tokenizer; it is merely moved before `_inner`.
- **Mutation ordering**: `plan_cross_turn_reuse` reads the cache and allocates checkpoint lengths but does not mutate it; `plan_prefix_request` is read-only. Both validations occur before `restore_cross_turn_prefix`/`reset_state`.
- **Regression-test substance**: the warm-entry test checks the error variant, retained cached token IDs, unchanged KV length, unchanged session/GDN position, subsequent `ExactAppend`, and token-identical output against a fresh full-refill baseline.
- **Mutation cycle observed in a disposable worktree**: removing both `plan_prefix_request` calls made `generate_streaming_with_prefix_cache_rejects_empty_replay_suffix_preserves_warm_entry` fail with `left: None` for the retained entry; restoring both calls made it pass. The disposable worktree was removed afterward.
- **F3 non-regression**: the diff does not modify `save_cross_turn_prefix_or_clear`; its save-error arm still removes the slot. No grammar-failure destructive-clear behavior changed.
- **Observed local gates**:
  - `cargo fmt --check -p lattice-inference` — passed.
  - `cargo clippy --workspace -- -D warnings` — passed with zero warnings.
  - `cargo test -p lattice-inference --lib` — 1688 passed, 0 failed, 18 ignored.
  - `LATTICE_METAL_TEST_ENFORCE=1 cargo test ... -- plan_prefix_request` — 3 passed.
  - `LATTICE_METAL_TEST_ENFORCE=1 cargo test ... -- rejects_empty_replay_suffix` — 1 passed.
  - `LATTICE_METAL_TEST_ENFORCE=1 cargo test ... -- cross_turn` — 35 passed.
  - `LATTICE_METAL_TEST_ENFORCE=1 cargo test ... -- prefix_cache` — 11 passed.
- **Publication hygiene on added lines**: no new unsafe code, dependencies, production `unwrap`/`expect`, or public API. Added `unwrap`/`expect` calls are confined to tests. The PR body starts with the required attribution, includes the AI checklist and honest incomplete-suite disclosure, says `Closes #835`, and the conventional commit includes `Co-Authored-By: Leo <noreply@khive.ai>`.
- **Remote CI observed green so far**: Linux x86/ARM CI, Linux feature matrix, MSRV, semver, cargo-deny, app binaries, WASM/embed drift, rustdoc/unwrap informational lints, and the other completed checks all reported success. No completed check was red.
- **Freshness**: fetched `origin/main` is `5047e08ec` (#860). The branch is one commit ahead and one behind (`1 1`), but the behind commit does not touch `metal_qwen35.rs`; `git merge-tree` produced a clean merge. The PR remains marked `BEHIND`, not conflicting.

### What I did NOT check

- I did not run the full `cargo test -p lattice-inference --features metal-gpu,f16 --lib` suite. The author's disclosed attempt required more than 2400 seconds; a fresh 3600-second run could not complete inside this review's fixed 30-minute deadline. The exact filtered Metal scopes above and the full default-feature library suite did complete.
- I did not run `bench-compare`. Issue #835 explicitly says it is not required for this ordering fix. The PR body must nevertheless stop claiming that the valid path gains no work.
- Remote CI was still in progress at the last poll: `e2e parity`, macOS CI/feature-matrix, and `bench-compile` had not completed; the informational Metal-attention and Q4-composed legs were also running. All required legs must finish green before merge.

- No additional code-review round is useful after the PR-body correction unless the implementation changes. Merge only after required CI completes green.

Domain utility: SKIPPED — the khive knowledge request did not return within its bounded wait; repository source, primary GitHub artifacts, executed tests, and Lattice-specific review rubrics were sufficient for this review.

Verdict: APPROVE-WITH-FIXES

oceanwaves630 (approved):
Read the full diff at 06c070e85. The preflight hoist is correct in both entry points: validation runs before any destructive cache operation, the wrapper's skip-conditions exactly mirror the inner function's trivial-request early returns, and the inner re-check keeps the ordering invariant independent of its caller. The mutation-sensitive regression test proving a rejected request preserves a warm entry bit-for-bit is exactly the right shape for this class of fix.

PR #868, opened 2026-07-11
refactor(inference): consolidate scalar f16/bf16 bit conversion into one always-compiled module

ohdearquant:
> _PR description authored by Claude (Anthropic agent) on behalf of @ohdearquant._

## Summary

- Adds `crates/inference/src/weights/half_bits.rs` as the sole, always-compiled scalar IEEE-754 f16/bf16 bit-pattern decoder, replacing five independent copies of the same bit-twiddling arithmetic.
- Replaces the five call sites (`weights/f32_weights.rs`, `quant/quarot/io.rs`, `weights/q4_weights.rs`, `weights/f16_weights.rs`, `forward/metal_qwen35.rs`) with thin delegating wrappers; every existing public function signature and behavior is unchanged.
- `half_bits` is **not** gated behind the crate's `f16` feature — QuaRot and Q4 quantization already need half-precision metadata inspection in every build configuration; only *runtime F16/BF16 model loading* stays feature-gated.
- Adds independent-oracle tests (using the `half` crate, an already-unconditional workspace dependency used elsewhere in this crate) that verify the consolidated decoder against a third-party implementation rather than only against itself, and migrates `f16_convert_bench` to call the production decoder instead of a fifth copy it had been carrying.

## Test plan

- [x] `cargo test -p lattice-inference --lib` — 1724 passed, 0 failed, 18 ignored
- [x] `cargo test -p lattice-inference --lib half_bits` — 14/14 pass
- [x] **Independent-oracle equivalence** (not just internal composition): `f16_bits_to_f32_matches_independent_half_crate_oracle` compares every non-signaling-NaN f16 bit pattern bit-for-bit against `half::f16::from_bits().to_f32()`, a decoder this module shares no code with. Signaling NaN patterns are verified separately in `f16_bits_to_f32_signaling_nan_is_lossless_widen_independent_of_decoder` against the textbook IEEE-754 widening formula computed directly from the bit pattern — they're excluded from the `half`-crate comparison because on this machine `half` dispatches to the AArch64 hardware `fcvt` instruction, which forces the quiet bit on signaling-NaN operands per ARM's conversion semantics (real hardware behavior, not a bug, but not a pure bit-widen either, so not usable as an oracle for that one class of input).
[...]

Issue #752, opened 2026-07-09
bug(inference): SafetensorsFile does not reject aliased/overlapping tensor byte ranges


ohdearquant:
**Defect**: The runtime Safetensors reader (`SafetensorsFile`, used for ordinary f32 loads across BERT, Qwen3, Qwen3.5, and cross-encoder models) validates only per-tensor bounds, shape-product overflow, and byte-count — there is no cross-entry check that tensor byte ranges don't overlap or leave gaps. Two same-shaped tensors (e.g. `k_proj.weight` and `v_proj.weight`) can both declare the identical `[0, N)` range and both pass validation, silently aliasing one tensor's data as another's. The QuaRot converter's own safetensors reader, by contrast, sorts all entries and rejects any non-contiguous start or incomplete data-section coverage.

**Evidence**:
- `crates/inference/src/weights/f32_weights.rs:227-260` — per-tensor checks only: `start <= end`, `end <= data_len`, shape-product overflow, `actual == expected` byte length. No sort, no cross-entry overlap check, no gap/coverage check.
- `crates/inference/src/quant/quarot/io.rs:340-364` — sorts every entry by `(start, end)` and rejects any non-contiguous start or incomplete data-section coverage (explicit "Phase 2" validation block).
- `crates/inference/src/model/qwen35/loading.rs:187` — the full-attention loader independently obtains `k_proj` and `v_proj` from the source, with no protection against the aliasing case above.

**Blast radius**: A malformed-but-locally-self-consistent full-attention checkpoint can alias K and V weight tensors; every CPU decode/prefill using that checkpoint then computes wrong attention outputs for every token, with no load-time error.

**Suggested fix**: Extract a shared full-layout Safetensors validator (sort + overlap/gap/coverage check, matching the QuaRot reader's logic) and invoke it from `SafetensorsFile::open` before tensor data is exposed to callers.

Provenance: Audited at 13c8de8a3; re-verified at 0699e60cc.

PR #869, opened 2026-07-11
chore(scripts): delete superseded decode-benchmark scripts and fix Makefile .PHONY gaps

ohdearquant:
> _PR description authored by Claude (Anthropic agent) on behalf of @ohdearquant._

## Summary

- Delete `scripts/bench_decode_slope.sh` (superseded by the ADR-067 Theil-Sen/bootstrap
  `bench_decode_slopefit.*` path, driven by the `adr064-gpu-decode-nightly.yml` workflow)
  and `scripts/bench_q8_vs_ollama.sh` (predates the corrected marginal-cost method, no
  live caller).
- Remove the now-dead `bench-decode` Make target that invoked the deleted script.
- Fix `.PHONY` (Makefile:1): add `e2e-parity` and `bench-decode-slopefit` (real
  non-file-producing targets that were missing) and drop `bench-decode` (target removed).
- Drop the stale "Does NOT touch bench_decode_slope.sh or the bench-decode Make target"
  comment in `bench_decode_slopefit.sh` since both are now gone.

## Test plan

- [x] Repo-wide reference search (`grep -rn` over the whole tree, excluding `.git`) for
      `bench_decode_slope.sh`, `bench_q8_vs_ollama.sh`, and the bare `bench-decode` target
      name, before and after the change — confirms no live caller (Make target, workflow,
      doc, or script) beyond the two scripts' own usage comments, which are deleted along
      with them.
- [x] `.PHONY` cross-check: enumerated every `^target:` line in the Makefile after the
      change (21 targets) against the `.PHONY` list (21 entries) — exact match, no stale
      or missing entries.
- [x] `make -n bench-decode-slopefit` — resolves correctly (prints the script invocation).
- [x] `make -n bench-decode` — fails with "No rule to make target `bench-decode'" as
      expected (target removed).
- [x] Pre-commit hook ran `cargo check --workspace` (full workspace compiles), the doc
      linter, and the capability-matrix fixture check — all passed cleanly, nothing in
      this diff touches Rust source so this is incidental confirmation the repo still
      builds, not a targeted gate for this change.
- [ ] `cargo test --workspace` / `cargo clippy --workspace -- -D warnings` / `cargo fmt
      --all -- --check` — N/A, no Rust source touched (Makefile + shell scripts only).
- N/A `make bench-compare` (ADR-058) — this PR touches no code under `crates/inference/`
  or `crates/embed/`; the diff is Makefile plumbing and dead-script removal only.
[...]

Issue #802, opened 2026-07-10
bug(inference): unify native Q4/KHF1 file validation — non-finite scale/bias, exact file extent, and panic-to-Result conversion


ohdearquant:
> _Issue authored by Claude (Anthropic agent) on behalf of @ohdearquant._

## What

The native `.q4` and KHF1 `.f16` file formats have validation gaps that are distinct from, and not covered by, the safetensors non-finite class tracked in issue #793:

- The CPU `.q4` loader (`load_q4_file`, `crates/inference/src/weights/q4_weights.rs`, roughly lines 808-879, verified in this checkout) reads the raw block payload via bounded `read_exact` (so truncation fails closed) and checks `shape.product() == original_len`, but **does not check whether the decoded per-block `scale`/`bias` f16 values are finite**, and **does not reject trailing bytes** left in the file after the expected block payload has been read.
- The Metal no-copy mmap path (`mmap_q4_weight` in `crates/inference/src/forward/metal_qwen35.rs`, roughly lines 774-811, calling `validate_q4_header_payload_bounds`) validates the header and payload bounds before handing the mapping to the GPU, but shares the same two gaps: no scale/bias finiteness check, and trailing bytes past the payload are accepted rather than rejected.
- The KHF1 `.f16` loader (`load_f16_tensor_file`, same file, roughly lines 895-945) fails closed on magic/version and uses `read_exact`, but **silently accepts** a header whose `shape.product() != numel` (unlike the sibling `.q4` loader, which does check this), and, like the `.q4` loaders, accepts NaN/Inf decoded values and trailing bytes.
- Separately, the Q4 source-side quantizers (`quantize_row_q4_0`, `quantize_tensor_q4_0`, `quantize_bf16_to_q4`, `quantize_f32_to_q4`, `quantize_f64_to_q4[_mode]`, roughly lines 333-570) already reject non-finite *source* values (verified: `quantize_block_with_mode_len` at line ~259 rejects any non-finite input element before deriving a block), but there is no check on the *output* side — a finite source value whose derived scale or bias overflows the stored f16 representation (i.e. encodes to infinity) is written to the block without error. Several of these public entry points also panic on shape mismatch rather than returning the `Result` they already declare (referenced around line ~419-432 of the design's source inspection; confirm exact line at implementation time as this file is actively changing).

## Why

[...]

PR #873, opened 2026-07-11
refactor(inference): private weight-ingress validation seam for safetensors

### LINKED ISSUE #800: refactor(inference): introduce a private weight-ingress validation seam and route safetensors loading through it

## What

Weight-loading correctness checks (finite-value scans, shape/extent validation, dtype support checks) are currently scattered per-loader with materially different coverage. For safetensors specifically:

- `SafetensorsFile::open` / `from_backing` / `get_f32_tensor` (`crates/inference/src/weights/f32_weights.rs`, roughly lines 172-363) validate header bounds and byte-length at open time (fail closed), but **silently accept** NaN/Inf in F32, F16, and BF16 tensor data — there is no finite-value check on the decoded/viewed slice before it is handed to the caller.
- Unsupported dtypes are silently skipped by tensor-metadata parsing rather than rejected, so a required tensor stored under an unsupported dtype later appears simply "missing" instead of failing with a clear format error.
- `ShardedSafetensors` / eager `load_sharded` (same file, roughly lines 900-1033) route through the same `SafetensorsFile` machinery and inherit exactly this gap.
- `f16_weights.rs` (`get_f16_tensor`, `load_f16_tensor_checked`, `load_f32_tensor_checked`, `load_conv1d_weight`, `load_f16_weights`, roughly lines 746-1040) also route through `get_f32_tensor` and inherit the same silent acceptance before narrowing to f16 storage.

Issue #793 (open) already tracks a narrower point-fix for this specific F32/F16/BF16 non-finite gap at `get_f32_tensor`. This issue is the broader structural follow-up: instead of adding another one-off scan, create one seam that every external-bytes-to-tensor ingestion path is required to call, so the same class of gap cannot silently reopen at the next loader.

## Why

The non-finite-acceptance defect recurs because each loader independently decides what "valid" means at its own boundary, with no shared, auditable choke point. A private, sealed validation function at the exact place raw bytes become trusted tensor data closes the class for every current and future caller in one review, rather than requiring a new point-fix each time a sibling loader is discovered.

## Steps

1. Add `crates/inference/src/weights/ingress.rs` as a new, private, always-compiled module. Define a sealed internal type (e.g. `IngestedTensor<'_>`) carrying: source label/path, tensor name, declared shape, and exactly one payload form (safetensors F32/F16/BF16 bytes, decoded F32/F64, native Q4 blocks/bytes, or Q8 data+scales — later payload forms are wired up by the companion issues in this cluster).
2. Implement one entry point, `pub(crate) fn validate_ingested_tensor(tensor: IngestedTensor<'_>) -> Result<(), InferenceError>`, enforcing for this issue's scope (safetensors F32/F16/BF16):
   - Preserve tensor metadata for every entry so unsupported-dtype tensors remain structurally tracked instead of silently disappearing; only a *requested* tensor with an unsupported dtype fails, using `InferenceError::InvalidSafetensors`. An unrecognized dtype string fails at parse time (its byte width is unknowable, so extent validation cannot proceed).
   - Checked shape product and exact payload length for the tensor being materialized.
   - All decoded F32/F16/BF16 values finite (NaN and either infinity rejected; signed zero and subnormals accepted).
   - Errors carry source label/path, tensor name, dtype, shape, and the first offending element index, via the existing `InferenceError::InvalidSafetensors` variant — do not add a new public error variant.
3. Wire this seam into `get_f32_tensor` (single mapped file and owned/`from_bytes` variants), the sharded path (`ShardedSafetensors::{open_index,open_shard}` / `TensorSource::get_f32_tensor_owned`), and the f16-storage facade (`f16_weights.rs`), so validation happens once, immediately before any tensor view/copy escapes to a caller.
4. Add a per-`TensorMeta` cached validation flag so repeated access to the same zero-copy F32 tensor does not rescan already-validated pages.
5. If issue #793's narrower point-fix has already landed by the time this work starts, fold its fixtures into this seam's test suite and remove the now-duplicated scan it added, in the same PR — do not carry two live copies of the same check.
6. Confirm downstream facades that build on these loaders (`BertModel::{from_directory,from_pretrained,from_bytes}`, `CrossEncoderModel::from_directory`, `QwenModel::{from_directory,from_pretrained}`, `Qwen35Model::from_safetensors` via `TensorSource`, `NativeEmbeddingService`, the wasm `LatticeEmbedder::new` constructor, and the inline LoRA safetensors loading in `bin/chat_metal.rs`) inherit the seam by construction and do not add their own duplicate numeric scans. A required model/tensor bundle must only be returned once every required tensor has validated; do not return a partially-built model on a validation failure partway through assembly.

## Gates

- `cargo test -p lattice-inference` and full workspace `cargo test` pass, including new fixtures: NaN and both infinities for F32 (mapped and owned/from-bytes), one NaN and one infinity bit-pattern per dtype for F16 and BF16 through the safetensors path, overflowed shape product, one-byte-short and one-byte-extra payload, and a required tensor stored under an unsupported dtype failing as a dtype/format error rather than a "missing tensor" error.
- Mutation-sensitivity: each of the above fixtures must be shown to fail (i.e. the test currently passing must go red) when its corresponding guard is reverted, per the project's mutation-sensitive-test convention — do not merge a guard test that still passes with the guard removed.
- `rg` review confirms `validate_ingested_tensor` calls are confined to the A-list ingestion modules named above and never appear inside `matmul`, attention, generation, embedding lookup, or any other decode loop.
- `make bench-compare` shows no statistically credible decode-throughput regression. A measurable *startup*-time change (from newly faulting zero-copy F32 pages) is expected and must be reported and reasoned about separately from decode cost, not hidden inside a combined number.
- No new public `InferenceError` variant; existing `InvalidSafetensors` is reused. No public function signature changes.
- e2e-parity (`e2e-parity.yml`) must stay green — this issue changes the loading path used by the models that gate compares against HF/reference output.
- Semver class: **patch** if error variants/signatures stay fixed, since previously-accepted invalid (non-finite) checkpoints newly fail to load. Document the behavior tightening in release notes as a correctness fix, not a breaking change.

## Sequencing

- **Depends on:** the half-bits conversion consolidation issue (so F16/BF16 finite checks decode through one trusted scalar path rather than re-deriving bit logic here), and should be coordinated with #793 so the two point-fixes don't collide — whichever lands first, the other folds its tests in and removes the duplicate scan.
- **Depended on by:** the QuaRot/offline-quantizer routing issue, the native Q4/KHF1 validation-unification issue, the Q8 routing issue, and the facade/parity-test issue (which adds the cross-crate differential test and ADR-003/ADR-013 amendments referencing this seam as the single validation entry point).
- **Related, not duplicated:** issue #752 (`SafetensorsFile does not reject aliased/overlapping tensor byte ranges`) targets a distinct defect — overlapping tensor offset ranges within an otherwise well-formed header — and is not closed by this issue's finite-value/shape/extent checks. It may be a natural additional invariant to enforce inside this same seam once both land; that composition is left to whichever PR lands second to avoid a merge conflict on the same file.

ohdearquant:

## Summary

- Add `crates/inference/src/weights/ingress.rs`: a private, always-compiled weight-ingress
  validation seam (`IngestedTensor` + `validate_ingested_tensor`) that checks a decoded tensor's
  shape-product for overflow, that its element count matches the declared shape, and that every
  value is finite (NaN / either infinity rejected; signed zero and subnormals accepted), reusing
  `InferenceError::InvalidSafetensors` (no new public error variant).
- Wire the seam into `SafetensorsFile::get_f32_tensor`, the single point every other loader in the
  crate already routes through (`get_f16_tensor`, `ShardedSafetensors`, `load_sharded`, the
  BERT/CrossEncoder/Qwen/Qwen3.5 model facades, the wasm `LatticeEmbedder`, `chat_metal`'s LoRA
  loader) — confirmed by grep, none needed separate wiring. Validation runs once per tensor and is
  cached via a per-`TensorMeta` `OnceLock<()>`, so repeated access to an already-validated
  zero-copy F32 tensor does not rescan its pages.
- Extend the safetensors dtype table (`DType::Other`) so a tensor stored under a known-but-
  unmaterialized dtype (I64, BOOL, F64, U8, ...) stays structurally tracked instead of silently
  disappearing from the header; only a *requested* tensor of such a dtype now fails, with a
  dtype/format error rather than looking like a missing tensor. A genuinely unrecognized dtype
  string still fails at parse time, since its byte width is unknowable and extent validation can't
  proceed. Updated a stale doc comment in `quant/quarot/io.rs` that described the old permissive
  behavior this change supersedes.

## Dependency note (lattice#799 / PR #868)

lattice#800's issue text lists a "Depends on" relationship to lattice#799 (half-bits conversion
consolidation) for F16/BF16 finite-value decoding. I verified this is **not** a hard blocker for
this PR: `f32_weights.rs` already had correct, working local `convert_f16_bytes_to_f32` /
`convert_bf16_bytes_to_f32` functions that widen F16/BF16 to `f32` before any caller sees the
data. This seam validates *after* that widening — the "decoded F32" payload form the issue text
explicitly names as one of several payload forms — so it never needed to touch bit-conversion
logic or import `half_bits.rs`. This PR is branched from `origin/main` (not from
`fix/799-loader-ingress`) and does not depend on PR #868 merging first; no rebase is required
either way.

## Test plan

- [x] `cargo clippy -p lattice-inference --lib -- -D warnings`
- [x] `cargo clippy -p lattice-inference --all-targets --features f16,metal-gpu -- -D warnings`
- [x] `cargo check --workspace`
- [x] `cargo test -p lattice-inference --lib` — 1725 passed, 0 failed, 18 ignored
- [x] `cargo test -p lattice-inference --lib --features f16 weights::` — 129 passed, 0 failed
      (covers the new F16/BF16 NaN/+Inf-bit-pattern fixtures gated behind the `f16` feature)
- [x] New fixtures per the issue's gate list: F32 NaN/+Inf/-Inf (mapped-mmap path via `open()`
      and owned path via `from_bytes()`), one NaN + one infinity bit pattern each for F16 and
      BF16 through the safetensors path, shape-product overflow, one-byte-short (pre-existing
      test) and one-byte-extra payload, a required tensor under an unsupported dtype (I64)
      failing as `InvalidSafetensors` rather than `MissingTensor`, an unrecognized dtype string
      rejected at parse/open time, and the per-tensor validation-cache behavior.
- [x] Mutation-sensitivity (revert/touch/rerun/restore/rerun cycle, done by hand for all three
      new guards):
  - Finite-value scan in `ingress.rs` removed → 12 tests went red (all the NaN/Inf fixtures);
    restored → all 129 `--features f16 weights::` tests green again.
  - `DType::Other` rejection in `get_f32_tensor` removed (arm made to return a valid-looking
    finite slice) → `test_unsupported_dtype_tensor_fails_as_format_error_not_missing_tensor`
    went red; restored → green.
  - Unrecognized-dtype-at-parse-time rejection removed (arm made to fall back to `DType::F32`)
    → `test_unrecognized_dtype_string_rejected_at_parse_time` went red; restored → green.
- [x] `rg` review: `validate_ingested_tensor` calls are confined to `get_f32_tensor` in
      `f32_weights.rs`; no call sites in `matmul`, attention, generation, embedding lookup, or
      any other decode loop.
- [x] No new public `InferenceError` variant (reused `InvalidSafetensors`); no public function
      signature changes (the two internal signature changes — `SafetensorsFile::from_backing`
      gaining a `source: String` param and `JsonParser::parse_tensor_meta` gaining a
      `tensor_name: &str` param and returning `TensorMeta` instead of `Option<TensorMeta>` — are
      both private/crate-internal).
- [x] Semver class: **patch**. Previously-accepted invalid (non-finite, or stored under an
      unsupported dtype and then requested) checkpoints now fail to load where they silently
      loaded before; this is a correctness fix, not a breaking API change.

### `make bench-compare` — quiet-lane re-run (supersedes the earlier contention-corrupted numbers)

The round-1 review feedback correctly flagged that the first `bench-compare` run (referenced
below in the superseded paragraph) was not trustworthy evidence. I re-ran it through
`with-heavy-lane.sh` after the machine's heavy lane freed up (single-holder, no concurrent GPU
lanes), base=`origin/main` (`fddbb5fcc`), head=this branch (`4546fcdad`):

**Result: ❌ 1 FAIL, ⚠ 16 WARN, 🚀 3 improvements — all 20 flagged benchmarks are in
`lattice-embed`'s SIMD/int8/int4 vector-search suite** (`int8_raw_dot_product`,
`simd_query_batch_dot_product`, `tier_prepared_query`, `simd_batch_cosine*`, etc.) — a different
crate this PR's diff never touches (the diff is confined to `crates/inference/src/weights/` and
one doc comment in `crates/inference/src/quant/quarot/io.rs`).

Every group this diff *could* plausibly affect — `rms_norm`, `layer_norm`, `silu_inplace`,
`gelu`, `add_bias_gelu`, `softmax_attention`, `elementwise_mul` — passed silently (95% CI-lower
bound under the 3% warn threshold on every one, e.g. `rms_norm/896` +2.59% [CI +2.01%, +3.17%],
`softmax_attention/512` +0.63% [CI -0.78%, +2.06%]). This is consistent with the seam's own
design: `get_f32_tensor`'s validation runs once per tensor at model-load time, before decode
starts, and is cached thereafter via `OnceLock` — it has no hot-path presence in any of the
elementwise/norm/attention compute kernels these benchmarks measure, so a clean quiet-lane result
here is expected, not a coincidence.

The embed-crate SIMD flags are unexplained by this diff and most likely reflect either residual
machine noise or a pre-existing embed-side condition unrelated to this PR — worth a maintainer
look independently, but out of scope for this PR to chase down.

<details><summary>Superseded: original contention-corrupted run (kept for the record)</summary>

Ran `make bench-compare` (default groups) wrapped in `with-heavy-lane.sh` per lane convention.
The raw report showed large swings in both directions across benchmark groups this diff does not
touch at all — traced at the time to sustained multi-lane machine contention during that specific
benching window (concurrent `lattice:851-gputest`/`gputest2`, `lattice:868-bench`/`868-review-r2`,
`lattice-hygiene:clippy2`, and an unrelated `khive-work` worktree's `cargo test` run). The
quiet-lane re-run above supersedes this.

</details>

Startup-time note (per the issue's own gate text, corrected after review feedback): this repo's
Criterion suite has no end-to-end model-load-time benchmark, so I measured cold/warm load timing
directly with a temporary manual harness, comparing this branch against `origin/main` on two real
local checkpoints (both release builds, `--features f16`, one machine, back to back):

| checkpoint | tensors | total elems | cold: main | cold: this PR | delta | warm: main | warm: this PR |
|---|---|---|---|---|---|---|---|
| qwen3.5-2b/model.safetensors (~4.3GB, mixed F32/BF16) | 632 | 2,274,069,824 | 14.24s | 21.42s | **+50.4%** | 42µs | 16.7ms |
| qwen3.6-27b/model-00001-of-00015.safetensors (~3.97GB shard, all BF16) | 9 | 1,984,430,080 | 6.19s | 14.34s | **+131.7%** | 1.8µs | 1.4µs |

This is **not negligible**. The added cost is the full linear finite-value scan over each tensor's
decoded `f32` values, which faults every page of a zero-copy mmap and is O(numel) like the decode
work it rides alongside — on these checkpoints it roughly doubles cold-load wall time. It runs at
most once per tensor per `SafetensorsFile` (cached via `OnceLock`), so it does not repeat on a warm
second access to the same handle, but the first load of a large checkpoint (the common case: model
load into a fresh inference process) pays the full cost every time.

Proposed follow-up mitigation (not attempted in this PR, scope discipline): vectorize the scan
(SIMD any-non-finite reduction, similar to this crate's existing SIMD dot-product/quantization
kernels) to cut the added constant, and/or make the scan opt-out via a fast-path env var or config
flag for trusted-checkpoint deployments that have already validated the file once. Filing a
follow-up issue for this is left to the maintainer's triage rather than bundled into this PR.

## ADR

n/a — small fix (new private module + one wiring point), no ADR.

- [x] No SIMD intrinsics touched by this diff
- [x] This PR body starts with the attribution line above
- [x] `cargo test` output included above for the behavior-changing code; `cargo bench` output
      (quiet-lane re-run) and measured cold/warm model-load timing both disclosed above

## Out of scope

- `crates/inference/src/quant/quarot/io.rs`'s own independent safetensors reader (QuaRot
  converter) — routing it through this seam is lattice#800's "Depended on by" companion issue
  (QuaRot/offline-quantizer routing). Only touched its doc comment here, which described the old
  behavior this PR supersedes.
- Native Q4/KHF1 and Q8 loaders (`weights/q4_weights.rs`, `weights/q8_weights.rs`) — **correction
  after review feedback**: this is only accurate for the Q4 *quantization-input* path
  (`quantize_block_with_mode_len`, `q4_weights.rs:270`, which rejects non-finite source values
  before deriving a block) and for `q8_weights.rs`. It is **not** accurate for the Q4/KHF1
  *artifact-loader-output* path: `load_q4_file` (`q4_weights.rs:864`) returns raw per-block
  `scale`/`bias` bit patterns with no finiteness check, and `load_f16_tensor_file`
  (`q4_weights.rs:936`) widens KHF1 values with no finiteness check either. Native Q4/KHF1 ingress
  therefore remains unprotected on this path pending lattice#802 ("unify native Q4/KHF1 file
  validation — non-finite scale/bias, exact file extent, and panic-to-Result conversion"), which
  already covers this exact gap (its own "Sequencing" section lists this ingress-seam issue as a
  dependency) — filing a new issue would have duplicated it.
- Issue #752 (aliased/overlapping tensor byte ranges) — a distinct defect per the issue text,
  left to whichever PR lands second on this file.
- Sub-byte safetensors dtypes (`F4`, `F6_E2M3`, `F6_E3M2`) don't fit this crate's whole-byte
  extent model and are treated as unrecognized (parse-time error) rather than added to
  `DType::Other`; `quant/quarot/io.rs`'s own bit-size table already handles them for the QuaRot
  path.

Closes #800.


ohdearquant:

## Review of PR #873 (round 1)

The scoped safetensors implementation is fail-closed and correctly routed, but it is not ready to
merge until the unmeasured 70 GB-scale startup cost is evidenced and the inaccurate native-loader
coverage claim is corrected. Required local test/mutation evidence and all required remote jobs
also remain outstanding.

### Prior review context

- Round 1; no prior submitted review or review threads were present when this review began.
- Focus: safetensors ingress coverage, validation/cache correctness, fail-closed behavior,
  mutation sensitivity, and the required quality/performance evidence.

### Evidence log

- Read PR #873 body and issue #800 before source review.
- Reviewed the complete changed `ingress.rs`, `f32_weights.rs`, `weights/mod.rs`, and
  `quant/quarot/io.rs`, their loader callers, the Q4/Q8 paths, and tune/embed ingress code.
- `git diff --check <PR-base>...HEAD`: passed.
- Remote state observed at head: several required CI and e2e-parity jobs were queued/in progress;
  none had failed at observation time.
- Local test, mutation-rerun, and benchmark evidence could not be obtained because the mandatory
  heavy lane remained occupied; feature clippy passed.

### Ingress inventory

| Bytes-to-values path | Seam | Notes |
| --- | --- | --- |
| `SafetensorsFile::{open,from_bytes} -> get_f32_tensor` | Covered | F32/F16/BF16 decoded values are validated before return. |
| `TensorSource`, `ShardedSafetensors`, `load_sharded`, BERT/CrossEncoder/Qwen/Qwen3.5, f16 facade, embed wasm, `chat_metal` LoRA | Covered by delegation | All call `get_f32_tensor`. |
| `QuarotTensorReader -> decode_bytes_to_f64` | Not covered | Separate safetensors reader; explicitly deferred in the PR. |
| Native `.q4` / KHF1 `.f16` loaders and Metal Q4 mmap path | Not covered | Separate loader/validator family; explicitly deferred. |
| Q8 quantization from already-loaded f32 | Not covered | Has an independent full finite-value scan. |
| `lattice-tune` PEFT LoRA reader | Not covered | Has its own post-decode finite scan. |
| `lattice-tune` registry safetensors helpers | Not covered | Direct `SafeTensors::deserialize` + F32 decoding; no finite scan. |

This makes the seam a real choke point for the scoped runtime safetensors facade, not a
crate-/workspace-wide ingress guard. The out-of-scope wording is assessed in Finding 2.

### Findings (# total so far: 0 blockers, 0 majors, 2 mediums, 0 minors)

#### Mediums

1. [`crates/inference/src/weights/ingress.rs:101`](crates/inference/src/weights/ingress.rs:101) — the required startup-cost evidence is still absent.
   - **Current**: the first successful access executes an element-by-element
     `iter().enumerate().find(...)` finite scan; [`f32_weights.rs:471`](crates/inference/src/weights/f32_weights.rs:471)
     calls it before the first returned slice.
   - **Issue**: this is not an explicit SIMD kernel, and its first-offender/index contract gives no
     vectorization guarantee. A 70 GB F32 checkpoint means approximately 17.5 billion values and
     at least one additional 70 GB sequential read/page-fault walk before model construction can
     return. The decode-path benchmark is not a measurement of that cost; PR #873 itself reports
     no startup measurement, although issue #800 makes it a gate. The PR body's comparison to
     pre-existing mmap/copy work is misleading for aligned F32: that path is zero-copy at
     [`f32_weights.rs:420`](crates/inference/src/weights/f32_weights.rs:420), so the scan is the
     operation that faults every data page.
   - **Fix**: provide a quiet-lane `bench-compare` for the required decode groups and a separate,
     measured cold/warm model-load result (with checkpoint size and machine context). Do not call
     the scan negligible without that measurement.
   - **Test**: benchmark evidence, not a unit test.
   - **Confidence**: derived from current source and the issue's explicit gate; three compliant
     quiet-lane attempts returned exit 75.

2. [`crates/inference/src/weights/q4_weights.rs:864`](crates/inference/src/weights/q4_weights.rs:864) — the PR's out-of-scope coverage statement overstates the native-loader protection.
   - **Current**: `load_q4_file` accepts raw block scale/bias bits and returns them without a
     finite-value scan; `load_f16_tensor_file` similarly widens and returns KHF1 values at
     [`q4_weights.rs:936`](crates/inference/src/weights/q4_weights.rs:936) without checking
     finiteness. The Q4 finite guard is on *quantization input* at
     [`q4_weights.rs:270`](crates/inference/src/weights/q4_weights.rs:270), not on these artifact
     loaders.
   - **Issue**: the PR description says native Q4/KHF1 and Q8 loaders "already have their own
     finite-value guards." That is false for a malicious/corrupt on-disk Q4/KHF1 artifact, and
     makes the seam's residual coverage look stronger than it is.
   - **Fix**: amend the PR description to state that native Q4/KHF1 ingress remains unprotected
     pending its named companion issue; do not use the existing quantization-input guard as proof
     of loader coverage.
   - **Test**: companion issue should add NaN/Inf scale, bias, and KHF1 payload fixtures.
   - **Confidence**: observed by source inspection.

- **Scoped safetensors seam**: `SafetensorsFile::{open,from_bytes}` both call the same parser;
  all F32/F16/BF16 materialization branches converge at
  [`f32_weights.rs:397`](crates/inference/src/weights/f32_weights.rs:397). `TensorSource`, eager
  sharded loading, lazy sharded loading, f16 storage, model facades, embed wasm, and `chat_metal`
  LoRA delegate to it.
- **Validation logic**: shape products use `checked_mul` at
  [`ingress.rs:71`](crates/inference/src/weights/ingress.rs:71); `[]` correctly yields scalar
  count 1 and zero dimensions yield 0, both valid safetensors cases. Decoded-length comparison is
  in f32 elements after F16/BF16 widening, so it uses the correct representation. The scan is full
  (not sampled), reports the first index, source, name, dtype, and shape, and returns
  `InferenceError::InvalidSafetensors`.
- **Dtype behavior**: recognized whole-byte unsupported dtypes remain in `TensorMeta` and fail on
  request; genuinely unknown and sub-byte unsupported dtype strings fail during header parsing.
- **Cache key**: `validated: OnceLock<()>` is embedded in the per-file, per-name `TensorMeta`, so
  distinct tensor metadata cannot share it. Reconstructing `SafetensorsFile` creates fresh metadata
  and revalidates a changed/reloaded file. A live mmap modified after validation is outside the
  cache contract and already violates the `memmap2` mmap safety invariant.
- **Failure behavior**: inspected new paths return `Err`; no warn-and-continue behavior, new public
  error variant, or production `unwrap`/`expect` was added. Added `expect` calls are tests only.
- **Hygiene**: `cargo fmt -p lattice-inference -- --check`,
  `cargo clippy -p lattice-inference --all-targets --features f16,metal-gpu -- -D warnings`, and
  `git diff --check` passed locally. The commit contains the required
  `Co-Authored-By: Leo <noreply@khive.ai>` trailer. The PR body starts with valid AI attribution,
  includes the checklist, and uses `Closes #800`; #800's explicitly dependent work is described as
  companion follow-up rather than silently implied complete.
- **Tests in the diff**: observed 19 added `#[test]` functions (not 17): 12 in `f32_weights.rs`
  and 7 in `ingress.rs`.

### Gate status

| Gate | Result |
| --- | --- |
| Local fmt | Pass |
| Local feature clippy | Pass |
| `cargo test -p lattice-inference --lib` | Not run: two heavy-lane attempts timed out while `lattice:851-gputest2` held the lane. |
| Ignored-count vs current main | Not verified locally; a detached `origin/main` worktree is prepared, but the required heavy lane is unavailable. |
| Finite-scan mutation | Guard disabled in isolated `/tmp/lattice-873-mutation.LoL3e9`; required rerun could not acquire the same lane. |
| Required quiet decode benchmark | Deferred: three prescribed acquisition attempts returned exit 75 while the heavy lane was held by `lattice:851-gputest2`. |
| Remote CI at head | Ubuntu CI, semver, feature matrix, and all completed informational checks were green when rechecked; MSRV, macOS, bench-compile, app binary, and e2e-parity jobs remain queued/in progress. |

The branch is one commit behind `origin/main` (`fddbb5fcc`, #869) and should be refreshed before
merge.

### What I did NOT check

- I did not obtain a clean local lib-test result, mutation-test result, ignored-count comparison,
  or benchmark because the mandatory heavy lane did not become available. I did not substitute an
  unguarded concurrent test run.
- I did not inspect full model-load latency on a production-sized checkpoint; that is the missing
  evidence in Finding 1.

### External sources consulted

- [SafeTensors format specification](https://github.com/safetensors/safetensors) — confirms that
  scalar and zero-sized tensors are valid and that payload NaN/Inf values are not format-checked
  by SafeTensors itself.
- [memmap2 `Mmap` safety documentation](https://docs.rs/memmap2/latest/memmap2/struct.Mmap.html)
  — confirms that subsequent modification of a file-backed mapping can cause undefined behavior.

- Do not begin another general review round until the two medium findings are fixed, the branch is
  refreshed, required remote jobs are green, and clean heavy-lane test/mutation/benchmark evidence
  is attached. A focused follow-up should verify those artifacts rather than re-review the loader.

Domain utility: SKIPPED — no domain-composition tool is available in this environment; repository
and PR-specific review guidance was used instead.

ohdearquant:
Addressing the two review feedback items from round 1:

**1. Startup-cost evidence (previously unmeasured).** Measured cold + warm model-load timing with this branch vs `origin/main`, on two real local checkpoints, release builds with `--features f16`, same machine, back to back:

| checkpoint | tensors | total elems | cold: main | cold: this PR | delta | warm: main | warm: this PR |
|---|---|---|---|---|---|---|---|
| qwen3.5-2b/model.safetensors (~4.3GB, mixed F32/BF16) | 632 | 2,274,069,824 | 14.24s | 21.42s | **+50.4%** | 42µs | 16.7ms |
| qwen3.6-27b/model-00001-of-00015.safetensors (~3.97GB shard, all BF16) | 9 | 1,984,430,080 | 6.19s | 14.34s | **+131.7%** | 1.8µs | 1.4µs |

This is not negligible — the added finite-value scan roughly doubles cold-load wall time on these checkpoints (it faults every zero-copy mmap page it walks, same order as the decode work it rides alongside). It's cached per-tensor via `OnceLock` so it doesn't repeat on a warm re-access to the same handle, but the first load of a large checkpoint (the common case) pays the full cost. Proposed follow-up (not attempted here, kept out of this PR's scope): vectorize the scan (SIMD any-non-finite reduction), and/or an opt-out fast path for already-validated checkpoints. Full numbers and discussion are now in the PR body's "Startup-time note" section.

**2. Body overstatement on Q4/KHF1 guards (corrected).** Confirmed directly against `q4_weights.rs`: the finite-value guard at line 270 (`quantize_block_with_mode_len`) is on the quantization *input* path only. The artifact-loader-output functions `load_q4_file` (line 864) and `load_f16_tensor_file` (line 936) return raw/widened values with no finiteness check at all — the original PR body's claim was wrong for these two. lattice#802 ("unify native Q4/KHF1 file validation") already covers this exact gap and lists this ingress-seam issue as a dependency in its own sequencing section, so no new issue was filed to avoid duplicating it. The PR body's "Out of scope" section is corrected to state this accurately and points at #802.

**3. Branch refreshed and quiet-lane bench re-run.** Branch updated to `4546fcdad` (merged latest `main`, was 1 commit behind). Re-ran `bench-compare` in a quiet lane (single-holder, no concurrent GPU contention) — the one FAIL and all 16 WARNs are entirely in `lattice-embed`'s SIMD/int8/int4 vector-search suite, a crate this PR's diff never touches. Every group this diff could plausibly affect (`rms_norm`, `layer_norm`, `silu_inplace`, `gelu`, `softmax_attention`, `elementwise_mul`) passed silently, consistent with the seam running once at load time and having no hot-path presence in any compute kernel. Full breakdown in the PR body.

CI is re-running at the new head; will confirm green once it completes.

ohdearquant:

# PR #873 — Round 2 Review

**Status:** Complete.

## Review scope

1. Verify the published cold/warm load-timing evidence and its methodology.
2. Verify the PR-body correction and the scope of lattice#802.
3. Check the post-round-1 content diff for unintended source changes.
4. Check head CI, branch freshness, and regressions in the round-1-approved substance.

## Prior review context

- Round 1 held two mediums: no full-scan startup-cost evidence, and an inaccurate claim that the native Q4/KHF1 artifact loaders already had finite guards.
- Round 2 examines only those fixes, the post-review content, and the new-head gate state.

## Evidence log

- The PR body and the 2026-07-11 follow-up comment contain a distinct manual load-timing table for two real checkpoints. They distinguish cold from warm re-access, compare the seam branch with the no-seam `origin/main` baseline, name release builds with `--features f16`, give checkpoint size/tensor/element counts, and correctly present the data as direct manual timing rather than a Criterion result. The 50.4% and 131.7% cold-load deltas support the limited conclusion that the first full finite scan is material startup work; the warm rows support the `OnceLock` cache claim. This is qualitative, two-checkpoint evidence, not a general throughput benchmark.
- The PR correction accurately confines `q4_weights.rs:270-276` to quantization input. `load_q4_file` retains raw Q4 scale/bias bits at `crates/inference/src/weights/q4_weights.rs:864-871`, and `load_f16_tensor_file` widens KHF1 values at `crates/inference/src/weights/q4_weights.rs:936-942`; neither path checks finiteness. The correction therefore accurately defers native artifact validation.
- Read GitHub issue #802. Its stated scope explicitly includes the missing CPU/Metal Q4 scale/bias finite checks, KHF1 decoded-value finite checks, exact extent, and KHF1 shape/numel validation; its sequencing depends on this ingress-validation seam. The PR body's deferral to #802 is accurate and non-duplicative.
- `git log --no-merges 8a7ea236..30ba0e8` contains no commit. The two intervening commits are clean merge commits (no combined-diff conflict resolution), and the current PR-vs-`origin/main` diff remains only the four round-1-reviewed files: `ingress.rs`, `f32_weights.rs`, `weights/mod.rs`, and the QuaRot doc comment. `git diff --check 829595843...HEAD` passed.
- Branch freshness: `origin/main...HEAD` is `0 3`; head `30ba0e8f` contains current `origin/main` (`829595843`).
- Local gates observed: `cargo fmt -p lattice-inference -- --check`; `cargo clippy -p lattice-inference --all-targets --features f16,metal-gpu -- -D warnings`; and `cargo test -p lattice-inference --lib --features f16 weights::` (129 passed, 0 failed) all passed.
- Remote required contexts observed green at head: `actionlint`, `cargo-deny`, `typos`, and the aggregate `ci-gate`. The latter closes successfully only after its `ci`, feature-matrix, bench-compile, MSRV, and semver prerequisites pass. At the final status check, optional/additional e2e and app-binary jobs were still draining; none had failed.

## Findings (0 blockers, 0 majors, 0 mediums, 0 minors)

No findings remain from round 1.

## What I checked and is fine

- **Startup-cost medium closed.** The evidence is published in the PR body and 2026-07-11 follow-up comment, rather than asserted without numbers. It identifies the before/after configurations, uses production-style release + `f16` builds, declares two real checkpoint inputs, distinguishes first cold load from warm re-access, and names the cache contract. The displayed cold deltas demonstrate a material first-load cost, while the warm rows demonstrate that the guard is not re-run after a successful per-tensor validation. The prose does not confuse the direct timing with the separate Criterion comparison.
- **Native Q4/KHF1 correction medium closed.** The correction exactly matches the live source: `quantize_block_with_mode_len` rejects only quantizer source values at `crates/inference/src/weights/q4_weights.rs:270-276`; `load_q4_file` exposes raw Q4 metadata at `crates/inference/src/weights/q4_weights.rs:864-871`; and `load_f16_tensor_file` widens KHF1 values without a finite guard at `crates/inference/src/weights/q4_weights.rs:936-942`.
- **Companion issue scope.** Issue [#802](https://github.com/ohdearquant/lattice/issues/802) specifically covers CPU and Metal Q4 scale/bias finiteness, KHF1 decoded-value finiteness, exact extent, and KHF1 geometry. Its sequencing identifies the ingress seam as its dependency. The PR correctly defers rather than duplicates that work.
- **No regression/smuggled PR changes.** The round-1 source files are byte-for-byte unchanged between `8a7ea236` and `30ba0e8`; the two intervening commits are mechanical merges with no non-merge commit or combined-diff conflict resolution. Current `origin/main...HEAD` still changes only the four round-1-reviewed files. `git diff --check 829595843...30ba0e8` passed.
- **Freshness and local validation.** Head contains `origin/main` (`0` commits behind, `3` ahead). `cargo fmt -p lattice-inference -- --check`, feature clippy, focused `weights::` tests (129 passed), and the broad `cargo test -p lattice-inference --lib` run all completed successfully.
- **Remote CI at head.** GitHub reports the head as `30ba0e8f315dc72b2075eaf031a3225a17d7758d`; required checks, including `ci-gate`, are green. The PR remains a draft and GitHub reports `REVIEW_REQUIRED`, which is expected before this review is posted and is not a code-quality failure.

## What I did not check

- I did not independently reproduce the two local production-checkpoint timings: the temporary harness and those local checkpoint files are not part of the PR artifact. I verified the published configuration, comparison framing, and that the conclusions do not exceed the reported data.
- I did not wait for the non-required e2e/app-binary tail after the required head contexts turned green. Those jobs must still not be ignored if any later reports a failure, but they do not warrant another source-review round while the diff remains unchanged.

## Suggested next round focus

No further review round is needed. The two round-1 mediums are closed, the approved implementation is unregressed, and required CI is green. Monitor the remaining asynchronous job tail; only a failed job or a source change should reopen review.

Domain utility: MEDIUM — the performance/reproducibility briefing supported the calibration that the published measurements establish a qualitative startup-cost conclusion but are not a universal throughput benchmark.

Verdict: APPROVE

oceanwaves630 (approved):
APPROVED at 53f1721a. Lane-C read was completed this morning (10:45, conditions met: #897 filed, warm-anomaly question recorded); the four commits since are all merge-from-main with zero content delta, checks 30/30 green. This is the button. — Leo (λ₀)

PR #877, opened 2026-07-11
feat(bench): schema-v2 contract, perf-policy.toml, gate-math corrections

ohdearquant:

## AI-assisted development disclosure

- [x] This PR was developed with AI assistance (Claude, an Anthropic agent).
- [x] A human reviewed the diff before requesting review.
- [x] Tests were run locally and pass (see Test plan).
- [x] No secrets, credentials, or internal-only content are included.

## Summary

Part of the benchmark-overhaul program, PR 1 of the phased implementation plan: "canonical contract and policy, in days." Extends `scripts/bench_decode_harness.py` to a schema-v2 run-record contract, additive to the existing v1 per-call observation/JSONL schema (unchanged, still fully readable):

- Phase events (`load_start`/`backend_ready`/`prefill_start`/`prefill_end`/`token_available`), resource samples (`phys_footprint_bytes` — schema fields only in this PR, the sampler itself is a later PR), lock receipts, provenance records, cell aggregates, and cell records — the queryable per-cell shape (stable `cell_id`, `metric_family`, `context_point`, `quant_tier`, `device`, aggregate + CI) needed to answer "largest stable gaps by metric/context/quant" without a further schema change. A deterministic ranking test (`RankCellsByGapTest`) proves the shape supports that query today.
- An expected-cell registry (`scripts/bench_expected_cells.toml`), expanded programmatically from compact axis groups, with a coverage check that fails closed on any expected cell that did not execute (never a silent skip) while still allowing an explicitly named `unsupported` cell.
- A fail-closed run-record validator covering: missing path proof, missing/invalid lock receipts (both heavy-lane and metal-gpu), missing resource samples, a validated phase-event sequence, non-finite metric, low valid-n derived from the registered per-metric policy, missing measured-CV, post-run threshold change (policy content drifted under the same/different version), and candidate-SHA mismatch.
- `scripts/perf-policy.toml`: the versioned regression-gate policy (family thresholds, gate-math method, promotion criteria, replacement cap), pinned by SHA so a later re-validation can detect drift.
- `scripts/bench_gate_math.py` (stdlib-only, no numpy/scipy): order-stratified bootstrap of the log-slowdown mean, a one-sided lower confidence bound, Holm step-down correction, an exact Clopper-Pearson bound via bisection, a registered per-metric policy lookup, and a complete per-family gate evaluator wiring all of the above together.

Gate-math corrections identified during design review are encoded as executable functions (and folded into the design document, internal, not part of this diff):

1. Required `n` is derived from the measured same-session residual CV via `perf-policy.toml`'s `[[cv_bands]]`, not fixed at n=7 — the validator refuses any cell lacking a `measured_cv` record. High-CV bands widen the FAIL margin instead of inflating n indefinitely (diminishing power returns past a point).
2. The small-n bootstrap resamples the MEAN of paired log-slowdowns, never the median — the median's bootstrap distribution is capped at a handful of distinct values at small n, too coarse for a Holm-corrected per-cell tail cut.
3. Promotion records report the achieved Clopper-Pearson bound computed from `(null_sessions, failures)` and are rejected if they assert a tighter, aspirational bound; promotion is gated on `null_sessions >= 20` plus a required minimum of live mainline sessions, not an unreachable upfront session count.

No performance claim in this PR: no adapters, no CI wiring, no engine invocation — no cell in the registry is executed here.

## Review fixes

All four majors and two mediums from the review are addressed:

| # | Finding | Fix | Test |
|---|---------|-----|------|
| Major 1 | `bootstrap_upper_bound` (the `(1-alpha)` quantile) was the function documented to feed `CellAggregate.corrected_lower_bound`, but a FAIL decision needs a *lower* one-sided bound above the threshold, not an upper one. | Added `bootstrap_lower_bound` (the `alpha` percentile — the correct tail), rewired `evaluate_family_gate`'s FAIL rule to consume it exclusively, and left `bootstrap_upper_bound` documented as a diagnostic-only upper bound that must never feed a FAIL rule. | `BootstrapLowerBoundTest` (near-null sample's lower bound sits at/below threshold; deterministic 10% slowdown's lower bound clears it) + `EvaluateFamilyGateTest.test_high_cv_margin_flips_fail_to_non_fail`. |
| Major 2 | `validate_run_record` treated `required_n` as optional submitter-supplied metadata and derived noise class from a hard-coded `metric_family in (...)` heuristic that silently mis-classified `embed.batch_p95` as class A instead of the registered class B. | Added `bench_gate_math.resolve_metric_policy` (looks up the registered per-metric policy table, flat or nested) and wired `validate_run_record` to re-derive `required_n` from the measured CV and resolved noise class, rejecting any record whose `required_n` is missing or does not match. | `SubmitterControlledRequiredNTest` (CV 3% + `required_n=None`; forged `required_n=7` at CV 3%; `embed.batch_p95` at n=7 vs n=9; class-C exemption). |
| Major 3 | Metal records only required the `metal-gpu` lock receipt (never `heavy-lane`), and phase-event / resource-sample proof was never validated at all. | `validate_run_record` now requires both continuously-held lock receipts plus non-empty `resource_samples` on `device=metal` cells, and validates a correctly-ordered `load_start -> backend_ready -> prefill_start -> prefill_end -> token_available(+)` phase sequence on every non-unsupported record (`_validate_phase_sequence`). | `MissingLockReceiptTest` (only-metal-gpu / only-heavy-lane / no-resource-samples all reject) + new `PhaseSequenceTest` (empty, missing-phase, misordered, backwards-`monotonic_ns`, non-increasing `token_index` all reject). |
| Major 4 | `validate_promotion_record` hard-coded `min_null_sessions=20` and never examined `mainline_sessions` or any replacement cap — a 20-null/0-mainline record was accepted. | `validate_promotion_record` now consumes the versioned policy, enforces `min_mainline_sessions` (10), and enforces a new registered `max_invalid_pair_replacements` cap (3) added to both `perf-policy.toml` and `PromotionRecord`. | `PromotionMainlineSessionsTest` (0 and 9 mainline sessions reject, 10 passes) + `PromotionReplacementCapTest` (within/exceeding cap, and a policy missing the cap key fails closed). |
| Medium 1 | Holm correction and `fail_margin_multiplier` existed only as isolated, unit-tested helpers with no caller. | Added `bench_gate_math.evaluate_family_gate`: a complete per-family evaluator that computes per-cell one-sided bootstrap p-values, Holm-corrects them across a family at `alpha=0.05`, and applies the CV band's `fail_margin_multiplier` to the FAIL threshold. **Integrated, not deferred.** | `EvaluateFamilyGateTest.test_holm_ordering_changes_the_verdict` (two cells individually below naive `alpha=0.05` are rejected by neither once Holm-corrected as a family) and `test_high_cv_margin_flips_fail_to_non_fail` (identical effect size FAILs at the raw margin, does not FAIL once measured CV lands in the widened-margin band). |
| Medium 2 | `scripts/perf-policy.toml` and `tests/test_bench_gate_math.py` named an internal actor and internal tooling filenames (publication-hygiene violation). | Replaced with self-contained wording ("cross-checked by an independent Monte-Carlo power simulation"). | Full-diff hygiene sweep (`git diff origin/main...HEAD | grep -inE '...'`) comes back clean. |

### Round-2 review fixes

Round 1's four majors and two mediums (above) verified good; round 2 found three new majors in the evaluator code the round-1 fix commit added. All three addressed:

| # | Finding | Fix | Test |
|---|---------|-----|------|
| Major 1 | `evaluate_family_gate` fixed every cell's lower-bound percentile at the strictest step-0 Bonferroni tail (`alpha/m`) before also requiring the Holm reject flag, so Holm could never change a verdict — a fixed, uniformly-stricter bound already gated every cell regardless of Holm's per-cell step-down assignment. The shipped fixture only asserted the internal `holm_reject` flag changed under mutation, never the verdict. | The lower bound is now computed at each cell's own Holm-assigned step-down tail (`alpha/(m-rank)`, the exact level `holm_reject` tests it at, keyed by its ascending-p-value rank), so the FAIL decision and the confidence bound are the same test at the same alpha. | `test_holm_naive_mutation_flips_a_real_verdict`: swaps `holm_reject` for a naive uncorrected `p <= alpha` check via `mock.patch.object` on the same two-cell family as the round-1 fixture — the real evaluator confirms neither cell FAILs, the naive mutation flips one cell's actual verdict from WARN to FAIL. Mutation-verified by reverting the per-cell alpha assignment to the old fixed `alpha/m` and confirming the new test fails, then restoring. |
| Major 2 | `validate_promotion_record` parsed `policy_version` as any int `>= 1` but never compared it against the loaded `policy_doc["policy_version"]` — a record claiming `policy_version=999` with v1-shaped threshold values was accepted and evaluated under v1's thresholds. | `validate_promotion_record` now fails closed unless the record's `policy_version` exactly matches the structurally-validated loaded policy's version, before any threshold comparison runs. | `test_policy_version_mismatch_fails_closed` (a `policy_version=999` record rejects) + `test_policy_version_match_passes` (a matching version still passes). Mutation-verified by removing the version-binding check and confirming the mismatch test fails, then restoring. |
| Major 3 | `evaluate_family_gate` accepted non-finite raw values (seven NaNs returned a `PASS` verdict with NaN bounds) and under-sampled cells (one pair with `required_n=7` returned `PASS`) — the only per-cell shape check was `len(values) == len(order_ab)`, never finiteness or the policy-derived minimum count. | Both are now rejected (raises `GateMathError`) before any bootstrap resampling: a cell short of its own policy-derived `required_n`, or containing a non-finite value, never reaches a verdict. This is fail-closed via exception, consistent with `validate_run_record`'s existing INFRA-FAIL treatment of the same two conditions on an already-aggregated record. | `test_undersampled_cell_rejected_before_resampling` + `test_non_finite_value_rejected_before_resampling`. Mutation-verified by removing both guards and confirming both new tests fail, then restoring. |

### Round-3 review fixes

Round 2's three majors (above) verified closed; round 3 found one remaining major in the same evaluator.

| # | Finding | Fix | Test |
|---|---------|-----|------|
| Major 1 | `evaluate_family_gate` assigned every p-value rank `k` the tail `alpha/(m-k)` and reported a `corrected_lower_bound` at that tail even for ranks AFTER the real Holm step-down had already stopped at an earlier rank's failed comparison — those later-rank cells were never actually tested at their less-stringent tail, so the exposed bound could read as FAIL-level evidence while the `holm_reject` flag correctly said "not rejected." Concrete two-cell case: p-values `0.040`/`0.037`, the step-down stops at the first cell, yet the second cell's exposed bound (`0.068087`) cleared the 7% fail threshold (`log(1.07)=0.067659`) despite never being tested at that tail. | A cell's bound is now computed only when the real step-down procedure actually evaluated that rank before stopping (all ranks up to and including the stopping rank itself, which was tested and failed; every rank after it was not). Untested ranks report `corrected_lower_bound=None` and can never reach FAIL regardless of their `holm_reject` value — the flag/bound disagreement is structurally impossible now. Testedness is computed independently of whichever reject-decision function is plugged in (mocked or real), so it reflects the actual step-down order in every case. | `test_stopped_holm_rank_never_exposes_a_bound`: reproduces the exact adversarial-review two-cell fixture and asserts the untested cell now reports `None` (not `0.068087`), plus a general coherence assertion that no non-rejected cell across the family ever exposes a threshold-clearing bound. `test_holm_naive_mutation_flips_a_real_verdict` (retained, re-tuned fixture so the flip is driven by the always-tested rank-0 cell rather than the now-`None` untested rank) still demonstrates the real Holm flag is decision-effective, and additionally asserts the untested cell stays un-FAIL-able even under a naive/liberal mutation. Mutation-verified by reverting the tested-rank gating in a disposable copy (bound computed unconditionally again) and confirming the new coherence test fails with the exact `0.068087 > 0.067659` disagreement, then restoring. |

### Round-4 review fixes

Round 3's stopped-rank fix verified correct; round 4 found one remaining major -- a coherence defect for a rank that *was* executed.

| # | Finding | Fix | Test |
|---|---------|-----|------|
| Major 1 | `evaluate_family_gate` computed the per-cell p-value (`one_sided_bootstrap_pvalue`) and the `corrected_lower_bound` (`bootstrap_lower_bound`) from TWO INDEPENDENT bootstrap resampling passes, so percentile duality was not guaranteed even for a rank the real step-down actually tested. Concrete case: a single tested cell, seven near-boundary values (`base=log(1.07)+0.005`, `spread=0.02`), `bootstrap_replicates=2000`, `random.Random(108)` gave `p_value=0.051` (correctly not rejected -- 0.051 > 0.05) yet a separately-drawn `corrected_lower_bound=0.0680872...`, which cleared the 7% threshold `log(1.07)=0.06765864847381486` anyway -- a tested, non-rejected cell exposing FAIL-level bound evidence, at the production default replicate count. | `evaluate_family_gate` now generates ONE retained bootstrap-mean sample per cell (`order_stratified_bootstrap_means`, called once), and derives both the p-value and (for tested ranks) the lower bound from that same sample via new `_pvalue_from_boot_means`/`_lower_bound_from_sorted_boot_means` extraction helpers (also now shared by the standalone `one_sided_bootstrap_pvalue`/`bootstrap_lower_bound` functions). For any tested rank this makes `p <= corrected_alpha` (reject) and `bound > tau` algebraically the same fact about the same sample -- they cannot disagree. Also closed the residual replicate-floor gap: the p-value floor `max(p, 1/b)` could itself sit above the smallest tail Holm's step-down will ever test (`alpha_familywise / len(cells)`), letting an under-resolved floor reintroduce the same incoherence with a genuinely-zero bootstrap count. `evaluate_family_gate` now fails closed (before any resampling) when `bootstrap_replicates` cannot guarantee the floor stays at or below that tail. | `test_shared_sample_pvalue_bound_duality_at_production_replicates`: exact seed-108 reproduction from the review, asserts the tested-not-rejected cell's bound is `None` or `<= tau` (was `0.0680872... > tau`). `test_insufficient_replicates_for_floor_coherence_rejected`: a single cell at 10 replicates (below the `alpha_familywise/1=0.05` -> 20-replicate floor requirement) and a 200-cell family at the production default of 2000 replicates (below the `alpha_familywise/200`-driven requirement) both fail closed. `test_holm_naive_mutation_flips_a_real_verdict` -> **rewritten** to `test_holm_naive_mutation_cannot_flip_a_coherent_verdict`: the shared-sample fix makes real-Holm-reject always a subset of naive `p <= alpha_familywise` reject (since `corrected_alpha <= alpha_familywise` for every tested rank), so by duality anything only the naive check additionally flags already has `bound <= tau` -- the previously-demonstrated flip is now mathematically unreachable. The rewritten test confirms the internal `holm_reject` flags still genuinely disagree under mutation (`b`'s real flag stays `False`, naive flips it to `True`) while asserting the shipped verdict is invariant. Mutation-verified: reverted to two independent bootstrap distributions in a disposable copy, confirmed the exact `0.0680872199023863 > 0.06765864847381486` disagreement reproduces and all three new/rewritten tests fail, then restored and reconfirmed green. |

## Test plan

- [x] `python3 tests/test_bench_decode_harness.py -v` — 117 passed (v1 contract regression, unchanged)
- [x] `python3 tests/test_bench_gate_math.py -v` — 63 passed (was 39 pre-round-1, 57 post-round-1, 60 post-round-2, 61 post-round-3; +2 round-4 for the shared-sample duality and replicate-floor coherence regressions; one round-1/2/3 test rewritten, not net-new, to assert the new mutation-invariance property instead of a now-unreachable flip)
- [x] `python3 tests/test_bench_run_record.py -v` — 75 passed (was 47 pre-round-1, 73 post-round-1; +2 round-2 for the `policy_version` binding)
- [x] All three run with plain `python3` (no `uv`, no third-party deps), matching the existing `bench-decode-harness-tests` CI job's stdlib-only, ubuntu-only constraint -- 255 total passed post-round-4 (63 + 75 + 117)
- [x] Fixtures explicitly prove fail-closed behavior for: missing cell, missing path proof, missing/incomplete lock receipts, missing resource samples, missing/misordered phase sequence, non-finite metric, low valid-n, policy-derived required-n mismatch, missing measured-CV, post-run threshold change, wrong SHA, insufficient mainline sessions, exceeded/unregistered replacement cap, promotion `policy_version` mismatch, and (evaluator-level) non-finite/under-sampled raw cells
- [x] Every review fix has a disposable-copy mutation test: the relevant check was removed in a scratch copy, the corresponding test observed to fail, then the copy was discarded and the real fix re-verified green
- [x] `bench-compare`: N/A — this PR touches only `scripts/` and a design document (internal), no `crates/` code, so there is nothing for `make bench-compare` to measure

Design document (internal) citation: benchmark-overhaul program, phased implementation plan row 1 ("canonical contract and policy, in days"), extends the existing decode-harness consolidation lineage. Does not create a parallel harness.

### DIFF
--- a/scripts/bench_decode_harness.py
+++ b/scripts/bench_decode_harness.py
@@ -133,15 +133,25 @@ import subprocess
 import sys
 import time
 import tomllib
+from collections.abc import Callable, Mapping, Sequence
 from dataclasses import dataclass
-from datetime import datetime, timezone
+from datetime import UTC, datetime
 from pathlib import Path
-from typing import Callable, Mapping, Protocol
+from typing import Protocol
 
 SCHEMA_VERSION = 1
 
 REPO_ROOT = Path(__file__).resolve().parents[1]
 DEFAULT_PROFILES_FILE = REPO_ROOT / "scripts" / "bench_decode_profiles.toml"
+
+# `bench_gate_math` is a sibling script module, not an installed package;
+# inserting this file's own directory onto `sys.path` lets `import
+# bench_gate_math` resolve both when this file is run directly
+# (`python3 scripts/bench_decode_harness.py ...`) and when it is loaded by
+# path via `importlib.util.spec_from_file_location` (the test-suite
+# convention -- see tests/test_bench_decode_harness.py).
+sys.path.insert(0, str(Path(__file__).resolve().parent))
+import bench_gate_math  # noqa: E402
 
 AGGREGATION_METHODS = ("median", "trimmed_mean")
 
@@ -897,7 +907,7 @@ def prompt_hash(prompt: str) -> str:
 
 
 def _utc_timestamp() -> str:
-    return datetime.now(timezone.utc).isoformat()
+    return datetime.now(UTC).isoformat()
 
 
 def run_profile(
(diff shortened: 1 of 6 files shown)


ohdearquant:

## Review of PR #877 (round 1)

### Verdict: REJECT

### Review scope

Reviewed head: `510690665262e9e7296a41c3c2d41416e7a75929`.

This round covered the schema-v2 contract and policy with particular attention
to gate mathematics, fail-closed validation, promotion evidence, and v1
compatibility. The artifact remains a live review record while the remote CI
is pending, but the findings below already prevent approval.

### Evidence log

- Observed: the PR body begins with the required AI-attribution line, contains
  the AI checklist, and all three commits at the reviewed head carry
  `Co-Authored-By: Leo <noreply@khive.ai>`.
- Observed: `git diff --check origin/main...HEAD` passed. The six changed
  paths are exactly the claimed `scripts/` and `tests/` surfaces; no crate or
  engine code is changed, so the stated `bench-compare` N/A rationale is
  appropriate.
- Observed: `python3 tests/test_bench_decode_harness.py -v`,
  `python3 tests/test_bench_gate_math.py -v`, and
  `python3 tests/test_bench_run_record.py -v` passed: 117 + 39 + 47 = 203
  tests. They require no network or third-party package. The legacy harness
  test intentionally uses isolated `tempfile` files; the new gate-math
  negative test probes `/nonexistent/perf-policy.toml`, so the suite does not
  satisfy a literal no-filesystem-outside-repo interpretation.
- Observed: every requested standard sabotage rejects: missing expected cell,
  path proof, receipt, finite metric, declared low `n`, measured CV,
  policy SHA, and candidate SHA. Removing the path-proof branch in a
  disposable source copy made `MissingPathProofTest` fail, proving that test
  is mutation-sensitive.
- Derived and observed: with paired log-ratio residual SD `sqrt(2) * CV`, a
  one-sided Holm first-step alpha of 0.01, and true slowdown 10% versus the
  7% threshold, normal approximation gives power 99.8% at CV 1%/n=7 and
  82.4% at CV 3%/n=25. The policy band values are therefore plausible and
  the high-CV band widens the margin rather than raising n.
- Observed: an exact bisection returns CP upper bounds 0.13910834 for 0/20
  and 0.01000243 for 0/298. A 20-session, zero-failure promotion record that
  reports 0.14 is accepted; one reporting 0.009 is rejected.
- Observed: synthetic component exercise is correctly oriented for
  `log_slowdown` and the p-value: seven deterministic 10% throughput
  slowdowns yield `p=0.0005` and a Holm-five rejection; near-threshold null
  data yield `p=0.4095` and no rejection.
- Observed: v1 `validate_observation`/`validate_jsonl` have no diff hunk and
  the claimed v1 regression test passes. The only pre-v2 edits are import
  modernization and equivalent UTC spelling in `_utc_timestamp`.
- Observed: `rank_cells_by_gap` orders absolute gaps descending, excludes
  `unsupported`, and ties by `cell_id`; its deterministic tests pass.
- Observed at review time: remote required CI at this exact head is still in
  progress/queued (macOS, ARM, bench-compile, and MSRV), and the draft branch
  is three commits behind `origin/main` (`0 3`). It is therefore not fresh or
  remotely green yet.

### Findings (4 majors, 2 mediums)

#### Majors

1. [scripts/bench_gate_math.py:169] The function exported as the gate
   confidence bound takes the wrong tail.

   - **Current:** `bootstrap_upper_bound` returns the
     `(1 - corrected_alpha)` bootstrap quantile at lines 177-187, while its
     result is to populate `CellAggregate.corrected_lower_bound`.
   - **Issue:** A slowdown is positive, and the claimed hypothesis is
     `slowdown > threshold`; a confirmed FAIL needs a *lower* one-sided bound
     above the threshold. The current code calculates an upper bound instead.
     On near-null synthetic data (mean `0.067730`, threshold `log(1.07) =
     0.067659`), it returned `0.068230`, while the 1% lower quantile was
     `0.067302`. Treating the former as the documented lower bound silently
     overstates evidence of a regression. NIST's description of the
     percentile interval likewise places the lower endpoint at the alpha
     percentile, not at `1-alpha` ([source](https://tsapps.nist.gov/publication/get_pdf.cfm?pub_id=917303)).
   - **Fix:** Implement and name a lower one-sided bound (and test its
     threshold orientation with noisy 10% and null fixtures), or rename this
     function and ensure it cannot be wired into the `corrected_lower_bound`
     FAIL rule.
   - **Test:** A near-null sample whose point estimate is just above the
     threshold must have lower bound at or below the threshold; a stable 10%
     slowdown must be above it.
   - **Confidence:** Observed by directly executing the exported functions.

2. [scripts/bench_decode_harness.py:2056] The validator does not derive or
   require policy `required_n`; it trusts optional record data.

   - **Current:** `parse_cell_aggregate` makes `required_n` optional at
     lines 1660-1665. `validate_run_record` only rejects low n when it is
     non-null at lines 2062-2067 and determines class with a hard-coded
     family shortcut.
   - **Issue:** A valid-looking Class-A record with CV 3%, `n_valid=0`, and
     `required_n=None` was accepted in direct execution, despite the shipped
     policy requiring 25 pairs for that CV. It also accepted an `embed`
     `batch_p95` record at CV 1% with seven pairs even though the policy marks
     that metric Class B (nine pairs). This turns the advertised fail-closed
     measured-CV policy into submitter-controlled metadata.
   - **Fix:** Load the pinned policy, resolve class from the registered metric
     policy (not a family heuristic), re-derive `(required_n, margin)` from
     measured CV, require an exact recorded value if it is retained for audit,
     and reject a missing/mismatching value.
   - **Test:** Add mutations for CV 3% with `required_n=None`, a forged
     `required_n=7`, and `embed.batch_p95` with n=7; all must reject.
   - **Confidence:** Observed by executing those malformed record variants.

3. [scripts/bench_decode_harness.py:2037] Metal and phase proof validation
   omits mandatory parts of the contract.

   - **Current:** Metal validation accepts any continuously-held
     `metal-gpu` receipt at lines 2042-2048, then proceeds directly to
     aggregate validation at line 2051. It never requires a heavy-lane
     receipt or any phase event.
   - **Issue:** A Metal record with a path proof and only the GPU receipt was
     accepted, although the contract requires both outer heavy-lane and GPU
     receipts. A non-unsupported record with empty `phase_events` and
     `resource_samples` was also accepted. Both conditions permit a future
     measurement to pass without the promised contention proof or measurement
     boundary proof.
   - **Fix:** Require continuously-held receipts for both named locks on
     Metal records, and require/validate the ordered phase sequence required
     for the record path before accepting its aggregate.
   - **Test:** Add a record with only `metal-gpu`, one with only
     `heavy-lane`, and one with an empty/malformed phase sequence; each must
     reject.
   - **Confidence:** Observed by executing both accepted invalid variants.

4. [scripts/bench_decode_harness.py:2083] Promotion validation enforces only
   the CP-reporting subset of the promotion policy.

   - **Current:** `validate_promotion_record` accepts `min_null_sessions`
     alone and never examines `mainline_sessions`; the shipped policy records
     `min_mainline_sessions = 10` at `scripts/perf-policy.toml:91-95`.
     The policy also has no replacement-cap value despite the registered
     maximum of three invalid-pair replacements.
   - **Issue:** A 20-null-session, zero-failure record with
     `mainline_sessions=0` was accepted. This can promote a shadow cell
     without the required mainline evidence. Omitting the replacement cap
     likewise makes the policy incomplete for an explicit infrastructure-fail
     condition.
   - **Fix:** Make promotion consume the versioned policy, require its
     minimum mainline sessions, and add/validate the registered replacement
     cap in the policy/record contract.
   - **Test:** Add 0/9-mainline rejection cases and a policy fixture that
     proves missing or exhausted replacement capacity fails closed.
   - **Confidence:** Observed by executing a 0-mainline promotion record.

#### Mediums

1. [scripts/bench_gate_math.py:195] Holm and the high-CV margin exist only as
   isolated helpers. A repository-wide search found no evaluator that obtains
   per-cell p-values and invokes Holm, and no caller consumes
   `fail_margin_multiplier`; there is also no p-value field in
   `CellAggregate`. The PR says it adds these gate semantics, but only
   supplies unit-tested helper code and does not explicitly disclose that
   applying them to required-cell families is deferred. Either integrate the
   per-family evaluator or state the deferral precisely in the PR body and
   contract documentation.

2. [scripts/perf-policy.toml:34] Added public code names internal review
   artifacts and an actor (`Leo`, `mathcheck sim.py`, `sim2_power_curve.py`);
   [tests/test_bench_gate_math.py:176] repeats the actor/tooling reference.
   This contradicts the PR checklist's no-internal-content assertion and the
   publication-hygiene requirement. Replace these with self-contained
   derivation/test wording.

- CP bisection direction and values are correct for the requested 0/20 and
  approximately-0/298 cases. It rejects a promotion asserting a tighter
  bound than its own `{failures, null_sessions}` allow.
- The CV lookup table itself has the requested n=7/9, n=25, and high-CV
  margin-widening bands. The independent normal approximation agrees with
  its low/moderate-CV power intent.
- The bootstrap resamples means (not medians) and maintains AB/BA stratum
  counts. `log_slowdown` and the one-sided p-value direction behave correctly
  on direct synthetic signal/null exercises.
- The requested standard fail-closed sabotages reject; their path-proof test
  is mutation-sensitive. This does not cover the additional holes above.
- Expected-cell coverage, CP honesty, v1 readability, deterministic ranking,
  policy SHA mismatch, and candidate SHA mismatch behave as claimed.

### What I did NOT check

- No engine/adapter/GPU execution or benchmark comparison: this PR contains
  neither engine code nor a runnable real measurement profile.
- Remote CI completion: required jobs were still queued/in progress at review
  time, and the branch is behind main.

Return after the four major correctness fixes, fresh main rebase, and green
remote CI. The next useful round should run the new policy-derived validator
and a complete per-family gate evaluator through adversarial fixtures.

Domain utility: SKIPPED — the repository design and executable artifacts were
the authoritative material for this review.

ohdearquant:
> _PR review authored by Codex (OpenAI agent) on behalf of @ohdearquant._

# Review of PR #877 (round 2)

**Head reviewed**: `7ba742ab1820a224728410e7a5a3a8c01d1298e9`

### Verdict: REJECT

**Findings**: 0 blockers, 3 majors, 0 mediums, 0 minors.

### Prior review context

- Round 1 rejected `510690665262e9e7296a41c3c2d41416e7a75929` with four
  majors (lower-tail direction, policy-derived `required_n`, Metal/phase
  proof, and promotion evidence) and two mediums (unintegrated family math
  and publication hygiene).
- The round-2 fix commit addresses much of that work. This review re-ran
  the prescribed adversarial checks at the supplied head, tested two
  validators by mutation on disposable copies, and checked the remaining
  evaluator and policy-version contracts.

### What this round examined

- The complete current diff, fix commit `7ba742ab1`, PR body, and the latest
  round-1 comment.
- `validate_run_record`, `validate_promotion_record`,
  `evaluate_family_gate`, their policy inputs, and their new tests.
- Plain-Python regressions, independent malformed inputs, mutation
  sensitivity, v1 isolation, hygiene, branch freshness, and head CI state.

### Majors

1. [scripts/bench_gate_math.py:527](scripts/bench_gate_math.py#L527) Holm is
   still not decision-effective for the claimed FAIL rule.

   - **Current:** every cell's lower bound is fixed at the Bonferroni tail,
     `alpha_familywise / len(prelim)`, before `FAIL` also requires the Holm
     result at [lines 531-536](scripts/bench_gate_math.py#L531).
   - **Issue:** the new fixture does not make Holm “change the verdict,” as
     the PR body claims. I replaced the call to `holm_reject` in a disposable
     copy with naïve `p <= 0.05`. For the shipped two-cell fixture, the
     internal flags changed from `[False, False]` to `[True, True]`, but both
     verdicts remained `WARN`: their lower bounds (`0.067516`, `0.067230`)
     are below `log(1.07) = 0.067659`. Thus the test fails only because it
     asserts the internal flag, not because removing Holm changes a decision.
     More generally, the fixed alpha/m lower bound is at least as stringent
     as every Holm step, so it makes Holm redundant for a cell that actually
     clears that bound. This leaves the advertised per-family Holm behavior
     unproven and not decision-effective.
   - **Fix:** use a coherent cell-specific step-down tail for the lower bound
     and hypothesis decision, or state and implement a Bonferroni-bound
     gate without claiming a Holm decision. Add a fixture in which removing
     Holm changes an actual `FAIL`/non-`FAIL` verdict.
   - **Confidence:** observed by mutating and running the shipped fixture;
     derived from the alpha selection and conjunction in the cited code.

2. [scripts/bench_decode_harness.py:2245](scripts/bench_decode_harness.py#L2245)
   Promotion validation does not bind a record to the versioned policy it
   claims to consume.

   - **Current:** `PromotionRecord.policy_version` is parsed only as an
     integer >= 1 at [lines 1813-1818](scripts/bench_decode_harness.py#L1813).
     `validate_promotion_record` loads `policy_doc` but never compares that
     field with `policy_doc["policy_version"]`.
   - **Issue:** a syntactically valid promotion carrying
     `policy_version=999` and all version-1 threshold values was accepted by
     the shipped validator. It can therefore be evaluated under v1's
     `min_mainline_sessions` and replacement cap while representing a future
     policy version, which contradicts the claimed versioned-policy
     contract.
   - **Fix:** fail closed unless the record's policy version exactly matches
     a structurally validated supplied/default policy version; add a
     mismatch regression test.
   - **Confidence:** observed by direct invocation of
     `validate_promotion_record` with `policy_version=999`.

3. [scripts/bench_gate_math.py:517](scripts/bench_gate_math.py#L517)
   The advertised complete evaluator accepts malformed and under-sampled raw
   inputs and can fail open.

   - **Current:** the only per-cell shape check is equal `values` and
     `order_ab` length. The function derives `req_n` at line 520, but never
     requires that many raw pairs or checks their finiteness before computing
     a verdict.
   - **Issue:** seven `NaN` values returned
     `CellGateResult(... point_estimate=nan, corrected_lower_bound=nan,
     verdict='PASS')`. One zero-slowdown pair similarly returned `PASS` while
     reporting `required_n=7`. The separate run-record validator rejects
     non-finite aggregates, but this function has no production caller that
     enforces that composition and is documented as the complete evaluator.
     A caller using its verdict directly can therefore accept malformed or
     insufficient evidence.
   - **Fix:** reject a non-finite value and `len(values) < req_n` before any
     resampling; cover both with direct evaluator tests.
   - **Confidence:** observed by direct seeded evaluator invocations.

- Major 1's tail direction is corrected. An independent noisy near-threshold
  input gave point estimate `0.071429`, lower bound `0.061429`, and `WARN`;
  stable 10% slowdown gave lower bound `0.095310` and `FAIL`, versus
  `log(1.07)=0.067659`. This agrees with NIST's percentile-bootstrap
  endpoint description: the lower endpoint is the alpha percentile, not the
  `1-alpha` percentile. [NIST, *Computing Confidence Intervals for Common
  IR Measures*](https://tsapps.nist.gov/publication/get_pdf.cfm?pub_id=917303)
- Major 2's three required malformed variants reject: CV 3% with missing
  `required_n`, CV 3% forged as 7, and `embed.batch_p95` Class B with seven
  pairs. The validator derives 25, 25, and 9 respectively from the
  registered metric policy.
- Major 3's only-`metal-gpu`, only-`heavy-lane`, no-resource, empty-phase,
  and misordered-phase records all reject. The valid sequence and both
  receipts remain accepted.
- Major 4's 0- and 9-mainline records, four replacements, and a policy
  missing the replacement cap all reject. The remaining version-binding gap
  is finding 2 above.
- The high-CV margin fixture genuinely changes `FAIL` to non-`FAIL`.
  The Holm fixture is mutation-sensitive on `holm_reject`, but not on its
  final verdict as explained in finding 1.
- Two independent mutation checks were performed in disposable `/tmp`
  copies: removing exact `required_n` equality accepted an otherwise-valid
  record reporting 26 instead of derived 25; removing the resource-sample
  guard accepted a Metal record with no samples. The unmodified source
  rejected both.
- All requested Python suites passed under `python3`: **117 + 57 + 73 =
  247**. `git diff --check origin/main...HEAD` passed. v1's observation
  validation has no semantic diff hunk and its 117-test regression suite
  passes.
- The branch is current with `origin/main`; the PR body includes its review
  fixes table; and the targeted added-line sweep found no removed internal
  actor/tooling references in `perf-policy.toml` or
  `tests/test_bench_gate_math.py`.

### CI state

Observed at review time, the checks completed at this exact head are green,
including `bench-decode-harness-tests`; `feature-matrix (macos-latest)` is
in progress and `bench-compile`, MSRV, and macOS CI remain queued. The head
is therefore not yet remotely green in full.

### What I did NOT check

- No engine, Metal, or heavy-lane benchmark was run; the requested scope was
  pure Python and the diff contains contract/validator code rather than a
  real measurement runner.
- I did not wait for the queued remote CI jobs to complete.

Fix the three majors, add the targeted regressions (especially a true Holm
verdict flip), then return after all required CI jobs are green. Another
round is useful because the missing contracts are local and readily
testable.

Domain utility: SKIPPED — executable repository artifacts and the PR's
stated contract were the authoritative review material.

ohdearquant:

## Review of PR #877 (round 3)

**Head reviewed:** `c276703b0465dbe51b4fe08d72a885b66dcbb85d`

### Verdict: REJECT

**Findings:** 0 blockers, 1 major, 0 mediums, 0 minors.

### Prior review context

- Round 1 found four major contract/evidence defects. The round-1 fix commit
  corrected the lower-tail direction, policy-derived `required_n`, Metal
  execution proof, and promotion evidence contracts.
- Round 2 rejected `7ba742ab1` with three majors: Holm was not
  decision-effective, promotion records were not bound to their declared
  policy version, and the family evaluator could return a verdict from
  non-finite or under-sampled raw evidence.
- This round verifies the round-2 commit `c276703b0`, including independent
  disposable-copy mutations and the requested direct malformed inputs.

### What this round examined

- The latest round-2 PR review comment, the complete round-2 fix diff, the
  current evaluator and promotion-validator paths, their added tests, and
  the PR body's round-2 fixes table.
- Holm rank/tail behavior, including a controlled tie ordering and the
  two-cell stop case used by the new mutation fixture.
- The three plain-Python suites, mutation sensitivity, round-1 regression
  coverage, base-branch relationship, diff whitespace, and checks at this
  exact head.

### Findings (1 major)

#### Majors

1. [scripts/bench_gate_math.py:556] A lower bound is reported at a Holm step
   the cell was never tested at after an earlier step stops the procedure.

   - **Current:** the evaluator unconditionally assigns every p-value rank
     `k` the tail `alpha / (m - k)` at lines 556-563, then calculates and
     exposes that cell's `corrected_lower_bound` at lines 565-570.
     `holm_reject` correctly stops after its first failed comparison at
     [scripts/bench_gate_math.py:244](scripts/bench_gate_math.py:244), so
     later ranks are not tested at their less-stringent tails.
   - **Evidence:** in the shipped two-cell fixture, the p-values are
     `a=0.040`, `b=0.037`; ascending rank is `[b, a]`, so the code assigns
     `a` the 0.05 tail and `b` the 0.025 tail. Holm stops on
     `b: 0.037 > 0.025` and returns `[False, False]`, but `a` is returned
     with `corrected_lower_bound=0.068087`, above its 7% threshold
     `log(1.07)=0.067659`. Thus `a`'s reported bound says FAIL-level
     evidence while its Holm flag says it is not rejected. This is exactly
     the flag/bound disagreement the round-3 request asked to exclude.
   - **Issue:** the docstring's claim at lines 500-510 that this is “the
     exact level `holm_reject` tested it at” is false for every rank after
     a stop. The new true-verdict mutation proves the flag is now
     decision-effective, but it achieves that by calculating a later
     hypothetical bound, not a bound from the same executed Holm test. The
     result exposes a mathematically misleading `corrected_lower_bound` and
     does not close the stated round-2 requirement that the decision and
     bound are the same test.
   - **Fix:** carry the actual step-down execution state into bound handling.
     Do not label a later-rank `alpha/(m-k)` bound as an executed Holm bound
     after an earlier failure; represent it as unavailable/not-tested, or
     implement an adjusted-confidence-interval construction whose bound is
     compatible with the final Holm decision. Add a stop-case assertion that
     no non-rejected cell reports a threshold-clearing corrected bound.
   - **Test:** retain the naive-mutation verdict-flip test, and add the
     above coherence assertion to its real Holm execution path.
   - **Confidence:** observed by direct seeded evaluation at this head;
     derived from the rank assignment at lines 556-563 and the stopping
     semantics at lines 244-250.
   - **Self-rebuttal:** the final verdict remains `WARN` because line 573
     conjoins the flag with the bound, so this is not a current false FAIL.
     That does not make the exposed `corrected_lower_bound` an actual Holm
     bound, nor satisfy the explicitly claimed same-test contract.

- **Round-2 major 2 is closed.**
  [scripts/bench_decode_harness.py:2253](scripts/bench_decode_harness.py:2253)
  structurally validates the loaded version and lines 2259-2265 require
  exact equality before threshold use. A direct promotion record with
  `policy_version=999` raised `RunRecordValidationError`; the matching
  version-1 record accepted. Removing the guard in a disposable copy made
  `PromotionRecordTest.test_policy_version_mismatch_fails_closed` fail.
- **Round-2 major 3 is closed.**
  [scripts/bench_gate_math.py:537](scripts/bench_gate_math.py:537) rejects
  `len(values) < required_n`, and
  [scripts/bench_gate_math.py:543](scripts/bench_gate_math.py:543) rejects
  non-finite input before resampling. Direct seven-NaN and one-pair
  invocations both raised `GateMathError`, not `PASS`. Removing either guard
  in a disposable copy made its corresponding new test fail.
- **The new Holm mutation is genuinely verdict-sensitive.** Replacing the
  real call with naïve `p <= 0.05` in a disposable copy changed the seeded
  fixture from `[(a, WARN), (b, WARN)]` to `[(a, FAIL), (b, WARN)]`; the
  current mutation test also fails when that guard is removed. This closes
  the narrow decision-inertness defect from round 2, but not the stopped-step
  bound defect above.
- **Tie/rank determinism is fine.** Both `holm_reject` and the tail map use
  Python's stable ascending sort. With p-values `[0.01, 0.01, 0.04]`, the
  observed stable order was `[0, 1, 2]`, tails were
  `[0.016666..., 0.025, 0.05]`, and all three Holm flags were true.
- **Tests:** observed under plain `python3`:
  `tests/test_bench_gate_math.py` 60 passed,
  `tests/test_bench_run_record.py` 75 passed, and
  `tests/test_bench_decode_harness.py` 117 passed — **252 total**. This
  also exercises the previously accepted round-1 fixes. The round-2
  disposable-copy mutation checks for Holm, policy-version, under-sampling,
  and non-finite values each made their target test fail as required.
- **Repository/PR state:** `git diff --check origin/main...HEAD` passed. The
  branch contains `origin/main` and is `0 behind, 5 ahead`; the PR body has
  the round-2 fixes table, including all three claimed fixes and its mutation
  assertions.
- **Remote CI:** at the reviewed head, completed required Linux and
  parity/app-binary jobs are green, including `bench-decode-harness-tests`.
  `CI (macos-latest)`, `feature-matrix (macos-latest)`,
  `msrv (1.93, f16+metal-gpu)`, and `bench-compile` were still queued, so
  the head is not remotely green in full yet.
- I consulted NIST's percentile-bootstrap description; its lower endpoint
  is the alpha percentile, consistent with the already-correct
  lower-tail direction and not a finding here.
  [NIST: *Computing Confidence Intervals for Common IR Measures*](https://tsapps.nist.gov/publication/get_pdf.cfm?pub_id=917303)

### What I did NOT check

- No engine, Metal, or heavy-lane benchmark run: this pure-Python contract
  PR provides no executable adapter/measurement lane in scope.
- I did not wait for the queued remote macOS, MSRV, and bench-compile jobs;
  their status above is an observation at review time, not a claim of
  eventual success.

Another round is needed. First correct the stopped-step bound semantics and
add the no-flag/bound-disagreement regression, then return after all required
CI checks at the resulting head are green. If that localized fix passes the
same adversarial fixture and no new issue appears, no further review round
should be necessary.

Domain utility: SKIPPED — the executable repository artifacts, policy, and
observed commands were the authoritative material for this review.

ohdearquant:

## Review of PR #877 (round 4)

### Verdict: REJECT

**Findings:** 0 blockers, 1 major, 0 mediums, 0 minors.

### Prior review context

- Round 3 rejected one major: after a first failed Holm comparison,
  later, untested ranks still exposed a lower bound calculated at a tail the
  procedure had never executed.
- This round reviewed the round-3 repair commit
  `1d27ed35875f4fc29e60d1d16e2860d7486ea4a2`. Its stopped-rank repair is
  correct, but a separate p-value/bound coherence defect remains for a rank
  that *was* executed.

### What this round examined

- The latest round-3 GitHub review, current PR body (including its round-3
  fixes table), the full round-3 fix diff, and the affected evaluator/test
  paths.
- The exact `a=0.040`/`b=0.037` fixture, independent step-down reference
  cases, a disposable sabotage, all three requested plain-Python suites,
  current base/head ancestry, and remote checks at the reviewed head.

### Findings (1 major)

#### Majors

1. [scripts/bench_gate_math.py:601](scripts/bench_gate_math.py:601) and [scripts/bench_gate_math.py:620](scripts/bench_gate_math.py:620) independently resample the p-value and lower bound, so a tested non-rejection can still report FAIL-level bound evidence.

   - **Current:** `one_sided_bootstrap_pvalue` consumes one `b`-replicate
     bootstrap distribution at lines 601 and 152–166. After Holm is run,
     `bootstrap_lower_bound` consumes a second, independent distribution at
     lines 619–622. The docstring says the bound and decision are “the SAME
     test” at the same per-cell alpha at lines 541–561, and the new test says
     no non-rejected cell may expose a threshold-clearing bound at
     [tests/test_bench_gate_math.py:638](tests/test_bench_gate_math.py:638).
   - **Evidence:** I evaluated a single, hence necessarily tested, cell with
     the PR's seven-value near-boundary shape (`base=log(1.07)+0.005`,
     `spread=0.02`), `bootstrap_replicates=2000`, and
     `random.Random(108)`. Observed result:
     `p_value=0.051`, `holm_reject=False`,
     `corrected_lower_bound=0.0680872199023863`, while the 7% threshold is
     `log(1.07)=0.06765864847381486`; verdict is `WARN`. Thus the cell was
     tested but not rejected, yet its exposed bound still says FAIL-level
     evidence. This is not the fixed untested-rank case and occurs at the
     production default of 2,000 replicates.
   - **Issue:** Separate Monte Carlo draws do not preserve the percentile
     duality the artifact claims. The supplied coherence assertion passes
     only for its two-cell seed-0 fixture; it does not establish the claimed
     property generally. A consumer can again observe a false
     flag/bound disagreement, despite the final verdict's `reject and bound`
     conjunction at [scripts/bench_gate_math.py:624](scripts/bench_gate_math.py:624)
     preventing a false FAIL in this particular path.
   - **Fix:** Generate and retain one bootstrap-mean distribution per cell;
     derive both its p-value and its lower percentile from that same sample
     after assigning the Holm tail. Also make the replicate-floor rule
     coherent with the smallest Holm tail (or reject insufficient replicate
     counts), so an allowed small `b` cannot itself make a passing p-value
     incompatible with a clearing bound.
   - **Test:** Add the exact one-cell, 2,000-replicate, seed-108 regression
     above and assert `not holm_reject` implies a `None` bound or a bound at
     or below that cell's fail threshold. A complementary low-replicate
     case should cover the `max(p, 1/b)` floor.
   - **Confidence:** Observed by direct execution at the claimed head;
     derived from the separate bootstrap invocations. Holm's sequential
     rejective stopping model is also consistent with the original
     [Holm 1979 paper](https://doi.org/10.2307/4615733).
   - **Self-rebuttal:** The branch does not produce a false `FAIL` here;
     line 624 still requires both conditions. That is insufficient because
     the PR explicitly exposes the bound as executed evidence and expressly
     claims no non-rejected cell can expose a threshold-clearing one.

- **The round-3 stopped-rank repair is correct.** The helper at
  [scripts/bench_gate_math.py:254](scripts/bench_gate_math.py:254) exactly
  matched an independent ascending-p reference for all-reject, all-accept,
  stop-at-first, stop-at-last, and tied-stop-boundary p-vectors. In
  particular, the exact historical `[a=0.040, b=0.037]` execution reported
  `a: corrected_lower_bound=None, WARN` and
  `b: corrected_lower_bound=0.06751579133095773, WARN`; the old
  `0.068087` is no longer exposed on untested `a`.
- **The new None guard works.** `CellGateResult` documents `None` correctly
  at [scripts/bench_gate_math.py:502](scripts/bench_gate_math.py:502), and
  the FAIL branch requires both a reject and a non-None bound at line 624.
  In a disposable copy, fabricating an untested tail/bound and allowing it
  to reach FAIL made
  `EvaluateFamilyGateTest.test_stopped_holm_rank_never_exposes_a_bound`
  fail with `0.0680872199023863 is not None`; the committed test passes.
- **Regression suites:** observed under plain `python3`:
  `tests/test_bench_gate_math.py -v` — 61 passed;
  `tests/test_bench_run_record.py -v` — 75 passed; and
  `tests/test_bench_decode_harness.py -v` — 117 passed (**253 total**).
  These cover the accepted rounds 1–3 fixes; the remaining major is a
  coherence case those tests do not exercise.
- **Artifact hygiene:** the PR body has the round-3 fixes table;
  `git diff --check origin/main...HEAD` passed; and no failed completed
  remote check was observed at the claimed head.

### Gate and repository state

- Head verified: `1d27ed35875f4fc29e60d1d16e2860d7486ea4a2`.
- The remote `main` advanced to `8cefeb9ed90da2bb8b1578f302157dc12c90e4a8`
  during this review, so the PR is now **1 behind, 6 ahead**. Its new commit
  is unrelated bare-`#[ignore]` reason text, but the branch must still be
  updated and its checks re-run before merge.
- At review time, the head's completed checks were green (including
  `bench-decode-harness-tests`, both completed feature matrices,
  `cargo-deny`, actionlint, semver, and parity gate). Required CI legs
  `CI (ubuntu-latest)`, `CI (macos-latest)`, `CI (ubuntu-24.04-arm)`,
  `bench-compile`, and `msrv (1.93, f16+metal-gpu)` remained pending; this
  is not a claim of full remote CI success.

### What I did NOT check

- No engine, Metal, or heavy-lane benchmark was run: the assigned round is
  pure Python and the PR contains no adapter or measurement implementation.
- I did not wait for the pending remote checks to finish, because the major
  above already requires a new head and a fresh CI run.

Fix the bootstrap-sample coherence major, add its deterministic regression,
update onto current `main`, and re-run the remote checks. A targeted final
review is necessary after that; the round-3 stop-rank fix alone would have
ended review, but this independently reproducible new discrepancy prevents
that stop rule from applying.

Domain utility: SKIPPED — the executable repository artifacts and observed
commands were the authoritative review material.

Verdict: REJECT

ohdearquant:

## Review of PR #877 (round 5)

### Verdict: APPROVE

**Findings:** 0 blockers, 0 majors, 0 mediums, 0 minors.

### Prior review context

- Rounds 1–2 fixed the contract, policy binding, evidence, and evaluator-input failures.
- Round 3 fixed the stopped-Holm-rank problem: unexecuted ranks no longer expose a bound.
- Round 4 rejected one remaining major: a tested rank drew its p-value and lower bound from independent bootstrap samples. The seed-108 fixture therefore produced `p=0.051`, no Holm rejection, but a bound above the 7% threshold.
- This targeted final round reviewed `89312935b`, merged into the current head `7ca28a2c26ad42a660beb1c327c2268a132b0c3d`.

### What this round examined

- The latest round-4 PR comment, the full fix commit, and the complete current `scripts/bench_gate_math.py` and gate-math tests.
- Shared-sample p-value/bound extraction, Holm tested-rank handling, the bootstrap-replicate floor, and the disclosed mutation-test rewrite.
- The three requested stdlib-only suites, a pre-fix in-memory sabotage check, branch ancestry, PR body, whitespace, and remote CI at the actual head.

- **Round-4 major is closed.** `evaluate_family_gate` generates exactly one `order_stratified_bootstrap_means` sample per cell at [scripts/bench_gate_math.py:679](scripts/bench_gate_math.py:679), derives its p-value from it at [scripts/bench_gate_math.py:688](scripts/bench_gate_math.py:688), and derives the tested-rank bound from its sorted copy at [scripts/bench_gate_math.py:711](scripts/bench_gate_math.py:711). It no longer calls either standalone function that would draw another sample.

- **Exact seed-108 regression passes.** I re-ran the round-4 shape with `B=2000` and `random.Random(108)`. It returned `p_value=0.051`, `holm_reject=False`, `corrected_lower_bound=0.06751579133095773`, and `WARN`; the threshold is `log(1.07)=0.06765864847381486`. The non-rejected tested cell's bound is therefore not threshold-clearing. The shipped regression is [tests/test_bench_gate_math.py:650](tests/test_bench_gate_math.py:650).

- **The required directional duality is now structural, not fixture luck.** Let `c` be the number of retained bootstrap means at or below threshold, `B` the replicate count, and `a` the executed Holm tail. The p-value is `max(c/B, 1/B)` and the bound is order statistic `ceil(aB)`. The new floor check guarantees `1/B <= a` for every possible executed tail. Thus `p > a` implies `c >= ceil(aB)`, so a tested non-rejection's bound is `<= threshold`; if the rank was not tested, the bound is `None`. I independently exercised 50 results across five seeds (`0, 1, 15, 108, 999`) and four family/cell shapes, and found no tested non-rejected cell with a threshold-clearing bound. This agrees with the sequential stop behavior in [Holm (1979)](https://www.jstor.org/stable/4615733).

- **The rewritten naïve-Holm mutation is logically sound and discriminating.** A real Holm rejection satisfies `p <= a <= alpha_familywise`, hence is also a naïve rejection. A naïve-only rejection is either after the real stop (bound `None`) or a tested real non-rejection (bound `<= threshold` by the preceding invariant); it cannot turn a non-FAIL verdict into FAIL. The test at [tests/test_bench_gate_math.py:488](tests/test_bench_gate_math.py:488) still proves the mutation is non-vacuous by changing `b`'s internal flag. I executed that exact current fixture against the pre-fix two-draw source in memory: real `b` was `WARN` with `holm_reject=False`, while the naïve mutation made it `FAIL` with `holm_reject=True`; the current test's verdict-invariance assertion failed as required.

- **Quantile-grid precision note, not a defect.** The PR's informal wording that rejection and a threshold-clearing bound are always the same fact is one direction too strong at an exact discrete boundary. If `aB` is integral and `c=aB`, then `p=a` rejects under `<=`, while the `ceil(aB)` lower bound is `<= threshold`. I verified this with a 20-mean retained sample at `a=0.05`. The direction needed by this fix—and used by the rewritten test—is intact: `bound > threshold` implies rejection, and non-rejection implies no clearing bound. The equality case merely remains `WARN` under the evaluator's documented conjunction at [scripts/bench_gate_math.py:720](scripts/bench_gate_math.py:720); it cannot recreate the round-4 false evidence or a mutation-driven verdict flip.

- **Replicate-floor check fails closed at the right boundary.** [scripts/bench_gate_math.py:645](scripts/bench_gate_math.py:645) rejects before resampling when `B * alpha_familywise / m < 1`. Independently: `B=10,m=1,alpha=0.05` raised `GateMathError` (minimum 20); `B=2000,m=100` accepted; `B=2000,m=101` raised `GateMathError` with minimum 2020. The shipped low-replicate regression covers both a one-cell low-`B` case and an oversized family at [tests/test_bench_gate_math.py:696](tests/test_bench_gate_math.py:696).

- **Rounds 1–3 remain covered and green.** Observed under plain `python3`:
  - `python3 tests/test_bench_gate_math.py -v` — 63 passed.
  - `python3 tests/test_bench_run_record.py -v` — 75 passed.
  - `python3 tests/test_bench_decode_harness.py -v` — 117 passed.
  - Total: **255 passed**. These are stdlib-only and include the accepted validator, policy, stop-rank, non-finite, under-sampling, and v1-contract coverage.

- **Artifact and integration state are good.** `git diff --check origin/main...HEAD` passed. After fetching, the head is **0 behind and 8 ahead** of `origin/main`. The PR body contains the round-4 fixes table and the rewritten-test disclosure. At this exact head, the completed `CI`, `app-binaries`, and `e2e-parity` workflow runs all succeeded; their required gates, including `bench-decode-harness-tests`, `bench-compile`, MSRV, macOS/ARM matrices, and `parity-gate`, are green.

### What I did NOT check

- No engine, Metal, or heavy-lane benchmark was run. The reviewed fix is pure Python gate mathematics and contains no executable measurement adapter.
- I did not run a full workspace Rust suite or performance benchmark; the diff does not touch Rust crates, and the required three Python suites plus remote CI cover this scope.

Review is complete. The sole round-4 major is closed, the disclosed test rewrite is both mathematically valid for the required invariant and mutation-sensitive against the old implementation, and no new material issue appeared. **No further review round is needed.**

Domain utility: SKIPPED — the current executable artifacts, the PR history, and independently observed test/CI evidence were authoritative for this targeted review.

Verdict: APPROVE

oceanwaves630 (approved):
Lane C full read + independent math verification (Leo). An analyst leg re-derived the statistics by calling this PR's own evaluate_family_gate under scipy/Monte-Carlo, not a reimplementation. Verdict on the shipped scope (schema + policy + gate-math library, zero cells executed, zero CI wiring): SOUND — the implemented math is correct and independently reproduces the PR's worked numbers (per-band power consistent; the 13.91% Clopper-Pearson claim matches bit-for-bit).

APPROVED on that scope, with BINDING CONDITIONS on the CI-wiring follow-on (PR 2+), recorded here so they cannot be lost:

1. CROSS-FAMILY MULTIPLICITY (load-bearing): Holm here is scoped to one metric-family table per call; perf-policy.toml registers 9 independently-gating tables. If several gate in one CI invocation at alpha=0.05 each, suite-level FPR ≈ 1-(0.95)^k — at k=9 that is ~37%, the retracted ADR-107 number reappearing one abstraction level up. The wiring PR must either apply a cross-family correction across all tables gating in one invocation, or structurally guarantee one family gates per invocation. This PR fixes the within-family layer only; the description's "gate-math corrections" framing should not be read as closing the whole class.

2. VARIANCE PROVENANCE: cv_bands boundaries are design-doc Monte-Carlo values, not same-SHA null-session data (measured_cv is schema-only here). The wiring PR must enforce that a non-promoted cell can only produce advisory verdicts — no blocking threshold without same-SHA calibration.

Non-blocking asks for this PR or a fast-follow doc commit:
3. HIGH-CV BAND HONESTY: the >5% CV band at n=25/margin=2.0 has ~0.1% power against the stated 10% target (it detects only ~20%+ regressions). Defensible trade-off, but the design doc should say "high-CV cells are advisory/trend-only in practice", not imply corrected power.
4. SMALL-N TYPE-I NOTE: plain percentile bootstrap at n=7 runs ~10.6% observed Type-I at the threshold boundary vs nominal 5% (measured against this PR's real evaluator; decays to 5.4% by n=25). Document it; consider BCa/bootstrap-t later.
5. Off-by-one: theoretical_sessions_for_bound_lt_1pct_at_95conf should be 299, not 298 (CP upper at n=298 is 1.0002%). Informational constant, fix at convenience.

oceanwaves630 (approved):
Cross-account approval — Leo full content read APPROVED (Lane C, independent scipy/Monte-Carlo recomputation), codex clean APPROVE round 5 posted on-PR, CI rerunning at refreshed head.

PR #865, opened 2026-07-11
chore(inference): deprecate crate::generate module and Qwen35Model::generate_with_batch_prefill on 0.5.x (remove in 0.6.0)

ohdearquant:
> _PR description authored by Claude (Anthropic agent) on behalf of @ohdearquant._

## Summary

- Marks `crate::generate`'s exported items (`GenerateConfig`, `GenerateOutput`, `generate`) and `Qwen35Model::generate_with_batch_prefill` `#[deprecated(since = "0.5.1", ...)]`, scheduled for removal in 0.6.0. Both paths are repository-dead (no production caller — bin, example, library, or app) and duplicate the canonical, actively maintained Qwen3.5 decode loop (`Qwen35Model::generate` / `generate_streaming`, which already delegates its prefill phase to the same batched-prefill core via PR #680).
- **No code removed, no behavior change.** This is a deprecation-notice-only PR: every deprecated item keeps compiling and working exactly as before during the deprecation window.
- Adds a module-level migration guide in `crates/inference/src/generate.rs` (verified field-by-field `GenerateConfig` mapping, read at source) and corrects `docs/architecture.md`'s prose/trace, which previously described `Qwen35Model::generate_with_batch_prefill` as the generation entry point — the actually-canonical entry points are `Qwen35Model::generate` / `generate_streaming`.
- Fixes two stale doc comments in `crates/inference/src/forward/batch_prefill.rs` left over from before PR #680's delegation landed (they said `generate_with_batch_prefill` was slated to become/replace `generate()`, which is now backwards).

## Migration mapping summary (full table in `generate.rs` module docs)

| `crate::generate::GenerateConfig` | Canonical (`crate::model::GenerateConfig`) | Notes |
|---|---|---|
| `max_new_tokens: usize` | `max_new_tokens: usize` | same |
| `sampling: SamplingConfig` | `temperature`, `top_k`, `top_p`, `repetition_penalty` | canonical flattens the nested config |
| `eos_token_id: Option<u32>` | *(none — read from `Qwen35Config::eos_token_id`)* | canonical EOS is fixed per loaded model; use `stop_token_ids` for extra per-request stops |
| `include_prompt: bool` | *(no equivalent)* | canonical `generate`/`generate_streaming` **always** exclude the prompt from `GenerateOutput::text`/`token_ids` — verified at source (`model/qwen35_config.rs`, `GenerateOutput::text` doc: "excluding prompt"). Callers must prepend the prompt themselves if they need it. |
[...]

PR #882, opened 2026-07-11
feat(bench): real CPU flagship load->prefill->decode smoke (row 2)

ohdearquant:

**AI-assisted development checklist**
- [x] I have read the changes and understand what they do
- [x] I have run the relevant tests locally
- [x] I have reviewed the diff for correctness, not just skimmed the summary

Part of the benchmark-overhaul program (see #877); Depends on #877.

## Summary

Row 2 of the benchmark-overhaul phased plan: "Real CPU flagship smoke" — an
adapter/driver around the production load → prefill → decode boundary, a
macOS supervisor that samples process memory and computes the five mandated
metrics, and a paired A/B session driver, all producing schema-v2 run
records (#877) in SHADOW mode (no CI gate wired in this PR).

- **Rust side** (`crates/inference/src/bin/qwen35_generate.rs
  --emit-phase-events`): extends the existing demo binary rather than
  writing a parallel forward implementation. Reuses
  `Qwen35Model::from_safetensors` — a default-off raw-event observer
  (`generate_streaming_with_observer` / `RawGenEvent`, see "Round-1 review
  fixes" below) drives the phase-event emission at the exact raw-token
  boundary the production CPU decode loop (`chat_metal` / `lattice_serve` /
  `lattice.rs` all still call the same `generate_streaming_with_cancel`)
  crosses. Prints monotonic `load_start`/`backend_ready`/`prefill_start`/
  `prefill_end`/`token_available` phase-event lines on stdout,
  field-shape-compatible with #877's `PhaseEvent` record. Builds an
  exact-token-length prompt for a requested `--context` via a two-phase
  word-then-character search against the real tokenizer (fails closed
  rather than approximating). Originally added `GenerateConfig.disable_eos`
  so a benchmark trial always decodes the exact requested token count via
  `should_stop_token` (the single shared stop predicate every CPU/Metal
  decode loop already calls) — **superseded in round 3 below**: that field
  cannot ship on `0.5.x` (`semver-checks` `constructible_struct_adds_field`,
  a required gate) and was replaced with `Qwen35Model::config_mut().eos_token_id
  = u32::MAX`, reaching the identical `should_stop_token` behavior without
  adding any public field.
- **Python side** (`scripts/bench_cpu_flagship_supervisor.py`): spawns one
  fresh child process per trial, samples `proc_pid_rusage`
  (`ri_phys_footprint`) every 10ms plus at each phase marker via `ctypes`
  FFI (kept out of Rust `unsafe`, per the design doc), computes the five
  mandated metrics, drives an AB/BA-balanced paired session sweep, and
  assembles a schema-v2 `CellRecord` that is run through #877's
  `validate_run_record` as the acceptance round-trip proof. Every trial's
  raw phase-event trace is now validated (order, monotonicity, contiguous
  token indices, count agreement with the child's own summary) before any
  aggregation happens, and a hung/misbehaving child is bounded by a real,
  enforced timeout with kill+reap (see "Round-1 review fixes" below).

## Round-1 review fixes (codex REJECT → this round)

Codex's round-1 review (2 blockers, 1 major; full text in the review
thread) found the phase-event boundary was measured at the wrong point and
the supervisor could accept a malformed/reordered trace or hang
indefinitely on a stuck child. All three are fixed in this round:

- **Blocker 1 — wrong phase-event boundary.** `prefill_end` and
  `token_available` were previously emitted through the incremental UTF-8
  detokenizer's *text* callback, which deliberately buffers incomplete
  multi-byte sequences — one text callback is not guaranteed to equal one
  sampled token, and `prefill_end` fired after the first sample rather
  than before it. Fixed by adding a default-off `RawGenEvent` observer
  (`Qwen35Model::generate_streaming_with_observer`) that fires exactly at
  the raw-token boundary: `PrefillEnd` immediately before the first
  `sample_token` call, and `RawToken{index}` immediately after each raw
  token is pushed (both the pre-loop first token and both decode-loop
  paths). `generate_streaming_with_cancel`'s existing 4-argument signature
  is unchanged and behaves identically when no observer is installed — it
  is now a thin wrapper around the observer-taking method. The adapter
  asserts the raw event count equals `GenerateOutput.generated_tokens`
  before printing the summary line. Two new mutation-sensitive
  `generation.rs` tests prove `prefill_end` precedes the first sample and
  that a buffered incomplete-UTF-8 first token still produces one raw
  event per generated token, independent of when the text callback fires.
- **Blocker 2 — supervisor accepted malformed/reordered traces.** The
  supervisor sorted `token_available` events by `token_index` before
  computing metrics (masking, not rejecting, an out-of-order stream),
  derived `generated_tokens` from the observed event count instead of the
  child's own summary, and — because `validate_run_record`'s
  `_validate_phase_sequence` check only runs for non-`unsupported`
  verdicts — never validated a trial's trace at all when the record would
  end up downgraded to `unsupported`, which is exactly this PR's own n=2
  demonstration path. Fixed by a new `_validate_trial_trace`, called
  unconditionally from `run_one_trial` for every trial before any
  aggregation or verdict decision: reuses `bench_decode_harness`'s own
  order/monotonicity validator, adds an explicit token-index-contiguity
  check (strictly-increasing alone still permits gaps), and cross-checks
  the observed `token_available` count against both
  `summary.generated_tokens` and `delta_matches_generated_tokens`.
  `compute_trial_metrics` no longer sorts (arrival order is now proven
  upstream, never repaired) and trusts the summary's token count with a
  defensive equality re-check. New tests cover every rejection case codex
  named by name (reversed events, repeated/missing phases, decreasing
  timestamps, duplicate token indices, gapped token indices,
  `delta_matches_generated_tokens=false`, summary/event-count mismatch)
  plus a dedicated test proving an underpowered/`unsupported`-bound record
  still gets its trace validated.
- **Major — timeout could not bound a hung child.** The supervisor's
  blocking `for raw_line in proc.stdout` loop ran before
  `proc.wait(timeout=...)`, so a child that hung with stdout still open
  could never reach the timeout, and the timeout path (when reached)
  neither terminated nor reaped the child. Fixed by draining stdout/
  stderr concurrently on two background daemon threads while
  `proc.wait(timeout=timeout_s)` runs on the main thread; on
  `TimeoutExpired` the child is killed and reaped before the function
  returns. New tests: a silent sleeping fake child is confirmed killed and
  reaped (`os.kill(pid, 0)` → `ProcessLookupError`) within a small bound,
  and a fake child that fills its stderr pipe while holding stdout open
  does not deadlock. Both tests run `run_one_trial` on a joined, bounded
  background thread precisely so that if this fix ever regresses, the
  test fails cleanly instead of hanging the suite — confirmed directly:
  reverting the `proc.kill()` call turns a bounded `TrialFailure` into an
  actual unbounded hang (the reader thread's blocked `read()` holds the
  stdlib io object's lock, so the main thread's later
  `proc.stdout.close()` blocks on that same lock forever), which is what
  the bounded-thread test harness is there to catch.

## Known, disclosed simplifications (in scope for this PR)

- **A/A self-consistency, not A/B regression.** This PR introduces the
  harness itself — there is no candidate-vs-base code diff to gate yet.
  The paired driver runs the identical binary/config on both arms
  (different seeds), proving the paired-comparison machinery produces a
  log-slowdown distribution centered near zero, not a claim about a code
  change. A future PR wiring this into CI supplies real candidate/base
  arms.
- **Thermal state is captured, not enforced.** `NSProcessInfo.thermalState`
  is read before/after every pair and recorded, but this PR does not
  implement the registered thermal-envelope invalidation/replacement
  protocol — that envelope belongs to the protected lane this PR does not
  provision.
- **Demonstration run, not the registered n=7 smoke.** The shared build
  machine was heavily contended during this PR's evidence runs (multiple
  queued heavy jobs ahead of this one, both in round 1 and this round).
  The evidence below is disclosed as `run_kind: "demonstration"` with
  `n_pairs=2`, not the registered `n=7` smoke session. An underpowered
  record is downgraded to `verdict: "unsupported"` before assembly —
  `validate_run_record`'s low-valid-n check only fires on non-`unsupported`
  verdicts, so the demonstration record validates cleanly rather than
  needing the validator to reject it; this is the intended, honest
  behavior for a small-n run, not a bug, and it is exactly the case
  Blocker 2's `_validate_trial_trace` fix now proves is still validated at
  the trace level even though the schema-level check is skipped for it.

## Evidence run (regenerated this round, with the round-1 fixes applied)

Real run against a local Qwen3.5-0.8B f16 checkpoint (weight file sha256
`04b1c301231dd422...`), CPU-only, `n_pairs=2` per context point. Raw report
JSON committed at `scripts/bench_evidence/pr882/report_ctx{512,1024}.json`
(publication-hygiene grep-checked clean before committing — no
internal-ops content, only local filesystem paths under the operator's own
home directory, which is the field's designed provenance content, not a
leak) so the phase traces and `validate_run_record` outcomes can be
replayed without a checkpoint.

Provenance note: `report_ctx1024.json`'s `provenance.dirty` is `false` at
this PR's fix commit (`9d0299f`); `report_ctx512.json` was captured a few
minutes earlier, against the identical already-built binary and Python
source, but before that commit landed, so its `provenance.dirty` is `true`
against the prior (parent) commit `e5dfceab2`. Both ran the exact same code
this PR ships — the dirty flag reflects commit-timing of the evidence
capture, not a code difference between the two reports — but it is
disclosed here rather than silently normalized.

| context | prompt_tokens | generated_tokens | load_ms | prefill_tok/s | TTFT ms | decode_tok/s | peak mem | measured_cv | required_n (n=2) | verdict |
|---|---|---|---|---|---|---|---|---|---|---|
| 512  | 512  | 128 | 6210.2 | 122.4 | 4184.1  | 13.95 | 6221 MB | 0.057 | 25 | `unsupported` |
| 1024 | 1024 | 128 | 5175.4 | 86.4  | 11858.0 | 8.31  | 6221 MB | 0.085 | 25 | `unsupported` |

(figures are pair 0's arm-A trial per context; `order_balance` is `[1,1]`
AB/BA-balanced at each context.) Every record is downgraded to
`verdict: "unsupported"` before assembly because `n_valid=2 <
required_n=25` for the measured same-session CV at each context — this is
`build_decode_cell_aggregate` and `validate_run_record` both working
exactly as designed on an underpowered n, not a defect.

**ctx=2048 not included this round.** A third context-point run was
started (same command, `--context 2048`) but the shared build machine was
contended enough during this window that it ran well past the ~55s/trial
this PR's own harness reported for that context in round 1, and it was
stopped before completing all 4 trials (2 pairs × 2 arms) rather than
letting the evidence-regeneration task consume unbounded session time; no
partial/incomplete report was written or committed. The two committed
contexts (512, 1024) already exercise the fixed phase-event boundary
(Blocker 1) end to end and are each independently proven by
`_validate_trial_trace` (Blocker 2) and the timeout/reap-hardened
`run_one_trial` (the Major); ctx=2048 exercises the identical code path
with a larger prompt, so its omission here is a coverage gap in the
evidence table, not in the fix.

`validate_run_record(record, expected_repo_sha=..., current_policy_sha=...)`
raised nothing for both regenerated records — the schema-v2 round-trip
acceptance proof this PR's row requires, now additionally backed by
`_validate_trial_trace` having independently proven each trial's raw
phase-event trace (order, monotonicity, contiguous token indices, count
agreement with the child's own summary) before that record was even
assembled.

Reproduce (informational-only `phys_footprint`/`thermal_state` off macOS):

```
cargo build --release --bin qwen35_generate -p lattice-inference --features f16
uv run scripts/bench_cpu_flagship_supervisor.py \
  --model-dir <path to a Qwen3.5-0.8B safetensors checkpoint> \
  --context 512 --max-tokens 128 --warmup-tokens 16 --n-pairs 2 --seed 42 \
  --run-kind demonstration --out report_ctx512.json
```

## Existing-path perf-neutrality (bench-compare)

Touches `crates/inference/src` (`generation.rs`'s `should_stop_token`,
`qwen35_config.rs`'s `GenerateConfig`). The added `disable_eos` field
defaults to `false`, byte-for-byte preserving the pre-existing behavior of
every caller that doesn't set it. `make bench-compare`'s default
`elementwise_cpu_bench`/`simd` groups don't exercise the decode-loop stop
predicate at all (it's a control-flow branch in the per-token stop check,
not an elementwise/SIMD kernel), so a same-worktree source-swap A/B
against this PR's merge-base was run instead (round 1; not rerun this
round — this section's underlying code, `should_stop_token`'s
`disable_eos` branch, is untouched by the round-1 fixes): build the
release binary at both HEAD and the merge-base commit, then run the
*existing* (unflagged) `qwen35_generate --prompt` demo path 3x per side
with the same prompt/seed/max-tokens, reading the binary's own reported
`tok/s`:

| run | speed (tok/s) |
|---|---|
| HEAD #1  | 11.4 |
| HEAD #2  | 10.7 |
| HEAD #3  | 13.9 |
| base #1  | 14.7 |
| base #2  | 14.1 |
| base #3  | 14.3 |

HEAD trends slower in this small (n=3/side) sample, but this comparison
ran on the same contended machine immediately after a 6-trial evidence
sweep (see above) and shares that noise floor — it is not a controlled,
isolated A/B and should not be read as a measured regression. The
stronger evidence is structural: the only runtime change on this path is
`disable_eos`'s single `if gen_cfg.disable_eos { return false; }` early
check inside `should_stop_token`, executed once per generated token
against per-token latencies of tens to hundreds of milliseconds here — a
~1ns branch is roughly seven orders of magnitude below the noise visible
in this table, and the field defaults to `false` so every existing caller
takes the identical code path it always has. No regression is expected or
supported by the mechanism of the change; the empirical numbers above are
disclosed as contended/inconclusive rather than omitted.

## Test plan

- `python3 -m pytest tests/test_bench_cpu_flagship_supervisor.py
  tests/test_bench_run_record.py tests/test_bench_gate_math.py -q` — 169
  passed (up from 157 at round 1; +12 new tests: the trace-validation
  rejection cases codex named, and the bounded timeout/reap tests) —
  fake-child-output parsing/aggregation tests, exact-value metric math,
  gate-math folding (sufficiently-powered vs undersized n), the
  `validate_run_record` round-trip proof for both verdict paths (also
  verified against three real trials' live phase-event output, not just
  fixtures), and no regression on #877's own suite.
- `cargo test -p lattice-inference --lib --features f16 generation::` — 28
  passed, 4 ignored (require a local checkpoint) — the 2 round-1
  `disable_eos` mutation-sensitive tests, round-2's 2 `RawGenEvent`/
  raw-observer mutation-sensitive tests (prefill/first-sample ordering;
  buffered-incomplete-UTF-8 token counting), plus this round's 1 new test
  closing the round-2 codex medium finding (below).
- `cargo fmt --all -- --check` — clean.
- `cargo clippy -p lattice-inference --features f16 --lib --bins -- -D
  warnings` — clean; also `--features f16,metal-gpu --all-targets -- -D
  warnings` (test-target lints) — clean.
- Mutation verification (this round, not just "tests pass"): each new
  Rust and Python check was independently reverted (targeted mutation,
  never `git checkout` over uncommitted work — backup/restore via `cp`),
  rebuilt, confirmed to fail the relevant new test, then restored and
  confirmed to pass again. Covered: both `generation.rs` raw-observer
  tests, the trace-validation contiguity/count-mismatch/delta-matches/
  reused-order-validator checks, and the timeout/kill/reap path (removing
  `proc.kill()` turns a bounded `TrialFailure` into a genuine unbounded
  hang, confirming the kill is load-bearing, not cosmetic).
- Manual smoke: `qwen35_generate --emit-phase-events --context 512
  --max-tokens 8 --warmup-tokens 4` against a local Qwen3.5-0.8B
  checkpoint — exact `prompt_tokens=512`, `generated_tokens=8` (EOS
  forced off), `delta_matches_generated_tokens=true`.

### Round 2 follow-up (codex APPROVE-WITH-FIXES, 1 medium)

Codex round 2 found `raw_observer_prefill_end_precedes_first_raw_token_monotonic_index`
proves `PrefillEnd` fires before the `RawToken` *callback*, but not before
`sample_token` itself — a `PrefillEnd` moved to just after
`generated_ids.push` (after sampling, before the `RawToken` callback) would
still pass that test while silently folding sampling time into the reported
prefill interval. Fixed with a narrowly-scoped `#[cfg(test)]`-only seam
(thread_local + three helper fns, zero production effect) that stamps
"first sample entered" at the exact point `sample_token` is called, plus a
new test `raw_observer_prefill_end_precedes_first_sample_not_just_first_raw_token`
comparing it against "`PrefillEnd` observed". Mutation-verified: applying
codex's exact mutation (move `PrefillEnd` to fire after
`generated_ids.push`, before the `RawToken` callback) fails the new test
(`Some(false)` vs `Some(true)`) while the pre-existing event-order test
stays green — reproducing exactly the blind spot codex identified; restored
and both pass.

Fixing this surfaced a separate, pre-existing build breakage at this PR's
prior head (`72d9c9c87`): an earlier commit in this branch (`969b7df15`,
CPU flagship phase-event adapter) added `GenerateConfig.disable_eos`
without updating 38 exhaustive struct-literal call sites across
`metal_qwen35.rs`, four `src/bin` binaries, one integration test, and one
bench — breaking `bench-compile`, `feature-matrix (macos-latest)`,
`msrv (1.93, f16+metal-gpu)`, and the app-shipped-binaries build. All 38
sites now set `disable_eos: false` (parity-safe, matches
`Default::default()`). Verified clean at `cargo check --workspace
--features f16,metal-gpu --all-targets` (both toolchain 1.94.1 and MSRV
1.93.1), all six `feature-matrix` steps, and `scripts/build-app-bins.sh`.

### Round 3 follow-up: restructured off `GenerateConfig` (semver-checks gate resolved)

Per maintainer ruling: `GenerateConfig.disable_eos` cannot ship on `0.5.x`
as a new field — `semver-checks` is a required gate and correctly flags
`constructible_struct_adds_field` (confirmed round 2 above; `#[non_exhaustive]`
does not fix it, it only trades one major violation for another). Rather
than wait on an 0.6.0 version-bump decision (which collides with the
ADR-069 vision-encoder breaking change already reserved for that slot),
this round removes the field entirely and reproduces the identical
benchmark behavior through the crate's own existing surface.

First: **which `GenerateConfig` did 969b7df15 touch?** `model::qwen35_config::GenerateConfig`
— the active config `Qwen35Model::generate`/`generate_streaming`/
`generate_streaming_with_cancel` take, and the one `qwen35_generate`
already imports and constructs. It is **not**
`crate::generate::GenerateConfig` (the module-level `#[deprecated(since =
"0.5.1", ...)]` struct scheduled for removal in 0.6.0 per #865, which
`969b7df15` never touched). So the field was on the *live* struct, making
the semver gate doubly correct to block it — there was no "it's already
deprecated" out.

**Restructure**: `should_stop_token` (the single shared stop predicate
every CPU/Metal decode loop calls) reverts to its exact pre-969b7df15
body — no `disable_eos` branch. `GenerateConfig.disable_eos` is deleted
(struct field, `Debug` impl, `Default` impl); all 38 call sites cf43bf800
touched revert to their pre-969b7df15 form (nothing to default — the
field doesn't exist). In its place, one new additive method:
`Qwen35Model::config_mut(&mut self) -> &mut Qwen35Config` — `Qwen35Model.config`
is a private field, so this is the only way an external caller can adjust
it post-load. A genuinely new `pub` item is minor-at-most under
`semver-checks`, never `constructible_struct_adds_field` (that lint only
fires on *existing* constructible structs gaining fields). `qwen35_generate
--emit-phase-events` now calls `model.config_mut().eos_token_id = u32::MAX`
before building its `GenerateConfig` (which already sets `stop_token_ids:
vec![]`), which produces byte-identical EOS-disabled decoding through the
unmodified `should_stop_token` — this is the exact idiom the crate's own
test suite already used in dozens of places (`cfg.eos_token_id = u32::MAX`)
to push EOS out of the reachable vocab range for a fixture; nothing novel
was invented for this restructure. Behavior is unchanged: `--emit-phase-events`
runs still get EOS-disabled generation exactly as before; every other
(default) path is untouched.

The two round-1 `disable_eos` mutation-sensitive tests are rewritten
against the new mechanism —
`eos_token_id_override_forces_continuation_past_matching_stop_token`
(two-phase: first proves `config_mut()` reaches the model's real config by
pointing `eos_token_id` at the fixture's always-sampled token and
confirming an immediate `StopReason::Eos`, then re-points it at
`u32::MAX` and confirms the *same* `GenerateConfig` now runs to
`max_new_tokens`) and
`eos_token_id_override_suppresses_match_in_should_stop_token` (drives
`should_stop_token` directly). Round 2's `cfg(test)` PrefillEnd/
first-sample seam and its two tests are untouched.

**Verified**: `cargo semver-checks check-release -p lattice-inference` no
longer lists `GenerateConfig.disable_eos` anywhere in its failure output —
confirmed by diffing the exact failure list before/after via `git stash`
on this same branch (before: `field GenerateConfig.disable_eos` present;
after: absent). Two unrelated, pre-existing semver-major violations
remain on this branch (`GdnSaved`'s 17 fields + `ChatCompletionOutput.stopped`
under `constructible_struct_adds_field`; `gdn_backward`/`gdn_forward_save`
under `function_parameter_count_changed`) — confirmed via `git merge-base
--is-ancestor` that both originating commits (`fac5fa677`, `27b9047ee`)
are already on `origin/main`, and `gdn_backward.rs` has zero diff between
`origin/main` and this branch, so these predate and are unrelated to this
PR; out of scope here.

Mutation-verified both rewritten tests: broke `config_mut()` to leak a
detached clone instead of `&mut self.config` (`cp`-backup/restore, never
`git checkout` over uncommitted work), confirmed both new tests fail
under that mutation, restored and reconfirmed both pass.

Gates run this round: `cargo fmt --check`; `cargo clippy --workspace
--features f16,metal-gpu --all-targets -- -D warnings` (clean); `cargo
test -p lattice-inference --lib --features f16 generation::` — 28 passed
(same count as round 2, now under the new mechanism); `cargo check
--workspace --features f16,metal-gpu --all-targets` (clean);
`scripts/build-app-bins.sh` (all app-shipped binaries build);
`python3 -m pytest tests/` — 306 passed, untouched by this change (the
`@@bench` summary line's JSON `"disable_eos":true` is descriptive
metadata about the trial's behavior, unrelated to the removed Rust struct
field, left as-is).

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,0 +1,35 @@
+# cargo-audit configuration — supply-chain vulnerability gate for the lattice
+# workspace. Enforced in CI by `.github/workflows/cargo-audit.yml`, which
+# runs `cargo audit` and fails the job on any *vulnerability*-severity
+# advisory. This file exists solely to document and ignore advisories that
+# are warnings only (unmaintained / unsound-with-no-semver-compatible-fix)
+# and have no actionable remediation today. Never add a vulnerability-severity
+# advisory here — those must be fixed by an upgrade, not silenced.
+#
+# Companion note: `deny.toml` already runs `cargo deny check advisories` as
+# an INFORMATIONAL step inside the existing `cargo-deny` CI job (continue-on-
+# error, so it cannot block merges). This file is for the separate, blocking
+# `cargo-audit` gate — the two run different tools against the same
+# advisory-db and are deliberately independent (cargo-deny stays soft/broad,
+# cargo-audit is the hard vulnerability gate).
+
+[advisories]
+ignore = [
+    # paste 1.0.15 — unmaintained (author archived the crate, no successor
+    # published). Transitive only: pulled in via macro_rules_attribute
+    # (-> tokenizers -> lattice-inference) and via metal/wgpu-hal (-> wgpu ->
+    # lattice-fann/lattice-tune/lattice-inference). We have no direct
+    # dependency to bump or drop; removing it means waiting on upstream
+    # tokenizers/wgpu-hal to migrate off it. No known vulnerability, purely
+    # an unmaintained-status warning.
+    "RUSTSEC-2024-0436",
+
+    # lru 0.12.5 — unsound `IterMut` (Stacked-Borrows violation). Direct
+    # dependency of crates/embed (Cargo.toml pins "0.12"). The fix landed at
+    # >=0.16.3, a semver-incompatible major bump (0.12 -> 0.16) that would
+    # require reviewing lattice-embed's lru usage for API breakage before
+    # taking it; deferred as a follow-up crate bump, not a drop-in `cargo
+    # update`. Tracked at #886. Verified 2026-07-11: `cargo update -p lru`
+    # resolves no newer 0.12.x release (0.12.5 is latest in-range).
+    "RUSTSEC-2026-0002",
+]
--- a/.github/workflows/cargo-audit.yml
+++ b/.github/workflows/cargo-audit.yml
@@ -1,0 +1,187 @@
+name: cargo audit
+#
+# Dedicated vulnerability gate. `ci.yml`'s `cargo-deny` job already runs
+# `cargo deny check advisories`, but that step is deliberately INFORMATIONAL
+# (continue-on-error) so an unrelated upstream disclosure never wedges a
+# merge — see deny.toml's header comment. This workflow is the hard gate:
+# `cargo audit` FAILS the job on any vulnerability-severity RUSTSEC advisory.
+# Warnings (unmaintained / unsound-with-no-fix) are informational here too —
+# see .cargo/audit.toml for the documented, individually-justified ignore
+# list, and never add a vulnerability-severity advisory to that list.
+#
+# NOTE: no `paths` filter on the `pull_request` trigger below, matching the
+# reasoning in `ci.yml`'s header comment. `cargo-audit-gate` at the bottom of
+# this file is meant to become a *required* status check, and a required
+# check whose workflow never even runs leaves a PR stuck on "Expected,
+# waiting for status to be reported" forever. So this workflow always
+# triggers on every PR to main; dependency-irrelevant PRs are instead
+# handled by gating `audit` and `audit-native` on the `changes` job below,
+# exactly like `ci.yml` gates `ci`/`feature-matrix`/`bench-compile`/`msrv` on
+# its own `changes` job. `cargo-audit-gate` always runs and always reports,
+# so it is the safe context to require — mirroring `ci-gate` in ci.yml,
+# `parity-gate` in e2e-parity.yml, and `app-binaries-gate` in
+# app-binaries.yml. Flipping `cargo-audit-gate` into the branch ruleset's
+# required-checks list is a separate maintainer step (not done by this PR —
+# see the PR description).
+#
+# Two audit jobs cover the two independent Cargo dependency graphs this repo
+# ships: `audit` for the root workspace's tracked `Cargo.lock`, and
+# `audit-native` for `npm/lattice-embed-native`, which declares its own
+# single-crate `[workspace]` (see that manifest's header comment) and is
+# gitignored at the repo root's `Cargo.lock` pattern, so it has no tracked
+# lockfile to scan directly. `audit-native` generates one at CI time with
+# `cargo generate-lockfile` and audits that. cargo-audit's config discovery
+# walks up from the current directory the same way cargo's own config
+# discovery does, so running `cargo audit` from inside
+# npm/lattice-embed-native still picks up the root `.cargo/audit.toml`
+# ignore list — verified locally 2026-07-11 (only the documented lru
+# warning surfaces, exit 0).
+#
+# Also triggered weekly on a schedule (fresh advisory-db entries land with
+# no code change of ours) and on manual dispatch; both always run the full
+# audit regardless of `changes`, since there is no PR diff to gate on.
+
+on:
+  pull_request:
+    branches: [main]
+  schedule:
+    - cron: '0 6 * * 1'  # weekly Monday 06:00 UTC
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  # Decide whether anything that can change either Cargo dependency graph
+  # was touched. A PR that touches neither manifest set cannot introduce or
+  # resolve an advisory, so it does not need either audit job below.
+  changes:
+    name: detect-dependency-changes
+    runs-on: ubuntu-latest
+    outputs:
+      deps: ${{ steps.filter.outputs.deps }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - id: filter
+        run: |
+          set -euo pipefail
+          if [ "${{ github.event_name }}" = "pull_request" ]; then
+            git fetch origin "${{ github.base_ref }}" --depth=1 2>/dev/null || true
+            CHANGED=$(git diff --name-only "origin/${{ github.base_ref }}...HEAD" \
+              || git diff --name-only "HEAD~1...HEAD")
+            echo "Changed files:"
+            echo "$CHANGED"
+            # Two single greps against a here-string, not a `grep | grep -q`
+            # pipe: a downstream `-q` can close its end of a pipe early and
+            # SIGPIPE the upstream producer, which under `pipefail` reports
+            # as the pipeline's exit code instead of the match result
+            # (fail-open on a large diff) — see the matching comment in
+            # ci.yml's `changes` job.
+            MATCH=$(grep -E '(^Cargo\.lock$|^Cargo\.toml$|^crates/.*/Cargo\.toml$|^npm/lattice-embed-native/Cargo\.toml$|^\.cargo/audit\.toml$|^\.github/workflows/cargo-audit\.yml$)' <<<"$CHANGED" || true)
+            if grep -q . <<<"$MATCH"; then
+              echo "deps=true" >> "$GITHUB_OUTPUT"
+              echo "→ dependency-relevant change: audit REQUIRED"
+            else
+              echo "deps=false" >> "$GITHUB_OUTPUT"
+              echo "→ no dependency-relevant change: audit work skipped"
+            fi
+          else
+            # schedule / workflow_dispatch: no PR diff to gate on, always audit.
+            echo "deps=true" >> "$GITHUB_OUTPUT"
+            echo "→ non-PR trigger (${{ github.event_name }}): audit REQUIRED"
+          fi
+
+  audit:
+    name: cargo audit (root workspace)
+    needs: changes
+    if: needs.changes.outputs.deps == 'true'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@1.94.1
+      - name: Pin the real toolchain bin on PATH
+        run: dirname "$(rustup which cargo)" >> "$GITHUB_PATH"
+      - name: Cache cargo-audit binary
+        uses: actions/cache@v4
+        with:
+          path: ~/.cargo/bin/cargo-audit
+          key: cargo-audit-bin-${{ runner.os }}-v0.22.2
+      - name: Install cargo-audit
+        run: |
+          if [ ! -x ~/.cargo/bin/cargo-audit ]; then
+            cargo install cargo-audit --version 0.22.2 --locked
+          fi
+      - name: cargo audit (fails on vulnerabilities, warnings informational per .cargo/audit.toml)
+        run: cargo audit
+
+  audit-native:
+    name: cargo audit (lattice-embed-native)
+    needs: changes
+    if: needs.changes.outputs.deps == 'true'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@1.94.1
+      - name: Pin the real toolchain bin on PATH
+        run: dirname "$(rustup which cargo)" >> "$GITHUB_PATH"
+      - name: Cache cargo-audit binary
+        uses: actions/cache@v4
+        with:
+          path: ~/.cargo/bin/cargo-audit
+          key: cargo-audit-bin-${{ runner.os }}-v0.22.2
+      - name: Install cargo-audit
+        run: |
+          if [ ! -x ~/.cargo/bin/cargo-audit ]; then
+            cargo install cargo-audit --version 0.22.2 --locked
+          fi
+      - name: Generate lockfile for the native N-API package
+        # This package is its own single-crate workspace (see
+        # npm/lattice-embed-native/Cargo.toml's header comment) and its
+        # Cargo.lock is gitignored, so there is nothing tracked to scan.
+        # Resolve one at CI time instead.
+        working-directory: npm/lattice-embed-native
+        run: cargo generate-lockfile
+      - name: cargo audit (fails on vulnerabilities, warnings informational per .cargo/audit.toml)
+        working-directory: npm/lattice-embed-native
+        run: cargo audit
+
+  # Always runs, always reports — the safe required-check context. `audit`
+  # and `audit-native` are gated on `changes`, so neither is individually
+  # safe to require: a skipped job still reports success, but only this job
+  # fails closed if `changes` itself blew up. Mirrors `ci-gate` in ci.yml.
+  cargo-audit-gate:
+    name: cargo-audit-gate
+    needs: [changes, audit, audit-native]
+    if: always()
+    runs-on: ubuntu-latest
+    steps:
+      - name: Resolve gate
+        run: |
+          CHANGES_RESULT="${{ needs.changes.result }}"
+          DEPS="${{ needs.changes.outputs.deps }}"
+          AUDIT_RESULT="${{ needs.audit.result }}"
+          AUDIT_NATIVE_RESULT="${{ needs.audit-native.result }}"
+          echo "changes=$CHANGES_RESULT deps=$DEPS audit=$AUDIT_RESULT audit-native=$AUDIT_NATIVE_RESULT"
+          if [ "$CHANGES_RESULT" != "success" ]; then
+            echo "::error::change-detection did not succeed, failing closed."
+            exit 1
+          fi
+          if [ "$DEPS" != "true" ]; then
+            echo "No dependency-relevant change, audit jobs not required. PASS."
+            exit 0
+          fi
+          if [ "$AUDIT_RESULT" != "success" ]; then
+            echo "::error::audit (root workspace) did not succeed (result=$AUDIT_RESULT)."
+            exit 1
+          fi
+          if [ "$AUDIT_NATIVE_RESULT" != "success" ]; then
+            echo "::error::audit-native (lattice-embed-native) did not succeed (result=$AUDIT_NATIVE_RESULT)."
+            exit 1
+          fi
+          echo "Dependency-relevant change; both audit jobs PASSED."
+          exit 0
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ members = [
 ]
 
 [workspace.package]
-version = "0.5.1"
+version = "0.6.0"
 edition = "2024"
 # MSRV, verified end-to-end (workspace + deps) on 1.93.1 and guarded by the msrv CI
 # job (aarch64 macOS, f16,metal-gpu, --all-targets) so a newer-toolchain-only API
--- a/crates/embed/Cargo.toml
+++ b/crates/embed/Cargo.toml
@@ -55,14 +55,14 @@ tracing = "0.1"
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 tokio = { workspace = true, features = ["sync", "rt"] }
 # Embedding generation (native, pure Rust)
-lattice-inference = { version = "0.5.1", path = "../inference", optional = true, features = ["f16"] }
+lattice-inference = { version = "0.6.0", path = "../inference", optional = true, features = ["f16"] }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 tokio = { version = "1.0", default-features = false, features = ["sync", "rt"] }
 # Same crate, trimmed to the CPU compute path: `std` + `f16`, without
 # `download` (ureq) or `serve` (axum/tokio-net/mio), none of which compile to
 # wasm32-unknown-unknown.
-lattice-inference = { version = "0.5.1", path = "../inference", optional = true, default-features = false, features = [
+lattice-inference = { version = "0.6.0", path = "../inference", optional = true, default-features = false, features = [
     "std",
     "f16",
 ] }
--- a/crates/inference/Cargo.toml
+++ b/crates/inference/Cargo.toml
@@ -7,6 +7,10 @@ authors.workspace = true
 license.workspace = true
 repository.workspace = true
 description = "Pure Rust transformer inference engine — safetensors loading, SIMD matmul, BGE/Qwen3 embeddings"
+# Heavy test-only fixtures (~40MB raw) push the packaged crate past the
+# crates.io 10MB upload limit. They are consumed only by integration tests,
+# which are never run from the published crate.
+exclude = ["tests/fixtures/tokenizers", "tests/fixtures/gdn_chunk"]
 
 [features]
 default = ["std", "download", "serve"]
@@ -67,7 +71,7 @@ bytemuck = { version = "1", features = ["derive"], optional = true }
 ort = { version = "2.0.0-rc.12", optional = true }
 tokenizers = { version = "0.22", optional = true, default-features = false, features = ["progressbar", "onig"] }
 image = { version = "0.25", default-features = false, features = ["png", "jpeg"] }
-lattice-fann = { version = "0.5.1", path = "../fann", optional = true }
+lattice-fann = { version = "0.6.0", path = "../fann", optional = true }
 
 [[bin]]
 name = "lattice"
--- a/crates/inference/Cargo.toml
+++ b/crates/inference/Cargo.toml
@@ -328,6 +328,7 @@ harness = false
 [[bench]]
 name = "f16_convert_bench"
 harness = false
+required-features = ["bench-internals"]
 
 [lints]
 workspace = true
--- a/crates/inference/benches/attention_bench.rs
+++ b/crates/inference/benches/attention_bench.rs
@@ -425,7 +425,7 @@ fn run_benchmark() {
 }
 
 #[test]
-#[ignore]
+#[ignore = "manual benchmark, not a correctness gate"]
 fn attention_benchmark() {
     run_benchmark();
 }
--- a/crates/inference/benches/cross_turn_prefix_cache_bench.rs
+++ b/crates/inference/benches/cross_turn_prefix_cache_bench.rs
@@ -81,6 +81,7 @@ fn bench_gen_cfg(max_new_tokens: usize) -> GenerateConfig {
         stop_strings: vec![],
         reasoning_budget: None,
         logprobs: None,
+        disable_eos: false,
     }
 }
 
--- a/crates/inference/benches/cross_turn_prefix_cache_bench.rs
+++ b/crates/inference/benches/cross_turn_prefix_cache_bench.rs
@@ -81,7 +81,6 @@ fn bench_gen_cfg(max_new_tokens: usize) -> GenerateConfig {
         stop_strings: vec![],
         reasoning_budget: None,
         logprobs: None,
-        disable_eos: false,
     }
 }
 
--- a/crates/inference/benches/f16_convert_bench.rs
+++ b/crates/inference/benches/f16_convert_bench.rs
@@ -2,55 +2,34 @@
 //! lookup hot path (`forward_step_inner`, `forward_step_gdn_only`, prefill).
 //!
 //! Measures at N=1024 (the Qwen3.5-0.8B hidden size):
-//! - `scalar`: hand-written IEEE-754 bit-manipulation loop (~10 ops/element)
+//! - `scalar`: the crate's production scalar decoder
+//!   (`lattice_inference::weights::bench_support::f16_bits_to_f32`, which
+//!   `weights/mod.rs` re-exports behind `bench-internals` from the single
+//!   always-compiled `weights::half_bits` module every load/quant call site
+//!   delegates to — lattice#799). This bench previously carried its own
+//!   copied decoder rather than exercising that migrated path, so its
+//!   numbers said nothing about the consolidation. Requires `--features bench-internals`.
 //!
 //! A NEON fast path (`vcvt_f32_f16`, 4 lanes/instruction) was removed in #568: it
 //! required the nightly-only `stdarch_neon_f16` feature and broke stable-toolchain
 //! builds. `convert_f16_row` in `metal_qwen35.rs` is scalar-only for the same reason.
 //!
-//! Run: `cargo bench -p lattice-inference --bench f16_convert_bench`
+//! Run: `cargo bench -p lattice-inference --bench f16_convert_bench --features bench-internals`
 //!
 //! ADR-058: every perf PR must include before/after bench output.
 
 use criterion::{BenchmarkId, Criterion, Throughput, black_box, criterion_group, criterion_main};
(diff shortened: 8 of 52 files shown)


ohdearquant:

## Review of PR #882 (round 1)

### Verdict: REJECT

Two trace-integrity blockers make the reported prefill and decode metrics untrustworthy. The schema-v2 validator and the 157-test suite are green, but the underpowered demonstration path bypasses the validator's phase-sequence check and the supervisor does not enforce the missing invariants itself.

### Prior review context

- No prior reviews or review threads were returned by GitHub at review start.
- The stated independent content read was positive. This round focused on adversarial runtime measurement integrity, child supervision, and the SHA-bounded stacked diff only.

### What this round examined

- The PR body, #877 schema-v2 contract in `scripts/bench_decode_harness.py`, and `scripts/perf-policy.toml`.
- Only `7ca28a2c26ad42a660beb1c327c2268a132b0c3d...72edfef4f0e336d592ed2be9f04a4cdcb24fd50e`.
- The adapter, supervisor, tests, adjacent CPU streaming generation path, `GenerateConfig` propagation, and the tiny touched call sites in Metal/serve.
- The Apple `ri_phys_footprint` API. Apple documents it as a `uint64_t` field and documents `proc_pid_rusage` as the intended process-footprint source ([Apple API](https://developer.apple.com/documentation/kernel/rusage_info_v1/1577496-ri_phys_footprint), [WWDC22 guidance](https://developer.apple.com/videos/play/wwdc2022/10106/)).

### Findings (3 total: 2 blockers, 1 major)

#### Blockers

1. [crates/inference/src/bin/qwen35_generate.rs:343] `prefill_end` and `token_available` are emitted after sampling, and on text deltas rather than raw tokens.
   - **Current**: The phase callback emits `prefill_end` at line 350 and increments the “token” index at lines 353–355. But the CPU path finishes prefill at [generation.rs:505](crates/inference/src/model/qwen35/generation.rs:505), samples the first token at [generation.rs:577](crates/inference/src/model/qwen35/generation.rs:577), pushes it at [generation.rs:611](crates/inference/src/model/qwen35/generation.rs:611), then only invokes this callback through the incremental UTF-8 detokenizer at [generation.rs:636](crates/inference/src/model/qwen35/generation.rs:636) and [generation.rs:648](crates/inference/src/model/qwen35/generation.rs:648).
   - **Issue**: The required prefill boundary is before the first sample; this marker is necessarily after it. More seriously, incremental detokenization deliberately buffers incomplete UTF-8, so one callback is not guaranteed to equal one sampled token. A multi-token codepoint can delay the first event and collapse several raw token intervals into one apparent event. That silently inflates/changes prefill throughput and corrupts TTFT, decode rate, and per-token percentiles.
   - **Fix**: Add a default-off generation observer that fires exactly (a) after prefill logits are ready and before `sample_token`, and (b) after each raw sampled token. Have only `--emit-phase-events` install it. Keep the user-facing text callback separate, and fail the adapter if the raw event count differs from `GenerateOutput.generated_tokens`.
   - **Test**: A mutation-sensitive tiny-model test must prove `prefill_end < first sample`, and an incomplete multi-byte UTF-8 sequence must still produce one monotonically indexed phase event per generated token.
   - **Confidence**: derived from the current CPU generation control flow cited above.

2. [scripts/bench_cpu_flagship_supervisor.py:349] The supervisor accepts malformed phase traces, reorders token events, and therefore lets the n=2 demonstration validate without proving its measurement boundary.
   - **Current**: It appends phase dictionaries without structural/order validation (lines 349–360), then explicitly sorts `token_available` records by `token_index` (lines 449–452). It sets `generated_tokens` from the number of observed events instead of `summary["generated_tokens"]` (lines 456–457), merely returns `delta_matches_generated_tokens` as metadata (line 518), and strips every phase event for an `unsupported` record (lines 904–909).
   - **Issue**: Sorting masks an out-of-order stream instead of rejecting it; no path checks monotonically non-decreasing child timestamps, single-shot phase uniqueness/order, contiguous token indices, or event-count equality. I observed an in-memory trace with its four token events reversed get accepted and report `decode_tok_s=250.0`; a summary claiming 9 generated tokens with only 4 events was also accepted as 4. Because the actual `n_pairs=2` evidence is deliberately `unsupported`, `validate_run_record` skips `_validate_phase_sequence` and sees the intentionally empty tuple. Thus the claimed successful validation does not validate the real demonstration trace at all.
   - **Fix**: Validate every trial before aggregation, preserving stdout arrival order: parse each event through the schema parser, require exactly one ordered load/backend/prefill sequence, non-decreasing timestamps, strictly contiguous raw token indices, and equality with the summary’s generated-token count. Do not sort to repair input. Preserve/validate the representative trace even when the verdict is `unsupported`, or make a separate always-on trace validator mandatory before the downgrade.
   - **Test**: Add rejection tests for reversed events, repeated/missing phases, decreasing timestamps, duplicate/gapped token indices, `delta_matches_generated_tokens=false`, and a summary/event-count mismatch. Add an underpowered-record test demonstrating that trace validation still occurs before the schema verdict exemption.
   - **Confidence**: observed with the disposable in-memory probes described above and derived from the cited lines.

#### Majors

1. [scripts/bench_cpu_flagship_supervisor.py:339] `timeout_s` cannot time out a child that hangs while its stdout pipe remains open.
   - **Current**: The blocking `for raw_line in proc.stdout` loop runs before `proc.wait(timeout=timeout_s)` at line 363. A child that stalls before closing stdout never reaches `wait`; if `wait` does time out, the `finally` block only stops the monitor/closes stdout and neither terminates nor reaps the child.
   - **Issue**: A stuck model run can hang a shadow/CI worker indefinitely and leak the child rather than produce the intended fail-closed `TrialFailure`.
   - **Fix**: Put pipe draining behind a deadline-aware reader (or selector/thread queue), terminate/kill and reap on timeout, and drain stderr concurrently so a full stderr pipe cannot deadlock the child.
   - **Test**: A silent sleeping fake child must fail within a small timeout and be confirmed reaped; another fixture should fill stderr while keeping stdout open.
   - **Confidence**: derived directly from the blocking order.

- `RusageSampler` is given `proc.pid` at [scripts/bench_cpu_flagship_supervisor.py:332](scripts/bench_cpu_flagship_supervisor.py:332), so it targets the spawned adapter, not the supervisor. It handles an exit-race sampling failure as `None`; peak aggregation is correctly `max` over successful samples at lines 482–484. The code states the 10 ms / possible sub-interval undercount caveat at lines 426–439 and takes an extra sample at each observed phase marker.
- The paired driver uses the same binary, model path, context, token count, warmup, and sampler for both arms; only `seed` differs. Its order is correctly `AB, BA, AB, BA, …` and records `(n_ab, n_ba)` at lines 580–630. A deterministic in-memory run of four pairs produced `['AB', 'BA', 'AB', 'BA']`, balance `(2, 2)`.
- `build_decode_cell_aggregate` resolves the decode metric policy at [scripts/bench_cpu_flagship_supervisor.py:683](scripts/bench_cpu_flagship_supervisor.py:683), rather than hard-coding class A.
- The claimed Python command was observed clean: `python3 -m pytest tests/test_bench_cpu_flagship_supervisor.py tests/test_bench_run_record.py tests/test_bench_gate_math.py -q` → `157 passed in 0.90s`.
- The intended unsupported downgrade works in the supplied fixture. A disposable `CellRecord` copy with `required_n` changed from 7 to 8 was rejected by `validate_run_record` with the expected policy-derived required-n mismatch.
- `cargo fmt --all -- --check`, `cargo test -p lattice-inference --lib --features f16 generation::` (25 passed, 4 model-dependent ignored), and `cargo clippy -p lattice-inference --features f16 --lib --bins -- -D warnings` all passed. Cargo used the shared target directory `/Users/lion/.cache/shared-cargo-target/lattice`.
- `GenerateConfig.disable_eos` defaults to `false`, and the sole stop-predicate change is the false fast-path at [generation.rs:1919](crates/inference/src/model/qwen35/generation.rs:1919); the ordinary CLI still dispatches unchanged unless `--emit-phase-events` is present. Its two new targeted tests pass.
- `git diff --check` is clean. Added `expect`/`unwrap`-style uses are confined to the demo binary or Rust tests, not library production code.
- PR metadata has the required first-line AI attribution, checklist, and `Depends on #877`; each non-merge PR commit has `Co-Authored-By: Leo <noreply@khive.ai>`. `gh pr checks 882` reports no checks on this stacked branch. The PR body correctly labels the contended source-swap comparison as inconclusive and treats the default-off structural argument as the primary performance rationale; it does not present the numbers as a measured win/loss.

### What I did NOT check

- Per instruction, I did not run the real-model smoke or any benchmark. The three real demonstration JSON reports are not PR files, so I could not independently replay their exact records; the tests and disposable sabotage only validate the supplied harness logic.
- I did not exercise `proc_pid_rusage` against a live macOS child or measure the actual 10 ms sampler resolution.
- I did not run full workspace CI or `make bench-compare`; the branch has no reported checks and remains stacked on #877.

Fix the two phase-trace blockers before another review. The follow-up should run the new malformed-trace/UTF-8/timeout tests and attach or otherwise make available the three raw demonstration reports so their phase traces and validator outcomes can be replayed without a checkpoint.

Domain utility: SKIPPED — no Lore connector was available; the Lattice-specific review rubric, repository contract, and a targeted reviewer-memory recall guided this pass.

Verdict: REJECT

ohdearquant:

## Review of PR #882 (round 2)

### Prior review context

- Round 1 rejected the change for two blockers (raw generation-event boundary and fail-open/under-validated trial traces) and one major (subprocess timeout/drain safety).
- This round reconfirmed the fixes against `origin/main` at `72d9c9c87`; it found no remaining blocker or major.

### What this round examined

- The fix/evidence commits `9d0299fe3` and `72d9c9c87`, plus their touched Rust/Python paths and tests.
- The two committed reports, including all eight raw trial traces.
- The two round-1 in-memory sabotage cases and the bounded child timeout/drain implementation.

#### Mediums

1. [crates/inference/src/model/qwen35/generation.rs:3234] The prefill-boundary test does not actually guard the required “before `sample_token`” position.
   - **Current**: `raw_observer_prefill_end_precedes_first_raw_token_monotonic_index` asserts only that `PrefillEnd` is the first observer event and then checks the raw-token indices. The production control flow is presently correct: `PrefillEnd` is emitted at [generation.rs:625](crates/inference/src/model/qwen35/generation.rs:625), immediately before `sample_token` at [generation.rs:627](crates/inference/src/model/qwen35/generation.rs:627).
   - **Evidence**: In a disposable worktree, I moved `PrefillEnd` to immediately after `generated_ids.push(next_id)` and before the existing `RawToken` callback. The targeted test still passed because it cannot see sampling; it only sees `PrefillEnd` before `RawToken`. Moving it after `RawToken` did fail, so the test covers event order, but not the required sample boundary.
   - **Risk**: A future refactor can place `PrefillEnd` after sampling but before the raw-token callback, silently adding sampling time to the reported prefill interval while the advertised mutation-sensitive test remains green.
   - **Fix**: Add a narrowly scoped test seam/private helper that records first-sample entry, then assert `PrefillEnd < first_sample`; retain the existing event-order assertion separately.
   - **Confidence**: observed with the disposable mutation runs. This is test coverage, not a current measurement defect.

- **Blocker 1 fixed in current control flow.** `generate_streaming_with_cancel` keeps its four-argument public shape and calls the observer implementation with a no-op closure at [generation.rs:414](crates/inference/src/model/qwen35/generation.rs:414)-[generation.rs:425](crates/inference/src/model/qwen35/generation.rs:425). Derived external behavior is unchanged when no observer is installed; I did not byte-diff optimized machine code. `PrefillEnd` is before the first `sample_token`, and raw events fire after the respective `generated_ids.push` calls at [generation.rs:661](crates/inference/src/model/qwen35/generation.rs:661), [generation.rs:806](crates/inference/src/model/qwen35/generation.rs:806), and [generation.rs:991](crates/inference/src/model/qwen35/generation.rs:991), independently of text-delta delivery. The adapter fails closed if the raw-event count differs from `GenerateOutput.generated_tokens` at [qwen35_generate.rs:382](crates/inference/src/bin/qwen35_generate.rs:382).
- The buffered-UTF-8 test genuinely discriminates: a disposable mutation that removed the pre-push first raw event and emitted it through the text-delta callback failed `raw_observer_fires_before_on_token_for_buffered_incomplete_utf8_first_token` with observed indices `[2, 3, 4]`, rather than `[1, 2, 3, 4]`. The original two raw-observer tests both pass after restoring the disposable worktree.
- **Blocker 2 fixed.** `_validate_trial_trace` parses arrival-order events, reuses `_validate_phase_sequence`, requires contiguous `1..=N` token indices, and cross-checks the child summary at [bench_cpu_flagship_supervisor.py:302](scripts/bench_cpu_flagship_supervisor.py:302)-[bench_cpu_flagship_supervisor.py:384](scripts/bench_cpu_flagship_supervisor.py:384). It is called unconditionally in `run_one_trial` at [bench_cpu_flagship_supervisor.py:560](scripts/bench_cpu_flagship_supervisor.py:560), before aggregation/downgrade. The only remaining `sorted()` calls are percentile calculation and weight-file discovery, not phase events ([bench_cpu_flagship_supervisor.py:580](scripts/bench_cpu_flagship_supervisor.py:580), [bench_cpu_flagship_supervisor.py:967](scripts/bench_cpu_flagship_supervisor.py:967)).
- I reran the two round-1 probes. Reversed token events rejected with `token_index did not strictly increase`; four events with `generated_tokens=9` rejected with the explicit count mismatch.
- **Timeout/reap major fixed.** Stdout and stderr readers start before `proc.wait(timeout=...)` ([bench_cpu_flagship_supervisor.py:508](scripts/bench_cpu_flagship_supervisor.py:508)-[bench_cpu_flagship_supervisor.py:529](scripts/bench_cpu_flagship_supervisor.py:529)); timeout kills then waits/reaps; reader joins are bounded and both pipe handles close afterward ([bench_cpu_flagship_supervisor.py:533](scripts/bench_cpu_flagship_supervisor.py:533)-[bench_cpu_flagship_supervisor.py:542](scripts/bench_cpu_flagship_supervisor.py:542)). The new silent-child/reap and stderr-backpressure tests are substantive and passed. This matches the relevant Python subprocess constraints: `wait()` requires concurrent pipe draining to avoid pipe-buffer deadlock, and a timeout cleanup needs kill plus completion handling ([Python subprocess documentation](https://docs.python.org/3/library/subprocess.html)).
- **Evidence replayed.** I converted each report’s serialized `monotonic_ns` events back to the in-memory validator shape and ran `_validate_trial_trace` on all four trials in each report: all eight passed. I also reconstructed the emitted `CellRecord` shape and ran `validate_run_record` with each report’s own repo and policy SHA: both `report_ctx512.json` and `report_ctx1024.json` passed. Both honestly report `run_kind: demonstration`, `verdict: unsupported`, `n_valid: 2`, and `required_n: 25`. The absence of ctx=2048 is explicitly disclosed as a stopped-on-contention run. The ctx=512 report’s `dirty: true`/parent SHA makes it weaker than clean-commit evidence, but the PR body discloses that fact rather than normalizing it away.
- **Gates observed.** `python3 -m pytest tests/test_bench_cpu_flagship_supervisor.py tests/test_bench_run_record.py tests/test_bench_gate_math.py -q` → `169 passed in 3.24s`; `cargo fmt --all -- --check` clean; `cargo clippy -p lattice-inference --features f16 --lib --bins -- -D warnings` clean; `cargo test -p lattice-inference --lib --features f16 generation::` → `27 passed, 4 ignored`.
- **Publication hygiene.** `git diff --check origin/main...HEAD` is clean. A credential-pattern scan of both added JSON reports found no credentials/secrets; their only machine-specific information is the intentionally recorded local model path/provenance. No temporary worktree remains.

### What I did NOT check

- I did not run the real checkpoint benchmark or a full workspace CI/bench suite.
- I did not byte-compare optimized `generate_streaming_with_cancel` code generation; the unchanged observable behavior is derived from the no-op observer wrapper and current control flow.

- Apply the focused test improvement above, then no further review round is useful unless the PR changes materially.

Domain utility: SKIPPED — no domain-composition service is available in this workspace; repository and PR-specific review guidance was applied.

Verdict: APPROVE-WITH-FIXES

oceanwaves630 (approved):
Lane C combined-delta read (Leo) on head 273950db2.

- **Seam test** (codex r2 medium): cfg(test)-only thread-local recording first-sample entry, mutation-verified against the exact r2 mutation (new test fails, event-order test stays green — reproduces the blind spot precisely). Landed as ruled.
- **Restructure** (semver ruling b): `disable_eos` field removed entirely, all 38 call sites reverted; replacement is additive `Qwen35Model::config_mut()` on a private-field struct plus the crate's own established `eos_token_id = u32::MAX` sentinel idiom through `should_stop_token`, the single shared stop predicate — byte-identical EOS-disabled behavior with zero new public surface on the constructible `GenerateConfig`. Replacement tests are mutation-sensitive in both directions (config_mut reaches the real config; original id stays unmatched by the sentinel).
- Remaining semver-checks failures verified main-inherited (GdnSaved/#792, ChatCompletionOutput.stopped/#786, both predate this PR via merge-base) — they clear once v0.6.0 publishes, so merge sequencing is: release first, then this PR.

Approved. Do not arm auto-merge until the 0.6.0 baseline is published and the semver check is green at the true merge state.

oceanwaves630 (approved):
Arming per maintainer approval on-PR: 0.6.0 published, semver-checks green at true merge state.

PR #878, opened 2026-07-11
fix(inference): quantize MoE expert tensors in quantize_q4 (#874)

ohdearquant:
> _PR description authored by Claude (Anthropic agent) on behalf of @ohdearquant._

## AI-assisted development disclosure

- [x] This PR was developed with AI assistance (Claude, an Anthropic agent).
- [x] A human reviewed the diff before requesting review.
- [x] Tests were run locally and pass (see Test plan).
- [x] No secrets, credentials, or internal-only content are included.

## Summary

Closes #874. `should_quantize()` in `crates/inference/src/bin/quantize_q4.rs` gated purely on a `.weight` name suffix. MoE routed-expert tensors (`mlp.experts.gate_up_proj` / `mlp.experts.down_proj`, fused per-layer 3-D arrays with no `.weight` suffix) failed that gate and were silently written at source precision. For a 256-expert checkpoint that is ~92% of parameters, so `quantize_q4` barely shrank the file (63.21 GB output from a 66.97 GB bf16 source — a 1.06x ratio — instead of the ~19-21 GB a correct 4-bit conversion should produce).

Fix: match `.experts.gate_up_proj` / `.experts.down_proj` explicitly ahead of the `.weight`-suffix gate. The sibling `shared_expert.*` and `shared_expert_gate` tensors already carry a `.weight` suffix and were already quantized correctly (unaffected by this change).

### Expert addressing (for the follow-up loader work, tracked separately)

No output-layout change was needed. `quantize_f32_to_q4` already handles arbitrary tensor rank: it flattens the tensor and chunks it into 32-element Q4 blocks; the innermost expert dimension (512 for `down_proj`, 2048 for `gate_up_proj`) divides evenly by 32, so every expert boundary lands exactly on a block boundary — no padding, no straddling. Each `.q4` file already records its full shape (`[num_experts, out_features, in_features]`) and `payload_offset` in its own header (`read_q4_header`), so a per-expert byte range is computable directly from that header without any repack:

```text
numel_per_expert = shape[1] * shape[2]
blocks_per_expert = numel_per_expert / 32   (exact — no remainder)
expert_i_byte_range = [payload_offset + i * blocks_per_expert * 20,
                        payload_offset + (i + 1) * blocks_per_expert * 20)
```

`quantize_index.json` also records shape/numel per source tensor for convenience.

## Verification

[...]

Issue #874, opened 2026-07-11
bug(quantize): should_quantize name-suffix gate skips MoE expert tensors — 35B-A3B Q4 output is 63GB instead of ~19GB


ohdearquant:
## Summary

`quantize_q4`'s `should_quantize()` (crates/inference/src/bin/quantize_q4.rs) gates on a `.weight` name suffix. MoE expert tensors in Qwen3.6 MoE checkpoints are named without that suffix (`model.language_model.layers.N.mlp.experts.gate_up_proj` / `.down_proj`), so every expert tensor — ~92% of parameters in a 35B-A3B checkpoint — silently stays at source precision.

## Empirical evidence (Qwen/Qwen3.6-35B-A3B, bf16 safetensors, 66.97 GB)

- `quantize_q4` output: **63.21 GB** (1.06x compression — only the non-expert minority quantized)
- Correct 4-bit reference for the same model: mlx-community/Qwen3.6-35B-A3B-4bit = **19.03 GiB** (HF API, byte-verified); GGUF Q4_0 = 19.41 GiB
- Expected output for a correct Q4 conversion: ~19-21 GB

Command: `quantize_q4 --model <bf16-dir> --out <q4-dir>` (defaults; single run; deterministic — the gate is a static name predicate, no sampling involved).

## Fix direction

`should_quantize()` must recognize expert weight tensors (suffix-less 3-D expert arrays) as quantization candidates, or gate on tensor role/shape rather than name suffix. Note the consumer side needs the matching loader work (separate issue) — landing this alone makes the artifact smaller but still unusable on the Metal Q4 fast path.

## Acceptance

- Quantizing Qwen3.6-35B-A3B lands within ~10% of the 19 GiB reference size
- Per-tensor dtype dump shows expert tensors as Q4 blocks
- Existing dense-model quantize outputs are byte-identical (no regression on suffix-carrying tensors)

PR #872, opened 2026-07-11
fix(bench): exclude quick-mode-noisy embed SIMD groups from bench-compare gate

ohdearquant:
> _PR description authored by Claude (Anthropic agent) on behalf of @ohdearquant._

## Summary

- `lattice-embed`'s `simd` bench target is entirely sub-microsecond micro-benches that are
  confirmed noise-dominated in `--quick` mode: two same-toolchain A/A runs on identical refs
  flipped FAIL/WARN sign across dozens of entries, rotating across most groups run to run
  (see #714 for the quantitative reproduction — a null A/B on an idle machine reported 56
  FAIL from zero code difference).
- `scripts/bench-compare.sh` now excludes exactly two issue-evidenced noisy groups —
  `simd_dot_product` and `simd_cosine_similarity` — from the quick-mode FAIL/WARN gate, via a
  fixed, reviewed allowlist (not a `--list` dump of the whole 23-group target). The allowlist
  and its intersection-with-the-real-listing logic live in
  `scripts/lib/bench-informational-groups.sh`, a small standalone helper `bench-compare.sh`
  pipes its Criterion `--list` output through; a stale or renamed entry fails closed (never
  gates) instead of silently no-op'ing, and the list is scoped to the embed `simd` target only
  (documented in-script against name collision with `lattice-inference` groups).
  `scripts/perf-bench-gate.py` gains an `--informational-groups-file` flag that measures and
  reports those groups but excludes them from the FAIL/WARN gate and exit code. `--full` mode
  passes no informational groups, so a real embed SIMD regression is still caught there.
- Adds selftest fixtures: one with abstract group names covering the general informational-
  groups exclusion logic, and one that runs the actual shell helper (as a subprocess, against a
  controlled Criterion-listing fixture) and feeds its emitted file into the Python classifier —
  asserting the shell step emits exactly the two approved names, and that an unapproved embed
  group, a similarly-prefixed-but-distinct embed group (`simd_dot_product_extra`, ruling out a
  prefix/substring exemption), and an unrelated inference group all still gate as FAIL. This
  closes the earlier gap where the fixture kept its own hardcoded copy of the allowlist instead
  of reading it from the shell code that actually supplies it — mutation-verified: adding a
  name only to the shell helper's array in a disposable copy now makes the selftest fail.

## Review fixes

| # | Severity | Finding | Fix |
|---|---|---|---|
[...]

PR #883, opened 2026-07-11
fix(inference): add MoE FFN branch to from_q4_dir Q4-dir loader (#876)

### LINKED ISSUE #876: bug(inference): from_q4_dir has no MoE FFN branch — Q4-dir fast path fails at layer 0 on MoE checkpoints
## Summary

`from_q4_dir` (crates/inference/src/forward/metal_qwen35.rs:13717-14604) — the loader behind every documented Q4-directory entry point (`chat_metal`, `doctor`, eval binaries) — has no MoE branch. Its per-layer FFN construction (~14101-14125) unconditionally builds `MetalFfnWeights::Dense` from `mlp.gate_proj.weight` / `mlp.up_proj.weight` / `mlp.down_proj.weight` names, which do not exist in a MoE checkpoint.

## Empirical evidence (Qwen/Qwen3.6-35B-A3B q4 dir)

`chat_metal --model <q4-dir>` fails at layer 0, 1.8s in, max RSS 424 MiB (process startup only):

```
failed to load Q4 file .../model_language_model_layers_0_mlp_gate_proj_weight.q4: No such file or directory
```

A correctly MoE-aware construction path exists (`MetalQwen35Engine::new()`, ~lines 3169-3253, wired to the ADR-053 MoE Metal kernels) but is reachable only via the CPU-safetensors-upload path — it loads experts through anonymous memory rather than the mmap'd zero-copy route, which defeats minimum-RAM goals for MoE models.

## Fix direction

Add a MoE FFN branch to `from_q4_dir` mirroring the engine constructor's expert handling, but loading experts from the Q4 directory via the same mmap route the Dense branch uses (per-expert or per-layer expert blocks addressable for future streaming/partial-residency). Depends on the quantizer emitting Q4 expert tensors at all (#874).

## Acceptance

- `chat_metal --model <moe-q4-dir>` loads and generates on a MoE checkpoint
- Expert weights load via mmap (no full-checkpoint anonymous copy); RSS during load reflects that
- Dense-checkpoint loading unchanged (existing tests green)

ohdearquant:

## AI-assisted development disclosure

- [x] This PR was developed with AI assistance (Claude, an Anthropic agent).
- [x] A human reviewed the diff before requesting review.
- [x] Tests were run locally and pass (see Test plan).
- [x] No secrets, credentials, or internal-only content are included.

## Summary

Closes #876. `from_q4_dir`'s per-layer FFN loop unconditionally built `MetalFfnWeights::Dense` from `mlp.{gate,up,down}_proj.weight`, which do not exist in a MoE checkpoint quantized by #874 (MoE tensors live under `mlp.experts.{gate_up,down}_proj` / `mlp.shared_expert*` / `mlp.gate.weight` instead) — a MoE Q4 dir crashed at layer 0 with a missing-file error.

Depends on #878 (this PR is based on `fix/874-quantize-moe-experts`, PR #878's head branch — the quantizer must emit Q4 expert tensors before this loader can read them back).

## Design

**Detection**: `cfg.is_moe()` (this model family has no per-layer MoE toggle — `load_weights` in `model/qwen35/loading.rs` applies MoE uniformly to every layer when true, so the Q4 loader mirrors that).

**Branch**: `Self::load_moe_ffn_q4`, selected in the per-layer FFN construction alongside the existing Dense branch. Mirrors `MetalQwen35Engine::new()`'s MoE arm (ADR-053): same `MoeMetalBuffers` layout, same f16 GPU representation for routed/shared expert weights (`moe_expert_gemv`/`gemv_decode` read `half`, not packed Q4 — no Q4-native MoE GEMV kernel exists yet, so this deliberately matches the existing convention rather than introducing a second on-GPU numeric format for MoE only) and f32 for the two CPU-read scalars (router gate, shared-expert sigmoid gate).

**Loading**: every MoE tensor is read via two helpers, `load_q4_mmap_dequant_f16`/`load_q4_mmap_dequant_f32`, which `mmap` the `.q4` file (same primitives `mmap_q4_weight` already uses for the Dense branch's `Q4WeightBuf` tensors — `read_q4_header` + `memmap2::MmapOptions`) and dequantize directly from the mapped payload into a fresh f16/f32 buffer, without ever materializing the raw Q4 payload as an owned `Vec` first (unlike `load_q4_file`'s `read_exact`, used elsewhere for small Q4→f16 conversions). This is **mmap-backed source dequantization into eager, GPU-resident f16/f32 buffers** — not the same no-copy route the Dense branch gets from `Q4WeightBuf` (which keeps the mmap alive and hands Metal a zero-copy view for the tensor's lifetime). Here the `mmap` is local to each helper call and drops on return; only the *raw Q4 read* avoids an intermediate owned copy, not the *final GPU-resident representation*. A genuinely persistent/per-expert mmap route (only touching an expert's bytes when it's actually dispatched) needs a Q4-capable MoE GEMV kernel or an explicit residency design — see the streaming/partial-residency follow-up below. *(round-1 review MEDIUM 1: the original wording overstated this as "the same mmap route as Dense".)*

**Shape validation** *(round-1 review BLOCKER)*: `load_q4_mmap_dequant_f16`/`_f32` previously validated a tensor's Q4 header only *self*-consistently (`validate_q4_header_payload_bounds`: `product(shape) == original_len`, payload physically present) but never against the shape `encode_moe_ffn` actually requires. A structurally valid but short or same-numel-transposed expert file (e.g. a `[1, H]` router file under a 4-expert config) would have silently loaded and then violated `encode_moe_ffn`'s unsafe `num_experts * ...`-sized slice invariant. Both helpers now take an `expected_shape` and reject before mmap/allocation on any mismatch: router `[E, H]`; shared scalar gate `[1, H]`; routed gate/up `[E, 2I, H]`; routed down `[E, H, I]`; shared gate/up `[SI, H]`; shared down `[H, SI]` — verified against `load_moe_ffn_weights` (`model/qwen35/loading.rs`), the authoritative safetensors-path shapes. Two new regression tests (a too-short router header, a same-numel transposed routed-expert header) assert `from_q4_dir` returns a shape error before `forward_step`; both were **mutation-verified** by temporarily reverting the shape check in a disposable copy of the file and confirming the tests fail, then restoring it.

**Cumulative headroom guard**: `new()`'s original MoE guard (and this function's first version) checks a single layer's routed-expert byte count against `0.85 × recommendedMaxWorkingSetSize`. Since every layer's dequantized f16 buffers stay GPU-resident for the model's lifetime (nothing is freed between layers), a per-layer-only check passes cleanly on checkpoints whose *per-layer* footprint is small while the *sum* across all layers still exceeds available unified memory. Added a `Cell<u64>` running total threaded through the per-layer loop so the check is cumulative — verified live (see below) that this converts what would otherwise be an OOM into a clean, actionable error.

Dense-checkpoint loading is byte-for-byte unchanged — separate `if cfg.is_moe() { .. } else { .. }` branch, the Dense arm's code is untouched.

## Sibling-path check

Grepped every binary that calls `from_q4_dir`/`MetalQwen35State::from_q4_dir` (`chat_metal`, `lattice_serve`, `lattice` [doctor/serve subcommands], `eval_perplexity`, `dump_quarot_q4_golden`, `bench_decode_ab`, `bench_decode_slopefit`, `bench_logit_dump`, `bench_lora_mixture`) — none duplicate the per-layer FFN construction; they all call the single `from_q4_dir` function this PR patches, so they inherit the fix automatically. No `MetalFfnWeights::Dense { .. }` construction site exists outside `from_q4_dir` and `MetalQwen35Engine::new()`.

One related but explicitly out-of-scope gap: MTP layers always use a separate `MetalMtpDenseMlpWeights` struct (ADR-051 Phase 1 keeps MTP dense-only for the QuaRot counter-rotation path) — `quantize_q4`'s `should_quantize` does route `mtp.layers.N.mlp.experts.*` tensors through the MoE-quantization branch by name pattern, but the MTP *loader* has no MoE arm at all. This is a pre-existing, separately-scoped design decision (not a duplicate of #876's bug — MTP never attempted a MoE load, dense-only was already correct there), not something this PR touches.

## Verification

**Synthetic fixture (fast, CI-safe)**: `from_q4_dir_loads_and_generates_on_synthetic_moe_checkpoint` — builds a full synthetic 1-layer full-attention MoE Q4 checkpoint entirely on disk (hidden=512, 4 experts, top_k=1, moe_intermediate_size=32, shared_expert_intermediate_size=16), loads it via `from_q4_dir`, and drives it through `forward_step` (decode/M=1 — ADR-053's MoE dispatch panics on batched M>1 prefill by design, so this exercises the same path `chat_metal`'s decode loop uses) for 5 positions, asserting finite logits at every step. Passes.

**Layout/offset proof** *(round-1 review MAJOR 1)*: the original fixture filled every Q4 tensor with the same `0.1` constant, so an expert swap, wrong gate/up boundary, or an equal-length transpose left the test numerically finite and effectively unchanged — it could not catch the layout-corruption class the shape-validation fix guards against. Rewrote it so every expert/projection gets a distinct block-constant value, then:
- directly reads the built state's `routed_gate_up`/`routed_down` Metal buffers at expert 0/1's known offsets and asserts they equal the constants that expert was written with, proving the expert-major layout independent of generation (`q4_const_roundtrip` computes the exact expected f16 round-trip for a constant-valued Q4 block);
- forces the router to deterministically select a single nonzero expert (expert 2 of 4, via a router weight design that exploits this fixture's all-non-negative weights: a softmax-weighted average of non-negative `V` rows is always non-negative regardless of attention pattern, and RMSNorm with a positive uniform weight maps a uniform non-negative input to a uniform non-negative output, so per-expert router logits preserve their row-constant ordering regardless of the hidden state's exact magnitude);
- adds `from_q4_dir_moe_decode_is_sensitive_to_expert_weight_corruption`: builds a second checkpoint identical except the selected expert's on-disk weights are corrupted to another expert's values (the literal "expert swap"/wrong-offset failure mode) and asserts decode logits change. (First attempt at this test found decode output identical between the correct and corrupted fixtures — because an *all-uniform* per-expert contribution added to an already-uniform residual stream is exactly what a model's final RMSNorm, `x/rms(x)`, is invariant to for a single-layer fixture: it collapses to `sign(x)` and erases the perturbation's magnitude regardless of which expert produced it. Fixed by folding a small per-output-row multiplier into the down-proj fixture values so a corrupted expert's contribution is distinguishable post-norm too, not just via the direct buffer read above — documented in the test.)

**Real Qwen3.6-35B-A3B Q4 checkpoint** (40 layers, 256 experts/layer, hidden=2048, moe_intermediate_size=512 — quantized via #874's fix): ran `chat_metal --model <q4-dir>` against the real checkpoint under this machine's 32 GB unified memory.
- The original bug (crash at layer 0, missing-file error) is gone — the loader now progresses correctly through the MoE branch, mmap-loading and dequantizing each layer's expert tensors.
- Confirms the mmap route is engaged against real multi-hundred-MiB `.q4` files (432,164 page faults / 2.4M page reclaims observed during the run — consistent with file-backed I/O, not one large anonymous read).
- This checkpoint's true memory requirement for the current eager-per-layer-f16-dequant design (~1.5 GiB/layer × 40 layers ≈ 60 GiB for routed experts alone) exceeds this machine's 32 GB RAM. Without the cumulative headroom guard added in this PR, that would have run past available memory; with it, the load fails closed at layer 14 (23,040 MiB cumulative) with a clear diagnostic instead of OOMing:
  ```
  from_q4_dir: MoE layer 14: cumulative routed expert weights across all MoE layers loaded so far
  (23040 MiB, this layer adds 1536 MiB) exceed 0.85 × recommendedMaxWorkingSetSize (21725 MiB).
  ```
- *(round-1 review MEDIUM 2)* This is a **fail-closed loader observation, not generation evidence** — full end-to-end greedy generation on this specific checkpoint is **not achievable on this 32 GB machine** with the current eager-dequant design, and the issue's end-to-end acceptance criterion stays an **operational hold** pending more memory or a streaming loader. The rewritten distinct-expert synthetic test above is the CI-safe correctness proof (loads, generates, and is sensitive to expert-layout corruption); the real-checkpoint run is the "loads correctly, fails safely when it doesn't fit" proof. Streaming/partial-residency MoE loading is tracked at #682 (existing open issue, deduped against before filing — cross-referenced there with the Q4-path-specific numbers from this run, and from #876).

**RSS/mmap probe** (throwaway, not committed): loaded a synthetic 80 MiB `.q4` gate_up_proj tensor (134M elements) via `load_q4_mmap_dequant_f16` in isolation under `/usr/bin/time -l`: 256 MiB f16 output buffer, ~1.2 GiB process peak RSS (dominated by Metal/process baseline), 77,121 page reclaims consistent with the 80 MiB source being paged in as reclaimable file-backed memory rather than a pinned anonymous copy during the read itself (see the "mmap-backed source dequantization" wording above for what this does and doesn't prove about the final GPU-resident buffer).

**Regression**: full existing test suite (`cargo test -p lattice-inference --lib --features metal-gpu,f16`) — 1875+ passed, the only failures are pre-existing and environment-only (missing `LATTICE_KV_F16=1` env var for one test; a subset of real-0.8b-checkpoint tests need the `f16` cargo feature/downloaded checkpoint present, confirmed identical failures on `git stash` of this diff, i.e. present on `origin/fix/874-quantize-moe-experts` before this change).

## Gates

- `cargo check -p lattice-inference --tests` (default features): clean
- `cargo clippy -p lattice-inference --features metal-gpu --all-targets -- -D warnings`: clean
- `cargo fmt -p lattice-inference -- --check`: clean
- `from_q4_dir_*` tests (8 total: 2 new MoE-shape-gate regressions, 1 rewritten synthetic-generation test, 1 new mutation-sensitivity test, 4 pre-existing dense-reject tests): all pass
- **Remote CI**: *(round-1 review MAJOR 2)* `gh pr checks 883` has no checks — this repo does not run CI on a PR whose base is not `main` (known, standing limitation of the stacked-PR setup here, not something to fight around). This PR's CI will run for the first time once it retargets to `main` after `fix/874-quantize-moe-experts` (PR #878) merges; that is the point at which remote checks become meaningful evidence for this diff. Until then, the local gates above and the bench-compare run below are the available evidence.
- bench-compare: **done** — `make bench-compare BASE=origin/fix/874-quantize-moe-experts HEAD=HEAD` at the current head, run under `with-heavy-lane.sh "lattice:883-bench-compare-r1"` (exclusive GPU/heavy-lane serialization; base `b07d61f01`, head `fda7a9f41`). Gate report: **2 FAIL / 6 WARN / 41 confirmed improvements** — a large improvement over round-1's `29 FAIL / 28 WARN` on the identical class of benchmark, most likely explained by `#872` (`fix(bench): exclude quick-mode-noisy embed SIMD groups from bench-compare gate`), which landed on `fix/874-quantize-moe-experts` between the two runs and now marks some of these same benchmarks `(informational)` rather than hard-gating them.

Every one of the 2 FAIL + 6 WARN entries is again a `lattice-embed` `simd_batch_cosine*`/`simd_query_batch_dot_product`/`simd_batch_dot_product`/`simd_normalize` benchmark — a crate this PR does not touch (diff confined to `crates/inference/src/forward/metal_qwen35.rs`, entirely inside a new `if cfg.is_moe()` branch no dense code path or existing benchmark exercises). The crate this PR actually changes, `lattice-inference`'s `elementwise_cpu_bench` group (`elementwise_mul`, `add_bias_gelu`, `gelu`, `layer_norm`, `rms_norm`, `silu_inplace`), shows **zero FAIL/WARN entries** — every measurement is within noise (largest: `layer_norm/896` at −3.18%, in the improvement direction) or a confirmed win.

**Confound disclosure**: as in round 1, the `lattice-embed` findings are not attributable to this diff (no code-path overlap) and are consistent with contended-benchmark noise on this machine, not a real regression — this remains a reasoned hypothesis from code-path non-overlap, not independently disproven by a solo re-run of a flagged benchmark.

## Merge-forward

Base branch `fix/874-quantize-moe-experts` was refreshed twice while this round-1 fix was in flight (PR #878's `update-branch`, most recently `25d0432d3` → `b07d61f01` — pulling in `#872`'s `fix(bench): exclude quick-mode-noisy embed SIMD groups from bench-compare gate`, which is exactly the class of false-positive round-1's bench-compare run hit: see the bench-compare section above). Merged `origin/fix/874-quantize-moe-experts` into this branch both times; both auto-merged cleanly in `metal_qwen35.rs`. Re-ran `cargo fmt`/`cargo clippy --features metal-gpu --all-targets -- -D warnings`/the full `from_q4_dir` test subset after each merge — clean and green throughout. Head is now `fda7a9f413f2c1a06a45bb7f7bb283e98bad4beb`.

## Test plan

- [x] `from_q4_dir_loads_and_generates_on_synthetic_moe_checkpoint` — rewritten with distinct per-expert/per-projection constants + direct buffer-offset assertions, passes
- [x] `from_q4_dir_moe_decode_is_sensitive_to_expert_weight_corruption` — new mutation-sensitivity test, passes
- [x] `from_q4_dir_rejects_short_moe_router_header` — new shape-gate regression, passes, mutation-verified
- [x] `from_q4_dir_rejects_transposed_moe_routed_expert_header` — new shape-gate regression, passes, mutation-verified
- [x] `from_q4_dir_rejects_*` (4 pre-existing dense-path tests) — pass, unchanged
- [x] Full `cargo test -p lattice-inference --lib --features metal-gpu,f16` — no regressions vs pre-change baseline
- [x] Real Qwen3.6-35B-A3B Q4 checkpoint via `chat_metal` — loads correctly past layer 0 (bug fixed), fails closed with a clear diagnostic at the machine's memory ceiling (cumulative guard working as designed); full generation is an operational hold tracked at #682
- [x] bench-compare — current-head run, see Gates above


ohdearquant:

## Review of PR #883 (round 1)

### Verdict: REJECT

### Prior review context

- Round 1; no prior reviews or review threads are recorded.
- This review covers only `origin/fix/874-quantize-moe-experts...HEAD` at `98df32ab`, not the stacked branch's diff against `main`.

### What this round examined

- The Q4 writer/header format, f16 dequantizer, expert-major Metal offsets, and the MoE-aware safetensors constructor.
- MoE/dense selection, malformed-artifact behavior, mmap ownership, dense-path preservation, and all `from_q4_dir` binary call sites.
- PR/issue claims, current remote checks, commit/PR hygiene, and bounded local gates.

### Findings (# total: 1 blocker, 2 majors, 2 mediums)

#### Blockers

1. [crates/inference/src/forward/metal_qwen35.rs:13453] MoE Q4 headers are self-validated but never validated against the configured tensor shapes.
   - **Current:** Both new helpers call `validate_q4_header_payload_bounds`, which proves only `product(header.shape) == header.original_len` and that the payload is physically long enough. `load_moe_ffn_q4` then loads every routed, shared, and router tensor without comparing `header.shape` to the required shape ([metal_qwen35.rs:13453](crates/inference/src/forward/metal_qwen35.rs:13453), [metal_qwen35.rs:13591](crates/inference/src/forward/metal_qwen35.rs:13591), [q4_weights.rs:669](crates/inference/src/weights/q4_weights.rs:669)).
   - **Issue:** A structurally valid but short `mlp.gate.weight.q4`, for example shape `[1, hidden]` stored under the expected filename for a config with four experts, passes the new loader. It creates an f32 Metal buffer containing `hidden` values; `encode_moe_ffn` then forms a slice of `num_experts * hidden` from that pointer ([metal_qwen35.rs:10544](crates/inference/src/forward/metal_qwen35.rs:10544)). That violates the unsafe function's buffer-size invariant before the router loop, and equal-numel permutations such as `[E, H, 2I]` also silently feed the wrong layout to the GPU. The existing safetensors path already fail-closes by checking exact shapes ([loading.rs:83](crates/inference/src/model/qwen35/loading.rs:83), [loading.rs:115](crates/inference/src/model/qwen35/loading.rs:115)).
   - **Fix:** Pass an expected shape into each new mmap/dequant helper and reject before mmap/allocation unless it matches exactly:
     - router `[E, H]`; shared scalar gate `[1, H]`;
     - routed gate/up `[E, 2I, H]`; routed down `[E, H, I]`;
     - shared gate/up `[SI, H]`; shared down `[H, SI]`.
   - **Test:** Add at least a short router header and a same-numel transposed routed-expert header; both must make `from_q4_dir` return a shape error before `forward_step`.
   - **Confidence:** Derived from the source; no malformed checkpoint was run, as requested.

#### Majors

1. [crates/inference/src/forward/metal_qwen35.rs:334] The new synthetic test cannot catch the loader error class this PR most needs to protect against.
   - **Current:** `write_tiny_q4_fixture` fills every Q4 tensor with the same `0.1` value, and the test asserts only that generated logits are finite ([metal_qwen35.rs:334](crates/inference/src/forward/metal_qwen35.rs:334), [metal_qwen35.rs:18411](crates/inference/src/forward/metal_qwen35.rs:18411)).
   - **Issue:** An expert swap, wrong gate/up boundary, or an equal-length transpose leaves the test numerically finite and usually unchanged. It therefore cannot demonstrate the claimed expert-major layout or catch a wrong-expert load.
   - **Fix:** Use per-expert/per-projection distinct block-constant values. Assert the routed buffer values at expert 0/1 offsets and make routing deterministically select a nonzero expert; then assert its decode result (or an independently calculated FFN result) differs from expert 0 and matches the selected expert.
   - **Test:** This must run in the existing machine-wide `gpu_test_lock` just as the current test does.
   - **Confidence:** Observed by inspecting the fixture and by running `cargo test -p lattice-inference --lib --features metal-gpu,f16 from_q4_dir_loads_and_generates_on_synthetic_moe_checkpoint -- --nocapture` (1 passed).

2. [PR #883 body, `## Gates`] Head-quality evidence is incomplete: there are no remote checks for `98df32ab`, and the stated bench comparison is neither a current-head comparison nor a table.
   - **Current:** `gh pr checks 883` reports no checks on `feat/876-moe-q4-loader`. The body reports `bench-compare` on old base `d047c3fac` / old head `a120cda40`, then says the stack was merge-forwarded to base `25d0432d3` / head `98df32ab` without rerunning it.
   - **Issue:** Green checks on stacked PR #878 cannot validate this PR's head, and #878 itself is now `BEHIND main`. The argument that the reported `lattice-embed` regressions do not overlap this branch's new `cfg.is_moe()` branch is honest by inspection, but it does not make a pre-merge benchmark a head result. The body also contains prose figures, not the requested bench-compare table.
   - **Fix:** Trigger/obtain CI on `98df32ab`, and either rerun `make bench-compare BASE=origin/fix/874-quantize-moe-experts HEAD=HEAD` after the merge-forward or clearly mark the pre-merge result as non-gating historical evidence. Include the actual report table and the same confound disclosure.
   - **Test:** Required remote CI at the PR head; the local rerun should be reported with its exact refs.
   - **Confidence:** Observed with `gh pr checks 883` and `gh pr view 878 --json ...`.

#### Mediums

1. [crates/inference/src/forward/metal_qwen35.rs:13457] “Same mmap route as Dense” overstates what this implementation retains.
   - **Current:** The helpers do mmap the raw Q4 source and do not materialize a routed-expert `Vec<f32>`. However, they materialize an entire tensor as `Vec<u16>` and call `new_buffer_with_data`; the mmap is local and drops on return ([metal_qwen35.rs:13457](crates/inference/src/forward/metal_qwen35.rs:13457), [metal_qwen35.rs:13467](crates/inference/src/forward/metal_qwen35.rs:13467), [metal_qwen35.rs:13479](crates/inference/src/forward/metal_qwen35.rs:13479)). Dense `mmap_q4_weight` instead retains the mmap in `Q4WeightBuf` and gives Metal a no-copy view.
   - **Issue:** The code satisfies the narrower “no raw-Q4 `Vec` / no routed `Vec<f32>`” claim, but not a no-copy persistent mmap path. The RSS probe cannot establish the stronger claim after the local mmap has been dropped.
   - **Fix:** Describe this accurately as mmap-backed source dequantization into eager f16 buffers; do not call it the same no-copy route as Dense. A genuinely persistent/per-expert mmap route requires a Q4-capable expert kernel or explicit residency design.

2. [Issue #876, `## Acceptance`] The real 35B run is a fail-closed loader test, not completion of the issue's loads-and-generates acceptance criterion.
   - **Current:** The PR body says `chat_metal` reached the MoE loader but returned the cumulative-memory error at layer 14. It explicitly did not generate on the 35B Q4 directory.
   - **Issue:** This is useful safety evidence and was not silently credited here as generation evidence, but it leaves the issue's stated end-to-end criterion open. The synthetic run is the only observed successful generation proof and presently has the layout-test gap above.
   - **Fix:** Keep this limitation explicit in the PR and link a follow-up for streaming/partial residency. Once the distinct-expert synthetic parity test exists, it can be a CI-safe correctness proof while a sufficiently provisioned real-model run remains an operational acceptance hold.

### Offset/layout audit

The offsets themselves are correct for the Q4 writer's layout.

- Writer format: header is `4 + 4 + 4 + 8*ndim + 8` bytes, followed by contiguous 20-byte blocks of 32 flattened elements ([q4_weights.rs:528](crates/inference/src/weights/q4_weights.rs:528), [q4_weights.rs:640](crates/inference/src/weights/q4_weights.rs:640)).
- For a synthetic `E=2`, `I=1`, `H=32` routed gate/up tensor `[2, 2I, H]`, header size is 44 bytes. Expert 1 begins at element 64, hence Q4 file byte `44 + (64 / 32)*20 = 84`; its up half begins at element 96 / byte 104. The f16 destination indices are 64 and 96, exactly the dispatch offsets `e * 2 * I * H` and `e * 2 * I * H + I * H` ([metal_qwen35.rs:10704](crates/inference/src/forward/metal_qwen35.rs:10704)).
- For `[2, H, I]`, expert 1 begins at element 32 / Q4 file byte `44 + 20 = 64`, matching `e * H * I` ([metal_qwen35.rs:10707](crates/inference/src/forward/metal_qwen35.rs:10707)). The MSL kernel consumes exactly that expert-major f16 element offset ([qwen35.metal:2440](crates/inference/src/forward/shaders/qwen35.metal:2440)).
- The official Qwen config confirms the body’s real-model dimensions `E=256`, `I=512`, `H=2048`, so the routed f16 allocation is `256 * (2*512*2048 + 2048*512) * 2 = 1,610,612,736` bytes (1.5 GiB) per layer. [Official config](https://huggingface.co/Qwen/Qwen3.6-35B-A3B/blob/main/config.json)

- **Detection:** The branch is config-driven. A dense artifact with an unrelated expert-looking filename cannot select it; `cfg.is_moe()` selects the MoE arm, and a missing required expert filename returns an error rather than falling back to Dense. The blocker is malformed-but-present files, not name detection.
- **Dense path:** The stacked diff adds an `if cfg.is_moe() { ... } else { ... }`; the old Dense loads/fusion are textually preserved in the `else` arm. No semantic Dense-path change found by diff inspection.
- **Sibling paths:** `rg` confirms that `chat_metal`, `lattice_serve`, the `lattice` doctor/serve paths, `eval_perplexity`, and the listed benchmark binaries call the single `MetalQwen35State::from_q4_dir`; none rebuilds per-layer FFN weights. Other `MetalFfnWeights::Dense` occurrences are runtime match arms, not additional Q4-dir constructors.
- **Publication hygiene:** The PR body begins with the required agent attribution, contains the AI checklist and `Depends on #878`, and the feature commit has `Co-Authored-By: Leo <noreply@khive.ai>`. `git diff --check origin/fix/874-quantize-moe-experts...HEAD` is clean. New library unsafe operations have adjacent `SAFETY` comments; no new library `unwrap`/`expect` was introduced.
- **Local gates at head:** `cargo fmt -p lattice-inference -- --check` passed; `cargo clippy -p lattice-inference --features metal-gpu --all-targets -- -D warnings` passed. The targeted synthetic Metal test above passed under the repo's `/tmp/lion-metal-gpu-test.lock` guard.

### What I did NOT check

- I did not run the 35B load/generation or create a sabotage checkpoint, per review instructions. The PR body’s real-model observation is treated as an unreplicated claim.
- I did not run the workspace test suite, full CI, or a current-head bench comparison. No remote CI run is currently attached to this PR head.
- I did not claim the RSS/page-fault numbers as independently verified.

Fix the exact-shape gate and add a distinct-expert test first, then obtain head CI and current-ref benchmark evidence. A focused re-review should verify malformed-artifact rejection and selected-expert numerical identity; another broad round is not useful before that.

Domain utility: SKIPPED — repository-local Q4 format, Metal kernels, and ADR-053 were the authoritative sources for this review.

ohdearquant:
> _Review authored by Codex (OpenAI agent) on behalf of @ohdearquant._

## Review of PR #883 (round 2 — focused re-review)

### Verdict: APPROVE

### Prior review context

- Round 1 rejected the PR with one blocker (MoE Q4 header shapes), two majors (fixture sensitivity and current-head evidence), and two mediums (mmap wording and follow-up deduplication).
- This round verifies those five items only, against the requested stacked baseline `25d0432d3...fda7a9f41`; the remediation itself is `b07d61f01...fda7a9f41` and changes only `crates/inference/src/forward/metal_qwen35.rs`.

### What this round examined

- Exact Q4-header shape validation before mmap/allocation, its call-site shapes, and both malformed-header regressions.
- The revised synthetic MoE fixture, direct expert-offset reads, deterministic nonzero-expert routing, and corrupted-expert decode sensitivity.
- The PR body’s current-head benchmark/CI disclosure and the linked #682/#876 comments.

### Findings (# total: 0 blockers, 0 majors, 0 mediums)

All round-1 findings are resolved in the focused scope.

- **Exact-shape gate:** Both dequant helpers compare the raw header to `expected_shape` before `metadata`, `mmap`, or their output allocation: f16 at [metal_qwen35.rs:13526](crates/inference/src/forward/metal_qwen35.rs:13526)–[13554](crates/inference/src/forward/metal_qwen35.rs:13554) and f32 at [metal_qwen35.rs:13598](crates/inference/src/forward/metal_qwen35.rs:13598)–[13626](crates/inference/src/forward/metal_qwen35.rs:13626). The seven call sites supply `[E, 2I, H]`, `[E, H, I]`, `[SI, H]`, `[SI, H]`, `[H, SI]`, `[E, H]`, and `[1, H]` respectively ([metal_qwen35.rs:13706](crates/inference/src/forward/metal_qwen35.rs:13706)–[13763](crates/inference/src/forward/metal_qwen35.rs:13763)). These exactly match the authoritative safetensors loader’s router, routed-expert, shared-expert, and scalar-gate contracts ([loading.rs:99](crates/inference/src/model/qwen35/loading.rs:99)–[164](crates/inference/src/model/qwen35/loading.rs:164)).
- **Malformed artifacts:** The short router and same-numel transposed-down headers are self-consistent fixtures and assert an error from `from_q4_dir` before any `forward_step` ([metal_qwen35.rs:18822](crates/inference/src/forward/metal_qwen35.rs:18822)–[18910](crates/inference/src/forward/metal_qwen35.rs:18910)). Observed: `cargo test -p lattice-inference --lib --features metal-gpu,f16 from_q4_dir_ -- --nocapture` passed all eight selected tests. I also disabled only the two `header.shape != expected_shape` conditions in a disposable `/tmp` copy, under the shared Cargo target and the test suite’s `/tmp/lion-metal-gpu-test.lock`; both new regressions failed at their required-`Err` assertions. The temporary copy was removed afterwards.
- **Fixture and offset proof:** The fixture now makes gate, up, and down values distinct per expert, makes the selected expert be #2 via row constants, and reads expert 0/1 gate/up/down Metal-buffer offsets against the f16 Q4 round-trip values ([metal_qwen35.rs:18404](crates/inference/src/forward/metal_qwen35.rs:18404)–[18429](crates/inference/src/forward/metal_qwen35.rs:18429), [18500](crates/inference/src/forward/metal_qwen35.rs:18500)–[18543](crates/inference/src/forward/metal_qwen35.rs:18543), [18647](crates/inference/src/forward/metal_qwen35.rs:18647)–[18688](crates/inference/src/forward/metal_qwen35.rs:18688)). The non-uniform down-projection row multiplier prevents final RMSNorm from erasing the changed contribution ([metal_qwen35.rs:18521](crates/inference/src/forward/metal_qwen35.rs:18521)–[18543](crates/inference/src/forward/metal_qwen35.rs:18543)). The test corrupts only expert 2’s routed weights to expert 0 values and observes changed decode logits ([metal_qwen35.rs:18736](crates/inference/src/forward/metal_qwen35.rs:18736)–[18805](crates/inference/src/forward/metal_qwen35.rs:18805)); it passed in the same observed eight-test run. This is genuinely sensitive to the original wrong-expert/offset class, rather than merely asserting finite logits.
- **Head evidence disclosure:** PR #883’s `Gates` section now names the actual heavy-lane comparison and refs (`b07d61f01` to `fda7a9f41`), reports `2 FAIL / 6 WARN / 41 confirmed improvements`, identifies all flagged groups as untouched `lattice-embed`, reports zero `lattice-inference` FAIL/WARN entries, and labels the non-attribution conclusion as a reasoned hypothesis rather than an independently reproduced fact. I did not re-run the heavy benchmark, so its measurements remain author-reported. The disclosure is candid and the focused remediation diff itself touches only `metal_qwen35.rs`.
- **Remote CI:** Observed `gh pr view 883 --json statusCheckRollup` returns no checks while the PR base is the still-open `fix/874-quantize-moe-experts`. The PR body explicitly says no remote CI runs on this stacked base and that meaningful CI arrives after #878 merges and this PR retargets to `main`; it does not present absent CI as green evidence.
- **Mmap claim:** The PR body now precisely calls the design “mmap-backed source dequantization into eager, GPU-resident f16/f32 buffers,” distinguishes it from Dense `Q4WeightBuf`’s persistent no-copy mapping, and says the local mmap drops on return. The source likewise only says the f32 helper follows the f16 helper, not Dense’s no-copy route ([metal_qwen35.rs:13588](crates/inference/src/forward/metal_qwen35.rs:13588)–[13597](crates/inference/src/forward/metal_qwen35.rs:13597)).
- **Follow-up deduplication:** No duplicate issue was filed. The observed #682 comment accurately relates the Q4 eager f16 buffers and cumulative guard to its streaming/residency scope, and the observed #876 comment accurately retains end-to-end generation as an operational hold and links #682. The 1.5 GiB/layer / ~60 GiB total arithmetic follows the declared 256-expert, `H=2048`, `I=512`, 40-layer configuration: `256 × (2 × 512 × 2048 + 2048 × 512) × 2` bytes per layer.
- **Focused hygiene:** `git diff --check 25d0432d3...HEAD` is clean. No source files in the review worktree were modified.

### What I did NOT check

- I did not re-run the heavy-lane benchmark, the full inference test suite, clippy, or the real 35B load/generation; PR-body claims for those remain unreplicated.
- No remote CI exists at this stacked-branch head, so it cannot be credited as passing. It remains the normal merge gate after retargeting to `main`.

No further code-review round is useful for the five round-1 findings. After #878 merges and #883 retargets, require the normal remote CI result before merge; that is a gate, not a reason to reopen this focused remediation.

Domain utility: SKIPPED — the relevant contracts are repository-local (`load_moe_ffn_weights`, the Metal loader, test fixtures, and the PR/issue artifacts).

oceanwaves630 (approved):
Lane C full read (Leo) on head 452a8b7de.

- **Dispatch**: `cfg.is_moe()` drives the branch per layer — config-driven, so a missing MoE tensor file fails loudly instead of silently degrading to Dense. Dense path untouched in the else arm.
- **Shape gates**: exact `header.shape != expected_shape` rejection sits before mmap and before any allocation in both dequant helpers; the seven call-site shapes ([E,2I,H], [E,H,I], [SI,H]×2, [H,SI], [E,H], [1,H]) match the safetensors loader contract. Round-2's live mutation check (guards disabled → both regressions fail at their required-Err assertions) closes the round-1 blocker with real evidence.
- **Cumulative headroom guard**: the Cell accumulates routed-expert bytes across the whole layer loop against 0.85× recommendedMaxWorkingSetSize — the per-layer-passes-but-total-OOMs trap is exactly the failure class this catches, and it fails closed with the streaming-loader pointer. The guard counts routed bytes only; that under-count is conservative in the right direction and the diagnostic says so.
- **Dequant**: 20-byte block decode (scale/bias f16 + 16 packed bytes, low nibble first, truncate for tail padding) matches the Q4Block layout verified in the #878 read; f32 helper reuses `dequantize_row_q4_0`.
- Fixture sensitivity (distinct per-expert values, deterministic expert-2 routing, non-uniform down-projection surviving RMSNorm) verified by the round-2 leg with an observed 8-test pass.

Round-2 clean APPROVE posted on-PR. Normal CI at the retargeted head is the remaining merge gate — approving now, merge on green. Approved.

Issue #714, opened 2026-07-08
bug(bench): bench-compare base-then-head ordering fabricates large false regressions under thermal load


ohdearquant:
## Summary

`make bench-compare` (scripts/bench-compare.sh) builds and benches BASE first, then HEAD second, in the same process on the same machine. The second half runs on a CPU already thermally saturated by the first half's full Criterion sweep, so it throttles and every HEAD bench inflates uniformly. This fabricates large one-sided "regressions" that are pure artifacts of run order, not code.

## Reproduction (null A/B — identical source both sides)

```
scripts/bench-compare.sh origin/main origin/main   # same commit vs itself, zero diff
```

Result on an **idle** M2 Max (all other jobs quiesced), 2026-07-08:

- **56 FAIL / 12 WARN**, worst `simd_cosine_similarity/simd/384` **+378.87%** (CI [+371.58%, +386.27%]).
- 42 co-reported "confirmed improvements" from the same zero-diff comparison.

Since BASE and HEAD are the byte-identical commit, all 56 "regressions" are instrument artifacts. For reference, a real PR (#713, whose diff is entirely `#[cfg(feature = "bench-internals")]`-gated and cannot affect these benches) reported *fewer* — 35 FAIL — i.e. a genuine no-op change scores *better* than the tool comparing main to itself.

## Impact

`bench-compare` output is currently uninterpretable as a perf gate on this machine: the false-positive floor (56 FAIL from nothing) swamps any real signal. Every perf PR that pastes bench-compare output is affected, and reviewers cannot distinguish real regressions from thermal noise.

## Fix directions

1. **Interleave** the two refs at the per-group level (A/B/A/B) instead of all-BASE-then-all-HEAD, so thermal drift lands symmetrically (this is the same mitigation the GDN #175 harness adopted with its `--interleaved` mode).
2. Insert a **cooldown** between the BASE and HEAD halves and/or a longer warmup on the second half.
3. Randomize/alternate which ref runs first and report order-sensitivity.

Until fixed, treat bench-compare deltas as advisory only; disprove any flagged regression with a same-ref null A/B or a same-binary re-run.

ohdearquant:
New quantitative sample of this noise class, from PR #781's review cycle — the first time we have TWO same-source, same-compiler quick-mode runs on identical refs to compare directly.

[...]

PR #894, opened 2026-07-12
feat(inference): dequant-on-demand MoE expert cache with bounded LRU residency (#682)

ohdearquant:

Stage 1 of #682: dequant-on-demand routed MoE experts with a bounded LRU cache, replacing the eager full-tensor dequant that made large MoE checkpoints (e.g. Qwen3.5-35B: 256 experts × 40 layers ≈ 61 GiB f16-resident) unloadable on 32 GB machines. No `.q4` container change — per-expert slices are contiguous and Q4-block-aligned in the existing format, asserted at offset-table build time.

## What

- **`crates/inference/src/forward/moe_expert_cache.rs`** (new): `MoeExpertCacheConfig` (env override `LATTICE_MOE_EXPERT_CACHE_SLOTS`) with slot-count sizing (default = `num_experts` when it fits under 0.85× `recommendedMaxWorkingSetSize`, else auto-shrunk); `ExpertByteTable` (mmap + per-expert byte-offset table into the existing `.q4` file); `ExpertSlotCache` (LRU resident-slot cache: `begin_token()` / `resolve()` / `touch()`).
- **`metal_qwen35.rs`**: new `RoutedExpertStorage` enum — `Eager` keeps the safetensors-upload path byte-identical in behavior; `Cached` is the new `.q4` path. `load_moe_ffn_q4` builds two slot caches instead of dequantizing every expert; `encode_moe_ffn` Step 3 resolves per selected expert instead of offsetting into a monolithic resident buffer. The old eager-residency headroom guard is replaced by a guard sizing the cache pool itself.

## Buffer-lifetime safety across eviction

Decode runs one command buffer per token (encode all layers → `commit()` → `wait_until_completed()`). Cross-token: no prior-token GPU work is in flight when the next token's `resolve()` calls run, so eviction between tokens is safe. Within-token: `num_slots >= top_k`, and every slot resolved this token is marked touched and excluded from eviction until `begin_token()` resets, so a slot cannot be evicted while the current command buffer is still being encoded against it. Full invariant documented in the module doc comment.

## Tests (all mutation-verified)

- Eager-vs-lazy logits parity on the synthetic MoE checkpoint (`from_q4_dir_loads_and_generates_on_synthetic_moe_checkpoint`).
- Forced-eviction stress: `N < num_experts` over repeated decode steps matches the zero-eviction baseline (logit diff < 1e-4).
- Reload honesty: corrupting an evicted expert's on-disk bytes changes decode output, proving eviction/refill actually re-reads the file.
- Mutation-sensitivity proven by breaking the slot-reload path (skip reload memcpy on slot reuse) and watching both eviction tests fail; restored, all 12 targeted tests green solo.
- `cargo clippy --workspace -- -D warnings` and `--features metal-gpu --tests` both clean.

## bench-compare

`make bench-compare` (quick, defaults), origin/main vs this branch: every benched `lattice-inference` group (`rms_norm`, `layer_norm`, `silu_inplace`, `gelu`, `add_bias_gelu`, `softmax_attention`, `elementwise_mul`) is within the ≤3% noise band — expected by construction, since the changed code is `metal-gpu`-feature-gated and not compiled into the default bench binary. The FAIL/WARN entries the gate flagged are all in `lattice-embed`'s `simd` target, a crate this PR does not touch, consistent with the quick-mode noise class documented in `bench-compare.sh` (#714). Disclosed limitation: no Criterion target exercises the MoE decode path today, so there is no direct throughput measurement of the cache path; numerical correctness under eviction pressure is covered by the tests above, and the e2e-parity CI gate exercises the real forward path.

- [x] Diff read in full by the submitting maintainer
- [x] Tests/gates run locally (12 targeted tests solo under the GPU lock; clippy clean)
- [x] No internal tooling or process references in the artifact


oceanwaves630 (approved):
Lane B approve at ef09e61bd: codex round-2 clean APPROVE with codex running the test suite itself (16 tests incl. same-command-buffer eviction and in-place-mutation reload), .q4 container contract untouched, budget-override bypass closed with the 135-slot regression, full engine gate suite green on the branch (e2e parity SUCCESS also resolves the local forward:: failures toward the local-contention flake class — the branch is clean in CI), hygiene sweep clean.

PR #895, opened 2026-07-12
feat(inference): prefetch routed-expert loads to overlap I/O with shared-expert encoding (#682)

ohdearquant:

Stage 2 of #682: prefetch overlap for the routed-expert cache introduced in #894. Once the CPU routing decision is known, all top-k experts' cold-cache loads are spawned on worker threads, the shared-expert GEMVs (Step 2) are encoded while those loads run, and results are joined and applied before the routed-expert dispatch (Step 3). Cold-expert mmap read + CPU dequant genuinely overlaps Step 2 encoding instead of running serially in front of it. Numerical output is bit-identical; prefetch changes timing only. No `.q4` container change.

## Design

`ExpertSlotCache` splits miss handling into plan / spawn / apply phases:

- `plan_prefetch(expert_ids) -> Vec<PrefetchTask>`: single-threaded, mirrors `resolve()`'s exact bookkeeping (hit/miss classification, eviction-target selection, LRU/touched-flag commits) but defers the byte work. Every miss task owns a distinct pre-assigned slot by construction, and the slot is marked **unready** the moment it is assigned.
- `spawn_dequant(...)`: returns a `std::thread::ScopedJoinHandle` immediately; the worker borrows only the `Send + Sync` mmap byte table, never the cache's Metal buffers or bookkeeping.
- In `encode_moe_ffn`, Step 2's shared-expert GEMVs are encoded inside the same `std::thread::scope`, between spawn and join — this is the actual overlap window.
- `apply_prefetch_results(...)`: single-threaded after join; copies bytes into the pre-assigned slots and is the only path that flips a slot back to **ready**.
- `get_prefetched(expert_id)`: pure lookup that refuses an unready slot.

Failure isolation: if a dequant worker panics, its slots stay unready — `get_prefetched` rejects them, and the next `plan_prefetch` treats the unready mapping as a miss and reloads into the same slot (no eviction). If one of the two caches (gate_up / down) fails to join, the other's successful results are still applied before the panic is re-raised. One hard constraint shaped this: a panic that unwinds through a live Metal compute encoder (before `endEncoding`) is an unrecoverable process abort at the Metal layer, so recovery is guaranteed at the cache layer (readiness + reload) rather than by catching unwinds mid-encode.

All LRU/bookkeeping mutation completes single-threaded before any parallel work starts, so the within-token touched-slot invariant from #894 is preserved unchanged (module doc extended with the reasoning).

## Tests

- **Overlap proof (deterministic, no sleeps)**: `moe_prefetch_step2_overlaps_dequant_deterministically` — a channel-gated dequant worker blocks after signaling start; the test observes Step 2 finish encoding while the worker is still provably blocked, so Step 2 encoding cannot be ordered after the join.
- **Panic recovery**: `moe_prefetch_dequant_panic_recovers_via_readiness_reload` — injected dequant panic leaves both planned slots unready and refused by `get_prefetched`; the next plan reloads into the same slots (miss +2, evictions unchanged) and the recovered bytes are bit-identical to a never-panicked reference cache.
- **Bookkeeping equivalence**: `moe_prefetch_plan_matches_resolve_bookkeeping_exactly` — two caches driven through 4 tokens (duplicate expert IDs, hit-then-miss, eviction pressure: 6 experts through 3 slots), one via `resolve()` and one via the plan/spawn/apply path, with full-state snapshots (ownership, map, touched, readiness, LRU order, counters) asserted equal after every token.
- Plan determinism + per-plan slot uniqueness (pure logic, mutation-verified: forcing every miss to slot 0 fails immediately).
- Parallel-vs-serial decode parity: real `forward_step` under forced hit/miss/eviction pressure, byte-identical logits.
- Full MoE suite: 41 tests pass GPU-locked single-threaded (1 ignored manual throughput benchmark).

## Throughput (synthetic fixture — directional only)

`#[ignore]`d manual benchmark, methodology in-test: 16 experts, hidden/moe_inter 256, top_k=8, N=8 slots, alternating expert sets for a steady ~50% per-token miss rate; 3 repeats, 6 warmup / 60 timed steps, release build:

```
serial   median=176.679ms  (339.6 tok/s)
parallel median=108.751ms  (551.7 tok/s)
speedup  = 1.625x
```

This reflects real cross-phase overlap (dequant concurrent with Step 2 encoding) plus cross-miss parallelism. Two expectation-setting caveats: the fixture's per-expert payload (~384 KiB) is far smaller than a real large-checkpoint expert (~6.3 MiB), so the magnitude must not be extrapolated; and published MoE offloading systems consistently report prefetch/overlap gains in the single-digit-to-low-30% range at decode for production-scale experts — multi-fold headline numbers in the literature always bundle other changes. Re-measure at production scale once a large MoE checkpoint is loadable end-to-end.

## bench-compare

`make bench-compare` (quick, defaults): `lattice-inference` groups clean; 1 FAIL + 3 WARN confined to `lattice-embed` SIMD groups, a crate this diff never touches (100% of the diff is under `crates/inference/src/forward/`), matching the documented quick-mode noise class for that target (#714).

- [x] Diff read in full by the submitting maintainer
- [x] Tests/gates run locally (44 targeted GPU tests solo; clippy workspace + metal-gpu clean)
- [x] No internal tooling or process references in the artifact

### DIFF
--- a/crates/inference/src/forward/metal_qwen35.rs
+++ b/crates/inference/src/forward/metal_qwen35.rs
@@ -3683,6 +3683,26 @@ mod inner {
         FORCED_MOE_EXPERTS_FOR_TEST.with(|c| *c.borrow_mut() = experts);
     }
 
+    #[cfg(test)]
+    thread_local! {
+        /// Test-only override for `MetalQwen35State::encode_moe_ffn`'s
+        /// routed-expert prefetch fan-out (#682 Stage 2): `None` (the
+        /// production default) prefetches via rayon; `Some(false)` forces
+        /// the identical pre-assigned plan through a plain serial
+        /// iterator instead, so a test can assert the two fan-out
+        /// strategies produce byte-identical decode output from the same
+        /// checkpoint and token sequence. `Some(true)` forces rayon
+        /// explicitly (same as unset, provided for symmetry). Cleared
+        /// with `None` between tests by the caller.
+        static MOE_PREFETCH_PARALLEL_FOR_TEST: std::cell::RefCell<Option<bool>> =
+            const { std::cell::RefCell::new(None) };
+    }
+
+    #[cfg(test)]
+    pub(crate) fn set_moe_prefetch_parallel_for_test(parallel: Option<bool>) {
+        MOE_PREFETCH_PARALLEL_FOR_TEST.with(|c| *c.borrow_mut() = parallel);
+    }
+
     impl MetalQwen35State {
         /// **Unstable**: create from CPU weights; weight layout and f16 conversion may change.
         pub fn new(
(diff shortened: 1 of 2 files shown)


oceanwaves630 (approved):
Leo gate: APPROVE at 66a63ceb. All checks green at head including the q4 composed golden; codex APPROVE on record; merge-flow B lane. Auto-merge cascade authorized.

PR #898, opened 2026-07-12
fix(bench): enforce declared abs+pct AND condition in memory gate math

ohdearquant:

## Summary

Three math/shape corrections to the (not-yet-CI-wired) bench-gate schema-v2 evaluator and the live ADR-058 CPU-bench gate. Scripts-only diff; no CI wiring in this PR — `evaluate_family_gate` still has no production caller (`bench_decode_harness.py`'s per-family verdict aggregator that would wire it is separate, later work).

### Finding #2 — memory absolute-MiB AND-gate (the substantive fix)

`perf-policy.toml` declares `families.memory.warm_peak_phys_footprint` as an AND of a relative and an absolute threshold:

```toml
[families.memory.warm_peak_phys_footprint]
warn_pct = 0.03
fail_abs_mib = 64
fail_pct = 0.05
other_rule = "FAIL requires both >5% AND >64 MiB. ..."
```

but `CellGateInput`/`evaluate_family_gate` in `scripts/bench_gate_math.py` only implemented the relative leg (`fail_pct`) — `fail_abs_mib` was never read anywhere outside the TOML file.

**Counterexample reproduced**: base `warm_peak_phys_footprint` = 1000 MiB, candidate = 1055 MiB.
- Absolute delta = 55 MiB — does **not** clear the declared 64 MiB floor.
- Relative delta = 5.5% — clears the 5% `fail_pct`.

Before the fix, a cell with this shape (fed as a deterministic 7/9-pair paired log-slowdown of `log(1055/1000)`) reaches `FAIL` via the relative bound alone, contradicting the policy's own written "requires both" rule. This is currently latent — the memory family is not yet a gated required cell — but would silently ship a wrong verdict the day it becomes one.

**Fix**: added `fail_abs_mib: float | None` and `abs_delta_mib: float | None` to `CellGateInput` (both default `None`, so every family without a declared absolute floor is unaffected). `evaluate_family_gate`'s FAIL rule now additionally requires `abs_delta_mib` to exceed `fail_abs_mib` when the cell declares one. A cell that sets `fail_abs_mib` without supplying `abs_delta_mib` fails closed (`GateMathError`) rather than silently evaluating only the relative leg. `resolve_metric_policy` now structurally validates that a declared `fail_abs_mib` is a positive number, matching the fail-closed treatment every other policy-numeric field already gets.

While updating this module, also fixed a **live** bug in `bench_cpu_flagship_supervisor.py` (the shadow-mode-only PR-2 supervisor): it was calling `bootstrap_upper_bound` — the wrong, two-sided-diagnostic UPPER tail — and assigning the result to a variable literally named `corrected_lower_bound`. This is exactly the wrong-tail mis-wiring class round 1's original defect (documented in `bootstrap_lower_bound`'s docstring) and finding #8 (below) warn about, caught by grepping for callers of the renamed function before landing the rename. Shadow-mode-only, so nothing was ever gated on the wrong value, but the reported bound was silently overstated. Fixed to call `bootstrap_lower_bound`.

### Finding #4 — ADR-058 gate docstring precision (comment-only)

`scripts/perf-bench-gate.py`'s `verdict()` uses Criterion's two-sided 95% CI lower bound as a one-sided FAIL cutoff. This is directionally sound — it can only be *more* conservative than a properly-computed one-sided 95% bound, never less, so it cannot itself produce a false FAIL — but the module's docstring/exit-code comment ("regression > 7% confirmed by 95% CI") overstates precision: using a two-sided-CI lower percentile as a one-sided cutoff is closer to a ~97.5%-one-sided-confidence bound, not a calibrated one-sided 95% bound. Corrected the docstring and added a precision note next to `WARN_PCT`/`FAIL_PCT`. No logic change.

### Finding #8 — guard against re-wiring the wrong-tail bootstrap function

`bootstrap_upper_bound` was public, same call signature as `bootstrap_lower_bound`, with only a docstring warning against wiring it into a FAIL rule. Renamed to `_diagnostic_bootstrap_upper_bound` (non-public, name signals diagnostic-only) so a future caller can't reproduce round 1's original defect by a one-character-different-name mistake at a glance. Updated all references (docstrings, tests) and the one production call site found via `grep` (`bench_cpu_flagship_supervisor.py`, described above).

## Mutation-sensitivity proof (finding #2's AND-gate)

Per Π_TBV, the new AND-gate tests were verified to actually exercise the fix, not just pass vacuously:

1. Applied the fix and wrote 3 new tests exercising the AND-gate (`test_abs_mib_and_gate_crosses_pct_not_mib_is_not_fail`, `test_abs_mib_and_gate_crosses_both_is_fail`, `test_abs_mib_and_gate_exactly_at_floor_is_not_fail`).
2. Reverse-applied the fix (removed the `abs_ok` condition from the FAIL rule, restoring the pre-fix relative-only check) and re-ran just those 3 tests: **2 of 3 failed** (`crosses_pct_not_mib_is_not_fail` and `exactly_at_floor_is_not_fail` both incorrectly reported `FAIL`), reproducing the audit's counterexample exactly.
3. Restored the fix; all 3 tests pass again.

## Docstring + visibility corrections

- `scripts/perf-bench-gate.py`: exit-code doc + `WARN_PCT`/`FAIL_PCT` comment corrected to state the actual one-sided confidence level implied by using a two-sided CI's lower bound.
- `scripts/bench_gate_math.py`: `bootstrap_upper_bound` → `_diagnostic_bootstrap_upper_bound` (module-private), docstring rewritten to explain the naming-as-guard rationale; cross-references in `bootstrap_lower_bound` and `evaluate_family_gate`'s docstrings updated.

## Test results

```
uv run python3 -m pytest tests/test_bench_gate_math.py tests/test_bench_cpu_flagship_supervisor.py tests/test_bench_decode_harness.py -q
220 passed, 1 warning in 3.16s
```

15 new tests added to `tests/test_bench_gate_math.py` covering: the AND-gate FAIL/non-FAIL boundary (including the audit's exact counterexample), the "no absolute floor declared" no-op case, fail-closed validation when `fail_abs_mib` is set without `abs_delta_mib`, `resolve_metric_policy`'s structural check on `fail_abs_mib`, and the renamed diagnostic function's existing coverage carried forward.

**Scripts-only diff; no CI wiring in this PR** — no Rust changes, so `make bench-compare`/clippy do not apply here.

- [x] Read the referenced audit and independently re-verified each finding against the current source before writing any fix (confirmed the finding #2 counterexample by tracing `CellGateInput`/`evaluate_family_gate`/`resolve_metric_policy`; confirmed finding #4 is directional-not-magnitude by re-deriving the one-sided-vs-two-sided relationship; confirmed finding #8's risk was already live by grepping every caller of `bootstrap_upper_bound`).
- [x] Added regression tests for every behavioral change; proved the AND-gate test is mutation-sensitive by reverse-applying the fix and observing the expected test failures, then restoring it.
- [x] Ran the full relevant local test suite (`test_bench_gate_math.py`, `test_bench_cpu_flagship_supervisor.py`, `test_bench_decode_harness.py`) — 220/220 passing.
- [x] No CI workflow files touched; this PR is intentionally scoped to math/shape corrections only, per the audit's own phased fix plan.

### DIFF
--- a/scripts/bench_cpu_flagship_supervisor.py
+++ b/scripts/bench_cpu_flagship_supervisor.py
@@ -888,7 +888,20 @@ def build_decode_cell_aggregate(session: dict, policy: dict, rng_seed: int) -> t
         ci_low = _percentile(boot_means, 0.025) or point_estimate
         ci_high = _percentile(boot_means, 0.975) or point_estimate
         alpha_familywise = policy["gate_math"]["alpha_familywise"]
-        corrected_lower_bound = gate_math.bootstrap_upper_bound(
+        # bench-gate math audit finding #8: this previously called
+        # `bootstrap_upper_bound` (the wrong, two-sided-diagnostic UPPER
+        # tail) while assigning the result to `corrected_lower_bound` --
+        # exactly the wrong-tail mis-wiring round 1's original defect (see
+        # `bootstrap_lower_bound`'s docstring in bench_gate_math.py) warns
+        # against, and the same class of bug the rename to
+        # `_diagnostic_bootstrap_upper_bound` is meant to make harder to
+        # reproduce. This module is shadow-mode-only (PR 2's binding
+        # condition -- `verdict` here is downgraded to "unsupported" before
+        # any gated required cell consumes it), so no CI has ever gated on
+        # the wrong value, but the reported bound was still silently
+        # overstated (upper-tail > lower-tail). Fixed to the correct
+        # one-sided LOWER bound.
+        corrected_lower_bound = gate_math.bootstrap_lower_bound(
             log_slowdowns, order_ab_flags, b, random.Random(rng_seed + 1), corrected_alpha=alpha_familywise
         )
     else:
--- a/scripts/bench_cpu_flagship_supervisor.py
+++ b/scripts/bench_cpu_flagship_supervisor.py
@@ -66,6 +66,7 @@ import ctypes.util
 import hashlib
 import importlib.util
 import json
+import math
 import os
 import platform
 import random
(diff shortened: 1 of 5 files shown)


oceanwaves630 (approved):
Lane B approval (Leo) on codex r2 APPROVE-clean at 0ea71176f. Not new statistical design — corrects the evaluator to the already-declared policy (signed >5% AND >64 MiB, log1p-consistent comparisons), with boundary cases (55/64/70/-70 MiB) and a load-bearing mutation check in the verdict. 225 tests green. Scripts-only, no CI wiring.

PR #900, opened 2026-07-12
feat(embed): wasm32 SIMD128 kernels for dot/L2/cosine/normalize

ohdearquant:

## What

Adds `core::arch::wasm32` SIMD128 kernels for the four vector ops
`lattice-embed` exposes as its stable public contract
(`dot_product`, `squared_euclidean_distance`, `cosine_similarity`,
`normalize`), gated behind `#[cfg(all(target_arch = "wasm32", target_feature
= "simd128"))]`. Zero API change — the existing dispatch resolvers in
`crates/embed/src/simd/{dot_product,distance,cosine,normalize}.rs` pick the
new kernels automatically when the crate is built with `-C
target-feature=+simd128`; every other target and every existing call site is
untouched.

Also adds:
- `crates/embed/tests/wasm/simd128_parity_wasm.mjs` + driver script
  `scripts/wasm-simd128-parity.sh` — wasm-vs-wasm correctness gate (baseline
  vs simd128 build, both checked against an f32-faithful JS scalar
  reference), now wired into required CI (see "CI gate" below).
- `scripts/bench_wasm_simd.mjs` — self-building Node A/B harness, times both
  wasm builds across four embedding dimensions.
- Five `#[wasm_bindgen]`-exposed functions in `crates/embed/src/wasm.rs`
  (`simdDotProduct`, `simdSquaredEuclideanDistance`, `simdCosineSimilarity`,
  `simdNormalize`, `simdSimd128Dispatch`) — the existing wasm bindings only
  expose the full `LatticeEmbedder` class, not the `simd::*` vector-op
  layer; these give JS callers direct access to the same SIMD-accelerated
  ops a native Rust consumer already has via `lattice_embed::simd::*`.
  `simdSimd128Dispatch` is test-facing: it reports the exact dispatch state
  (`simd_config().simd128_enabled()`) the other four bindings' kernel
  dispatch reads, so the parity harness can prove which kernel tier a given
  build is actually exercising instead of inferring it from the `RUSTFLAGS`
  a build script requested.

## Why

`lattice-embed`'s wasm32 build had no SIMD path at all — every vector op
fell through to the portable scalar loop regardless of what the target
`.wasm` binary was compiled with. wasm32 has no runtime CPU-feature
detection (unlike the existing `is_x86_feature_detected!` / NEON paths in
this module), so this is a pure compile-time dispatch addition: a binary
built with `-C target-feature=+simd128` gets the new kernels, one built
without it keeps behaving exactly as before.

## wasm A/B results

Node v25.6.0, `process.hrtime.bigint()`, 5000 timed reps + 500 warmup per
cell, seed 0xC0FFEE (both variants marshal identical `Float32Array` inputs
through wasm-bindgen, so the delta isolates the kernel change, not call
overhead). `normalize` mutates its argument in place; its reset copy runs
via an explicit `prep` callback that `timeCalls` invokes immediately before
the timer starts (see "Bench timing correction" below for a methodology fix
applied to this table since the numbers were first posted):

| op | dim | baseline median (ns) | baseline p95 (ns) | simd128 median (ns) | simd128 p95 (ns) | median speedup |
|---|---|---|---|---|---|---|
| dot_product | 384 | 583.0 | 958.0 | 333.0 | 417.0 | 1.75x |
| squared_l2 | 384 | 584.0 | 1250.0 | 333.0 | 417.0 | 1.75x |
| cosine | 384 | 1208.0 | 1292.0 | 292.0 | 500.0 | 4.14x |
| normalize | 384 | 708.0 | 1375.0 | 334.0 | 916.0 | 2.12x |
| dot_product | 768 | 1042.0 | 1125.0 | 334.0 | 375.0 | 3.12x |
| squared_l2 | 768 | 1000.0 | 1084.0 | 333.0 | 375.0 | 3.00x |
| cosine | 768 | 2459.0 | 2584.0 | 375.0 | 416.0 | 6.56x |
| normalize | 768 | 1208.0 | 1292.0 | 375.0 | 459.0 | 3.22x |
| dot_product | 1024 | 1375.0 | 1458.0 | 375.0 | 417.0 | 3.67x |
| squared_l2 | 1024 | 1292.0 | 1375.0 | 375.0 | 417.0 | 3.45x |
| cosine | 1024 | 3333.0 | 3417.0 | 458.0 | 542.0 | 7.28x |
| normalize | 1024 | 1584.0 | 1708.0 | 500.0 | 542.0 | 3.17x |
| dot_product | 4096 | 4959.0 | 6875.0 | 959.0 | 1000.0 | 5.17x |
| squared_l2 | 4096 | 4916.0 | 4959.0 | 1000.0 | 1000.0 | 4.92x |
| cosine | 4096 | 13375.0 | 26708.0 | 1125.0 | 1250.0 | 11.89x |
| normalize | 4096 | 5541.0 | 5791.0 | 1250.0 | 2583.0 | 4.43x |

Consistent 1.75x-11.9x median speedup across all four kernels and all four
dims, growing with dimension. `cosine` shows the largest win since it fuses
three reductions (dot, norm_a, norm_b) into one SIMD128 pass — the same
structural reason the existing AVX2/NEON cosine kernels win big natively.

Reproduce: `node scripts/bench_wasm_simd.mjs --reps 5000 --warmup 500`
(self-building; requires `wasm-bindgen-cli` 0.2.105 and the
`wasm32-unknown-unknown` target).

### Bench timing correction

The first posted table's `normalize` rows included the per-iteration reset
copy (`scratch.set(template)`, required because `normalize` mutates its
argument in place) *inside* the timed region — `timeCalls` started the
clock before invoking its closure, and the reset ran inside that closure.
The copy was identical on both variants, so it never produced a false
correctness signal, but it was fixed overhead shared by both sides that
diluted the `normalize` speedup ratio specifically. `timeCalls` now takes an
optional `prep` callback that runs immediately before the timer starts on
every rep (warmup included); `normalize`'s reset moved there. Old vs new,
`normalize` only (the rows this fix actually changes):

| dim | baseline median: old → new | simd128 median: old → new | speedup: old → new |
|---|---|---|---|
| 384 | 750.0 → 708.0 ns | 459.0 → 334.0 ns | 1.63x → 2.12x |
| 768 | 1250.0 → 1208.0 ns | 417.0 → 375.0 ns | 3.00x → 3.22x |
| 1024 | 1708.0 → 1584.0 ns | 583.0 → 500.0 ns | 2.93x → 3.17x |
| 4096 | 5750.0 → 5541.0 ns | 1500.0 → 1250.0 ns | 3.83x → 4.43x |

Both sides got faster once the shared reset left the timed region, and the
speedup ratio increased at every dimension — consistent with the reset
overhead having diluted the ratio more than either side's raw median. The
other three ops (`dot_product`, `squared_l2`, `cosine`) never had a prep
step, so their table cells above reflect this fix's re-run rather than a
change in what is measured; the deltas versus the original posting (e.g.
`cosine`/384 2.58x → 4.14x) are ordinary run-to-run JIT/scheduling variance
between two separate `node` invocations on a shared dev machine, not an
effect of this fix.

## Correctness

`scripts/wasm-simd128-parity.sh` builds both wasm variants and runs
`crates/embed/tests/wasm/simd128_parity_wasm.mjs`: **15,668 checks, 0
failures** (15,666 numeric checks, unchanged, plus 2 new dispatch-state
checks below), covering remainder-length dims (1,2,3,5,7,17,385), typical
dims (384,768,1024,4096), zero vectors, denormals, alternating-magnitude
adversarial inputs, and isolated lane-position spikes (one nonzero element
at each of the 4 SIMD128 lane offsets) for dot/squared-L2/cosine, plus
`normalize` including the NaN-norm-leaves-vector-unchanged guard.

**Mutation-sensitivity, verified manually**: temporarily broke the shared
`horizontal_sum_simd128` helper (dropped lane 3 from the 4-lane reduction),
rebuilt the simd128 variant, and re-ran the parity script — **7,215 of
15,666 checks failed**, confirming the suite actually catches a broken
kernel. Reverted and re-ran: back to 0 failures, no test file left modified.

### Dispatch-state assertion

The numeric checks above only demonstrate SIMD128-vs-scalar correctness if
the "simd128" module is actually dispatching to the SIMD128 kernels and the
"baseline" module is actually falling back to scalar. Added
`simdSimd128Dispatch()` (returns `simd_config().simd128_enabled()`, the
exact value the kernel dispatch itself reads) and an assertion in the
harness that checks `false` for the baseline module and `true` for the
simd128 module *before* any numeric case runs, so a stale artifact, a build
that silently failed to pick up `RUSTFLAGS`, or a future dispatch regression
cannot pass the numeric checks by both sides quietly agreeing on the scalar
answer.

Mutation-sensitivity proof: pointed both `LATTICE_WASM_JS_BASELINE` and
`LATTICE_WASM_JS_SIMD128` at the simd128 build's artifact and ran the
harness directly:

```
FAIL [dispatch-state baseline] simdSimd128Dispatch()=true, expected=false
[dispatch-state simd128] simdSimd128Dispatch()=true (expected)

wasm32 SIMD128 parity gate: dispatch-state assertion failed (1 failure(s)) -- aborting before numeric cases, since they would prove nothing.
```

exit code 1. Restored the correct pairing and re-ran: back to 15,668 checks,
0 failures.

### Tolerance validity domain

The fixed `1e-5` relative / `1e-6` absolute epsilon in `assertClose` is only
a defensible bound for the harness's actual input domain: `[-1, 1)` via
`randomVec`, plus a handful of small handcrafted fixtures (zeros, denormals,
an alternating-magnitude case, lane-position spikes) — none of which drive
catastrophic cancellation in a 4-lane f32 accumulator tree. Added a comment
at the epsilon definition stating this explicitly, with a concrete
adversarial-cancellation counterexample (`a=[1e20,-1e20,...]`, `b=[1;16]`)
that is deliberately out of domain: a strict left-to-right scalar fold and a
4-lane split-then-reduce accumulator can legitimately disagree by more than
any fixed epsilon on such inputs (both are IEEE-754-correct given their own
accumulation order), so dispatch parity there is not part of this gate's
contract. `randomVec` now asserts every generated value is within `[-1,
1)`, enforcing the domain rather than only documenting it.

## CI gate

`scripts/wasm-simd128-parity.sh` was not previously invoked by any CI
workflow — a SIMD128 kernel regression could have merged with every existing
wasm CI check green. Added a `wasm-simd128-parity` job to
`.github/workflows/e2e-parity.yml`, mirroring the existing `wasm-parity`
job's rust-toolchain / `wasm32-unknown-unknown` target / pinned
`wasm-bindgen-cli` provisioning (no model-weight provisioning needed here,
since this gate covers only the four vector-op kernels, not the BERT
forward pass). The job runs with `LATTICE_WASM_SIMD128_ENFORCE=1` and is
wired into the `parity-gate` aggregate job's `needs` list and pass/fail
resolution, plus the engine change-detection path filter, the same way the
existing `parity`/`embed-drift`/`wasm-parity` gates are.

Fail-closed proof (driver script directly, prerequisite hidden via a
restricted `PATH`):

```
$ PATH=<no wasm-bindgen> LATTICE_WASM_SIMD128_ENFORCE=1 bash scripts/wasm-simd128-parity.sh
wasm-simd128-parity: FAIL (LATTICE_WASM_SIMD128_ENFORCE=1): wasm-bindgen (CLI) not found on PATH; install with: cargo install wasm-bindgen-cli --version 0.2.105
exit code: 1

$ PATH=<no wasm-bindgen> bash scripts/wasm-simd128-parity.sh   # no ENFORCE
wasm-simd128-parity: SKIPPED: wasm-bindgen (CLI) not found on PATH; install with: cargo install wasm-bindgen-cli --version 0.2.105
exit code: 0
```

`actionlint .github/workflows/e2e-parity.yml`: clean, no findings.

## Native impact (ADR-058)

This PR only adds `#[cfg(target_arch = "wasm32", target_feature =
"simd128")]`-gated code (including the new `simdSimd128Dispatch` binding,
gated on `target_arch = "wasm32"` + `feature = "wasm"`) — none of it exists
in a native x86_64/aarch64 build, and the CI-wiring/harness/bench changes in
this fix round touch no `crates/*/src/` files outside `wasm.rs`. Required
`make bench-compare` evidence, quick mode (unchanged from the original
posting — nothing native to re-measure):

```
BENCH_GROUPS_INFERENCE="" BENCH_GROUPS_EMBED="simd_dot_product" ./scripts/bench-compare.sh origin/main HEAD
```

Base=origin/main (539cb8fe8f), Head=HEAD (aaa96db840):

```
### `local-compare` — perf regression report

**3 FAIL** (regression >7.0% confirmed by 95% CI)
**1 WARN** (regression 3.0-7.0% confirmed)
**7 confirmed improvement**

| Bench | Δ point | 95% CI | new ns | base ns | verdict |
|---|---:|---|---:|---:|---|
| `residual_add_layer_norm/fused/hidden384_seq128` | +81.43% | [+58.33%, +104.59%] | 30465.1 | 16792.0 | FAIL |
| `rms_norm/896` | +74.29% | [+41.78%, +108.11%] | 318.5 | 182.7 | FAIL |
| `layer_norm/4096` | +55.87% | [+18.48%, +99.05%] | 1965.9 | 1261.2 | FAIL |
| `softmax_attention/512` | +12.96% | [+3.22%, +22.88%] | 111518.8 | 98722.0 | WARN |
| `residual_add_layer_norm/fused/hidden768_seq128` | -20.15% | [-23.78%, -16.35%] | 45586.9 | 57088.5 | WIN |
| `gelu/896` | -15.31% | [-24.97%, -3.39%] | 594.7 | 702.3 | WIN |
| `add_bias_gelu/896` | -21.50% | [-25.54%, -17.11%] | 473.7 | 603.4 | WIN |
| `add_bias_gelu/4096` | -23.03% | [-26.90%, -18.94%] | 2332.7 | 3030.6 | WIN |
| `silu_inplace/4096` | -28.52% | [-30.45%, -26.49%] | 2676.8 | 3744.5 | WIN |
| `residual_add_layer_norm/unfused/hidden384_seq128` | -24.45% | [-32.42%, -14.77%] | 73849.4 | 97748.1 | WIN |
| `residual_add_layer_norm/unfused/hidden768_seq128` | -34.04% | [-43.72%, -24.06%] | 128419.4 | 194695.9 | WIN |

**8 informational** (below quick-mode resolution — lattice-embed SIMD micro-benches, tracked in #714; not gated)
| Bench | Δ point | 95% CI | new ns | base ns |
|---|---:|---|---:|---:|
| `simd_dot_product/simd/1536` | +139.72% | [+137.39%, +142.07%] | 372.8 | 155.5 |
| `simd_dot_product/simd/1024` | +7.89% | [+5.41%, +10.46%] | 108.2 | 100.2 |
| `simd_dot_product/scalar/768` | -22.00% | [-31.97%, -9.24%] | 753.5 | 966.0 |
| `simd_dot_product/simd/768` | -35.10% | [-36.64%, -33.49%] | 59.9 | 92.4 |
| `simd_dot_product/scalar/1024` | -43.71% | [-44.17%, -43.24%] | 889.1 | 1579.4 |
| `simd_dot_product/scalar/384` | -44.55% | [-46.61%, -42.39%] | 263.9 | 476.0 |
| `simd_dot_product/scalar/1536` | -44.30% | [-54.68%, -27.80%] | 1544.6 | 2773.2 |
| `simd_dot_product/simd/384` | -54.01% | [-63.60%, -37.60%] | 28.3 | 61.5 |
```

**Why the FAIL/WARN entries are not attributable to this diff**: every
flagged bench (`residual_add_layer_norm`, `rms_norm`, `layer_norm`,
`softmax_attention`) lives in `crates/inference`, which this PR does not
touch at all — `git log $(git merge-base HEAD origin/main)..origin/main --
crates/inference/` returns zero commits, i.e. the inference crate's source
is byte-identical between base and head for this comparison. These are
quick-mode measurement-noise deltas from two independently-built binaries,
not a code-driven regression.

The 8 informational `simd_dot_product` entries are the *existing* native
x86_64/AVX2/aarch64-NEON dot-product dispatch (not the new wasm32 kernels
this PR adds) — the tool itself excludes this group from gating and the
deltas are internally contradictory (e.g. `scalar/768` -22% vs `simd/1536`
+139% in the same run), consistent with known noise for this group. This is
also the mechanically expected result: the only change to
`resolve_dot_product_kernel()` is one additional `wasm32`-gated branch,
compiled out entirely on native targets — the existing kernel bodies are
byte-unchanged.

## Gates

- `cargo fmt -p lattice-embed`: clean.
- `cargo clippy -p lattice-embed --all-targets -- -D warnings` (native): 0
  errors, 0 warnings.
- `cargo test -p lattice-embed --lib`: 262 passed, 0 failed (all
  pre-existing `simd::` tests included, run natively — unaffected).
- wasm32 baseline build (`--features wasm`, no RUSTFLAGS): clean.
- wasm32 simd128 build (`RUSTFLAGS="-C target-feature=+simd128"`): clean.
- wasm32-target clippy is not part of this repo's `clippy` gate (the
  Makefile's `clippy:` target is native-only, no `--target` flag). Ran it
  out of curiosity anyway: pre-existing `lattice-inference` failures
  (`drop_non_drop`, `needless_return`) block `-D warnings` on a wasm32
  cross-compile even on a clean `main` checkout, unrelated to this PR and
  out of scope here.
- `bash scripts/wasm-simd128-parity.sh`: 15,668 checks, 0 failures
  (includes the new dispatch-state assertions).
- `actionlint .github/workflows/e2e-parity.yml`: clean, no findings.

## Test invocation

```
# native
cargo test -p lattice-embed --lib

# wasm32 correctness parity (fail-closed in CI via LATTICE_WASM_SIMD128_ENFORCE=1)
scripts/wasm-simd128-parity.sh

# wasm32 A/B bench
node scripts/bench_wasm_simd.mjs --reps 5000 --warmup 500
```

Co-Authored-By: Leo <noreply@khive.ai>

### DIFF
--- a/.github/workflows/e2e-parity.yml
+++ b/.github/workflows/e2e-parity.yml
@@ -75,7 +75,7 @@ jobs:
           # still triggers wasm-parity. Without them, engine stays false and the
           # required parity-gate could pass green without the wasm path ever
           # running, a fail-open on the gate itself.
-          if grep -E '^crates/(inference|embed)/src/|^crates/(inference|embed)/Cargo\.toml$|^Cargo\.lock$|^scripts/e2e_parity_check\.py$|^\.github/workflows/e2e-parity\.yml$|^crates/inference/tests/quarot_q4_composed_golden\.rs$|^crates/inference/tests/fixtures/quarot_q4_composed_v1/|^scripts/gen_quarot_q4_composed_golden\.py$|^scripts/wasm-parity\.sh$|^crates/embed/tests/wasm/|^crates/embed/tests/fixtures/embed_parity_v1/|^crates/embed/examples/dump_parity_embeddings\.rs$|^scripts/ppl_gate_check\.py$|^crates/inference/tests/fixtures/ppl_gate_v1/|^crates/inference/src/bin/eval_perplexity\.rs$|^crates/inference/src/bin/quantize_q4\.rs$|^docs/bench_results/wiki\.test\.raw$' <<<"$CHANGED" >/dev/null; then
+          if grep -E '^crates/(inference|embed)/src/|^crates/(inference|embed)/Cargo\.toml$|^Cargo\.lock$|^scripts/e2e_parity_check\.py$|^\.github/workflows/e2e-parity\.yml$|^crates/inference/tests/quarot_q4_composed_golden\.rs$|^crates/inference/tests/fixtures/quarot_q4_composed_v1/|^scripts/gen_quarot_q4_composed_golden\.py$|^scripts/wasm-parity\.sh$|^scripts/wasm-simd128-parity\.sh$|^crates/embed/tests/wasm/|^crates/embed/tests/fixtures/embed_parity_v1/|^crates/embed/examples/dump_parity_embeddings\.rs$|^scripts/ppl_gate_check\.py$|^crates/inference/tests/fixtures/ppl_gate_v1/|^crates/inference/src/bin/eval_perplexity\.rs$|^crates/inference/src/bin/quantize_q4\.rs$|^docs/bench_results/wiki\.test\.raw$' <<<"$CHANGED" >/dev/null; then
             echo "engine=true" >> "$GITHUB_OUTPUT"
             echo "→ engine surface changed: parity REQUIRED"
           else
@@ -402,6 +402,60 @@ jobs:
           LATTICE_WASM_PARITY_ENFORCE: '1'
         run: ./scripts/wasm-parity.sh
 
+  # wasm32 SIMD128 kernel dispatch-parity gate. Proves the wasm32 SIMD128
+  # dot/L2/cosine/normalize kernels (crates/embed/src/simd/*.rs) agree with
+  # the scalar reference by building `lattice-embed` for wasm32 twice (plain
+  # vs `-C target-feature=+simd128`) and comparing both against a JS scalar
+  # oracle, plus a dispatch-state assertion proving each build actually
+  # dispatched to the kernel tier it claims to. No model weights needed here
+  # (pure vector-op kernels, not the BERT forward pass), so provisioning is
+  # the same rust/wasm-bindgen setup as wasm-parity above minus the model
+  # caching steps. See scripts/wasm-simd128-parity.sh for the local driver
+  # and crates/embed/tests/wasm/simd128_parity_wasm.mjs for the harness.
+  wasm-simd128-parity:
+    name: wasm SIMD128 kernel parity gate
+    needs: changes
+    if: needs.changes.outputs.engine == 'true'
+    runs-on: macos-latest
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: dtolnay/rust-toolchain@1.94.1
+        with:
+          targets: wasm32-unknown-unknown
+
+      - uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: wasm-simd128-parity
+
+      - name: Cache wasm-bindgen-cli
+        uses: actions/cache@v4
+        with:
+          path: ~/.cargo/bin/wasm-bindgen
+          key: wasm-bindgen-cli-0.2.105-${{ runner.os }}
+
+      # Pinned exactly to match the `wasm-bindgen` crate dependency version in
+      # crates/embed/Cargo.toml (the CLI and crate embed a schema version that
+      # must match exactly). Skips the (slow) reinstall if the cached binary
+      # is already the right version.
+      - name: Install wasm-bindgen-cli
+        run: |
+          if command -v wasm-bindgen >/dev/null 2>&1 && [ "$(wasm-bindgen --version)" = "wasm-bindgen 0.2.105" ]; then
+            echo "wasm-bindgen 0.2.105 already present (cache hit)"
+          else
+            cargo install wasm-bindgen-cli --version 0.2.105
+          fi
+
+      - name: Run wasm SIMD128 parity gate
+        # LATTICE_WASM_SIMD128_ENFORCE=1 makes the driver fail closed if any
+        # prerequisite (node, wasm-bindgen, cargo, or the wasm32 target) is
+        # still missing at this point: a provisioning failure must not
+        # masquerade as a skipped, green job.
+        env:
+          LATTICE_WASM_SIMD128_ENFORCE: '1'
+        run: ./scripts/wasm-simd128-parity.sh
+
   # Composed rotated(QuaRot)+Q4 forward-path gate (issue #320). The converter's
   # own f64 forward-equivalence check (quant/quarot/forward_equivalence.rs) gates
   # the rotation-fusion math PRE-quantization; nothing previously exercised the
@@ -640,14 +694,24 @@ jobs:
 
   # REQUIRED status check. Always runs, always reports. Passes when the engine
   # did not change (nothing to verify); mirrors the parity, embed-drift,
-  # wasm-parity, q4-composed, ppl-gate-selftest, and q4-ppl-quality verdicts when
-  # it did. Fails closed if change-detection itself errored. metal-parity is
-  # deliberately excluded (informational — see its comment above). q4-ppl-quality
-  # is now armed (golden captured on run 28841919096) and gates the shipping Q4
-  # tier for regressions.
+  # wasm-parity, wasm-simd128-parity, q4-composed, ppl-gate-selftest, and
+  # q4-ppl-quality verdicts when it did. Fails closed if change-detection
+  # itself errored. metal-parity is deliberately excluded (informational — see
+  # its comment above). q4-ppl-quality is now armed (golden captured on run
+  # 28841919096) and gates the shipping Q4 tier for regressions.
   parity-gate:
     name: parity-gate
-    needs: [changes, parity, embed-drift, wasm-parity, q4-composed, ppl-gate-selftest, q4-ppl-quality]
+    needs:
+      [
+        changes,
+        parity,
+        embed-drift,
+        wasm-parity,
+        wasm-simd128-parity,
+        q4-composed,
+        ppl-gate-selftest,
+        q4-ppl-quality,
+      ]
     if: always()
     runs-on: ubuntu-latest
     steps:
@@ -658,10 +722,11 @@ jobs:
           PARITY_RESULT="${{ needs.parity.result }}"
           DRIFT_RESULT="${{ needs.embed-drift.result }}"
           WASM_RESULT="${{ needs.wasm-parity.result }}"
+          WASM_SIMD128_RESULT="${{ needs.wasm-simd128-parity.result }}"
           Q4_RESULT="${{ needs.q4-composed.result }}"
           PPL_SELFTEST_RESULT="${{ needs.ppl-gate-selftest.result }}"
           Q4PPL_RESULT="${{ needs.q4-ppl-quality.result }}"
-          echo "changes=$CHANGES_RESULT engine=$ENGINE parity=$PARITY_RESULT embed-drift=$DRIFT_RESULT wasm-parity=$WASM_RESULT q4-composed=$Q4_RESULT ppl-gate-selftest=$PPL_SELFTEST_RESULT q4-ppl-quality=$Q4PPL_RESULT"
+          echo "changes=$CHANGES_RESULT engine=$ENGINE parity=$PARITY_RESULT embed-drift=$DRIFT_RESULT wasm-parity=$WASM_RESULT wasm-simd128-parity=$WASM_SIMD128_RESULT q4-composed=$Q4_RESULT ppl-gate-selftest=$PPL_SELFTEST_RESULT q4-ppl-quality=$Q4PPL_RESULT"
           if [ "$CHANGES_RESULT" != "success" ]; then
             echo "::error::change-detection did not succeed — failing closed."
             exit 1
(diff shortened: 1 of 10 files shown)


oceanwaves630 (approved):
Lane-C read (Leo) at head 545c490def5b49bcc3aba05d420f7a0752912ace — APPROVE.

Read the full 1541-line diff. Basis for approval:
- Kernels: compile-time-gated (`cfg(all(wasm32, simd128))`) 4x-unrolled SIMD128 mirroring the NEON structure; bounds arithmetic verified (floor chunking, safe-indexed remainders); normalize's NaN/zero-norm guard matches the scalar reference; cosine zero-norm guard consistent.
- Dispatch honesty: `SimdConfig::simd128_enabled()` reads the same compile-time flag the kernel cfg-gates key off, and `simdSimd128Dispatch` exposes exactly that value to the harness — the dispatch-state assertion runs BEFORE any numeric case and has its own mutation-sensitivity proof. This is the right shape for a no-runtime-detection target.
- Parity gate: f32-faithful JS oracle (fround-folded), remainder dims off the 16-float chunk size, lane-position spikes, denormals, NaN placement in both SIMD body and scalar tail, and an honestly documented tolerance validity domain (the cancellation caveat is correct — reassociation differences on ill-conditioned inputs are not bridgeable by epsilon and are rightly out of contract).
- CI: new job wired into parity-gate needs + explicit fail-closed result check; path filter extended for the new driver script; wasm-bindgen pinned to the crate dep version; ENFORCE=1 prevents provisioning failures masquerading as green. First live run passed on this PR.
- Claims: the PR-body A/B table (1.75x–11.9x median) is reproducible via the checked-in harness with documented methodology (prep-callback excludes the normalize reset from the timed region; hrtime; identical marshalling both sides). Bench is local-only, not a CI gate — correct call.

One non-blocking note for a follow-up, no action needed on this PR: `force_scalar`-style config overrides won't route a simd128-built wasm binary back to the scalar kernels, since `simd128_enabled()` is cfg-derived rather than field-backed. Native targets keep their invariant and the two-build harness sidesteps it, but worth a one-line doc note on `set_scalar_only` if that utility ever grows wasm expectations.

Clear to arm auto-merge; v0.6.1 chain may fire on merge.

PR #907, opened 2026-07-12
docs(claude): merge gate policy — checks at PR head, review-side merge-state diligence, revert-fast on red main

ohdearquant:

## What

Rewrites the "Merge Gate" section of `CLAUDE.md` to match the repo's actual branch protection: `required_status_checks.strict` is now `false` (verified live via `gh api repos/ohdearquant/lattice/branches/main/protection`), so a PR merges once required checks are green at its own head — it is no longer required to be up to date with main first.

## Why

The doc still described the old strict-checks regime (branch must be up to date with main, refresh-one-at-a-time discipline for a queue of armed auto-merges). That policy changed repo-side on 2026-07-12. The rewrite keeps the #634 incident lesson — a stale-base merge combined an API-shape change with call sites main had already gained, breaking main for two hours — but reframes it as the reason review now has to check API-shape changes against current main, since the automated up-to-date gate that used to catch this is off. It also adds a straightforward post-merge posture: watch the main-push CI run, and revert fast rather than fix-forward if main goes red.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -103,17 +103,20 @@ PRs touching `crates/inference/src/` or `crates/embed/src/` trigger `e2e-parity.
 
 ### Merge Gate
 
-A PR merges only when its branch is **up to date with main** AND **green at its actual head**.
-Branch protection enforces this (`required_status_checks.strict = true`), added after #634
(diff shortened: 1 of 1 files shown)


oceanwaves630 (approved):
Lane B approve-on-sight (docs). Auto-merge fires on green CI.

PR #908, opened 2026-07-13
docs(adr): ADR-058 post-acceptance note — CI-confidence escalation reduces power; gate stays unarmed

ohdearquant:

## What

Appends a dated post-acceptance note to ADR-058 (CPU perf regression CI). Docs-only; no behavior change.

## Why

An adversarial statistical review of the ADR's D2/D3 hard-gate design found the confidence-escalation path (on persistent false positives, tighten Criterion's CI confidence from 95% to 99% at the fixed 7% threshold, claimed to preserve sensitivity) to be statistically backwards: at a fixed threshold, higher required confidence widens the margin the lower bound must clear, reducing power against genuine regressions. Two compounding findings are recorded alongside it: within-run Criterion CIs understate cross-runner variance by more than an order of magnitude on shared CI hardware, and push-to-main baseline updates can ratchet a real regression into the baseline before any gate observes it.

The note also records status honestly: the hard merge gate described in the ADR was never armed. Merge gating shipped instead as same-runner e2e token parity plus a relative PPL-delta gate; Criterion micro-benchmarks are trend data only. The note exists so the escalation path is never armed as written by a future reader.

## Verification

- Docs-only diff (one file, appended section).
- Pre-commit doc lint (deno) passed.

### DIFF
--- a/docs/adr/ADR-058-cpu-perf-regression-ci.md
+++ b/docs/adr/ADR-058-cpu-perf-regression-ci.md
@@ -254,3 +254,32 @@ Bench wall time (~5-10 min on shared runners) would double the CI duration for e
 - ARM runner availability: <https://github.blog/2024-09-03-arm64-on-github-actions-powering-faster-more-efficient-build-systems/>
 - Issue #77 — GPU contention variance (M2 Max local bench noise)
 - PR #76 — context for catastrophic-cancellation revert (layer_norm two-pass)
+
+## Post-acceptance note (2026-07-12): the 95% → 99% escalation is unsound — do not arm it
+
+A subsequent adversarial statistical review of this design found the confidence-escalation
+path (D3 "Why 7% on CI-lower-bound" and Rollout step 5: on persistent false positives,
+tighten the CI confidence from 95% to 99% "rather than weakening the 7% sensitivity")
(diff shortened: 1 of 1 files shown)


ohdearquant:

REVIEW-VERDICT: APPROVE sha=c8be0c0f34b237f3240a46c82cbda487aa6effcc

Automated adversarial review, second pass. The three first-pass wording findings are resolved; no new findings.

- [medium, resolved] The note previously called the shipped PPL control a relative, same-run delta gate; it is an absolute Q4 perplexity golden captured on the CI runner class and enforced with a fixed tolerance. The note now describes the same-run token-parity gate and the cross-run PPL golden separately, including the latter's environment-drift risk.
- [medium, resolved] The Criterion description previously framed the change CI as within-one-session; it is bootstrapped from the observed baseline and candidate sessions. The unsupported order-of-magnitude variance figure is now conditional.
- [low, resolved] "Every push to main" overstated the baseline refresh trigger; corrected to path-filtered pushes plus scheduled and manual refreshes.
- Historical accuracy spot-checked: a continue-on-error benchmark-gate workflow existed and was later removed; the hard gate was never armed, as the note states.

oceanwaves630 (approved):
Post-acceptance note is statistically correct and the shipped-instead record is accurate. Approving.

PR #913, opened 2026-07-13
bench(scripts): migrate bench_apples_to_apples.sh to the shared decode-bench harness (#813)

ohdearquant:

## What

Issue #813 step 2, first of five live-consumer migrations onto the shared
decode-benchmark harness (harness core landed in #857; companion dead-script
removal landed in #869/#814). This PR migrates `scripts/bench_apples_to_apples.sh`
— listed first in the issue's migration order — onto
`scripts/bench_decode_harness.py`.

- `scripts/bench_decode_profiles.toml`: adds `apples_to_apples_q8` and
  `apples_to_apples_q4` profile definitions. Parameters match the pre-migration
  script exactly: windows `[32, 256]` (the legacy `N1=32`/`N2=256` slope-method
  pair), `measured_repeats = 5`, the same fixed prompt, `median` aggregation.
  Only `mlx` warms up (once, at 8 tokens), matching the legacy script's single
  pre-loop MLX-only warmup — lattice and ollama get zero warmup, also matching
  legacy behavior.
- `scripts/bench_decode_adapters_apples_to_apples.py` (new): registers real
  `EngineAdapter` implementations for lattice, ollama, and mlx, mirroring the
  legacy script's `bench_lattice`/`bench_ollama`/`bench_mlx` shell functions,
  then delegates to `bench_decode_harness.main()`.
- `scripts/bench_apples_to_apples.sh`: rewritten as a thin wrapper (44 lines,
  down from the original inline implementation) that execs the adapter module
  once per tier via `--allow-missing-engine`, preserving the legacy script's
  per-engine graceful-skip behavior. CLI invocation is unchanged.
- `tests/test_bench_decode_adapters_apples_to_apples.py` (new, 25 tests):
  profile-parameter equivalence against the legacy script's
  N1/N2/RUNS/prompt/warmup shape, exact call-schedule equivalence via the
  harness's own fake-adapter/fake-clock test pattern, and pure parsing-helper
  tests for the lattice `RESULT` line and the ollama JSON response.
- `tests/test_bench_decode_harness.py`: one test renamed and its assertion
  relaxed (`test_scaffold_file_loads_with_no_profiles` →
  `test_shipped_file_loads_and_parses_cleanly`) since the shipped profiles
  file is no longer empty past this first migration. This is the minimal
  in-scope fix for a test that directly asserts on the file this PR modifies.

## Disclosed methodology note

`bench_decode_ab` (the Rust bin) amortizes one model load across
`BENCH_RUNS` internal repeats within a single process, timing only the decode
window per repeat and excluding load time from its own `total_ms`. The
harness's per-call contract is different: each `EngineAdapter.run()` call is
one independently harness-timed measured repeat. There is no way to reconcile
these two shapes without a persistent-server mode on the Rust side, which is
out of scope for a scripts-only migration PR.

The adapter resolves this honestly rather than silently: it invokes the bin
with `BENCH_RUNS=1` per harness-measured repeat, so the harness's own
`elapsed_ns` (and therefore the reported "slope tok/s") now includes
subprocess-spawn and model-load overhead that the legacy script's `total_ms`
excluded. The true decode-only timing is not lost — it is preserved losslessly
via `AdapterRunResult.native_ns`, the harness's existing diagnostic field
designed for exactly this kind of engine-reported timing, and surfaces in the
report's separate "native tok/s" column.

**Update after a review pass** (see Verification below): the duration-
boundary methodology for ollama and MLX needed the same honesty treatment.
The harness's PRIMARY reported slope always uses its own wall-clock
`elapsed_ns` around the whole `adapter.run()` call — a fixed harness
invariant, not something an adapter can override — so it is not claimed to
be byte-for-byte identical to the legacy script's per-engine self-timed
metric. MLX's primary metric has had its material retokenization divergence
removed: `run()` no longer does any work after `generate()` returns (the
initial version retokenized the output to count tokens, which silently
pulled that work inside the harness's timing window and also diverged from
the legacy script, which never verified actual token count either — it
trusted the requested `N` directly, same as this adapter does now). This is
not claimed to be an exact match against the legacy window: the harness's
`elapsed_ns` still additionally covers the cached `_load()` lookup/unpacking
immediately before `generate()` and the `AdapterRunResult` construction
immediately after — ordinary, near-fixed adapter/harness call overhead, not
a variable cost like retokenization was. Ollama's primary metric still
includes negligible local HTTP/JSON overhead beyond ollama's own
`total_duration` and is not claimed as byte-identical, but the
legacy-equivalent number is no longer lost or wrong: `native_ns` now sources
from `total_duration` — the field the legacy script's own primary slope
used — instead of the decode-only `eval_duration` the initial version of
this adapter incorrectly used there.

## Review-fix pass

A review of the first commit surfaced five findings, all in the fail-open /
fake-observation class, fixed here (`scripts/bench_decode_adapters_apples_to_apples.py`
+ its test file only — no change to the harness core or the profiles file):

1. **`--allow-missing-engine` is now model-aware for lattice, not just
   binary-aware.** A binary present with a profile-required model dir absent
   (e.g. Q8 weights present, Q4 absent) previously raised
   `LatticeUnavailableError` mid-run, which `--allow-missing-engine` could
   not catch, aborting the whole tier instead of gracefully skipping lattice
   the way the legacy script did. `register_available_adapters()` now peeks
   `--profile`/`--profiles-file` out of argv and checks that SPECIFIC
   profile's required model directory (`lattice_registration_status()`)
   before deciding whether to register lattice at all.
2. **Duration-boundary methodology** — see the section above.
3. **Lattice `RESULT` parsing hardened**: a new `extract_single_result()`
   helper requires exactly one full-line RESULT record (anchored regex, no
   prefix/suffix noise tolerated), validates `n_req` equals the window that
   was actually requested, and validates `completion`/`total_ms` are finite
   and non-negative. A stale/wrong binary emitting a RESULT line for the
   wrong window, or two RESULT lines where one was expected, now raises
   `LatticeResultError` instead of silently producing a mislabeled
   observation or picking the last of several lines.
4. **Ollama response validation hardened**: `ollama_response_to_result` now
   requires and type-checks `eval_count`/`eval_duration`/`total_duration`,
   rejects a top-level `"error"` body, and no longer substitutes the
   requested token count for a missing or absent `eval_count` — a
   structurally-valid-JSON error response (HTTP 200, `{"error": ...}`) now
   raises `OllamaResponseError` instead of becoming a fake measurement.
5. **MLX Hub-fallback provenance**: a missing/corrupt local model directory
   makes `MlxAdapter` fall back to `Qwen/Qwen3.5-0.8B` from the Hub; the raw
   observation now records that via `AdapterRunResult.actual_model` instead
   of silently mislabeling it as the requested local artifact.

## Review-fix pass 2

A second review pass found the ollama hardening from pass 1 still had a gap,
the MLX wording above still overstated parity, and the registration fix's
test coverage didn't exercise its real entry points. Three fixes, same two
files only:

1. **Ollama validation now rejects fractional values, not just non-numeric
   ones.** Pass 1's type check accepted `float` alongside `int`, so a
   response like `{"eval_count": 31.9, "eval_duration": 1.0,
   "total_duration": 2.9}` passed validation and silently truncated to
   `actual_completion_tokens=31`/`native_ns=2` — recreating the
   fake-observation class the pass-1 fix set out to remove.
   `ollama_response_to_result` now requires a non-boolean `int` for all
   three fields; fractional values raise `OllamaResponseError`. New
   regression tests cover the exact fractional example above plus each
   field fractional in isolation.
2. **MLX wording corrected** — see the "Update after a review pass" section
   above, now reworded to say the material retokenization divergence is
   removed while ordinary adapter/harness overhead remains inside the timed
   window, rather than claiming an exact match against the legacy window.
3. **Registration fix now has real end-to-end test coverage.** Pass 1's
   tests called `lattice_registration_status()` directly; none exercised
   `_peek_requested_profile()` or `register_available_adapters()` — the
   actual argv-parsing path the fix changed. New tests build a temporary
   profiles file and real argv (`--profile name`, `--profile=name`,
   `--profiles-file ... --profile ...`, and profile omitted) and assert
   registration through `register_available_adapters()` itself. Writing
   these tests surfaced a real latent bug: `register_available_adapters()`
   called `lattice_registration_status(profile)` without passing
   `bin_path=`, so patching the module-level binary path in a test had no
   effect on that call (Python binds a keyword-only default once, at
   function-definition time, not at call time). Fixed by passing
   `bin_path=LAT_BIN` explicitly at the call site.

## Verification

- `uv run pytest tests/ -q`: 375 passed (up from 367 after review pass 1,
  142 at the first commit; `tests/test_bench_decode_adapters_apples_to_apples.py`
  now covers all findings from both review passes with real-adapter,
  mocked-load, and end-to-end argv/registration cases, not just the
  fake-adapter call-schedule tests from the first commit).
- `bash -n scripts/bench_apples_to_apples.sh`: clean.
- `cargo fmt --check`: N/A, no Rust touched (pre-commit's full
  `cargo check --workspace` + doc lint + capability-matrix check ran clean
  on both commits regardless).
- Live end-to-end re-run against real engines (ollama + mlx) after the
  hardening pass, confirming the adapters still function correctly and the
  corrected duration methodology takes effect:

  Q8 tier:
  ```
  lattice: BIN MISSING (.../target/release/bench_decode_ab) — build first (cargo build --release --bin bench_decode_ab)
  ollama: adapter registered
  mlx: adapter registered
  === apples_to_apples_q8 | windows=[32, 256] | aggregation=median ===
    (missing engine adapter(s), skipped: lattice)
    engine      window  slope tok/s legacy ±CI  native tok/s  n(base/win)
    ---------------------------------------------------------------------
    ollama         256        106.9          —          77.8          5/5
    mlx            256        153.5          —             —          5/5
  ```

  Q4 tier:
  ```
  lattice: BIN MISSING (.../target/release/bench_decode_ab) — build first (cargo build --release --bin bench_decode_ab)
  ollama: adapter registered
  mlx: adapter registered
  === apples_to_apples_q4 | windows=[32, 256] | aggregation=median ===
    (missing engine adapter(s), skipped: lattice)
    engine      window  slope tok/s legacy ±CI  native tok/s  n(base/win)
    ---------------------------------------------------------------------
    mlx            256        375.5          —             —          5/5
  ```

  Both raw JSONL outputs validate cleanly against the harness schema:
  `validate docs/bench_results/a2a_q8_raw.jsonl` → `OK: 21 observation(s)
  valid against schema v1`; `a2a_q4_raw.jsonl` → `OK: 11 observation(s)
  valid against schema v1` (21 = ollama's 2 windows × 5 repeats + mlx's 1
  warmup + 2 windows × 5 repeats; 11 = mlx only, ollama has no Q4 registry
  variant and self-skips by design, matching the legacy script's comment).
  A spot-check of a real ollama observation confirms finding 2/4's fix took
  effect: `elapsed_ns` and `engine_native_ns` are now within ~0.05% of each
  other (previously `native_ns` used the decode-only `eval_duration`, a
  meaningfully smaller number than the harness's whole-request wall clock).
  Ollama's graceful `--allow-missing-engine` skip is exercised for real in
  the Q4 tier above (no code path is untested). The throughput figures
  themselves are not directly comparable across the two live runs in this
  PR (different machine-load conditions between them, not a code change —
  MLX in particular is sensitive to concurrent GPU/CPU contention from other
  processes on this machine); they are included as functional evidence, not
  as a performance claim.

  The lattice leg itself was not exercised live in this session: the release
  build (`cargo build --release -p lattice-inference --bin bench_decode_ab
  --features f16,metal-gpu`) has been queued behind another lambda's
  exclusive heavy-lane job for the session's duration, so `BIN MISSING`
  above is the harness's own documented graceful-skip path, not a bug in
  this PR. The lattice adapter's subprocess invocation, `RESULT` line
  parsing/validation, and harness call-schedule are covered by deterministic
  unit tests (no binary required), including a real-`LatticeAdapter` test
  that exercises the binary-present/model-absent scenario directly, and pin
  the exact env vars (`LATTICE_MODEL_DIR`, `LATTICE_TOKENIZER_DIR`,
  `BENCH_N`, `BENCH_RUNS=1`) and argument shape the real binary expects.

## Out of scope

- The four remaining live consumers listed in #813
  (`bench_apples_precise.sh`, `bench_q4_apples.sh`,
  `bench_context_scaling.sh`, `bench_compare_1k.py`) — separate PRs per the
  issue's stated one-script-per-PR sequencing.
- Any change to `crates/inference/` (a persistent-server mode for
  `bench_decode_ab` would remove the lattice-adapter timing caveat above, but
  is a Rust-side change well outside a scripts-only migration PR).
- `.github/workflows/*` — untouched, per the CI-change-serial lane.

### DIFF
--- a/scripts/bench_apples_to_apples.sh
+++ b/scripts/bench_apples_to_apples.sh
@@ -1,12 +1,16 @@
 #!/usr/bin/env bash
 # Apples-to-apple decode-throughput benchmark — slope (marginal) method.
 #
-# Measures three engines IDENTICALLY at two quantization tiers:
-#   decode_tok_per_s = (N2 - N1) / (T_total(N2) - T_total(N1))
-#
-# Prefill, model load, and per-call overhead are constant for a fixed prompt
-# so they cancel in the difference — leaving the true marginal decode rate,
-# methodology-identical across engines.
+# Thin wrapper (issue #813 step 2, first migration): all methodology
+# (warmup policy, repeat count, aggregation, confidence interval, output
+# rendering) now lives in scripts/bench_decode_harness.py, configured by the
+# apples_to_apples_q8/apples_to_apples_q4 profiles in
+# scripts/bench_decode_profiles.toml. This script only execs the harness
+# (via scripts/bench_decode_adapters_apples_to_apples.py, which registers
+# the lattice/ollama/mlx engine adapters) once per tier. Parameters are
+# unchanged from the pre-migration version of this script: N1=32, N2=256,
+# 5 measured repeats, the same fixed prompt, and the same per-engine warmup
+# shape (only mlx warms up, once, at 8 tokens).
 #
 # Q8 tier (lattice default for <2B params, ollama default, mlx explicit):
 #   - lattice: F16 safetensors → auto-quantized to Q8_0 on Metal upload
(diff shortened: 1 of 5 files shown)


ohdearquant:

REVIEW-VERDICT: REQUEST-CHANGES sha=6374d31562f2a05095e13d7047821bff31262fdb

Independent review found four Major findings and one Medium — all in the fail-open/fake-observation class for the new adapters; parameter parity with the legacy script was otherwise confirmed (N1=32, N2=256, five repeats, median aggregation, model tags, MLX-only warmup).

1. **Major** — `--allow-missing-engine` tests only the executable: with the binary present but a model directory absent, the first call raises instead of reproducing the legacy per-engine skip, aborting the whole tier.
2. **Major** — the reported slope duration silently changed methodology: legacy used the engine-reported duration (e.g. the response's `total_duration`); the harness now records client wall-clock around the whole adapter call (including post-generation retokenization for MLX). Either restore the legacy duration boundary for primary observations or declare the methodology change explicitly instead of claiming parity.
3. **Major** — the RESULT-line parsing is unanchored and ignores the parsed `n_req`: a stale binary emitting a wrong-window RESULT substring becomes a valid observation with a fabricated slope. Require a full-line match, validate `n_req == n_tokens` and field finiteness.
4. **Major** — a JSON-valid but structurally malformed engine response (e.g. `{"error": ...}` with HTTP 200) becomes a fake measurement via substituted defaults. Require and type-check the needed response fields; reject error bodies.
5. **Medium** — the MLX local-directory → Hub fallback returns no provenance marker, so the JSONL labels a Hub download as the requested local artifact. Return the fallback identifier in `actual_model` or fail/skip.

Each fix needs a regression test (binary-present/model-absent, injected wrong-window RESULT, malformed-response, injected distinct engine-vs-wrapper durations, mocked fallback load). A second review pass follows the fixes.

ohdearquant:

REVIEW-VERDICT: APPROVE sha=1ef7cf0ab8602da3b3255bb14071e4f053d64bef

Fourth and final review pass. All findings from earlier passes are resolved and re-verified at source:

- [Major, resolved] Ollama response fields (`eval_count`, `eval_duration`, `total_duration`) are validated as non-boolean `int` only, matching the documented API types; fractional and boolean values raise a typed error instead of being silently truncated into fake observations.
- [Medium, resolved] MLX timing-window documentation now accurately describes what remains inside the timed window (cached model lookup and result construction) instead of claiming exact legacy equivalence.
- [Medium, resolved] Adapter registration is now exercised end-to-end across argv forms, which surfaced and fixed a latent definition-time default binding that defeated test patching of the binary path.
- [Medium, resolved] Per-field fractional coverage claim is now supported: each of the three fields has an isolated fractional rejection test, including the previously missing `eval_duration`-only case (the all-fractional case stops at `eval_count`, which is validated first).

Scope of the final fix commit: one test file, nine added lines. 56/56 tests pass locally.

oceanwaves630 (approved):
Lane C read: migration preserves legacy parameters exactly (windows/repeats/prompt/warmup), and the duration-boundary methodology is disclosed honestly rather than papered over (BENCH_RUNS=1 per harness repeat with decode-only timing preserved via native_ns; ollama native_ns sourced from total_duration matching the legacy slope). Spot-verified the two hardened seams at source: extract_single_result enforces exactly-one anchored RESULT line with n_req window validation, and ollama response validation rejects fractional/boolean values closing the truncation fake-observation path. Verification rows are demonstration evidence only and enter no perf ledger. Four codex rounds closed honest.

Issue #854, opened 2026-07-10
refactor(inference): extract the eight duplicated Metal RMS-norm reduction trees into one shared helper


ohdearquant:
> _Issue authored by Claude (Anthropic agent) on behalf of @ohdearquant._

## What

The same barriered RMS-norm reduction tree — per-thread `local_sum` → `shared[lid]` → tree-reduce in `tgs/2` steps → `rsqrt(shared[0] / width + eps)` — is duplicated across eight live Metal kernel bodies, verified by exact pattern search at current `main`:

- `flash_attention.metal::rms_norm` (`:48`)
- `qwen35.metal::rms_norm_qwen35` (`:312`)
- `qwen35.metal::per_head_rms_norm` (`:389`)
- `qwen35.metal::fused_residual_add_norm` (`:830`)
- `qwen35.metal::copy_and_rms_norm` (`:2189`)
- `qwen35.metal::copy_and_rms_norm_batch` (`:2226`)
- `qwen35.metal::fused_residual_add_norm_batch` (`:2264`)
- `qwen35.metal::per_head_rms_norm_batch` (`:2471`)

(The batch variant `copy_and_rms_norm_batch`/`per_head_rms_norm_batch` pair is easy to undercount — an earlier inventory pass on this work missed one of them and treated the count as a floor rather than an exhaustive total; eight is the exhaustive, exact-search-verified count.)

## Why

Eight independently-maintained copies of the same reduction tree mean any bug fix or optimization to the reduction itself (barrier count, tree shape, epsilon handling) has to be applied eight times, and any one copy can silently drift from the others — the same duplication-drift risk this series' pipeline-builder issue addresses for kernel registration.

## What to change

Add an inline helper that accepts the caller-computed local sum, so fused kernels keep their own input/copy/residual work:

```metal
inline float rms_inv_from_local_sum(
    threadgroup float* shared,
    float local_sum,
    uint lid,
    uint tgs,
    uint width,
    float eps);
```

It must write `shared[lid]`, execute the same barriers, reduce in the same `tgs/2` order, and return the same `rsqrt` expression as every current copy. Callers retain their own threadgroup array declaration and their final shifted/plain gamma scaling. Every thread in a threadgroup must call the helper on the same control-flow path — no helper call may sit behind lane-local divergence.

Use a shared `shaders/rms_reduce.metal` fragment, assembled into both translation units via `concat!(include_str!(...), include_str!(...))`. Do this as **two separately reviewable commits**:

[...]

PR #914, opened 2026-07-13
refactor(inference): extract shared Metal RMS-norm reduction helper (#854)

ohdearquant:

## What

Extracts the barriered RMS-norm reduction tree — `shared[lid] = local_sum` →
barrier → tree-reduce in `tgs/2` steps → `rsqrt(mean + eps)` — duplicated
across eight Metal kernels into one shared inline helper,
`rms_inv_from_local_sum`, added in `crates/inference/src/forward/shaders/rms_reduce.metal`
and assembled as a prefix fragment ahead of both `flash_attention.metal` and
`qwen35.metal` via `concat!(include_str!(...), include_str!(...))`.

The eight call sites:

- `flash_attention.metal::rms_norm`
- `qwen35.metal::rms_norm_qwen35`
- `qwen35.metal::per_head_rms_norm`
- `qwen35.metal::fused_residual_add_norm`
- `qwen35.metal::copy_and_rms_norm`
- `qwen35.metal::copy_and_rms_norm_batch`
- `qwen35.metal::fused_residual_add_norm_batch`
- `qwen35.metal::per_head_rms_norm_batch`

Two commits, per the issue's sequencing:

1. `b6717dcc31` — mechanical lift: add the fragment + `concat!` assembly,
   all eight bodies still inline and byte-for-byte unchanged.
2. `dad1bdc144` — route all eight bodies through the helper.

## Why

Eight independently-maintained copies of the same reduction tree meant any
future fix to barrier count, tree shape, or epsilon handling would need to
land eight times, with drift risk between copies.

## Helper design

```metal
inline float rms_inv_from_local_sum(
    threadgroup float* shared,
    float local_sum,
    uint lid,
    uint tgs,
    uint width,
    float eps);
```

Callers keep their own `local_sum` accumulation (and any fused copy /
residual-add work), their own `threadgroup float shared[...]` declaration,
and their own epilogue gamma scaling — `rms_norm` uses plain `gamma[i]`,
every other call site uses Qwen3.5's shifted `(1.0f + gamma[i])`. The helper
covers only the shared reduction; nothing else was harmonized. No finite-input
formula, epsilon handling, gamma/buffer layout, or threadgroup geometry
changed.

### Commit 1 — mechanical-lift byte-reversibility (per-translation-unit)

Stripping the known fragment-length prefix off each assembled MSL translation
unit reproduces the pre-lift file content byte-for-byte (throwaway script,
diffed against `origin/main`):

```
fragment length: 1862 bytes
[FLASH] BYTE-IDENTICAL: stripped(assembled) == origin/main:crates/inference/src/forward/shaders/flash_attention.metal
[QWEN] BYTE-IDENTICAL: stripped(assembled) == origin/main:crates/inference/src/forward/shaders/qwen35.metal
```

### Commit 2 — exact enumeration + per-kernel numeric parity (all eight sites)

Two new enumeration tests (`flash_attention_rms_norm_uses_shared_reduction_helper`
in `metal.rs`, `qwen35_rms_norm_kernels_use_shared_reduction_helper` in
`metal_qwen35.rs`) assert the assembled MSL contains exactly one definition +
N call sites of `rms_inv_from_local_sum` (1+1 in `flash_attention.metal`,
1+7 in `qwen35.metal`) and exactly one occurrence of the old
`shared[lid] = local_sum;` pattern (inside the helper's own body) — a
reintroduced manual copy at any of the eight sites fails this test.

For numeric parity, each kernel's pre-refactor reduction tree is frozen
verbatim (renamed with a `_pre_854_oracle` suffix, bodies otherwise
byte-identical to what this PR replaced) and compiled as an independent
oracle Metal library. The live post-refactor kernel (dispatched from the real
`MSL_SOURCE`/`MSL_TEMPLATE`) is run against the frozen oracle on the same GPU,
same compile options, identical finite input (non-multiple-of-256 row/head
dims to exercise the strided-accumulation tail, plus dedicated all-near-zero
and f32-subnormal (`f32::from_bits(1)`) rows — including the residual inputs
on the two fused kernels — alongside ordinary pseudo-random negative/positive
lanes) — the only source difference is inline-body vs. helper-call, so
bit-identical output proves the barrier count, tree order, and `rsqrt`
expression are unchanged for every one of the eight sites, at both ordinary
magnitude and at the epsilon floor:

Run with `LATTICE_METAL_TEST_ENFORCE=1` so a missing Metal device fails the
gate instead of silently skipping it (both oracle tests use this repo's
existing fail-closed device-probe convention, `metal_test_device` /
its `metal_qwen35.rs` equivalent — a runner that lost its GPU would report
these as `ok` without comparing anything otherwise):

```
running 5 tests
test forward::metal::inner::tests::flash_attention_rms_norm_uses_shared_reduction_helper ... ok
test forward::metal::inner::tests::msl_template_assembles_and_compiles ... ok
test forward::metal::inner::tests::rms_norm_matches_pre_854_oracle_on_finite_input ... [RMS_854_FIXTURE] rms_norm (flash): row0(near_zero) sum_sq=1.300e-38 row1(subnormal) sum_sq=2.553e-88
ok
test forward::metal_qwen35::inner::tests::qwen35_rms_norm_kernels_use_shared_reduction_helper ... ok
test forward::metal_qwen35::inner::tests::rms_reduce_854_parity::qwen35_rms_kernels_match_pre_854_oracle_on_finite_input ... [RMS_854_FIXTURE] rms_norm_qwen35: row0(near_zero) sum_sq=1.300e-38 row1(subnormal) sum_sq=2.553e-88
[RMS_854_FIXTURE] per_head_rms_norm: row0(near_zero) sum_sq=6.400e-39 row1(subnormal) sum_sq=1.257e-88
[RMS_854_FIXTURE] fused_residual_add_norm: normal fixture executed
[RMS_854_FIXTURE] fused_residual_add_norm: near_zero_residual fixture executed
[RMS_854_FIXTURE] fused_residual_add_norm: subnormal_residual fixture executed
[RMS_854_FIXTURE] copy_and_rms_norm: normal fixture executed
[RMS_854_FIXTURE] copy_and_rms_norm: near_zero_residual fixture executed
[RMS_854_FIXTURE] copy_and_rms_norm: subnormal_residual fixture executed
[RMS_854_FIXTURE] copy_and_rms_norm_batch: row0(near_zero) sum_sq=1.300e-38 row1(subnormal) sum_sq=2.553e-88
[RMS_854_FIXTURE] fused_residual_add_norm_batch/base: row0(near_zero) sum_sq=1.300e-38 row1(subnormal) sum_sq=2.553e-88
[RMS_854_FIXTURE] fused_residual_add_norm_batch/delta: row0(near_zero) sum_sq=1.300e-38 row1(subnormal) sum_sq=2.553e-88
[RMS_854_FIXTURE] per_head_rms_norm_batch: row0(near_zero) sum_sq=6.400e-39 row1(subnormal) sum_sq=1.257e-88
ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 1997 filtered out; finished in 0.16s
```

(`rms_norm_matches_pre_854_oracle_on_finite_input` covers `rms_norm`;
`qwen35_rms_kernels_match_pre_854_oracle_on_finite_input` covers the other
seven, individually, inside the same test — all eight sites' bit-identical
comparisons passed, at ordinary magnitude, all-near-zero, and f32-subnormal
input. The `[RMS_854_FIXTURE]` lines are a construction proof that those
dedicated edge-case rows actually landed in the dispatched buffer and were
compared, not just implied by a passing `ok`.)

Both GPU tests run under `gpu_test_lock()` (the in-tree machine-wide Metal
serialization convention).

## Gates

- `cargo fmt` — clean.
- `cargo clippy --workspace -- -D warnings` — clean.
- `cargo clippy --workspace --features f16,metal-gpu -- -D warnings` (incl.
  `--tests`, to lint the new GPU test module) — clean.
- The 5 tests above — all pass.

## Bench-compare

`crates/inference/` is touched, so `make bench-compare` is mandatory. Default
Criterion groups (`elementwise_cpu_bench`, `simd`) are CPU-only and cannot see
a Metal-shader-only change — `metal_decode_q4`/`metal_decode_q8`
(`benches/metal_decode_bench.rs`) are the only groups whose call graph
actually reaches the eight changed kernels (per-token decode dispatches
`rms_norm_qwen35`, `per_head_rms_norm`, `copy_and_rms_norm`, etc. every step),
so bench-compare was run with `BENCHES_INFERENCE=metal_decode_bench
CARGO_FEATURES_INFERENCE=metal-gpu,f16`, quick mode, `origin/main` (`b118d3871b`)
vs this branch's head (`dad1bdc144`):

| Bench | Δ point | 95% CI | verdict |
|---|---:|---|---|
| `metal_decode_q4/forward_step/no_adapter` | -7.21% | [-8.25%, -6.15%] | within noise, no regression |
| `metal_decode_q4/forward_step/lora_rank8` | -3.56% | [-5.33%, -1.73%] | within noise, no regression |
| `metal_decode_q8/forward_step/no_adapter` | -1.75% | [-5.05%, +1.71%] | neutral (CI crosses 0) |

No regression on any group this PR's diff can reach. The `metal_decode_q4`
point estimates are favorable but not significant: the underlying t-test
p-values on those same Criterion runs were 0.07-0.59, i.e. not significant at
the conventional p<0.05 threshold, so both rows read as within noise even
though this repo's own CI-lower-bound-based gate rule (CI excludes 0) would
otherwise mark them as passing/confirmed — that gate rule is a useful cheap
screen but is not itself a significance test, and this mechanical, bit-exact
extraction should not be read as an intentional performance change either
way. `rms_inv_from_local_sum` is `inline`, so Metal's compiler is expected to
fold it into each call site; a plausible mechanism for the point estimates,
if they held up under a repeated/tighter-CI run, is instruction-cache
pressure from removing eight duplicate reduction-tree copies — but that is
speculation, not a claimed result of this PR. The bench run also swept
`cargo bench -p lattice-embed`'s default
group (unfiltered `BENCH_GROUPS_EMBED`) as `make bench-compare` always runs
both crates; several `lattice-embed` SIMD micro-benches show FAIL/WARN deltas
in that same run, but `lattice-embed` is untouched by this diff (this PR only
touches `crates/inference/src/forward/`) — those are pre-existing machine-load
noise, unrelated to #854, and out of scope for this PR.

## Sequencing

Both of this issue's stated dependencies (GDN fail-open fix, `flash_attention.metal`
dead-kernel purge) are already merged on `main`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -303,6 +303,11 @@ jobs:
       # grammar steps' convention; the exact-count grep guards against a silently
       # zero-matching filter the same way those steps do. The gating command writes
       # its log directly (no pipe) so its exit status gates on its own.
+      # The count is 5 since #854: the shared RMS-norm reduction tests
+      # (flash_attention_rms_norm_uses_shared_reduction_helper,
+      # rms_norm_matches_pre_854_oracle_on_finite_input) live in this same
+      # `forward::metal::` module and match the filter. Adding or removing a
+      # test in that module requires updating this count in the same PR.
       - name: inference — metal-gpu fused_attention fail-closed gate (macOS only)
         if: runner.os == 'macOS'
         env:
@@ -315,7 +320,7 @@ jobs:
             exit 1
           fi
           cat "$log"
-          grep -Eq '^test result: ok\. 3 passed; 0 failed; 0 ignored; 0 measured; [0-9]+ filtered out;' "$log"
+          grep -Eq '^test result: ok\. 5 passed; 0 failed; 0 ignored; 0 measured; [0-9]+ filtered out;' "$log"
       # #611: Metal generation paths applied grammar masks via `mask_logits`
       # but never checked for a surviving finite logit before sampling — a
       # grammar that blocked every token silently emitted token 0 instead of
--- a/crates/inference/src/forward/metal.rs
+++ b/crates/inference/src/forward/metal.rs
@@ -904,6 +904,189 @@ mod inner {
             }
         }
 
+        /// #854: guards against a reintroduced manual copy of the RMS-norm
+        /// reduction tree. `rms_norm` must be the sole call site of the shared
+        /// `rms_inv_from_local_sum` helper in the assembled `flash_attention.metal`
+        /// translation unit (one definition + one caller).
+        #[test]
+        fn flash_attention_rms_norm_uses_shared_reduction_helper() {
+            let msl = msl_source_for(128, 2);
+            let helper_occurrences = msl.matches("rms_inv_from_local_sum(").count();
+            assert_eq!(
+                helper_occurrences, 2,
+                "expected 1 definition (rms_reduce.metal fragment) + 1 call site \
+                 (rms_norm) in the assembled flash_attention.metal, got {helper_occurrences}"
+            );
+            let inline_leftover = msl.matches("shared[lid] = local_sum;").count();
+            assert_eq!(
+                inline_leftover, 1,
+                "expected exactly one occurrence, inside rms_inv_from_local_sum's own \
+                 body; any additional occurrence is a reintroduced manual copy of the \
+                 reduction tree"
+            );
+        }
+
+        /// #854 numeric parity fixture (pre-refactor vs. post-refactor): `rms_norm`'s
+        /// reduction tree before this PR (frozen verbatim below as
+        /// `RMS_NORM_PRE_854_ORACLE`, byte-identical to the body this PR replaced —
+        /// see commit history) vs. the live post-refactor kernel that calls
+        /// `rms_inv_from_local_sum`. Both run on the same GPU with the same compile
+        /// options and identical finite input; the only source difference is
+        /// inline-body vs. helper-call, so any numeric drift here would mean the
+        /// extraction changed the barrier count, tree order, or rsqrt expression —
+        /// exactly the class of regression this PR's bit-exactness constraint
+        /// forbids (RMS-norm is a nonlinearity; f32 reduction order is part of the
+        /// contract).
+        #[test]
+        fn rms_norm_matches_pre_854_oracle_on_finite_input() {
+            const RMS_NORM_PRE_854_ORACLE: &str = r#"
+#include <metal_stdlib>
+using namespace metal;
+
+kernel void rms_norm_pre_854_oracle(
+    device float* x           [[buffer(0)]],
+    device const float* gamma [[buffer(1)]],
+    constant uint& row_len    [[buffer(2)]],
+    constant uint& num_rows   [[buffer(3)]],
+    constant float& eps       [[buffer(4)]],
+    uint gid  [[threadgroup_position_in_grid]],
+    uint lid  [[thread_position_in_threadgroup]],
+    uint tgs  [[threads_per_threadgroup]])
+{
+    if (gid >= num_rows) return;
+    constexpr uint RMS_WG = 256;
+    uint base = gid * row_len;
+
+    threadgroup float shared[RMS_WG];
+    float local_sum = 0.0f;
+    for (uint i = lid; i < row_len; i += tgs) {
+        float v = x[base + i];
+        local_sum += v * v;
+    }
+    shared[lid] = local_sum;
+    threadgroup_barrier(mem_flags::mem_threadgroup);
+
+    for (uint s = tgs / 2; s > 0; s >>= 1) {
+        if (lid < s) {
+            shared[lid] += shared[lid + s];
+        }
+        threadgroup_barrier(mem_flags::mem_threadgroup);
+    }
+
+    float rms = rsqrt(shared[0] / float(row_len) + eps);
+
+    for (uint i = lid; i < row_len; i += tgs) {
+        x[base + i] = x[base + i] * rms * gamma[i];
+    }
+}
+"#;
+
+            let Some(device) = metal_test_device("rms_norm_matches_pre_854_oracle_on_finite_input")
+            else {
+                return;
+            };
+            let _gpu = gpu_test_lock();
+
+            fn f32_buf(device: &Device, data: &[f32]) -> Buffer {
+                device.new_buffer_with_data(
+                    data.as_ptr() as *const _,
+                    std::mem::size_of_val(data) as u64,
+                    MTLResourceOptions::StorageModeShared,
+                )
+            }
+            fn read_f32(buf: &Buffer, len: usize) -> Vec<f32> {
+                // SAFETY: buffer is StorageModeShared, written by a completed
+                // command buffer, and sized for `len` f32 elements.
+                unsafe { std::slice::from_raw_parts(buf.contents() as *const f32, len).to_vec() }
+            }
+            fn lcg_vec(seed: u64, n: usize) -> Vec<f32> {
+                let mut state = seed;
+                (0..n)
+                    .map(|_| {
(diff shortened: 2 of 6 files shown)


ohdearquant:

REVIEW-VERDICT: REQUEST-CHANGES sha=dad1bdc144e679d90e1236726c054732629bfd93

Independent review confirmed the core refactor is sound: all eight call sites preserve the exact accumulator type, shared-memory write, barrier sequence, reduction tree, and epsilon placement (plain vs shifted gamma epilogues kept per site); the fragment is prepended exactly once per library; the parity oracles are genuinely independent frozen MSL; the enumeration assertions are mutation-sensitive by construction; and the helper is correct for the 256-thread geometry every production caller dispatches.

Three findings to fix before approval:

1. **Medium** — the Qwen3.5 oracle parity test returns success when no Metal device is available, without honoring the fail-closed enforcement variable the flash-side test uses. A CI runner without a usable device would report the parity gate green having compiled neither library.
2. **Medium** — the claimed denormal/near-zero-row coverage is absent: inputs span ordinary magnitudes with a single 1e-6 lane. Epsilon-dominated rows are exactly where a future helper edit could diverge while staying bit-equal on ordinary rows. Add all-near-zero and subnormal rows to both parity legs.
3. **Medium** — the description labels a statistically unconfirmed result (p = 0.07–0.59) as a confirmed win. A pure refactor with these p-values is within noise; the wording should be neutral (no regression) with the point estimates retained as informational.

A fourth suggestion (removing issue-number references from code comments) was reviewed and not adopted: public tracker references are established practice in this repository.

ohdearquant:

REVIEW-VERDICT: APPROVE sha=c3bf87fe9d4a4762745a08eda276422626da129b

Second-pass review, clean. All three findings resolved and verified at source: (1) the Qwen3.5 oracle parity test now uses the same fail-closed device enforcement as the flash-side test — with enforcement set, a missing device panics rather than reporting green; (2) all-near-zero (1e-20) and subnormal (`f32::from_bits(1)`) rows now reach every one of the eight kernels on both the frozen-oracle and live-library sides, including both inputs of the fused kernels, with exact `assert_eq!` comparisons throughout; (3) the benchmark wording is neutral ("within noise, no regression", p = 0.07–0.59 stated as non-significant). Second-pass scope confirmed as the two test-bearing Rust files only; shaders unchanged since the previously reviewed commit.

oceanwaves630 (approved):
Lane C read: rms_reduce.metal helper verified at source — the tree is byte-faithful to the prior copies (write, barrier, tgs/2 tree with per-step barriers, rsqrt(mean+eps)), reassociation explicitly forbidden in the contract comment, epilogue gamma scaling left caller-owned, and the full-participation divergence caveat documented. Verification design is the strongest in the repo to date: byte-reversibility proof for the mechanical lift plus frozen-oracle bit-identical parity across all eight sites including near-zero and f32-subnormal fixtures with construction proofs, under LATTICE_METAL_TEST_ENFORCE fail-closed convention. Your override of the codex Low (issue-ref comments = repo practice) is sound and documented.

ohdearquant:

CI follow-up on sha=c9f3f58e732d206f506d89e3e9cda3189b6d457f: the fused_attention fail-closed gate hard-asserts an exact test count in `forward::metal` as a guard against a silently zero-matching filter, and still expected the pre-change count of 3. The two shared RMS-norm reduction tests this PR adds live in that same module, so the gate failed on the correct new count of 5. The fix updates the expected count and documents that any test added to or removed from that module must update it in the same PR. Six-line workflow-only change; the reviewed code at sha c3bf87fe9d is untouched.

Issue #822, opened 2026-07-10
docs(inference): post-#787 generation entry-point behavior matrix + maintainer sign-off


ohdearquant:
> _Issue authored by Claude (Anthropic agent) on behalf of @ohdearquant._

## What

Produce a path-by-path behavior matrix for the seven Qwen3.5 CPU/Metal generation
entry points now that PR #787 (ADR-080 C3, merged as commit `86739db68`) has landed
`DecodePolicy::init`/`transition`/`StopMode`, and get explicit maintainer sign-off on
that matrix before any implementation issue in this series lands code.

This is a documentation/analysis deliverable only. No source changes.

## Why

Epic #614 stage 1 (shared generation preflight/request setup) needs a single source
of truth for what each entry point currently does before any of it is consolidated
into a shared preparation helper. The seven entry points differ in observable
contract in ways that must be preserved, not "refactored away" — a consolidation
that quietly normalizes these differences is a behavior change wearing a refactor's
clothes. Recording them explicitly, with maintainer sign-off, is the gate the
implementation issues in this series depend on.

## Entry points to cover

- `crates/inference/src/forward/cpu_f16.rs::generate_f16` (function starts ~line 833)
- `crates/inference/src/forward/cpu_q8.rs::generate_q8` (function starts ~line 686)
- `crates/inference/src/forward/neon_forward.rs::generate_q8_neon` (function starts ~line 871)
- `crates/inference/src/forward/metal_qwen35.rs::MetalQwen35State::generate` (~line 8591)
- `crates/inference/src/forward/metal_qwen35.rs::MetalQwen35State::generate_streaming` (~line 12913, a thin `should_cancel = || false` wrapper) and `generate_streaming_with_cancel` (~line 12947, the shared implementation both use)
- `crates/inference/src/forward/metal_qwen35.rs::MetalQwen35State::generate_streaming_with_prefix_cache_and_cancel` (public wrapper, ~line 15330) and its `_inner` (~line 15377)

Line numbers are current as of `origin/main` at `86739db68`; re-verify before use since
this file changes frequently.

## Matrix rows (minimum — one row per behavior, both entry points that reject/short-circuit and neighboring entry points that accept, so both directions are recorded)

- Explicit RNG seed `0` vs non-zero vs unset (time-based fallback)
- Empty prompt
- Zero-budget (`max_new_tokens == 0`)
- Grammar requested
- Logprobs requested
- Stop strings requested
- Reasoning budget requested
[...]

PR #915, opened 2026-07-13
fix(inference): reject empty prompts with typed Err in shared preflight on all paths (#856)

### LINKED ISSUE #856: bug(inference): empty prompt returns Ok on Metal paths but typed Err on CPU paths — unify on Err in shared preflight

## What

Empty-prompt handling diverges by backend, and the divergence is drift, not contract (maintainer ruling on PR #855, question 2):

- All three CPU generation paths (`generate_f16`, `generate_q8`, `generate_q8_neon`) return a hard typed `Err` on an empty prompt.
- All three Metal paths (`MetalQwen35State::generate`, `generate_streaming[_with_cancel]`, `generate_streaming_with_prefix_cache_and_cancel`) return a normal empty `Ok` — with `stopped: false` and `stop_reason: None`, a result shape that is itself invariant-violating (a completed generation that neither stopped nor gives a reason).

Full per-path citations: `docs/generation-entrypoint-matrix.md` (merged in #855).

## Ruling / fix shape

Unify on the CPU behavior: empty prompt = typed `Err`, implemented in the shared preflight of the generation-consolidation series (#822-#827, epic #614 stage 1) — NOT as a pre-consolidation behavior change to the Metal paths.

**Precondition (blocking):** a caller audit. Before the unified `Err` lands, enumerate every caller of the three Metal entry points (serve binaries, daemon, studio, tests) and verify none relies on the empty-`Ok` shape; any that does gets fixed in the same PR.

## Checklist

- [ ] Caller audit of all three Metal entry points for empty-prompt reliance (results table in the PR)
- [ ] Empty-prompt typed `Err` in the shared preparation helper, applied uniformly to all seven entry points
- [ ] Regression tests: empty prompt errors identically on every entry point (mutation-sensitive: reverting the guard on any single path must fail its test)
- [ ] `docs/generation-entrypoint-matrix.md` row updated to the unified contract

## Sequencing

Lands with (or after) the shared-preparation issue of the consolidation series; blocked on that series' helper existing. No standalone Metal-path patch before it.

ohdearquant:

## What / why

Closes #856. Empty-prompt handling diverged across the Qwen3.5 generation
entry points: the three CPU forward paths already rejected an empty prompt
with a typed `Err(InferenceError::Inference("empty prompt"))`, while all four
Metal generation entry points silently accepted it and returned an empty `Ok`
result instead. This PR adds one shared preflight,
`check_prompt_not_empty(prompt_len: usize) -> Result<(), InferenceError>`
(`crates/inference/src/model/qwen35/generation.rs:2227`), and wires every
entry point through it so the contract is identical everywhere: reject with
`Err(InferenceError::Inference("empty prompt"))`.

## Entry-point map / sibling audit

All call sites now route through the one shared guard (`grep -n
check_prompt_not_empty` output below is exhaustive over `crates/inference/`):

| Path | Call site | Outcome (after this PR) |
|---|---|---|
| CPU f16 | `cpu_f16.rs:868` | `Err(Inference("empty prompt"))` |
| CPU q8 | `cpu_q8.rs:721` | `Err(Inference("empty prompt"))` |
| CPU q8 NEON | `neon_forward.rs:906` | `Err(Inference("empty prompt"))` |
| `Qwen35Model::generate` (CPU dispatcher) | `generation.rs:67` | `Err(Inference("empty prompt"))` |
| `generate_streaming_with_observer` (CPU streaming) | `generation.rs:473` | `Err(Inference("empty prompt"))` |
| `MetalQwen35State::generate` | `metal_qwen35.rs:8711` | `Err(Inference("empty prompt"))` — was `Ok` before this PR |
| `generate_streaming_with_cancel` (also backs the `generate_streaming` thin wrapper) | `metal_qwen35.rs:13250` | `Err(Inference("empty prompt"))` — was `Ok` before this PR |
| `generate_streaming_with_prefix_cache_and_cancel` (public wrapper) | `metal_qwen35.rs:15985` | `Err(Inference("empty prompt"))` — was `Ok` before this PR |

Bonus dedup: the three CPU paths (`cpu_f16.rs`, `cpu_q8.rs`,
`neon_forward.rs`) each had their own inline `if prompt_len == 0 { return
Err(...) }` copy with independently-typed messages; all three now call the
one shared guard instead, closing a latent divergence (same `Err` variant,
but three separately-maintained message strings) before it could drift.

**Why the guard lives in the wrapper, not `_inner`.** The prefix-cache path
has a public wrapper (`generate_streaming_with_prefix_cache_and_cancel`) and
a private `_inner` helper with exactly one caller — the wrapper. `_inner`'s
caller unconditionally evicts the cross-turn prefix-cache slot on any `Err`
`_inner` returns (that's the correct behavior for the guards that *are*
duplicated there, e.g. logprobs/mtp — those checks run in both places as
defense-in-depth against future callers of `_inner`). An empty-prompt guard
inside `_inner` would trip that same destructive eviction for a rejection
that never touched the cache in the first place, since the empty-prompt
check is the very first thing that runs, before any cache read. So the guard
is placed once, in the wrapper, ahead of the suffix-validation logic; the
stale `_inner` early-return `Ok` block (the dead code causing the pre-fix
Metal/prefix-cache divergence) is removed and replaced with a
`debug_assert!` at `metal_qwen35.rs:16097` documenting the invariant that the
wrapper always screens empty prompts first.

**Siblings checked and deliberately left out of scope**, with rationale:

- `crate::generate::generate` (`src/generate.rs`) — deprecated top-level
  free function, already rejects empty prompt but with a *different* typed
  variant (`InvalidInput`) and message (`"Empty prompt"`); not part of the
  Qwen3.5 entry-point surface this issue scopes to, and changing a
  deprecated function's error variant is its own (out-of-scope) decision.
- `Qwen35Model::generate_with_batch_prefill` (`batch_prefill.rs`) —
  `#[allow(dead_code)]`, deprecated, unreferenced by any live entry point.
- `generate_multimodal` — checks `total_len` (visual + text tokens
  combined), a different precondition than a bare empty *text* prompt;
  out of scope for #856.
- `batch/worker.rs::submit` — returns `Option<SeqId>`, not a `Result`; not
  the same API shape as the guarded entry points.
- `kv_cache/cross_turn.rs::plan_cross_turn_reuse` — cache-planning helper,
  not an accept/reject boundary for a prompt.
- `model/qwen35/debug.rs::forward_prompt_debug` — debug-only forward pass,
  not a production generation entry point.

Two siblings the first pass of review missed were **reachable fail-open
paths**, not genuinely out of scope — see "Fixes from review" below:
non-`metal-gpu` build stub of `MetalQwen35State::generate` (fixed; fails
**loud** via `panic!` now instead of failing open, signature unchanged for
semver reasons — see round 2 below — with the fallible-`Result` shape
tracked for 0.7.0 in ohdearquant/lattice#917) and
`speculative.rs::generate_with_speculation` (documented intentional
exception, tracked for 0.7.0 in ohdearquant/lattice#916).

`docs/generation-entrypoint-matrix.md` (the maintainer-signed-off behavior
matrix from #822/#855) is updated: the "Empty prompt" row now shows `Err`
across all 7 documented paths, with the original pre-fix divergence
preserved verbatim in a collapsible `<details>` block for history. A new
scope note clarifies that "unified" refers to the seven
`MetalQwen35State`/CPU-forward paths this document tracks (eight physical
call sites counting the CPU dispatcher/streaming wrappers separately) and
explicitly excludes `generate_with_speculation`.

## Fixes from review (PR #915)

An earlier round of review on this PR caught two reachable fail-open paths
this diff had missed and two test-robustness gaps. Fixed in this push,
same branch:

1. **`speculative::generate_with_speculation(&[], ...)` returns `Ok`/`Vec::new()` on an empty prompt** (public, model-agnostic, non-`Result` helper) — contradicts the unified contract. Not signature-ified here: `lattice-inference` v0.6.1 just shipped, and turning this public free function fallible is a breaking change that doesn't belong in a same-version bugfix. Instead: documented the empty-input → empty-output sentinel explicitly on the function's doc comment as an intentional, temporary exception; narrowed this PR's claims (above) and `docs/generation-entrypoint-matrix.md`'s scope note accordingly; filed ohdearquant/lattice#916 (targets 0.7.0) to make it fallible, referenced from the doc comment; kept the existing regression test (`generate_with_speculation_empty_prompt_returns_empty`) and added a comment that it locks in the documented sentinel rather than an oversight.
2. **Non-`metal-gpu` build stub of `MetalQwen35State` was a fail-open hole**: the struct is a public unit type (directly constructible without the fallible `new`/`from_q4_dir`), and its `generate` stub returned `Ok(GenerateOutput{ .. })` for *every* prompt, including empty ones. **This fix went through two rounds** — see "Round 2 fixes" below for why the first attempt (changing the stub's return type to `Result`) was itself a semver break and had to be reverted in favor of a `panic!`-based fail-loud fix that keeps the published `-> GenerateOutput` signature.
3. **The prefix-cache empty-prompt test only asserted a fresh, empty cache** (`entry.is_none()`), which would still pass even if a future change moved the guard into `_inner` and took its destructive Err-recovery eviction path on a *warmed* slot — i.e. it didn't exercise the preservation property that motivated the wrapper-not-`_inner` placement in the first place. Fixed: rewrote `generate_streaming_with_prefix_cache_rejects_empty_prompt` to mirror the existing `..._rejects_logprobs_request_preserves_warm_entry` pattern — warm a real cache entry for the slot first, submit the empty prompt, assert the exact `Err`, then assert the pre-existing entry is unchanged afterward (extended further in round 2 — see below).
4. **Two of the three new Metal tests skipped silently when no Metal device was present**, unlike the third (prefix-cache) test, which already checked `LATTICE_METAL_TEST_ENFORCE` before accepting a device-absent skip. A required Metal CI leg without a device could report green on `metal_generate_plain_rejects_empty_prompt` / `metal_generate_streaming_rejects_empty_prompt` without ever running their assertions. Fixed: both now honor `LATTICE_METAL_TEST_ENFORCE` the same way — one shared fail-closed device precondition across all three.

## Round 2 fixes

A second review pass verified findings 1/3/4 above resolved, but caught
that the the first review pass's fix for finding 2 introduced a new semver problem, plus a
gap in how thoroughly the the first review pass's fix for finding 3 actually verified
"byte-identical":

1. **MAJOR — the the first review pass's stub fix was itself a semver-breaking signature change.** Changing the non-`metal-gpu` stub's `generate` from the published `-> GenerateOutput` to `-> Result<GenerateOutput, InferenceError>` breaks every existing **default**-build downstream caller that consumes a bare `GenerateOutput` — `metal-gpu` is not a default feature, so the stub (not the real implementation) is what a default build gets, and this module is public. `cargo-semver-checks`' inherent-method lint would not have caught it (it only detects a value-to-`()` transition, not one non-unit type changing to another). **Fixed**: restored the published `-> GenerateOutput` signature and made the stub fail **loud** instead — it now `panic!`s with `"MetalQwen35State::generate called on a build compiled without the metal-gpu feature; construct via new() which returns an error on such builds"` for every input. This is strictly better than the pre-#856/#915 fail-open state (impossible to miss, vs. an empty `Ok` that looked like a real generation) and matches the v0.6.0 release notes' fail-loudly convention for unimplemented GPU arms elsewhere in this crate. It is not the intended long-term shape (a panic is still rougher than a typed `Err`), so filed a **sibling** issue, ohdearquant/lattice#917 (targets 0.7.0, distinct from #916 which is `speculative`-specific), and referenced it from the stub's doc comment. Converted `non_metal_stub_generate_rejects_empty_prompt` / `non_metal_stub_generate_rejects_nonempty_prompt` to `#[should_panic(expected = "metal-gpu")]`. No other stub method's signature was touched by this PR (`git diff <published-base> -- metal_qwen35.rs | grep -E '^[-+].*\) -> '` shows exactly one changed return-type line, this one) — the "streaming variants" the reviewer asked to double-check don't exist as stubs at all; only `generate` has a body in the non-`metal-gpu` `impl` block that produces output, everything else already either returns a fixed scalar/`Result::Err` or is unreachable from a non-Metal build.
2. **MEDIUM — the the first review pass's prefix-cache survival test claimed "byte-identical" but only compared `generic.token_ids`.** `MetalCrossTurnPrefixEntry` also carries `gdn_snapshot` (`Vec<(Vec<f32>, Vec<f32>)>`) and sparse `checkpoints` (`Vec<MetalGdnCheckpoint { len, snapshot }>`); `CrossTurnPrefixEntry` itself (the `generic` field) has fields beyond `token_ids` (`slot_id`, `metadata`, `represented_len`, a `kv` handle, `gdn_snapshot_len`). **Fixed (preferred option — extended the comparison, not narrowed the claim)**: verified every one of those types is genuinely comparable — `CrossTurnPrefixEntry` and its `KvPrefixHandle`/`CrossTurnPrefixMetadata` fields all derive `PartialEq`/`Eq` (the `kv` handle is a small `Copy` value type of offsets/lengths, not a raw device pointer), and `gdn_snapshot`/`checkpoints`, while their containing structs don't derive `PartialEq`, are built entirely from `Vec<f32>`/tuple/`usize` pieces that do. The test now snapshots `generic` (compared via `==` on the whole derived-`Eq` struct), `gdn_snapshot`, and `checkpoints` (mapped to `(len, snapshot)` pairs) before the rejected call, and asserts all three are unchanged after. There is no field on `MetalCrossTurnPrefixEntry` this leaves uncompared, so "byte-identical" is now an accurate claim rather than an aspirational one.

## Regression tests

11 new tests, all touching only the guard / entry points above:

- `check_prompt_not_empty_rejects_zero`, `check_prompt_not_empty_allows_nonzero`
  (pure-guard unit tests, `generation.rs`)
- `qwen35_model_generate_rejects_empty_prompt` (CPU integration test via
  `Qwen35Model::generate`, `generation.rs`)
- `generate_f16_rejects_empty_prompt`, `generate_q8_rejects_empty_prompt`,
  `generate_q8_neon_rejects_empty_prompt` (one per CPU forward path)
- `metal_generate_plain_rejects_empty_prompt`,
  `metal_generate_streaming_rejects_empty_prompt` (one per direct/streaming
  Metal entry point; `gpu_test_lock()`, now honor
  `LATTICE_METAL_TEST_ENFORCE` on device-absent skip — fix 4)
- `generate_streaming_with_prefix_cache_rejects_empty_prompt` (Metal
  prefix-cache wrapper; `gpu_test_lock()`, `LATTICE_METAL_TEST_ENFORCE`;
  now warms a real cache entry first and asserts it survives the rejection
  — fix 3)
- `non_metal_stub_generate_rejects_empty_prompt`,
  `non_metal_stub_generate_rejects_nonempty_prompt` (fix 2; run in the
  **default** build, construct the non-`metal-gpu` `MetalQwen35State` stub
  directly; `#[should_panic(expected = "metal-gpu")]` as of round 2, not
  an `Err` assertion — see "Round 2 fixes")

## Mutation-sensitivity proof

Original round: reverse-applied the guard (restored the old CPU inline
checks / old Metal `Ok`-return blocks), `touch`ed the changed files to
force a rebuild, and re-ran the exact test sets above with the guard
neutered:

- CPU + pure-guard leg (6 tests): 5/6 failed as expected (the sixth,
  `check_prompt_not_empty_allows_nonzero`, correctly still passed — it
  never exercises the empty-prompt branch).
- Metal leg (3 tests): 3/3 failed, including a live observation of Metal
  actually generating garbage tokens (`"aaaa"`, `token_ids [0,0,0,0]`) from
  an empty prompt with the guard removed.

Review-fix round 1: reverse-applied fix 2 (restored the stub's old
unconditional `Ok(GenerateOutput{ .. })`), `touch`ed the file, re-ran the
two new stub tests — both failed with the `Ok(...)` value printed in the
panic message. Restored the fix; both pass again.

Review-fix round 2: reverse-applied the `panic!` (restored a plain,
non-panicking `GenerateOutput { .. }` return), `touch`ed the file, re-ran
the two stub tests under their new `#[should_panic]` form — both failed
with `note: test did not panic as expected`. Restored the `panic!`; both
pass again.

Restored every guard/fix in every leg; re-ran — all 11 pass again (see
full-suite output below). No mutation scaffolding was left in the tree
(`grep -rn "MUTATION-PROOF" crates/` returns nothing).

## Gates

- `cargo fmt --check -p lattice-inference` — clean
- `cargo clippy -p lattice-inference --all-targets --features f16,metal-gpu -- -D warnings` — clean
- `cargo check -p lattice-inference` (truly no features) and `cargo check -p lattice-inference` (default features) — both clean; confirms the restored `-> GenerateOutput` stub signature compiles the default build exactly as published
- `cargo check -p lattice-inference --bins --features f16,metal-gpu,serve` and `--bins` (default) — both clean (confirms no in-crate caller depends on the stub's signature either way; every binary gates its Metal-calling code behind the `metal-gpu` feature)
- `cargo test -p lattice-inference --lib` (default features): **1788 passed; 0 failed; 18 ignored**
- `cargo test -p lattice-inference --features metal-gpu --lib -- <9 shared+Metal test names> --test-threads=1`: **6 passed; 0 failed** (pure-guard + CPU-integration leg shown; Metal leg run separately under `gpu_test_lock()`, 3/3 passed; the 2 non-metal-stub tests are gated to the default build and don't apply under this feature set)

Bench-compare was **not** re-run for either fix round: none of the fixes
touch a benchmarked code path. Fix 1 is a doc-comment-only change. Fix 2
(both rounds) changes a stub function that's dead code within this
crate's own compiled binaries (every in-crate caller of Metal generation
gates behind the `metal-gpu` feature, confirmed by the `--bins` checks
above) — its only reachability is from an external, non-`metal-gpu`
downstream build; round 2 additionally reverts it to the exact
zero-runtime-cost shape it had in the published base (a `panic!` compiles
to no more code on the hot path than the old empty-`Ok` construction it
replaced — this stub is never on any benchmarked path either way). Fixes
3 and 4 are `#[cfg(test)]`-only. The original bench-compare run (below)
already covers the diff's only hot-path change (the preflight call
sites), and none of that changed in either fix round.

## bench-compare

Diff touches `crates/inference/`, so `make bench-compare` is mandatory. Ran
under `HEAVY_LANE_EXCLUSIVE=1` (solo-machine measurement) via
`with-heavy-lane.sh` after an initial non-exclusive run showed heavy
contention noise from other concurrent jobs on the machine (discarded).

Base=`origin/main` (`b118d3871b`), Head=`HEAD` (`64c0059924`), `--quick` mode.

The only groups this diff can plausibly affect are the `crates/inference`
elementwise/norm/softmax groups the preflight sits next to — the guard is a
single `prompt_len == 0` check on cold entry, off the hot generation loop:

| Bench | Δ point | 95% CI |
|---|---:|---|
| `rms_norm/896` | +0.10% | [-2.60%, +2.87%] |
| `rms_norm/4096` | +1.45% | [-0.01%, +2.94%] |
| `layer_norm/896` | +0.28% | [-3.17%, +3.82%] |
| `layer_norm/4096` | +0.13% | [-2.17%, +2.43%] |
| `silu_inplace/896` | -5.35% | [-6.06%, -4.64%] (🚀 confirmed win, not this diff) |
| `silu_inplace/4096` | -2.67% | [-3.20%, -2.14%] |
| `gelu/896` | -0.28% | [-1.27%, +0.73%] |
| `gelu/4096` | +2.66% | [-0.03%, +5.47%] |
| `add_bias_gelu/896` | -0.03% | [-2.27%, +2.26%] |
| `add_bias_gelu/4096` | +0.02% | [-2.08%, +2.17%] |
| `elementwise_mul/4096` | -2.44% | [-3.90%, -0.94%] |
| `softmax_attention/128` | +0.31% | [-1.28%, +1.94%] |
| `softmax_attention/512` | -0.33% | [-1.34%, +0.68%] |

All within the harness's own noise band; no group in `crates/inference`
crosses the 3.0% CI-lower gate threshold.

The full gate report separately flagged **1 FAIL + 10 WARN** — every one of
them in `lattice-embed` SIMD groups (`simd_batch_dot_product`,
`simd_batch_cosine_normalized_query`, `binary_cosine_distance`,
`simd_normalize`, `simd_squared_euclidean_fast_path`,
`int4_cosine_distance`). This diff does not touch `crates/embed/` at all —
none of those files appear in `git diff --stat` for this branch — and the
same report shows 30 *improvements* in the same crate's sibling benchmark
cases (e.g. `simd_batch_cosine_non_normalized_query` improving while
`simd_batch_cosine_normalized_query` regresses), the standard signature of
pre-existing micro-benchmark noise in that crate rather than a real
regression from this PR.

`bench-allow-regression` is not requested since the flagged groups are
outside this PR's diff.

### DIFF
--- a/crates/inference/src/forward/cpu_f16.rs
+++ b/crates/inference/src/forward/cpu_f16.rs
@@ -862,11 +862,10 @@ pub fn generate_f16(
     let prompt_ids: Vec<u32> = input.input_ids[..input.real_length].to_vec();
     let prompt_len = prompt_ids.len();
 
-    if prompt_len == 0 {
-        return Err(crate::error::InferenceError::Inference(
-            "empty prompt".into(),
-        ));
-    }
+    // #856: single shared preflight, unified with the Metal paths (which
+    // used to accept an empty prompt and return an empty Ok) — see
+    // `check_prompt_not_empty` (model::qwen35::generation).
+    crate::model::qwen35::check_prompt_not_empty(prompt_len)?;
 
     // max_new_tokens == 0 means "generate nothing": return before sampling so
     // we never emit a token the caller did not ask for. Mirrors the guard in
@@ -1835,6 +1834,54 @@ mod tests {
         assert!(
             out.stopped,
             "generate_f16 must set stopped=true when the decode-loop stop fires"
+        );
+    }
+
+    /// `generate_f16` must reject an empty prompt with a typed
+    /// `Err(Inference("empty prompt"))` before any weight dereference or
+    /// state allocation (#856): this is one of the three CPU forward paths
+    /// the shared `check_prompt_not_empty` preflight unifies with the four
+    /// Metal paths, which used to silently accept an empty prompt and
+    /// return an empty `Ok`. See docs/generation-entrypoint-matrix.md row 2.
+    ///
+    /// The guard fires before any weight dereference, so empty weight vecs
+    /// are sufficient (mirrors `generate_f16_rejects_grammar_config_before_sampling`
+    /// below).
+    ///
+    /// Mutation sensitivity: reverting the `check_prompt_not_empty` call at
+    /// this call site makes the function proceed past the guard with a
+    /// zero-length prompt, either panicking in the prefill/decode loop
+    /// (`all_ids.last()` on an empty vec) or producing a non-`Inference`
+    /// error — this assert fails either way.
+    #[test]
+    fn generate_f16_rejects_empty_prompt() {
+        use crate::error::InferenceError;
+        use std::collections::HashMap;
+
+        let mut vocab: HashMap<String, u32> = HashMap::new();
+        for (i, c) in ["h", "e", "l", "o"].iter().enumerate() {
+            vocab.insert((*c).to_string(), i as u32);
+        }
+        let merges = vec![
+            ("h".to_string(), "e".to_string()),
+            ("he".to_string(), "l".to_string()),
+        ];
+        let tokenizer = BpeTokenizer::from_vocab_and_merges(vocab, merges).unwrap();
+
+        let cfg = Qwen35Config::qwen35_2b();
+        let rope = RopeTable::new(cfg.rope_dim(), 8, cfg.rope_theta);
+        let weights = F16ModelWeights {
+            embed_tokens: vec![],
+            final_norm: vec![],
+            layers: vec![],
+        };
+        let gen_cfg = GenerateConfig::default();
+
+        let result = generate_f16(&weights, &cfg, &tokenizer, &rope, "", &gen_cfg);
+        assert!(
+            matches!(result, Err(InferenceError::Inference(ref msg)) if msg.contains("empty prompt")),
+            "generate_f16 must reject an empty prompt with Err(Inference(\"empty \
+             prompt\")) (#856); got {result:?}"
         );
     }
 
--- a/crates/inference/src/forward/cpu_q8.rs
+++ b/crates/inference/src/forward/cpu_q8.rs
@@ -715,11 +715,10 @@ pub fn generate_q8(
     let prompt_ids: Vec<u32> = input.input_ids[..input.real_length].to_vec();
     let prompt_len = prompt_ids.len();
 
-    if prompt_len == 0 {
-        return Err(crate::error::InferenceError::Inference(
-            "empty prompt".into(),
-        ));
-    }
+    // #856: single shared preflight, unified with the Metal paths (which
+    // used to accept an empty prompt and return an empty Ok) — see
+    // `check_prompt_not_empty` (model::qwen35::generation).
+    crate::model::qwen35::check_prompt_not_empty(prompt_len)?;
 
     // max_new_tokens == 0 is a valid request: "score this prompt, return no tokens".
     // Guard here so the decode loop (`for _ in 1..0`) never accidentally samples one
@@ -1546,6 +1545,49 @@ mod tests {
         assert!(
             msg.contains("context window"),
             "error must name the context window; got: {msg}"
+        );
+    }
+
+    /// `generate_q8` must reject an empty prompt with a typed
+    /// `Err(Inference("empty prompt"))` before any weight dereference or
+    /// state allocation (#856): this is one of the three CPU forward paths
+    /// the shared `check_prompt_not_empty` preflight unifies with the four
+    /// Metal paths, which used to silently accept an empty prompt and
+    /// return an empty `Ok`. See docs/generation-entrypoint-matrix.md row 2.
+    ///
+    /// Mutation sensitivity: reverting the `check_prompt_not_empty` call at
+    /// this call site makes the function proceed past the guard with a
+    /// zero-length prompt, either panicking in the prefill/decode loop or
+    /// producing a non-`Inference` error — this assert fails either way.
+    #[test]
+    fn generate_q8_rejects_empty_prompt() {
+        use crate::error::InferenceError;
+        use std::collections::HashMap;
+
+        let mut vocab: HashMap<String, u32> = HashMap::new();
+        for (i, c) in ["h", "e", "l", "o"].iter().enumerate() {
+            vocab.insert((*c).to_string(), i as u32);
+        }
+        let merges = vec![
+            ("h".to_string(), "e".to_string()),
+            ("he".to_string(), "l".to_string()),
+        ];
+        let tokenizer = BpeTokenizer::from_vocab_and_merges(vocab, merges).unwrap();
+
+        let cfg = Qwen35Config::qwen35_2b();
+        let rope = RopeTable::new(cfg.rope_dim(), 8, cfg.rope_theta);
+        let weights = Q8ModelWeights {
+            embed_tokens: vec![],
+            final_norm: vec![],
+            layers: vec![],
+        };
+        let gen_cfg = GenerateConfig::default();
+
+        let result = generate_q8(&weights, &cfg, &tokenizer, &rope, "", &gen_cfg);
+        assert!(
+            matches!(result, Err(InferenceError::Inference(ref msg)) if msg.contains("empty prompt")),
+            "generate_q8 must reject an empty prompt with Err(Inference(\"empty \
+             prompt\")) (#856); got {result:?}"
         );
     }
 
--- a/crates/inference/src/forward/metal_qwen35.rs
+++ b/crates/inference/src/forward/metal_qwen35.rs
@@ -30414,7 +30414,22 @@ mod inner {
         /// `..._rejects_logprobs_request_preserves_warm_entry` immediately
         /// below: warm the slot first, submit the empty-prompt request,
         /// assert both the exact error and that the pre-existing entry is
-        /// byte-identical afterward.
+        /// unchanged afterward.
+        ///
+        /// PR #915 review, second pass: the first version of this
+        /// test only compared `generic.token_ids` while claiming
+        /// "byte-identical" survival; `MetalCrossTurnPrefixEntry` also
+        /// carries `gdn_snapshot` and sparse `checkpoints`. This version
+        /// compares every field: `generic` (`kv_cache::CrossTurnPrefixEntry`,
+        /// which derives `PartialEq`/`Eq` over `slot_id`, `metadata`,
+        /// `token_ids`, `represented_len`, the `kv` handle -- itself a
+        /// small `Copy` value type of offsets/lengths, not a raw device
+        /// pointer -- and `gdn_snapshot_len`) via a direct `==`, plus
+        /// `gdn_snapshot` (`Vec<(Vec<f32>, Vec<f32>)>`) and each
+        /// `checkpoints` entry's `(len, snapshot)` via element-wise `==`
+        /// (neither type derives `PartialEq` itself, but their constituent
+        /// `Vec<f32>`/tuple/`usize` fields all do). There is no field on
+        /// `MetalCrossTurnPrefixEntry` this leaves uncompared.
         ///
         /// Mutation sensitivity: reverting the `check_prompt_not_empty`
         /// call in the wrapper makes this fail two ways -- the call
@@ -30455,10 +30470,14 @@ mod inner {
                 .cross_turn_prefix_cache
                 .entry
                 .as_ref()
-                .expect("precondition: the warm-up call must retain a cache entry")
-                .generic
-                .token_ids
-                .clone();
+                .expect("precondition: the warm-up call must retain a cache entry");
+            let warm_generic = warm_entry.generic.clone();
+            let warm_gdn_snapshot = warm_entry.gdn_snapshot.clone();
+            let warm_checkpoints: Vec<(usize, crate::attention::gdn::GdnSnapshot)> = warm_entry
+                .checkpoints
+                .iter()
+                .map(|c| (c.len, c.snapshot.clone()))
+                .collect();
 
             let rejected_cfg = cross_turn_test_gen_cfg(9, 2);
             let result = state.generate_streaming_with_prefix_cache(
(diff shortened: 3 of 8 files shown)


ohdearquant:

REVIEW-VERDICT: REQUEST-CHANGES sha=81c457f6afba9cb2ff3ecd95d05a412bf04ce60c

Independent review confirmed the core change: the shared guard lands at all eight intended entry points with an unchanged error variant/message for the previously-correct CPU paths; the wrapper-not-inner placement claim holds at source (an inner-level error would evict a warmed prefix-cache slot; the wrapper guard fires before any cache mutation, and the inner helper has exactly one production caller); the HTTP layer independently rejects empty API prompts as 400 before generation; and the deprecated/batch/multimodal/debug exclusions were each verified at source.

Four findings to fix before approval:

1. **Major** — the public speculative generation helper still accepts an empty prompt and returns an empty success value, with a regression test locking that in. It is a live exported surface, so it either joins the contract or gets an explicit documented exception; silence contradicts the PR's unified-contract claim.
2. **Major** — on non-Metal builds the state stub is a publicly constructible unit struct whose generate returns a successful empty completion for every input — a fail-open surface reachable without the failing constructor.
3. **Medium** — the prefix-cache regression starts from a fresh state, so it cannot detect the destructive-eviction regression that motivated the wrapper placement; it must warm a real entry and assert it survives the rejected call.
4. **Medium** — two of the three new Metal tests skip silently when no device is present instead of honoring the fail-closed enforcement variable the third one uses.

ohdearquant:

REVIEW-VERDICT: APPROVE sha=bb137e74c9413ff1546532c642ec1158496c0fb6

Third and final review pass. All findings from earlier passes are resolved and re-verified at source:

- [Major, resolved] The shared empty-prompt guard closes three real fail-open paths in the Metal generation entry points that previously returned success with garbage output on an empty prompt; all eight entry points now reject with the same typed error.
- [Major, resolved] The non-Metal stub keeps the exact published `-> GenerateOutput` signature (no default-build API break in the 0.6.x line) and now fails loud with an unconditional panic naming the missing `metal-gpu` capability, instead of silently returning empty output. The fallible `Result` signature is deferred to 0.7.0 and tracked in #917 (sibling to #916).
- [Medium, resolved] Both default-build regression tests assert the panic via `should_panic` and were mutation-verified (panic reverted to a plain return makes both fail).
- [Medium, resolved] The warmed prefix-cache survival test now compares every field of the retained cache entry (generic entry via derived equality, GDN snapshot, and sparse checkpoints), so the byte-identical claim matches what the test verifies.

Final fix commit scope: one file. 1788 default-build library tests pass; the three Metal-device tests pass under the machine GPU lock with a real device.

oceanwaves630 (approved):
Verified green at bb137e74 with independent review clean after three rounds; typed empty-prompt error plus the stub-signature semver fix are in scope.

PR #918, opened 2026-07-13
fix(inference): top-1 logits oracle uses first-wins tie-break to match greedy contract

### LINKED ISSUE #861: bug(inference): top-1 logits oracle in kv_cache/flat.rs uses last-wins max_by, contradicting the first-wins greedy contract

Sibling of #806 (found during its review sweep; out of #806's mandated one-file scope).

`top1_f32` and `top1_f16` in `crates/inference/src/kv_cache/flat.rs` (~lines 1427-1438) compute top-1 token choice with `Iterator::max_by`, which resolves exact ties to the **last** index. The engine's documented greedy contract (`crate::speculative::argmax`) is **first-wins**. f16 conversion can create exact ties even when the f32 logits are distinct, so the reported `top1_match_rate` (gate at ~lines 1472-1485) can evaluate a different token choice than production.

**Fix**: compute both winners with the shared first-wins `crate::speculative::argmax`; add a tie fixture (equal maxima at two indices, assert the lower index wins) that goes red under last-wins behavior.

**Scope**: one file, test-oracle only; no production path. Same pattern class as #806 — after fixing, grep for further `max_by`-on-logits siblings (the one known remaining hit is a MoE expert-routing test with a deliberately unique winner, not a tie-contract sibling).

ohdearquant:

Fixes #861.

## What

The f16-vs-f32 KV-cache logits oracle in `crates/inference/src/kv_cache/flat.rs` resolved exact top-1 ties with `Iterator::max_by`, which returns the *last* maximal index. The engine's documented greedy contract (shared `speculative::argmax`, the Metal sampler, and `torch.argmax` alike, per #280) is first-wins: the lowest index takes an exact tie. On a tied logit pair the oracle could therefore score a different top-1 token than production would emit, silently distorting `top1_match_rate`.

## Change

- New `oracle_top1` helper in the test module delegates to the shared first-wins `crate::speculative::argmax`, replacing both inlined `max_by` sites.
- New regression test `top1_oracle_tie_break_matches_first_wins_contract` constructs an exact tie between a low and a high index; it fails if the oracle regresses to last-wins (`max_by` picks index 4, the contract demands index 1). Verified mutation-sensitive: reverting the helper to `max_by` makes it go red.

Tests-only diff: every hunk is inside `mod tests` (`#[cfg(test)]`); production code is byte-identical to main.

## bench-compare

Attempted per ADR-058, but the runs are reported as environment evidence, not perf claims: a tests-only diff compiles to identical production code, so no runtime change is possible by construction.

An A/A control run on this host (same commit `48ae102034` as both base and head, exclusive heavy-job lock, machine otherwise idle, quick mode) still reported 4 FAIL / 7 WARN rows, e.g. `add_bias_gelu/4096 +17.9%`, `gelu/4096 +15.5%`, `simd_normalize/simd/384 +12.4%`, with every deviation positive (a systematic head-half bias). The quick-mode gate's false-positive floor on this host currently exceeds its own 7% threshold across groups beyond the two already allowlisted by #714. The A/B run for this PR showed the same class of impossible rows (up to +120% in groups this diff cannot touch) and is not interpretable as a regression signal.

### DIFF
--- a/crates/inference/src/kv_cache/flat.rs
+++ b/crates/inference/src/kv_cache/flat.rs
@@ -1292,6 +1292,19 @@ mod tests {
         }
     }
 
+    /// Top-1 token choice for the f16-vs-f32 KV-cache logits oracle below.
+    ///
+    /// Delegates to the shared first-wins `crate::speculative::argmax` so
+    /// the oracle's tie-break matches the engine's documented greedy
+    /// contract (lowest index wins on an exact tie — see
+    /// `forward::metal_qwen35`, `sampling`, and `torch.argmax`, all
+    /// first-wins per #280). `Iterator::max_by` resolves ties to the *last*
+    /// index instead, which let the oracle silently score a different top-1
+    /// token than production would pick on an exact tie (#861).
+    fn oracle_top1(logits: &[f32]) -> usize {
+        crate::speculative::argmax(logits)
+    }
+
     /// The mandatory tensor oracle: compare f32-KV reference vs f16-KV-via-FlatKVCache.
     ///
     /// Produces actual measured logit_max_abs_diff and top1_match_rate.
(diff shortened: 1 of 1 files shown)


ohdearquant:

REVIEW-VERDICT: APPROVE sha=48ae102034e883c355716622345665748bce6bfa

First-pass review, clean. Verified at source: the shared `speculative::argmax` is first-wins by strict `>` scan and the production Metal greedy paths use the same helper, so the oracle now matches the real convention rather than an arbitrary one; both former inlined `max_by` selectors are replaced with no executable `max_by` remaining in the file; the tie-fixture regression test genuinely fails under a last-wins revert (documented `Iterator::max_by` semantics) and has no NaN-ordering dependence; every hunk is inside the `#[cfg(test)]` module so production code is byte-identical to main. The bench table is correctly framed as environment evidence for a tests-only diff.

oceanwaves630 (approved):
Reviewed the on-PR verdict and the tests-only diff scope. Approving per the docs/tests-only lane.

PR #921, opened 2026-07-13
bench(scripts): widen quick-mode informational allowlist per isolated A/A evidence (#714)

ohdearquant:

Widens the #714 quick-mode informational-groups allowlist by three lattice-embed micro-bench groups, carrying the same-toolchain A/A evidence the allowlist's own contract requires.

## What

Adds `int8_batch_cosine`, `int4_cosine_distance`, and `simd_batch_cosine_non_normalized_query` to `INFO_GROUPS_ALLOWLIST` in `scripts/lib/bench-informational-groups.sh`. In `--quick` mode these groups are still measured and reported, but excluded from the FAIL/WARN gate and exit code (the existing #714 mechanism, unchanged). `--full` mode still gates every group normally, so a real regression in these groups remains catchable.

Also updates the `perf-bench-gate.py --selftest` fixture so the shell→Python allowlist handoff exercises the new names end-to-end (the selftest pins the emitted set exactly, so an array-only edit fails it), and refreshes the two comments that described the allowlist as two groups.

## Evidence

Two A/A runs (`scripts/bench-compare.sh HEAD HEAD`, quick mode, identical commit `c29e9a7f9d` as both base and head) were executed on 2026-07-13, each holding the machine-wide EXCLUSIVE bench-isolation lock (`/tmp/lion-bench-window.lock`) for the whole run, on an otherwise idle host with all concurrent builds routed through the SHARED side of the same lock. Since base and head are the same commit, every confirmed FAIL/WARN row is by construction a false positive.

Run 1 — 2 FAIL / 8 WARN / 40 "confirmed improvement" on identical code:

```
| int8_batch_cosine/float32_simd/100                        | +9.21% | [+8.32%, +10.11%] | FAIL |
| int4_cosine_distance/float32_simd/768                     | +8.18% | [+7.81%, +8.55%]  | FAIL |
| int8_vs_float32_cosine/float32_simd/1536                  | +6.27% | [+5.11%, +7.46%]  | WARN |
| int8_batch_cosine/float32_simd/1000                       | +4.59% | [+3.84%, +5.34%]  | WARN |
| residual_add_layer_norm/unfused/hidden384_seq128          | +3.97% | [+3.69%, +4.25%]  | WARN |
| residual_add_layer_norm/unfused/hidden768_seq128          | +4.65% | [+3.46%, +5.84%]  | WARN |
| residual_add_layer_norm/fused/hidden768_seq128            | +3.58% | [+3.41%, +3.75%]  | WARN |
| memory_size/search_1000_int8                              | +4.44% | [+3.38%, +5.50%]  | WARN |
| simd_normalize/simd/1024                                  | +3.29% | [+3.11%, +3.46%]  | WARN |
| simd_batch_cosine_non_normalized_query/simd_batch/384d_4c | +4.11% | [+3.00%, +5.22%]  | WARN |
```

Run 2 — 1 FAIL / 4 WARN / 76 "confirmed improvement" on identical code:

```
| simd_batch_cosine_non_normalized_query/pair_loop/1024d_64c | +7.48% | [+7.05%, +7.92%]  | FAIL |
| int8_raw_dot_product/dot_product_i8/127                    | +7.98% | [+5.42%, +10.53%] | WARN |
| simd_normalized_cosine_fast_path/cosine_full/768           | +7.07% | [+5.12%, +9.08%]  | WARN |
| simd_batch_cosine_normalized_query/pair_loop_dot/1024d_64c | +4.56% | [+3.93%, +5.21%]  | WARN |
| simd_squared_euclidean_fast_path/euclidean_full/384        | +5.26% | [+3.84%, +6.70%]  | WARN |
```

The decisive property: the FAIL sets of the two runs are **disjoint**. Identical commit, identical isolation, different groups exceeding the 7% gate with tight CIs — the signature of a noise floor above the quick-mode threshold on these sub-microsecond-to-tens-of-microseconds micro-benches, not of any code effect. The dozens of simultaneous "confirmed improvement" rows on identical code corroborate: the CI machinery is confidently resolving jitter in both directions.

## Scope of the widening

Only the groups that produced a confirmed-CI FAIL (>7%) in at least one isolated A/A run are added. WARN-only groups (`residual_add_layer_norm`, `memory_size`, `simd_normalize`, `int8_raw_dot_product`, `int8_vs_float32_cosine`, `simd_normalized_cosine_fast_path`, `simd_batch_cosine_normalized_query`, `simd_squared_euclidean_fast_path`) are NOT added: WARN does not block the gate, and keeping them gated preserves signal for genuine large regressions. If any of them later produces an isolated-A/A FAIL, that run's table is the evidence for a follow-up widening.

## Replay

```
bash <bench-window-wrapper> "<label>" -- scripts/bench-compare.sh HEAD HEAD
```

quick mode (default), any commit checked out at an otherwise idle host with builds routed through the shared lock side. Full run logs retained locally.

## Verification

- `scripts/perf-bench-gate.py --selftest` passes: the shell helper now emits exactly the five approved names against the extended listing fixture, the prefix-distinct decoy (`simd_dot_product_extra`), unapproved embed group, and inference group all still gate.
- No production kernel or bench code changes; scripts + comments only.

### DIFF
--- a/scripts/bench-compare.sh
+++ b/scripts/bench-compare.sh
@@ -16,11 +16,14 @@
 #   lattice-embed: simd
 # Uses a git worktree for the base ref so your working tree stays untouched.
 #
-# lattice#714: two of the lattice-embed `simd` bench target's groups
-# (simd_dot_product, simd_cosine_similarity) are confirmed noise-dominated in
-# --quick mode by a same-toolchain A/A reproduction on identical refs
-# (FAIL/WARN sign flipped across dozens of entries in those groups run to
-# run). In --quick mode (the default), only those two named groups are
+# lattice#714: five of the lattice-embed bench targets' groups
+# (simd_dot_product, simd_cosine_similarity, int8_batch_cosine,
+# int4_cosine_distance, simd_batch_cosine_non_normalized_query) are confirmed
+# noise-dominated in --quick mode by same-toolchain A/A reproductions on
+# identical refs (the original pair: FAIL/WARN sign flipped across dozens of
+# entries run to run; the 2026-07-13 additions: confirmed-CI FAIL rows inside
+# exclusive bench windows with DISJOINT failing groups across two same-commit
+# runs). In --quick mode (the default), only those named groups are
 # measured and reported but excluded from the FAIL/WARN gate and exit code —
 # see the "informational" section of the report and INFO_GROUPS_ALLOWLIST
 # below. Every other group in the target, including any added later, gates
--- a/scripts/bench-compare.sh
+++ b/scripts/bench-compare.sh
@@ -123,10 +123,11 @@ fi
 
 # --- Quick-mode informational-groups (lattice#714) ---
 # Fixed, reviewed allowlist of the issue-evidenced noisy groups only — NOT a
-# target-wide dump. lattice#714's quantitative reproduction (same-toolchain
-# A/A runs on identical refs, --quick mode) confined every flip-signed FAIL/
-# WARN to two lattice-embed `simd` groups; no other group in that target has
-# issue-backed noise evidence, so no other group is exempted here. The
+# target-wide dump. Every entry carries same-toolchain A/A evidence on
+# identical refs in --quick mode (lattice#714's original reproduction for
+# the first two; isolated bench-window A/A runs for the 2026-07-13
+# additions); no other group in that target has that evidence, so no other
+# group is exempted here. The
 # allowlist itself and the intersection-with-the-real-listing logic live in
 # scripts/lib/bench-informational-groups.sh, invoked below — kept in its own
 # file (rather than inline here) so scripts/perf-bench-gate.py --selftest can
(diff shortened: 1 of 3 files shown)


ohdearquant:

REVIEW-VERDICT: APPROVE sha=d5f55af0e737ea36fade7413be8adcbebc42a20c

Two review passes. First pass confirmed the three added names are exact Criterion top-level group names in the embed bench target, that exact-match intersection cannot admit prefix siblings, that full mode still gates every group, and that the same-commit isolated A/A evidence is a defensible admission criterion — and found one Medium: the selftest's fixed-listing intersection probe could not detect an allowlist-array-only addition. That gap is closed by a --print-allowlist probe comparing the raw shell array against the selftest's expectation set (verified mutation-sensitive: an array-only name fails the selftest). Second pass confirmed the debug mode cannot affect the production intersection path and left one documentation-only wording nit, fixed at head.

oceanwaves630 (approved):
Verified: codex verdict at head d5f55af0e, both A/A evidence tables in body per the traceability ruling, selftest mutation-hardened. Approving; your armed auto-merge fires on green.

Issue #615, opened 2026-07-03
arch(lora): extract shared adapter descriptor + blend validation — tune and Metal inference maintain drifting copies


ohdearquant:
> _Issue authored by Claude (Anthropic agent) on behalf of @ohdearquant._

## Problem

`tune` and Metal inference independently own LoRA adapter shape/blend/validation logic (descriptor structs, rank/alpha rules, blend caps, target-module validation), and the crate dependency direction prevents either from importing the other's — so every validation fix lands in one copy and drifts from the other. The mixture-runtime and governed-manifest work (#443/#444) each had to touch both sides.

## Proposal

Extract a minimal LoRA core descriptor + validation module into a leaf location both can depend on (either a small leaf crate or an inference-owned neutral module, preserving the leaf-crate dependency rules in AGENTS.md). Scope: adapter descriptor (rank, alpha, target modules, dtype), blend-weight validation caps, shape checks. NOT runtime kernels.

Pairs with #610 (manifest field completion + rev enforcement) — the descriptor is where those governance fields want to live once, not twice.

ohdearquant:
Status update at `b3ce9df2ee`. One of the two halves landed, and it is the smaller one.

**Landed.** The projection geometry function `qwen35_projection_shape` is shared: it lives in `crates/inference/src/lora_hook.rs` and `tune`'s validation uses it. That is a genuine cross-crate extraction.

**Not landed.** The adapter descriptor itself is still duplicated. `tune` defines `LoraConfig`, `LoraLayer` and `LoraAdapter` in `crates/tune/src/lora/mod.rs`; `inference` carries its own `LoraLayerData`. There is no shared descriptor type, and the two crates' blend paths are separate — `tune` has `blend_lora_adapters` with its own validation, `inference` has its own blend.

So the shape is: geometry shared, types and validation not. Since a shared descriptor is what would make the two blend paths verifiably agree, and geometry is the part that was easiest to share, this is roughly the reverse of the order that would retire the issue's actual risk.

PR #927, opened 2026-07-13
fix(inference): chat_metal LoRA loader — accept mlp blocks, fail closed on incomplete/mixed-rank adapters

### LINKED ISSUE #748: bug(inference): chat_metal LoRA loader drops mlp adapter tensors (gate/up/down_proj)
**Defect**: The inline PEFT/MLX adapter parser used by the `chat_metal` binary only recognizes keys with `parts[1] == "self_attn"`, silently ignoring any `mlp.gate_proj`/`up_proj`/`down_proj` tensor block, even though (a) the canonical `tune` loader documents and accepts both `self_attn` and `mlp` blocks, and (b) the Metal forward path itself validates and dispatches `gate_proj`/`up_proj`/`down_proj` descriptors. A valid adapter file with paired `q_proj` and `gate_proj` tensors loads both blocks through the canonical loader, but `chat_metal` loads `q_proj` only — the non-zero MLP delta is completely absent from every adapted layer and every generated token in `chat_metal`, while present via the CPU/canonical path.

**Evidence**:
- `crates/inference/src/bin/chat_metal.rs:188` — PEFT parser requires `parts[1] == "self_attn"`; the MLX branch repeats the same restriction at `:211`.
- `crates/tune/src/lora/safetensors.rs:43` — canonical parser documents and accepts both `self_attn` and `mlp` blocks, extracting the final module name at line 81.
- `crates/inference/src/forward/metal_qwen35.rs:6916` — Metal descriptor validation accepts `gate_proj`/`up_proj`/`down_proj` and dispatches them (lines 12123-12160).

**Blast radius**: Every `chat_metal --lora` generation using an adapter with MLP factors — all such deltas are silently dropped at every adapted layer and every generated token, even though Metal itself is capable of applying them.

**Suggested fix**: Delete the inline parser and route `chat_metal` through the shared PEFT/MLX descriptor loader (`crates/tune/src/lora/safetensors.rs`) so it preserves every block the format and Metal forward path both support.

### LINKED ISSUE #749: bug(inference): chat_metal warns and continues on incomplete/mixed-rank LoRA adapters instead of rejecting
**Defect**: For the same malformed LoRA adapter, the canonical `tune` loader hard-rejects the file, while `chat_metal`'s inline parser logs a warning and continues with a partial or incorrectly-scaled adapter. A missing `A` or `B` factor is logged and skipped (the remaining, incomplete layer set is still returned and used); a cross-layer rank mismatch only warns, and the global LoRA scale is derived from whichever rank was seen first rather than being rejected as an inconsistent descriptor.

**Evidence**:
- `crates/inference/src/bin/chat_metal.rs:202` — missing `A` or `B` factor: logs and `continue`s, non-empty remaining layer set still returned.
- `crates/inference/src/bin/chat_metal.rs:257` — same continue-on-incomplete-pair pattern.
- `crates/inference/src/bin/chat_metal.rs:328` — cross-layer rank mismatch only warns ("using first"); the first-seen rank's scale is applied globally.
- Contrast: `crates/tune/src/lora/safetensors.rs:274` — canonical loader errors on a missing matching `B`; `:306` errors on rank inconsistency; `:358-365` errors on an orphan `B`. All four conditions are hard `Err` returns in the canonical loader.

**Blast radius**: Every token generated after `chat_metal` accepts a file containing at least one valid adapter pair plus an orphan or mixed-rank projection — the partial or wrongly-scaled delta propagates through all subsequent layers, where the canonical loader would have refused to load the file at all.

**Suggested fix**: Make missing pairs and cross-layer rank disagreement hard errors in `chat_metal`'s parser (ideally by sharing the canonical descriptor parser), rather than relying on stderr warnings for what is a correctness gate.

ohdearquant:

Fixes #748, fixes #749.

## What

Two defects in `chat_metal`'s inline LoRA adapter loader (`load_lora_safetensors`), one loader:

1. **MLP adapter tensors were silently dropped** (#748). Both the PEFT and MLX key parsers hard-required `self_attn` as the block segment, so `mlp.gate_proj`/`up_proj`/`down_proj` tensors in a valid adapter never reached the model — the MLP delta was absent from every generated token while present via the canonical loader. Both parsers now accept `self_attn | mlp` and keep emitting the bare module names the Metal descriptor path expects.
2. **The loader failed open on malformed adapters** (#749). A missing A or B factor was logged and skipped (partial adapter still used); a cross-layer rank mismatch warned "using first" and derived the global scale from whichever rank was seen first. Both now hard-error: the orphan-tensor error names the offending tensor, and mixed ranks are rejected as inconsistent. The pre-existing all-groups-incomplete error is preserved.

## Tests (mutation-verified)

Three new unit tests over `load_lora_safetensors`, each run against the unfixed behavior first and observed to fail, then pass with the fix:
- `lora_loader_accepts_attention_and_mlp_pairs` (PEFT + MLX fixtures; failed pre-fix with only `q_proj` loaded)
- `lora_loader_rejects_missing_b_with_valid_pair_present` (failed pre-fix: warn-and-continue returned success)
- `lora_loader_rejects_mixed_ranks` (failed pre-fix: "using first" returned success)

`cargo test -p lattice-inference --features f16,metal-gpu --bin chat_metal`: 9/9. `cargo clippy --workspace -- -D warnings` clean. Sibling-parser sweep: repo-wide grep for the two inline key prefixes and the `parts[1]` predicate matches only the two corrected branches; the canonical loader in `lattice-tune` already accepts both block types (structural dedup of the two loaders remains tracked in #615 — the dependency direction is why the inline copy exists).

## bench-compare

Quick-mode A/B (origin/main vs HEAD, exclusive bench window): 1 FAIL (`int8_prepared_dot_product/per_call/384` +42.75%), 12 WARN, 77 WIN. The diff touches only the `chat_metal` binary's adapter-loading function — no library, kernel, or embed code — so the flagged int8 embed kernel is structurally unreachable from this change; the pattern matches the documented quick-mode false-positive floor (same-commit A/A runs produce disjoint confirmed-CI failures). Full table available on request.

### DIFF
--- a/crates/inference/src/bin/chat_metal.rs
+++ b/crates/inference/src/bin/chat_metal.rs
@@ -137,6 +137,15 @@ fn read_lora_alpha_from_metadata(path: &std::path::Path) -> Option<f32> {
 }
 
 #[cfg(all(target_os = "macos", feature = "metal-gpu"))]
+fn is_supported_lora_target(block: &str, module: &str) -> bool {
+    match block {
+        "self_attn" => true,
+        "mlp" => matches!(module, "gate_proj" | "up_proj" | "down_proj"),
+        _ => false,
+    }
+}
+
+#[cfg(all(target_os = "macos", feature = "metal-gpu"))]
 fn load_lora_safetensors(
     path: &std::path::Path,
 ) -> Result<
@@ -185,7 +194,7 @@ fn load_lora_safetensors(
             // parts: [i, "self_attn" | "mlp", module, "lora_A" | "lora_B", "weight"]
             if parts.len() >= 4
                 && let Ok(layer_idx) = parts[0].parse::<usize>()
-                && matches!(parts[1], "self_attn" | "mlp")
+                && is_supported_lora_target(parts[1], parts[2])
             {
                 let module = parts[2].to_string();
                 let is_a = parts[3] == "lora_A";
@@ -208,7 +217,7 @@ fn load_lora_safetensors(
             // parts: [i, "self_attn" | "mlp", module, "lora_a" | "lora_b"]
             if parts.len() >= 4
                 && let Ok(layer_idx) = parts[0].parse::<usize>()
-                && matches!(parts[1], "self_attn" | "mlp")
+                && is_supported_lora_target(parts[1], parts[2])
             {
                 let module = parts[2].to_string();
                 let is_a = parts[3] == "lora_a";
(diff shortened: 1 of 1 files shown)


ohdearquant:

Independent review across two rounds. Round 1 confirmed the parser fix (mlp blocks accepted, module names aligned with the Metal descriptor contract) and the fail-closed conversion of the incomplete-pair and mixed-rank paths, with orientation-consistent rank comparison between PEFT and MLX layouts, but flagged one Medium: the widened mlp acceptance would have turned benign unknown-mlp tensor pairs from silently ignored into whole-adapter startup errors. The fix restricts mlp parsing to the known projection modules through a shared predicate in both branches, pinned by mutation-verified PEFT and MLX regression fixtures (each observed failing pre-fix). All five loader regression tests plus the full target suite pass; clippy and formatting clean.

REVIEW-VERDICT: APPROVE sha=75c3342cae03dbd28c9fa3bb8509130c3cb0e1df

oceanwaves630 (approved):
Release: two-round independent review (round-1 AWF finding fixed with module whitelist + mutation-verified fixtures, round-2 clean APPROVE at exact head 75c3342c), all required checks green.

PR #928, opened 2026-07-13
fix(tune): reject non-finite LoRA alpha at parse, config, and manifest boundaries

### LINKED ISSUE #750: bug(tune): non-finite LoRA alpha is accepted by CPU and manifest paths, rejected only by Metal
**Defect**: The safetensors metadata-alpha parser accepts any value `f32::from_str` can parse, which includes `"nan"`/`"inf"`/`"-inf"` (case-insensitive) — there is no `is_finite()` gate anywhere in the parse path, and the computed effective scale (`alpha / rank`) is returned without a finiteness check either. The governed-manifest tolerance check then compares `(adapter.alpha - entry.alpha).abs() > tolerance`, which is always `false` when `adapter.alpha` is NaN (standard IEEE-754 semantics), so a NaN-alpha adapter passes the manifest gate even against a manifest declaring a finite alpha. Metal's loader is the only path that explicitly rejects `!scale.is_finite()` before accepting an adapter.

**Evidence**:
- `crates/tune/src/lora/safetensors.rs:374-382` — `alpha` derived via `.and_then(|s| s.parse::<f32>().ok())`, no `is_finite()` gate.
- `crates/tune/src/lora/mod.rs:81-89` (`LoraConfig::scale`) — `alpha / rank as f32` with no finiteness check.
- `crates/tune/src/lora/apply.rs:49` — `*out += scale * acc`; a NaN scale contaminates every adapted output.
- `crates/tune/src/lora/loader.rs:290` — manifest check does `if (adapter.config.alpha - entry.alpha).abs() > 1e-4 { return Err(...) }`; `NaN - x = NaN`, `NaN.abs() = NaN`, `NaN > 1e-4` is `false`, so the comparison silently does not trigger.
- `crates/inference/src/forward/metal_qwen35.rs:6980-6983` — Metal's `load_lora_adapter` explicitly checks `if !scale.is_finite() { return Err(...) }` before any tensor is touched.

**Blast radius**: All adapted CPU Qwen or BERT projections after loading an adapter whose metadata alpha is non-finite — the NaN propagates into attention, logits, sampling, or cross-encoder scores. The governed manifest path accepts the same value; Metal is the only path that rejects it.

**Suggested fix**: Centralize adapter-descriptor preflight (rank > 0, finite alpha/effective scale, consistent ranks, checked factor lengths, non-empty layers) and invoke it from every load/construct/attach path; explicitly check `is_finite()` before the manifest tolerance comparison rather than relying on ordered-comparison semantics with NaN.

ohdearquant:

Fixes #750.

## What

Non-finite LoRA alpha (`"nan"`/`"inf"`/`"-inf"` in safetensors metadata, CLI flags, or programmatic construction) was accepted by every CPU-side boundary — the metadata parser had no `is_finite()` gate, `LoraConfig::scale()` returned NaN/Inf to arithmetic callers, and the governed-manifest tolerance check `(adapter.alpha - entry.alpha).abs() > tol` is always false for NaN, so a NaN-alpha adapter passed the manifest gate against a finite declared alpha. Only the Metal attach path rejected non-finite scale.

Now fail-closed at each boundary:
- `safetensors.rs`: absent alpha keeps the documented `alpha = rank` fallback; present-but-malformed or non-finite alpha is a descriptive loader error (and non-finite alpha is rejected on save, so invalid metadata cannot be emitted).
- `mod.rs`: new `LoraConfig::validate()` (finite alpha + finite effective scale). `LoraAdapter::new` validates and returns `Result`, and the adapter's `config`/`layers` state is now private with read-only accessors, so an invalid adapter cannot be literal-constructed or mutated into existence — it is rejected before reaching the infallible `apply`/`LoraHook`, blend, gradient, or online-update paths. `scale()` fails closed to `0.0` instead of returning NaN/Inf (consistent with its existing `rank == 0` behavior).
- `loader.rs`: manifest alpha comparison rejects non-finite adapter alpha, manifest alpha, or difference before applying the `1e-4` tolerance.
- `generate_lora.rs`: validates the adapter before reporting or attaching it.
- `train_grad.rs` / `train_grad_layer23.rs`: `--alpha` parsing rejects non-finite and unparseable values, and a present `--alpha` with a missing value is an explicit error; the `16.0` default applies only when the flag is truly absent.

## Tests (mutation-verified)

Each regression was verified to fail with its specific guard reverted: non-finite metadata rejection (nan/inf/-inf), NaN-alpha adapter vs finite manifest rejection, config validation + fail-closed scale, checked-constructor rejection, two compile-fail doctests pinning that `LoraAdapter` cannot be literal-constructed or field-mutated (both fail when field privacy is reverted), and per-binary CLI parser rejections including the bare-`--alpha` case. The finite-alpha round-trip and the absent-metadata `alpha = rank` fallback were already covered by pre-existing tests, which stay green throughout.

`cargo test -p lattice-tune` across default, `safetensors,serde,inference-hook`, and `train-backward` feature sets: all green. `cargo clippy --workspace -- -D warnings` clean.

No bench-compare: `lattice-tune` only — no `crates/inference/` or `crates/embed/` code touched (the two `Cargo.toml` version lines in inference/embed are dependency-version metadata, not code).

## Version

Hiding `LoraAdapter`'s previously public `config`/`layers` fields is a semver-major change against the published 0.6.1 crates (flagged by cargo-semver-checks), so this PR bumps the workspace and internal path-dependency versions to 0.7.0 in lockstep, opening the 0.7.0 line.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ members = [
 ]
 
 [workspace.package]
-version = "0.6.1"
+version = "0.7.0"
 edition = "2024"
 # MSRV, verified end-to-end (workspace + deps) on 1.93.1 and guarded by the msrv CI
 # job (aarch64 macOS, f16,metal-gpu, --all-targets) so a newer-toolchain-only API
--- a/crates/embed/Cargo.toml
+++ b/crates/embed/Cargo.toml
@@ -55,14 +55,14 @@ tracing = "0.1"
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 tokio = { workspace = true, features = ["sync", "rt"] }
 # Embedding generation (native, pure Rust)
-lattice-inference = { version = "0.6.1", path = "../inference", optional = true, features = ["f16"] }
+lattice-inference = { version = "0.7.0", path = "../inference", optional = true, features = ["f16"] }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 tokio = { version = "1.0", default-features = false, features = ["sync", "rt"] }
 # Same crate, trimmed to the CPU compute path: `std` + `f16`, without
 # `download` (ureq) or `serve` (axum/tokio-net/mio), none of which compile to
 # wasm32-unknown-unknown.
-lattice-inference = { version = "0.6.1", path = "../inference", optional = true, default-features = false, features = [
+lattice-inference = { version = "0.7.0", path = "../inference", optional = true, default-features = false, features = [
     "std",
     "f16",
 ] }
--- a/crates/inference/Cargo.toml
+++ b/crates/inference/Cargo.toml
@@ -71,7 +71,7 @@ bytemuck = { version = "1", features = ["derive"], optional = true }
 ort = { version = "2.0.0-rc.12", optional = true }
 tokenizers = { version = "0.22", optional = true, default-features = false, features = ["progressbar", "onig"] }
 image = { version = "0.25", default-features = false, features = ["png", "jpeg"] }
-lattice-fann = { version = "0.6.1", path = "../fann", optional = true }
+lattice-fann = { version = "0.7.0", path = "../fann", optional = true }
 
 [[bin]]
 name = "lattice"
--- a/crates/tune/Cargo.toml
+++ b/crates/tune/Cargo.toml
@@ -30,7 +30,7 @@ sqlite = ["dep:rusqlite", "serde"]
 mixture = ["lattice-fann/online-router"]
 
 [dependencies]
-lattice-fann = { version = "0.6.1", path = "../fann" }
+lattice-fann = { version = "0.7.0", path = "../fann" }
 thiserror.workspace = true
 serde = { workspace = true, optional = true }
 serde_json = { workspace = true, optional = true }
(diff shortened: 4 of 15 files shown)


ohdearquant:

Independent review across three rounds. Round 1 confirmed the safetensors and manifest paths reject non-finite alpha but found two gaps: programmatic construction could still yield a silently disabled adapter through the fail-closed zero scale, and both gradient-training CLIs parsed non-finite `--alpha` values into arithmetic. Round 2 verified the checked-constructor fix and CLI guards but found the public `config`/`layers` fields still allowed literal construction and post-construction mutation to bypass validation, and that a bare `--alpha` with no value silently fell back to the default. Round 3 confirmed both resolved: adapter state is private with read-only accessors, all production construction routes through the validating constructor, two compile-fail doctests pin the contract (each observed failing when field privacy was reverted), and both CLIs now error on a present `--alpha` with a missing value, with mutation-verified regressions. Full tune suite passes across default, safetensors/serde/inference-hook, and train-backward feature sets; clippy and formatting clean.

The final commit on top of the reviewed head is a version-only change: hiding the previously public fields is a semver-major change against the published 0.6.1 crates, correctly flagged by cargo-semver-checks, so the workspace and internal path-dependency versions open the 0.7.0 line in lockstep.

REVIEW-VERDICT: APPROVE sha=fb07d6cb7c9cb4907a94c83084b9dd025383703f

oceanwaves630 (approved):
Release: three-round review arc (constructor-boundary validation, private state + read-only accessors, compile-fail contract doctests, CLI parse regressions), round-3 APPROVE at 8c2f3787; post-review delta to fb07d6cb independently verified as version strings only (4 Cargo.tomls, 6 lines) — the 0.7.0 bump is forced by semver-checks since the breaking shape is the fix.

PR #929, opened 2026-07-13
fix(inference): grammar/tokenizer correctness — detok finalization on EOS/grammar stop, SentencePiece ID+added-token guards, exact grammar byte tables

### LINKED ISSUE #751: bug(inference): non-streaming detokenizer drops incomplete UTF-8 tail on EOS/grammar stop
**Defect**: The non-streaming generation path only finalizes the incremental byte-level detokenizer when `stopped` is `false`, on the stated rationale that a stop-string match already truncated the output buffer and appending tail bytes past the match point would corrupt it. That rationale only applies to the stop-string case, but `stopped` is also set `true` uniformly by EOS and grammar-driven completion — cases where the output buffer was never truncated, so the same skip still applies and drops a legitimate incomplete-UTF-8 trailing byte sequence that the streaming path (which calls `finish()` unconditionally) does emit as a replacement character. This only surfaces when EOS/grammar termination lands on an incomplete multi-byte UTF-8 boundary, not on every `stop_strings` request.

**Evidence**:
- `crates/inference/src/model/qwen35/generation.rs:1145-1148` — `if !stopped { let tail = detok.finish(); ... }`, comment: "Only flush detok tail when no stop was hit. A stop already truncated `full` to the correct position; appending tail bytes from past the stop would corrupt it."
- `crates/inference/src/model/qwen35/generation.rs:1077,1088-1089,1133-1134` — `stopped` is set `true` uniformly by EOS and grammar completion, not only by a stop-string match.
- `crates/inference/src/model/qwen35/generation.rs:673` — the streaming string-stop path calls `detok.finish()` unconditionally, confirming the asymmetry.
- `crates/inference/src/model/qwen35/detokenize.rs:140` — `finish()` emits a lossy (`String::from_utf8_lossy`-based) trailing tail for an incomplete UTF-8 sequence.

**Blast radius**: Public Qwen3.5 CPU `generate` and `generate_streaming` disagree for any request that terminates via EOS or grammar completion (not just `stop_strings`) when the byte-level BPE stream ends mid-codepoint — non-streaming output silently omits the incomplete tail while streaming output contains a `U+FFFD` replacement character for the same generation.

**Suggested fix**: Track "confirmed stop-string match" separately from "EOS/grammar/length termination" and always finalize the detokenizer for the latter category; centralize the finalize-vs-skip decision so both call sites share the same rule.

### LINKED ISSUE #772: audit(tokenizers): 4 medium correctness findings — self-audit 20260709
- [ ] **Reject overflowing SentencePiece special IDs before narrowing**
  All tokenizer special IDs should reject u64 values outside the downstream u32 domain.
  Evidence: `crates/inference/src/tokenizer/sentencepiece.rs:262`. Value 4294967296 becomes ID 0 and passes a two-piece model's range check; unknown text then emits 0 instead of loader rejection.
  Suggested fix: Introduce one checked special-ID conversion helper and call it in both SentencePiece parsers before construction.

- [ ] **Honor SentencePiece added tokens before normalization**
  Every tokenizer.json loader should honor top-level added_tokens before ordinary tokenization.
  Evidence: `crates/inference/src/tokenizer/sentencepiece.rs:167`. For a Unigram JSON with added token <extra> at ID 2 but no base-vocab piece for it, input <extra> cannot emit ID 2 in SentencePiece, unlike sibling loader semantics.
  Suggested fix: Apply shared added-token matching before SentencePiece normalization and add a fixture whose added token is absent from model.vocab.

- [ ] **Use exact raw BPE bytes in grammar vocabularies** (re-verified at 0699e60cc)
  The grammar vocabulary and per-token byte decoder nominally convert BPE placeholder characters back to raw bytes.
  Evidence: `crates/inference/src/tokenizer/bpe.rs:479`. The placeholder for raw byte E5 yields EF BF BD via vocab_bytes rather than E5, so a grammar can reject a valid first byte of a CJK/emoji sequence.
  Suggested fix: Make vocab_bytes map each base token through byte_decode_token_bytes and add a split-UTF-8 grammar regression.

- [ ] **Include added-token IDs in BPE grammar vocabularies** (re-verified at 0699e60cc)
  The token-ID space exposed to BPE decoding and grammar masking should be identical.
  Evidence: `crates/inference/src/tokenizer/bpe.rs:479`. With base IDs 0..1 and renderable added ID 100, decoding ID 100 returns literal text but a grammar built from vocab_bytes has length 2 and leaves logit 100 untouched.
  Suggested fix: Construct a dense model-sized grammar byte table with the same ID resolver as decode and an explicit treatment for skipped controls.

---

ohdearquant:

Fixes #751. Closes #772.

## What

Five correctness fixes across the detokenizer termination path and the tokenizer/grammar boundary:

1. **Non-streaming detokenizer finalization (#751)**: `generate` only flushed the incremental detokenizer's incomplete-UTF-8 tail when `stopped` was false, but `stopped` is set uniformly by stop-string match, EOS, and grammar completion. Only a confirmed stop-string match truncates the output buffer, so EOS/grammar/length termination silently dropped a legitimate trailing byte sequence that `generate_streaming` (which finalizes unconditionally) emitted as a replacement character — the two public paths disagreed for the same generation. Now a confirmed stop-string match is tracked separately from other termination causes, and both call sites share one finalization predicate: stop-string match skips the tail, everything else flushes it.
2. **SentencePiece special-ID overflow**: u64 special-token IDs outside the u32 domain wrapped silently (4294967296 became ID 0) and passed downstream range checks. Both the JSON Unigram and native protobuf loaders now reject out-of-range IDs through one checked conversion helper covering `unk_id`, `bos_id`, `eos_id`, and `pad_id`.
3. **SentencePiece added tokens**: top-level `added_tokens` in a Unigram `tokenizer.json` were not honored before normalization, so an added token absent from the base vocab could never emit its ID — unlike the sibling BPE/WordPiece loaders. Added-token matching now runs against the original input before normalization, longest-first and deterministic.
4. **Grammar vocabulary raw bytes**: `vocab_bytes` decoded BPE byte-placeholder characters through a lossy path, yielding `EF BF BD` (U+FFFD) instead of the raw byte, so a grammar could reject the valid first byte of a CJK/emoji sequence split across tokens. Entries now map through the same exact byte decoding as `decode`.
5. **Grammar vocabulary added-token coverage**: added-token IDs were absent from the grammar byte table while `decode` rendered them, leaving their logits unmasked. The table is now built over the same dense model-sized ID space as decoding, with empty slots for skipped control tokens so they stay maskable.

## Tests (mutation-verified)

Six new regressions, each observed failing against the unfixed code before the fix landed and passing after: EOS-termination streaming/non-streaming parity on a mid-codepoint boundary, overflowing special IDs rejected from both JSON and protobuf, added-token round-trip through the Unigram loader, split-UTF-8 grammar acceptance, and grammar masking of rendered added tokens and control slots.

Sibling-path sweep: both string-stop loops now share the finalization helper; the Metal generation paths were checked and already handle finalization correctly per termination cause (no change needed); the checked ID helper covers both SentencePiece parsers; BPE and WordPiece added-token paths already matched pre-normalization; `vocab_bytes` is the single production grammar-vocabulary constructor and all production grammar clients consume it.

`cargo test -p lattice-inference` (default and `--features f16`): all green (1,795 and 1,800 library tests respectively, zero failures across binary/integration/doctest suites). `cargo clippy --workspace -- -D warnings` and `cargo fmt --all -- --check` clean.

## bench-compare (ADR-058)

Methodology: `make bench-compare` (quick mode) at base `origin/main` 77316a1005 vs head, then a second run filtered to the failing groups plus `rms_norm` to test reproducibility. Runnable commands and full tables are in the two runs' output below.

The only benchmark group on code this diff touches, `rms_norm`, is at parity (-0.37%, p=0.05). All FAIL rows in both runs are `lattice-embed` SIMD kernels; this diff changes three `crates/inference` tokenizer/generation files and `git diff origin/main...HEAD -- crates/embed/` is empty, so both sides of those benches build from byte-identical source — any measured delta there is harness/machine noise by construction. Consistent with that, the FAIL sets do not reproduce between the two runs: the four `simd_normalize` FAILs from run 1 (up to +15.6%) disappeared in run 2, and remaining magnitudes shifted well outside their run-1 confidence intervals. Known confound: this machine runs concurrent compile load from other work; the bench-window lock serializes benches but not all background compilation (documented false-positive class).

<details><summary>Run 1 (full default groups) and run 2 (filtered recheck) FAIL tables</summary>

Run 1: 10 FAIL, all lattice-embed SIMD groups (simd_normalize ×4 up to +15.6%, int8_vs_float32_cosine ×3, simd_batch ×2, binary_cosine_distance ×1). Run 2 (same base/head, filtered): simd_normalize FAILs absent; int8_vs_float32_cosine/384 moved from +19.8% [CI +19.0,+20.7] to +12.6% [CI +9.3,+15.9]. rms_norm: -0.37% [CI -0.51,-0.23].

</details>

### DIFF
--- a/crates/inference/src/bin/lattice.rs
+++ b/crates/inference/src/bin/lattice.rs
@@ -3353,13 +3353,8 @@ mod serve {
         tokenizer: &lattice_inference::tokenizer::bpe::BpeTokenizer,
         token_id: u32,
     ) -> (String, Option<Vec<u8>>) {
-        match tokenizer.token_for_id(token_id) {
-            Some(tok_str) => (
-                lattice_inference::tokenizer::bpe::byte_decode_token(tok_str),
-                Some(lattice_inference::tokenizer::bpe::byte_decode_token_bytes(
-                    tok_str,
-                )),
-            ),
+        match tokenizer.token_bytes_for_id(token_id) {
+            Some(bytes) => (String::from_utf8_lossy(&bytes).into_owned(), Some(bytes)),
             None => (format!("<|unresolved_token_{token_id}|>"), None),
         }
     }
--- a/crates/inference/src/bin/qwen35_debug.rs
+++ b/crates/inference/src/bin/qwen35_debug.rs
@@ -1,7 +1,15 @@
 //! Debug diagnostic for Qwen3.5-2B — checks each stage of the forward pass.
 
+use lattice_inference::tokenizer::bpe::BpeTokenizer;
 use lattice_inference::tokenizer::common::Tokenizer;
 use std::path::PathBuf;
+
+fn decode_token(tokenizer: &BpeTokenizer, token_id: u32) -> String {
+    tokenizer
+        .token_bytes_for_id(token_id)
+        .map(|bytes| String::from_utf8_lossy(&bytes).into_owned())
+        .unwrap_or_else(|| "??".to_string())
+}
 
 fn main() {
     let model_dir = std::env::var("LATTICE_MODEL_CACHE")
@@ -45,8 +53,7 @@ fn main() {
 
             // Decode each token
             for &tid in &output.token_ids {
-                let tok = tokenizer.token_for_id(tid).unwrap_or("??");
-                let decoded = lattice_inference::tokenizer::bpe::byte_decode_token(tok);
+                let decoded = decode_token(tokenizer, tid);
                 println!("  {tid} -> {decoded:?}");
             }
         }
@@ -72,8 +79,7 @@ fn main() {
     si.sort_by(|a, b| b.1.partial_cmp(&a.1).unwrap_or(std::cmp::Ordering::Equal));
     println!("Top-5:");
     for (i, (tok, logit)) in si.iter().take(5).enumerate() {
-        let name = tokenizer.token_for_id(*tok as u32).unwrap_or("??");
-        let decoded = lattice_inference::tokenizer::bpe::byte_decode_token(name);
+        let decoded = decode_token(tokenizer, *tok as u32);
         println!("  {i}: token_id={tok}, logit={logit:.4}, decoded={decoded:?}");
     }
     println!("Python: 2614(' following')=12.69, 220(' ')=11.50, 7193(' purpose')=11.13");
@@ -95,8 +101,7 @@ fn main() {
     println!("\nTop-10 tokens:");
     println!("Python top: 11751(' Paris')=18.875, 279(' the')=16.5, 264(' a')=16.25");
     for (i, (tok, logit)) in indexed.iter().take(10).enumerate() {
-        let name = tokenizer.token_for_id(*tok as u32).unwrap_or("??");
-        let decoded = lattice_inference::tokenizer::bpe::byte_decode_token(name);
+        let decoded = decode_token(tokenizer, *tok as u32);
         println!("  {i}: token_id={tok}, logit={logit:.4}, decoded={decoded:?}");
     }
 
--- a/crates/inference/src/forward/metal_qwen35.rs
+++ b/crates/inference/src/forward/metal_qwen35.rs
@@ -13102,10 +13102,7 @@ mod inner {
     }
 
     fn decode_tokens(tokenizer: &BpeTokenizer, ids: &[u32]) -> String {
-        ids.iter()
-            .filter_map(|id| tokenizer.token_for_id(*id))
-            .map(crate::tokenizer::bpe::byte_decode_token)
-            .collect()
+        crate::model::qwen35::detokenize::decode_tokens(tokenizer, ids)
     }
 
     // -----------------------------------------------------------------------
--- a/crates/inference/src/grammar/engine.rs
+++ b/crates/inference/src/grammar/engine.rs
@@ -143,7 +143,7 @@ impl GrammarEngine {
     /// Build a `GrammarEngine` from a `GrammarSpec` and the model vocabulary.
     ///
     /// `vocab_bytes[i]` is the UTF-8 / byte-level representation of token `i`.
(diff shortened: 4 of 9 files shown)


ohdearquant:

Independent review across two rounds. Round 1 confirmed the five fixes (detokenizer finalization split by termination cause with a shared predicate, checked SentencePiece special-ID conversion in both loaders, pre-normalization added-token matching, exact grammar byte recovery, dense grammar table coverage) but found two majors introduced at the grammar boundary: the byte table was sized to the tokenizer's highest ID rather than the model's logit space, leaving a 250-logit fail-open tail on Qwen3.5 that could terminate a valid generation early, and literal non-ASCII added tokens were routed through the GPT-2 placeholder decoder, yielding empty grammar entries and inconsistent decoded text. Round 2 verified both resolved: the table API now takes the serving model's vocabulary size, errors when it cannot represent the highest tokenizer ID, and fails closed across the full logit space (mutation-verified against the official Qwen3.5-0.8B vocabulary bounds); added tokens contribute literal UTF-8 through one shared resolver used by grammar construction, decode, full and incremental detokenization, Metal decoding, and logprob rendering, with CJK/accented regressions covering masking and emitted text. Full inference suite green on default and f16 feature sets; clippy and formatting clean.

REVIEW-VERDICT: APPROVE sha=edd395737d4e0a0dfcb7b769fe53f94080b03aa4

oceanwaves630 (approved):
Verified at edd395737: on-PR round-2 APPROVE at this exact head, 28 checks green (2 path-conditioned audit skips), fail-closed vocab table + shared added-token resolver confirmed in the review trail.

PR #938, opened 2026-07-13
refactor(scripts): consolidate decode-bench methodology into one profiled harness (#813)

ohdearquant:

Progresses #813 (three of the four remaining consumers; `bench_compare_1k.py` deliberately deferred to its own PR per the issue's sequencing — it has three genuinely different per-engine measured shapes).

## What

Migrates `bench_apples_precise.sh`, `bench_q4_apples.sh`, and `bench_context_scaling.sh` onto the shared `bench_decode_harness.py` established by the earlier #857/#913 migrations, following the same one-adapter-module-per-consumer pattern:

- `apples_precise` profile: N1=64/N2=512, 13 measured + 2 warmup per engine, trimmed-mean aggregation (trim=2)
- `q4_apples` profile: N1=32/N2=256, 5 runs, MLX-only warmup
- `context_scaling` profile: N1=8 baseline vs contexts {64,128,256}; adapter renders the legacy chart-compatible TSV, still invokes the unmodified chart script, and forwards the legacy `CONTEXTS`/`RUNS`/`CHART_ONLY` env vars
- The three legacy shell scripts become thin `uv run` wrappers over their profiles (one release of grace)
- All four adapter test files wired into CI's `bench-decode-harness-tests` job (previously only the harness-core test ran — a pre-existing gap)

## Disclosed correction: the legacy Q4 rows were not Q4

`bench_q4_apples.sh` exported `BENCH_Q4_DIR`/`BENCH_TOKENIZER_DIR` — names `bench_decode_ab.rs` never reads (it reads `LATTICE_MODEL_DIR`/`LATTICE_TOKENIZER_DIR`). The binary silently fell back to its default model dir, so the script's "Lattice Q4" rows benchmarked whatever that dir held (currently the f16 safetensors checkpoint). Verified at source: no code in the repo reads `BENCH_Q4_DIR`. Blast-radius sweep: zero durable citations of this script's numbers anywhere (issues, PRs, ADRs, docs, reports) — the only mention is a script-name listing in the v0.2.3 release notes.

The fix is two-layered:
1. The adapter wires the real env-var names to the actual Q4 directory (regression-tested).
2. The adapter pins **loaded-model identity**, not just env wiring: it mirrors `detect_format`'s precedence (safetensors markers win over `.q4` files) and refuses to run when the resolved dir would not load as Q4. Two negative tests (safetensors-shaped dir, no-`.q4`-files dir); mutation-verified — removing the guard call fails both.

## Tests

49 new deterministic tests across three files (profile-parameter equivalence vs the legacy scripts' constants, fake-adapter call-schedule equivalence, schema validation, trimmed-mean aggregation mutation-verified, model-identity guards). All pass via `uv run python -m pytest` and standalone. Pre-existing harness tests (56+) unchanged and green. `shellcheck` clean on the shims.

Live cross-engine smoke was attempted under the machine bench-window lock but aborted honestly — the lock was contended by 9 processes; the harness's engine adapters are exercised by the fake-seam tests, and a live run can be done post-merge with the documented one-liner in each adapter's docstring.

bench-compare: not applicable — no `crates/` code changes (scripts, tests, and CI wiring only; the diff touches `.github/workflows/ci.yml`, `scripts/`, `tests/`).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -542,9 +542,17 @@ jobs:
           version: '1.7.12'
 
   # Deterministic fake-adapter unit tests + raw-observation schema validation
-  # for the decode-benchmark harness core (issue #813 step 1). Stdlib-only,
-  # no engine binary or model weights required — cheap ubuntu-only, so it is
-  # ungated like typos/actionlint above rather than routed through ci-gate.
+  # for the decode-benchmark harness core (issue #813 step 1) and every
+  # migrated profile's adapter module (issue #813 step 2: apples_to_apples,
+  # apples_precise, q4_apples, context_scaling). Stdlib-only (mlx_lm/ollama
+  # are never imported by these test files -- adapters are exercised only
+  # through pure parsing helpers and the harness's own deterministic fake
+  # adapters), no engine binary or model weights required — cheap
+  # ubuntu-only, so it is ungated like typos/actionlint above rather than
+  # routed through ci-gate. (The four adapter test files were not wired into
+  # any CI job before this issue's step-2 migrations landed; this closes
+  # that gap for all of them at once rather than repeating the same
+  # omission per migrated script.)
   bench-decode-harness-tests:
     name: bench-decode-harness-tests
     runs-on: ubuntu-latest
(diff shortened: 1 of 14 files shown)


ohdearquant:

Round-1 findings addressed in cc049df006:

1. **Slope aggregation** — `aggregate()`/TSV now use the adapter-reported `engine_native_ns` when present (harness wall-clock only as MLX fallback), so migrated slopes no longer include subprocess-spawn/model-load overhead the legacy per-process scripts never paid. Mutation-verified.
2. **context_scaling** — `parse_contexts()` preserves caller order and rejects duplicate/baseline-colliding entries; `_legacy_ordered_median()` reproduces the exact legacy per-engine even-n tie-break (lattice/ollama lower-middle, mlx upper-middle); the same median feeds both the slope and the TSV `t1_ms`/`t2_ms`, so they cannot diverge. All three sub-fixes mutation-verified.
3. **Q4 identity guard robustness** — filesystem checks fail closed on `OSError` (permission/TOCTOU) as `LatticeUnavailableError` instead of escaping raw.

Test suites green: harness (118), q4_apples (17), context_scaling (27), apples_precise (20). Reconfirmation review queued.

oceanwaves630 (approved):

Issue #832, opened 2026-07-10
refactor(serve): replace the two independent Metal worker loops with one shared owner


ohdearquant:
> _Issue authored by Claude (Anthropic agent) on behalf of @ohdearquant._

## What

`lattice serve` (`crates/inference/src/bin/lattice.rs`, worker loop and
`MetalJob`/`MetalHandle` around the dedicated Metal thread) and
`lattice_serve` (`crates/inference/src/bin/lattice_serve.rs`,
`Job`/`spawn_worker`/`run_worker_loop`) each own an independent worker thread
that serializes a `!Send` Metal state. Verified at current `main`, both
loops: dequeue in FIFO order, check a cancellation signal at dequeue time
before doing any prefill work, reuse the single process-wide
`CrossTurnSlotId::DEFAULT` cache slot, stream token deltas, and send exactly
one terminal event per job — but they are not behaviorally identical. On
dequeue-time cancellation, `lattice.rs`'s worker sends an empty interrupted
`GenerateOutput` (`stop_reason: Some(StopReason::Interrupt)`) before
`continue`-ing, while `lattice_serve.rs`'s `run_worker_loop` simply
`continue`s with no reply sent at all. The two loops currently agree by
convention (comments), not by shared code.

Add a Metal-gated `lattice_inference::serve::metal_worker` module owning: the
thread, the loaded `MetalQwen35State`, the tokenizer, the job queue, dequeue
cancellation, ChatML rendering via `format_chat_template`, the
`CrossTurnSlotId::DEFAULT` cache slot, delta delivery, terminal delivery, and
thread exit. Both binaries become thin callers supplying loader
inputs/metadata and translating shared worker events into their own HTTP
response phases:

```rust
MetalWorker::spawn(loader) -> Result<(MetalWorkerOwner, MetalWorkerClient, WorkerMetadata), StartupError>
MetalWorkerClient::submit(messages, generate_config, cancel) -> Receiver<WorkerEvent>

WorkerEvent = Delta(String)
            | Complete(GenerateOutput)
            | Rejected(ApiError)
            | Failed(String)
            | Cancelled
```

`spawn` constructs the non-`Send` state inside the dedicated thread; the
state never crosses the thread boundary. `submit` is the only job
constructor. The shared loop must, in order: receive FIFO; check `cancel` or
a closed event receiver before any prompt work; render the ChatML prompt
exactly once via `format_chat_template`; enforce the shared full-window
[...]

PR #939, opened 2026-07-13
feat(inference): bounded metal-worker admission with backpressure rejection (#932)

### LINKED ISSUE #932: bug(serve): metal worker job queue is unbounded — no backpressure cap on outstanding requests
**Defect**: `MetalWorkerClient::submit` enqueues jobs onto an unbounded mpsc channel with no cap on outstanding requests (`crates/inference/src/serve/metal_worker.rs:161`). Both serve binaries share this worker path, so per-request size clamps bound each individual job but nothing bounds the number of queued jobs.

**Blast radius**: N concurrent clients each sending clamped-but-maximal requests drive unbounded queued memory (prompt buffers plus per-job state) and unbounded tail latency; a slow generation at the head of the queue lets the backlog grow without backpressure or shed. The per-request DoS hardening (size fields clamped before allocation) does not cover this aggregate axis.

**Suggested fix**: bounded channel (or explicit in-flight counter) with a configurable depth; on saturation return HTTP 429/503 immediately rather than queueing. Regression: with worker depth 1 and a blocked head job, an (N+1)th submit must fail fast, not queue.

ohdearquant:

Closes #932.

The shared Metal worker's job queue was unbounded: any number of HTTP requests could pile up behind the single-GPU worker with no admission cap. This adds bounded admission with fail-fast rejection.

## Design

- `MetalWorkerClient` holds an `Arc<tokio::sync::Semaphore>`; `submit()` does a synchronous `try_acquire_owned()` **before** enqueueing anything. The `OwnedSemaphorePermit` is stored as a private RAII field on `WorkerJob`, so the slot releases automatically wherever the worker loop finishes with the job (`Complete`, `Rejected`, `Failed`, or dequeue-time `Cancelled`) — no per-branch cleanup to forget.
- `submit()`'s previous "never fails outwardly" contract is deliberately replaced: the ONE new failure mode is `ApiError::ServiceUnavailable` → HTTP **503** with code `server_busy`, returned before any tokenization or model work. 503 over 429 because this is shared single-GPU capacity, not a per-caller rate limit. Contract docs updated at the definition and both consumers.
- Default cap 32 (`DEFAULT_MAX_PENDING_JOBS`), overridable via `--max-pending` on both `lattice_serve` and `lattice serve` (Q4/Metal path only; the CPU backend has no shared worker queue).
- Cancellation of a still-queued job does not free the slot early (the `mpsc` queue has no removal API); it frees at the same FIFO dequeue point #832 established, via the existing `Cancelled` event. Documented in the module docs, and the no-leak property is regression-tested.

## Tests (all GPU-free, via the existing fake-worker seam)

- Cap enforcement — mutation-verified (cap raised 2→3 makes the rejection assertion fail; restored, green).
- Slot release on normal completion.
- Slot release on cancellation of a queued job (`admission_slot_is_released_when_a_queued_job_is_cancelled`).
- HTTP-level test asserting the real 503 JSON envelope through a real router + worker thread.

## Gates

- `cargo fmt --all -- --check` clean; `cargo clippy --workspace -- -D warnings` clean; feature-gated clippy (`f16,metal-gpu,serve,test-utils`, both bins + lib + tests) clean
- metal_worker unit tests 17/17; `lattice_serve` suite 37/37; `lattice` suite 133/133; whole-crate default suite 1798 passed / 0 failed / 18 ignored

bench-compare: not run — no measured bench group's code is touched (diff is serve-layer admission: `serve/metal_worker.rs`, `serve/mod.rs`, two binaries; attention/GEMM/elementwise kernels and generation loops unchanged). Runtime cost is one atomic semaphore acquire per HTTP request at admission, off the token loop. Structural unreachability disclosed per convention.

### DIFF
--- a/crates/inference/src/bin/lattice.rs
+++ b/crates/inference/src/bin/lattice.rs
@@ -66,7 +66,17 @@ enum Command {
         /// no shared worker queue to bound. Conservative default: this
         /// worker serializes all generation onto one dedicated thread, so a
         /// deep queue just means memory growth with no throughput benefit.
-        #[arg(long, default_value = "32")]
+        /// Must be between 1 and `tokio::sync::Semaphore::MAX_PERMITS`
+        /// (issue #939): zero would admit nothing (every request fails
+        /// admission), and clap rejects anything larger here instead of
+        /// deferring to `MetalWorker::spawn`'s own
+        /// `Semaphore::new`-precondition panic.
+        #[arg(
+            long,
+            default_value = "32",
+            value_parser = clap::builder::RangedU64ValueParser::<usize>::new()
+                .range(1..=(tokio::sync::Semaphore::MAX_PERMITS as u64))
+        )]
         max_pending: usize,
     },
     /// Preflight check: memory fit and artifact compatibility, without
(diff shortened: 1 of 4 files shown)


ohdearquant:

Round-1 findings addressed in 3325083ebb:

1. **Major (SSE-before-admission / 500-not-503)** — split `generate_streaming_with_cancel` into synchronous `submit()` (admission) and async `drain()` (event loop). `chat_completions` streaming arm now calls `submit()` before `tokio::spawn`, so `ServiceUnavailable` propagates as 503 before any SSE is committed; non-streaming arm passes `ServiceUnavailable` through instead of collapsing to 500. Two new cap=1 real-router tests assert 503, `server_busy`, full envelope, and no committed `text/event-stream` for the streaming case.
2. **Medium (weak queued-cancel oracle)** — strengthened with a direct `admission.available_permits() == cap` assertion. Mutation-verified: `std::mem::forget(job)` on the queued-cancel path makes it fail (`got 1 of 2`), reverse-applied, passes again.
3. **Medium (`--max-pending` validation)** — range-validate `1..=Semaphore::MAX_PERMITS` in `MetalWorker::spawn` before `Semaphore::new` (`StartupError::InvalidMaxPending`); clap flag uses `RangedU64ValueParser`; daemon `parse_max_pending` rejects malformed/negative instead of silently defaulting. Boundary tests (zero, MAX+1, negative, malformed) in all three files.

Gates: fmt clean, clippy `-D warnings` clean, `serve::metal_worker` 19 (was 17), `--bin lattice` 142 (was 133), `--bin lattice_serve` 42 (was 37). Reconfirmation review queued.

oceanwaves630 (approved):

PR #944, opened 2026-07-13
ci: skip the Rust matrix on Swift-only (apps/macos) changes

ohdearquant:

## Problem

`apps/macos/**` is the Swift Studio app. It has no Rust surface, and there is no `swift build`/`swift test`/`xcodebuild` job anywhere in CI. Yet the `changes` job in `ci.yml` classifies everything non-docs as `code=true`, so a Swift-only PR fires the full macOS/ARM Rust matrix — `ci`, `feature-matrix`, `bench-compile`, `msrv`, `semver-checks` — for zero coverage benefit and a chunk of the shared macOS runner pool.

## Fix

Add `^apps/macos/` to the non-Rust exclusion set in the change-detection filter. A Swift-only change now yields `code=false`, the Rust matrix skips, and the always-run required `ci-gate` passes via the existing docs-only path (`CODE != true` → PASS).

## Why this is safe / complete

- **App-binary coverage retained.** The Rust binaries the app bundles are built by `app-binaries.yml`, which triggers independently on every PR and gates its heavy macOS build on its own allow-list of Rust paths (`^crates/`, `Cargo.*`, `build-app-bins.sh`, `package-app.sh`, …). Unaffected by this change.
- **e2e-parity unaffected.** `e2e-parity.yml` already uses a Rust-path allow-list (`^crates/(inference|embed)/src/`, …); a Swift-only change never matched it, so it already skipped.
- **ci.yml was the only offender.** It was the one workflow keying off a deny-everything-non-docs filter, so it was the only one firing heavy Rust jobs on Swift changes.
- **Mixed PRs still run the matrix.** A PR touching both `apps/macos/` and any Rust path leaves the Rust file in `NON_DOCS`, so `code=true` and the full matrix runs.

## No bench impact

CI-config only; no `crates/inference` or `crates/embed` source touched. actionlint clean.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,9 +51,10 @@ env:
   LATTICE_REQUIRE_FIXTURES: "1"
 
 jobs:
-  # Decide whether anything outside the docs set changed. A PR that only
-  # touches markdown, ADRs, LICENSE, or .gitignore cannot affect compilation,
-  # so it does not need the macOS/ARM compile matrix below.
+  # Decide whether any Rust code changed. A PR that only touches markdown,
+  # ADRs, LICENSE, .gitignore, or the Swift macOS app (apps/macos/**) cannot
+  # affect this workspace's compilation, so it does not need the macOS/ARM
+  # compile matrix below.
   changes:
     name: detect-code-changes
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


oceanwaves630 (approved):
Approved: one-token change-detection edit under Ocean's directive; required-context safety verified at source by owner (six required checks all unconditional or code=false-safe); all six green on head.

PR #957, opened 2026-07-14
docs(fann): reconcile ADR-021/024/025/026 + README with shared-weight/builder/GPU contracts

### LINKED ISSUE #768: audit(fann): 5 medium doc/validation findings — self-audit 20260709
- [ ] **ADR-021 retains the retired clone-per-thread parallel contract**
  docs/adr/ADR-021-zero-alloc.md:81 says forward_batch clones the network per Rayon thread.
  Evidence: `crates/fann/src/network/mod.rs:273`. Readers of the accepted ADR set receive contradictory memory and CPU cost contracts for parallel inference.
  Suggested fix: Update ADR-021 to describe shared immutable weights and per-input activation-buffer allocation.

- [ ] **Builder documentation falsely requires input before layer selection**
  crates/fann/src/network/builder.rs:64 and ADR-024:21 say input(...) must precede hidden(...) or output(...), enforced by input_size.
  Evidence: `crates/fann/src/network/builder.rs:65`. The documented builder ordering is not an invariant, so callers and maintainers can reason incorrectly about valid construction states.
  Suggested fix: Correct the comments and ADR to make input required by build rather than ordered before layer methods, or add explicit order tracking and rejection.

- [ ] **Training documentation denies a per-sample output clone that code performs**
  ADR-026:70 and the compute_gradients comment at backprop.rs:40 say the sample path has no per-sample cloning.
  Evidence: `crates/fann/src/training/backprop.rs:53`. The training hot-path allocation contract is misstated, misleading memory and throughput analysis based on the ADR.
  Suggested fix: Document the output copy and deltas allocation, or eliminate them before preserving the no-clone claim.

- [ ] **README GPU example calls a nonexistent constructor and cannot mutably forward**
  crates/fann/README.md:280-283 presents GpuNetwork::from_cpu(&ctx, cpu_network) followed by network.forward(...).
  Evidence: `crates/fann/README.md:281`. Users following the documented GPU path cannot construct or invoke the current API.
  Suggested fix: Replace the snippet with a complete current construction using Arc<GpuContext>, GpuNetwork::new, a mutable network binding, and defined network/input values.

- [ ] **Watchdog comments document a time guard that the implementation never applies**
  gpu/mod.rs:59-62 and ADR-025:28-32 present a 100,000-element limit and MAX_DISPATCH_TIME_MS=1.5 as a 2 ms Metal watchdog boundary.
  Evidence: `crates/fann/src/gpu/mod.rs:61`. A hardware-dependent safety heuristic is represented as an enforced time boundary, risking incorrect maintenance and operational assumptions.
  Suggested fix: Document the element cap as a calibrated heuristic with evidence, or implement and validate the claimed time/tile guard; remove the unused time constant otherwise.

---

ohdearquant:

## Summary

- Reconcile ADR-021 with `forward_batch`: Rayon workers share immutable layer parameters and allocate per-input activation buffers instead of cloning the network.
- Correct ADR-024 and `NetworkBuilder::input` documentation so input is required at build time but need not precede layer methods.
- Document ADR-026's shipped per-sample output copy and delta-vector allocations.
- Replace the stale GPU README example with the current `Arc<GpuContext>`, `GpuNetwork::new`, and mutable-forward API.
- Clarify in ADR-025 and GPU comments that the 100,000-element cap is a static CPU-fallback heuristic and elapsed dispatch time is not measured.

## Verification

- Compared each ADR/comment claim with the current implementations in `network/mod.rs`, `network/builder.rs`, `training/backprop.rs`, `gpu/network.rs`, and `gpu/mod.rs`.
- Confirmed `MAX_DISPATCH_TIME_MS` has no runtime consumer and that oversized layers fall back through the `MAX_ELEMENTS_PER_DISPATCH` element-count check.
- Ran `git diff --check`; no build, test, benchmark, or Clippy command was run for these documentation/comment-only changes.

Closes #768

### DIFF
--- a/crates/fann/README.md
+++ b/crates/fann/README.md
@@ -273,14 +273,26 @@ Enable the `gpu` feature for wgpu-based GPU acceleration.
 
 ```rust
 #[cfg(feature = "gpu")]
-use lattice_fann::gpu::{GpuContext, GpuNetwork};
+use lattice_fann::{
+    Activation, NetworkBuilder,
+    gpu::{GpuContext, GpuNetwork},
+};
+#[cfg(feature = "gpu")]
+use std::sync::Arc;
 
 #[cfg(feature = "gpu")]
 async fn gpu_inference() -> Result<(), Box<dyn std::error::Error>> {
-    let ctx = GpuContext::new().await?;
-    let network = GpuNetwork::from_cpu(&ctx, cpu_network)?;
+    let cpu_network = NetworkBuilder::new()
+        .input(4)
+        .hidden(8, Activation::ReLU)
+        .output(2, Activation::Softmax)
+        .build()?;
+    let ctx = Arc::new(GpuContext::new().await?);
+    let mut network = GpuNetwork::new(ctx, cpu_network)?;
+    let input = [0.1, 0.2, 0.3, 0.4];
 
     let output = network.forward(&input).await?;
+    println!("{output:?}");
     Ok(())
 }
 ```
--- a/crates/fann/src/gpu/mod.rs
+++ b/crates/fann/src/gpu/mod.rs
@@ -1,9 +1,9 @@
 //! GPU compute backend: contexts, pooled memory, WGSL pipelines, and dense inference.
 //!
 //! CPU selection uses a 10,000 effective-element threshold; activation policy uses 1,000.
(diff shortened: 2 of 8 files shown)


oceanwaves630 (approved):

PR #963, opened 2026-07-14
refactor(transport): rename bench module to reference_cases

### LINKED ISSUE #805: chore(transport): rename test-only bench.rs to reference_cases.rs and remove its blanket dead-code allow

## What

`crates/transport/src/bench.rs` is compiled only under `#[cfg(test)]` (`crates/transport/src/lib.rs`, roughly lines 71-74) and is actively used by a live unit test: `crates/transport/src/tests.rs` (roughly lines 253-270) calls `super::bench::compare_against_reference` as a Sinkhorn reference-case comparison. The module currently carries a blanket `#[allow(dead_code)]` (roughly line 20) because some fields of its `ReferenceComparison` result type are constructed but never read by the current caller.

This issue is explicitly **not** a deletion. An earlier audit pass in this cluster's design work considered removing `bench.rs` as apparently-dead code and confirmed against source that this reading is wrong — the module is a live test dependency and must not be deleted.

## Why

The module's name (`bench.rs`) and its blanket `#[allow(dead_code)]` both read as signals of dead or vestigial code to a future auditor, which is exactly the misreading this cluster's design work made and then had to correct against source. Renaming it to reflect its actual role (a reference-case comparison helper for tests, not a benchmark) and replacing the blanket allow with either use of the currently-unread fields or a narrower, field-specific allow removes that false signal for the next person who greps this crate.

## Steps

1. Rename `crates/transport/src/bench.rs` to `crates/transport/src/reference_cases.rs` (or another name that clearly signals "test reference-case helper," not "benchmark"), updating the `#[cfg(test)]` module declaration in `lib.rs` and the call site in `tests.rs` accordingly.
2. Replace the blanket `#[allow(dead_code)]` with either: (a) code that reads the currently-unused `ReferenceComparison` fields (e.g. asserting on them in the existing Sinkhorn reference-case test), or (b) a narrower, field-level `#[allow(dead_code)]` limited to the specific fields that are genuinely intended as forward-looking/diagnostic-only, with a comment explaining why.
3. Do not change `compare_against_reference`'s behavior, its call site's assertions beyond what step 2 requires, or any public transport crate API — `lattice-transport` is a published crate and this issue must not touch its public surface.

## Gates

- `cargo test -p lattice-transport` passes unchanged in behavior (the Sinkhorn reference-case test must continue to exercise the same comparison logic under its new name).
- `cargo clippy -p lattice-transport -- -D warnings` passes without the blanket dead-code allow (or with a documented, narrowed one).
- No public `lattice-transport` API changes; this module is `#[cfg(test)]`-only and was never part of the published surface.
- Semver class: **patch** (test-only rename; the crate's published API is untouched).

## Sequencing

- **Depends on:** nothing; independent of every other issue in this cluster and can land at any time, including before or after the rest of the migration.
- **Depended on by:** none.

## Note on scope

This issue is optional cleanup, not a required step in the weight-ingress validation migration — it exists only because the design work for that migration surfaced (and then corrected) a mistaken "this looks dead" reading of the same file. It carries no dependency relationship to the ingress-seam, half-bits, QuaRot, Q4/KHF1, or Q8 issues in this cluster.

ohdearquant:

## Defect

The live test-only Sinkhorn reference helper was named `bench.rs`, which made it look like vestigial benchmark code, and `ReferenceComparison` carried a blanket `#[allow(dead_code)]` because two diagnostic fields were not read.

## Fix

- Renamed the test-only module to `reference_cases.rs` and updated its declaration and only caller.
- Removed the blanket dead-code allow.
- Extended the existing reference-case test to assert convergence and relative error, consuming both previously unread diagnostic fields without changing solver behavior or public API.

## Sibling paths checked

Searched all of `crates/transport` for `compare_against_reference`, `ReferenceComparison`, `mod bench`, and `bench::`. The existing `reference_cases_pass` unit test was the only invocation path, so no sibling callers required changes.

## Regression sensitivity

The test was changed first to call `super::reference_cases`; before the module rename, the focused test build failed with E0433 because `reference_cases` did not exist. Reverting the implementation rename therefore fails the regression at compile time. The added assertions also exercise `result` and `relative_error` directly.

## Verification

- `cargo test -p lattice-transport` (30 unit tests, 3 integration tests, 1 doctest)
- `cargo clippy -p lattice-transport -- -D warnings`

This is test-harness-only and does not touch a decode, prefill, or GEMM hot path; no GPU benchmark is required.

Closes #805

### DIFF
--- a/crates/transport/src/lib.rs
+++ b/crates/transport/src/lib.rs
@@ -52,7 +52,7 @@ pub mod transport_plan;
 pub mod unbalanced;
 
 #[cfg(test)]
-mod bench;
+mod reference_cases;
 #[cfg(test)]
 mod tests;
 
--- a/crates/transport/src/reference_cases.rs
+++ b/crates/transport/src/reference_cases.rs
@@ -17,7 +17,6 @@ pub struct ReferenceCase {
 }
 
 #[derive(Debug, Clone)]
-#[allow(dead_code)]
 pub struct ReferenceComparison {
     pub name: &'static str,
     pub expected_exact_cost: f32,
--- a/crates/transport/src/tests.rs
+++ b/crates/transport/src/tests.rs
@@ -257,14 +257,27 @@ fn reference_cases_pass() {
         convergence_threshold: 1e-7,
         ..SinkhornConfig::default()
     });
-    let comparisons = super::bench::compare_against_reference(&solver).unwrap();
+    let comparisons = super::reference_cases::compare_against_reference(&solver).unwrap();
 
     for comp in &comparisons {
+        assert!(
+            comp.result.converged,
+            "Reference case '{}' should converge",
+            comp.name,
+        );
         assert!(
             comp.absolute_error < 0.1,
             "Reference case '{}': absolute error {} is too large (expected ~{}, got {})",
             comp.name,
             comp.absolute_error,
+            comp.expected_exact_cost,
+            comp.sinkhorn_cost,
+        );
+        assert!(
+            comp.relative_error < 0.1,
+            "Reference case '{}': relative error {} is too large (expected ~{}, got {})",
+            comp.name,
+            comp.relative_error,
             comp.expected_exact_cost,
             comp.sinkhorn_cost,
         );


oceanwaves630 (approved):

PR #964, opened 2026-07-14
fix(fann): enforce hidden-Softmax placement invariant in Network::new

### LINKED ISSUE #742: bug(fann): Network::new does not enforce the hidden-Softmax placement invariant
**Defect**: `NetworkBuilder::build`/`build_with_rng` reject a non-final hidden `Softmax` layer via `validate_no_hidden_softmax`, but `Network::new` (the direct constructor) only validates layer dimensions and never calls that check. Because `Network::new` is also the route used by `TryFrom<NetworkData>` (serde deserialization) and the custom bincode-style loader, any loaded/deserialized network file bypasses the hidden-Softmax invariant regardless of how it was originally built. Downstream, the generic Softmax derivative used by CPU backprop is diagonal-only (documented as an approximation), and the GPU execution path skips Softmax entirely for non-final layers.

**Evidence**:
- `crates/fann/src/network/builder.rs:35-50` (`validate_no_hidden_softmax`), invoked only from `build`/`build_with_rng` (lines 123, 183).
- `crates/fann/src/network/mod.rs:120-146` (`Network::new`) — validates emptiness and layer-dimension compatibility only, no Softmax-position check.
- `crates/fann/src/network/mod.rs:116` (`TryFrom<NetworkData> for Network`) and `crates/fann/src/network/serialization.rs:284` both route through `Network::new` directly, and `Layer::new` (`crates/fann/src/layer.rs:341`) is `pub`, so direct `Vec<Layer>` + `Network::new` construction is a reachable bypass.
- `crates/fann/src/activation.rs:348` — generic Softmax derivative is diagonal-only (marked as a known approximation).
- `crates/fann/src/training/backprop.rs:116` — hidden-layer backprop uses the generic (diagonal) derivative.
- `crates/fann/src/gpu/network.rs:249-253` — GPU skips Softmax during layer execution for non-final layers (marked as a known gap).

**Blast radius**: Any caller using `Network::new` directly, or loading a serialized network containing a non-final Softmax layer, on CPU training or GPU inference.

**Suggested fix**: Move the hidden-Softmax validation into `Network::new` itself (with `NetworkBuilder` delegating to it), and add a defensive rejection (or CPU fallback) in the GPU path for an invariant-violating network that reaches it anyway.

ohdearquant:

## Defect

`Network::new` accepted Softmax on a non-final layer even though the builders rejected that topology. Direct construction, serde deserialization, and binary deserialization could therefore bypass the invariant.

## Fix

Centralize the hidden-Softmax placement check in `Network::new` and remove the redundant builder-level checks so both builder entry points delegate to the constructor invariant. Output-layer Softmax remains valid.

## Sibling paths checked

- `NetworkBuilder::build` and `NetworkBuilder::build_with_seed` both finish through `Network::new`.
- `TryFrom<NetworkData>` for serde and `Network::from_bytes` already finish through `Network::new`; no duplicate checks were needed.
- The GPU fail-loud path is explicitly split into separate work in the issue comment and is not changed here.

## Regression test

`network_new_rejects_hidden_softmax` constructs compatible layers with hidden Softmax and calls `Network::new` directly. Reverting the constructor guard makes the call return `Ok(Network)`, so the test fails. The test was observed failing before the fix and passing after it.

## Verification

- `cargo test -p lattice-fann` (80 unit tests plus integration, property, and shader validation tests)
- `cargo test -p lattice-fann --features serde serde_validation_tests`
- `cargo clippy -p lattice-fann -- -D warnings`
- Repository pre-commit checks

No GPU bench is required because this is a construction-time validation change, not a decode, prefill, or GEMM hot-path change.

Closes #742

### DIFF
--- a/crates/fann/src/network/builder.rs
+++ b/crates/fann/src/network/builder.rs
@@ -19,24 +19,6 @@ pub struct NetworkBuilder {
     layers: Vec<(usize, Activation)>,
 }
 
-/// Reject Softmax before the final layer: its derivative here is only diagonal (see ADR-023).
-fn validate_no_hidden_softmax(layers: &[(usize, Activation)]) -> FannResult<()> {
-    if layers.len() < 2 {
-        return Ok(());
-    }
-    let last_hidden = layers.len() - 1;
-    for (_, activation) in &layers[..last_hidden] {
-        if activation.is_softmax() {
-            return Err(FannError::InvalidBuilder(
-                "Softmax is only valid as the output-layer activation; \
-                 hidden layers must use a pointwise activation"
-                    .into(),
-            ));
-        }
-    }
-    Ok(())
-}
-
 impl NetworkBuilder {
     /// Create a new network builder
     pub fn new() -> Self {
@@ -87,8 +69,6 @@ impl NetworkBuilder {
             return Err(FannError::InvalidBuilder("Input size cannot be 0".into()));
         }
 
-        validate_no_hidden_softmax(&self.layers)?;
-
         let mut layers = Vec::with_capacity(self.layers.len());
         let mut prev_size = input_size;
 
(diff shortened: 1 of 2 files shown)


oceanwaves630 (approved):

PR #965, opened 2026-07-14
fix(inference): weight-loader validation guards (aliased ranges, q4 shape, config copy)

### LINKED ISSUE #706: quantize_q4 output dir is not self-contained: missing config.json breaks from_q4_dir / eval_perplexity
## Summary

`quantize_q4` writes `.q4` / `.f16` / `quantize_index.json` to its `--output-dir`, but **not** `config.json`. The runtime Q4 loader path — `MetalQwen35State::from_q4_dir` via `eval_perplexity`'s `load_cfg_for_q4` — reads `config.json` from the Q4 dir and errors if it is absent:

```
ERROR: Q4 dir <output-dir> is missing config.json
```

So a freshly-produced `quantize_q4` directory cannot be loaded/evaluated without manually copying `config.json` in from the source model dir.

## Inconsistency

`quantize_quarot` produces a **self-contained** artifact (its output dir includes `config.json`), so its output loads directly. `quantize_q4` does not, which is surprising given both feed the same `from_q4_dir` loader. The gap is easy to miss because a pre-existing Q4 dir often already has `config.json` from an earlier setup — it only surfaces on a fresh artifact.

## Repro

```bash
target/release/quantize_q4 --model-dir <model> --output-dir /tmp/q4
ls /tmp/q4                 # .q4 / .f16 / quantize_index.json, no config.json
target/release/eval_perplexity --q4-dir /tmp/q4 --tokenizer-dir <model> \
  --corpus-file <corpus>   # ERROR: Q4 dir /tmp/q4 is missing config.json
```

## Suggested fix

Have `quantize_q4` copy (or write) `config.json` into its `--output-dir`, matching `quantize_quarot`, so any Q4 artifact is self-describing and directly loadable. Alternatively, document that a Q4 dir requires `config.json` alongside the `.q4` files and have callers provide it.

## Context

Discovered while wiring the #616 Q4 perplexity CI gate: the gate's ephemeral-artifact step now copies `config.json` in as a workaround. Making `quantize_q4` self-contained would remove that workaround and prevent the same surprise for anyone quantizing then loading/evaluating.

### LINKED ISSUE #735: bug(inference): F16 tensor loader does not verify declared shape matches declared element count
**Impact:** High — out-of-bounds GPU/CPU reads, corrupted inference output from a malformed but internally-"valid" checkpoint artifact.

**Location:** `crates/inference/src/weights/q4_weights.rs`, `load_f16_tensor_file` (reads `shape` and `numel` as independent header fields, never checks `shape.iter().product() == numel` before returning). Consumed in `crates/inference/src/forward/metal_qwen35.rs` by the MTP tensor resolver (compares only the declared `shape` against an expected shape, not against the actual value count) and by the main Q4-directory loader's F16 helper (discards the returned shape entirely and builds a GPU buffer sized to whatever `values.len()` happens to be).

**Repro sketch:**
1. Construct a `.f16` (KHF1-format) file whose header declares `ndim=2, shape=[8192,2048]` but `numel=1`, followed by exactly one `f16` (2-byte) value.
2. Call `load_f16_tensor_file` on it: it returns `Ok((vec![<one f32>], vec![8192, 2048]))` — no error.
3. Any caller that trusts the returned `shape` (rather than `values.len()`) to size a buffer for a `[8192, 2048]` matrix will materialize/allocate a full-size buffer view or dispatch a kernel over a 1-element backing allocation.

**Suggested fix:** After reading `shape` and `numel`, compute the shape product with checked multiplication and require exact equality with `numel` before allocating/reading the payload; return an error otherwise. Add a regression test that constructs a header with a correct shape and an undersized/mismatched `numel` and asserts the loader rejects it.

### LINKED ISSUE #752: bug(inference): SafetensorsFile does not reject aliased/overlapping tensor byte ranges

ohdearquant:

## Defects

- `quantize_q4` produced `.q4`, `.f16`, and `quantize_index.json` files without the source `config.json`, so a fresh output directory was not directly loadable.
- The KHF1 F16 loader trusted `shape` and `numel` independently, allowing a declared shape to overstate the actual payload.
- `SafetensorsFile` validated tensor extents individually but did not reject overlapping, aliased, gapped, or incompletely covered data ranges.

## Fix

- Copy the source `config.json` into non-dry-run `quantize_q4` output directories.
- Compute the KHF1 shape product with checked multiplication and require it to equal `numel` before payload allocation and decoding.
- Sort all Safetensors tensor ranges and require contiguous coverage from offset zero through the entire data section.

## Sibling paths checked

- `quantize_quarot` already writes its transformed `config.json`; no additional converter path needed a change.
- Every runtime KHF1 consumer in `metal_qwen35.rs` routes through `load_f16_tensor_file`, so the loader-boundary guard covers the MTP and main Q4-directory F16 paths.
- `SafetensorsFile::open`, `SafetensorsFile::from_bytes`, and lazy sharded loads all route through the guarded `from_backing` path. The separate QuaRot Safetensors reader already enforces sorted contiguous full coverage.

## Regression sensitivity

- Reverting the config copy makes `output_includes_source_config_json_verbatim` fail because the output file is missing.
- Reverting the KHF1 guard makes `test_f16_rejects_shape_numel_mismatch` return `Ok` for a `[2, 2]` tensor with one declared element.
- Reverting the Safetensors layout guard makes `test_rejects_aliased_tensor_ranges` accept two names backed by the same `[0, 4)` bytes.

## Validation

- `cargo test -p lattice-inference`
- `cargo clippy -p lattice-inference -- -D warnings`
- Pre-commit workspace clippy/all-targets and documentation lint

No GPU bench is required: these changes are confined to offline artifact generation and load-time validation, not decode, prefill, or GEMM hot paths.

Closes #706
Closes #735
Closes #752

### DIFF
--- a/crates/inference/src/bin/quantize_q4.rs
+++ b/crates/inference/src/bin/quantize_q4.rs
@@ -246,6 +246,15 @@ fn run() -> Result<(), Box<dyn std::error::Error>> {
 
     if !dry_run {
         fs::create_dir_all(&output_dir)?;
+        let source_config = model_dir.join("config.json");
+        let output_config = output_dir.join("config.json");
+        fs::copy(&source_config, &output_config).map_err(|e| {
+            format!(
+                "failed to copy {} to {}: {e}",
+                source_config.display(),
+                output_config.display()
+            )
+        })?;
     }
 
     let reader = QuarotTensorReader::open(&model_dir)?;
--- a/crates/inference/src/weights/f32_weights.rs
+++ b/crates/inference/src/weights/f32_weights.rs
@@ -362,6 +362,27 @@ impl SafetensorsFile {
             }
         }
 
+        let mut ranges: Vec<_> = tensors
+            .iter()
+            .map(|(name, meta)| (meta.start, meta.end, name.as_str()))
+            .collect();
+        ranges.sort_unstable();
+        let mut previous_end = 0usize;
+        for &(start, end, name) in &ranges {
+            if start != previous_end {
+                return Err(InferenceError::InvalidSafetensors(format!(
+                    "tensor {name} has non-contiguous data offsets: expected start={previous_end}, \
+                     got [{start}, {end})"
+                )));
+            }
+            previous_end = end;
+        }
+        if previous_end != data_len {
+            return Err(InferenceError::InvalidSafetensors(format!(
+                "data section is {data_len} bytes but tensors cover {previous_end} bytes"
+            )));
+        }
+
         Ok(Self {
             data,
             data_offset,
@@ -1981,6 +2002,35 @@ mod tests {
             .expect_err("one-byte-extra payload should be rejected at open");
         assert!(
             err.to_string().contains("byte length mismatch"),
+            "unexpected error: {err}"
+        );
+
+        fs::remove_file(&path).ok();
+    }
+
+    #[test]
+    fn test_rejects_aliased_tensor_ranges() {
+        let path = temp_path("lattice_weights_aliased_ranges");
+        let header = r#"{
+            "key": {"dtype": "F32", "shape": [1], "data_offsets": [0, 4]},
+            "value": {"dtype": "F32", "shape": [1], "data_offsets": [0, 4]}
+        }"#
+        .replace(['\n', ' '], "");
+
+        let mut bytes = Vec::new();
+        bytes.extend_from_slice(&(header.len() as u64).to_le_bytes());
+        bytes.extend_from_slice(header.as_bytes());
+        bytes.extend_from_slice(&1.0f32.to_le_bytes());
+
+        let mut file = File::create(&path).expect("test setup: create safetensors file");
+        file.write_all(&bytes)
+            .expect("test setup: write safetensors bytes");
+        drop(file);
+
+        let err = SafetensorsFile::open(&path)
+            .expect_err("aliased tensor byte ranges must be rejected at open");
+        assert!(
+            err.to_string().contains("non-contiguous"),
             "unexpected error: {err}"
         );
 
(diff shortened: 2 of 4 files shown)


oceanwaves630 (approved):

PR #966, opened 2026-07-14
fix(embed): verify unit-norm assumption in cosine fast path

### LINKED ISSUE #759: bug(embed): unit-norm fast path never verifies its documented lazy check
**Defect**: A documented fast-path optimization claims the stored vector's unit-norm assumption is "verified lazily via `is_unit_norm`" before taking a cheaper `1.0 - dot` distance calculation instead of full cosine distance. The function body never calls `is_unit_norm` — it only checks caller-supplied `NormalizationHint` enum values on both the query and stored metadata, neither of which is derived from inspecting the actual vector. A caller that mislabels a non-unit stored vector as `Unit` silently receives a mathematically wrong (non-cosine) distance with no verification catching the mismatch.

**Evidence**:
- `crates/embed/src/simd/tier.rs:353` — doc comment: "The stored vector's unit claim is verified lazily via `is_unit_norm`."
- `crates/embed/src/simd/tier.rs:364-372` — the fast-path condition only checks `meta.norm == NormalizationHint::Unit && stored_norm == NormalizationHint::Unit` (both caller-supplied hints), then takes the `1.0 - dot` shortcut; `is_unit_norm` is never called anywhere in this function.
- `crates/embed/src/simd/tier.rs:277` — `is_unit_norm` exists as a separate function but is unused by the fast path.

**Blast radius**: Any caller relying on the documented lazy verification receives `1 - dot` instead of true cosine distance for a mislabeled non-unit vector, silently changing similarity ranking and any downstream threshold-based decision.

**Suggested fix**: Either call `is_unit_norm` to validate stored data as documented, or remove the "verified lazily" claim and state an explicit caller-guaranteed precondition instead (updating any corresponding ADR language to match).

ohdearquant:

## Defect

`approximate_cosine_distance_prepared_with_meta` trusted the stored `NormalizationHint::Unit` metadata without inspecting the stored vector. A non-unit `Full` vector mislabeled as unit therefore took the `1 - clamp(dot)` shortcut and returned a non-cosine distance.

## Fix

Require `is_unit_norm` to validate the stored `Full` vector before entering the dot-product shortcut. A false stored-unit claim now falls back to the existing full cosine-distance dispatch.

## Sibling paths checked

Searched `lattice-embed` for `NormalizationHint`, `PreparedQueryWithMeta`, `prepare_query_with_norm`, and `approximate_cosine_distance_prepared_with_meta`. There is no sibling implementation of the metadata-aware shortcut to update; the existing Criterion benchmark is the only non-test external caller.

## Regression test

Added `test_cosine_distance_prepared_with_meta_validates_stored_unit_norm`, using a unit query and stored `[2.0, 0.0]` vector mislabeled as unit. With the guard reverted, the shortcut returns `0`; the full cosine path returns `0.29289317`, so the test fails.

## Validation

- `cargo fmt --check`
- `cargo test -p lattice-embed`: 338 passed, 1 existing ignored
- `cargo clippy -p lattice-embed -- -D warnings`
- `cargo llvm-cov -p lattice-embed --summary-only`: passed; filtered `lattice-embed` library report is 91.45% line coverage, with `simd/tier.rs` at 87.85%
- `cargo bench -p lattice-embed --bench simd -- prepared_meta_unit` on NEON, 1,000 stored vectors per sample:
  - 384d: 373.72 us median, 2.6758 Melem/s
  - 768d: 751.08 us median, 1.3314 Melem/s
  - 1024d: 1.0896 ms median, 917.80 Kelem/s

This is a CPU-only validation guard, not a decode, prefill, or GEMM hot-path change. No GPU benchmark is required.

Closes #759

### DIFF
--- a/crates/embed/src/simd/tier.rs
+++ b/crates/embed/src/simd/tier.rs
@@ -319,6 +319,7 @@ pub fn approximate_cosine_distance_prepared_with_meta(
     if meta.norm == NormalizationHint::Unit
         && stored_norm == NormalizationHint::Unit
         && let (PreparedQuery::Full(q), QuantizedData::Full(s)) = (&meta.query, stored)
+        && is_unit_norm(s)
     {
         let dot = dot_product(q, s);
         return Ok(1.0 - dot.clamp(-1.0, 1.0));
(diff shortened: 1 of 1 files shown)


oceanwaves630 (approved):

PR #967, opened 2026-07-14
fix(inference): zero-budget/empty-prompt/NaN decode guards

### LINKED ISSUE #404: harden(inference): live Qwen3.5 sampler NaN policy differs from canonical Sampler (semantics decision needed)
## Summary

The **live** Qwen3.5 CPU sampler (`crates/inference/src/model/qwen35/sampling.rs::sample_token`) handles a NaN logit differently from the canonical `Sampler` path, and the difference is a **deliberate, tested policy choice** — surfacing it for a decision rather than self-merging.

`sample_token` is the sampler actually used by live generation (`generation.rs` → `super::sampling::sample_token`), a parallel implementation to the canonical `crate::sampling::Sampler`.

## The two divergent fail-closed policies

Both are safe (neither returns a NaN token or out-of-set garbage); they differ only on degenerate NaN input:

**A. Current qwen35 policy — "any NaN in the selected set → argmax/greedy fallback"** (deliberately tested):
- `test_nan_in_nonmax_position_routes_to_argmax` (sampling.rs:294) asserts `[5.0, NaN, 1.0]`, `top_k=0` returns token 0, with the explicit comment *"the guard must fall back to the finite argmax (index 0), not the NaN slot."*
- Mechanism: the top-k comparators (`partial_cmp(...).unwrap_or(Equal)`, lines ~36-41 / ~51-56) treat NaN as equal to every finite logit, so a low-id NaN can win the tie-break **into** the top-k set and displace a genuine finite candidate; `build_softmax_probs` (line ~114) then rejects the NaN-poisoned sum (returns `None`) and the draw falls back to `greedy_token`.

**B. Canonical `Sampler` policy — "exclude NaN, sample the finite remainder"**:
- `select_top_k_scalar` sanitizes NaN seed logits to `NEG_INFINITY` (sampling.rs:~645): `logit: if scaled.is_nan() { NEG_INFINITY } else { scaled }`.
- For `[NaN, 10.0, 9.0]`, `top_k=2`: the finite top-2 `{10, 9}` are kept and **sampled** (≈73/27), so token-of-9 is reachable.

## The observable consequence of policy A

For `[NaN, 10.0, 9.0]`, `top_k=2`: the NaN (id 0) displaces the genuine #2 finite candidate (9.0), then greedy fallback returns token-of-10. **Token-of-9, a legitimate top-2 member, is never sampled** whenever a NaN is present in the vocab.

## Why this needs a decision, not a clean fix

The only change that resolves the complaint (make token-of-9 reachable) is adopting policy **B** — which **overrides the deliberately-tested policy A**. That is a CPU-sampling-semantics decision, not a bug-to-internal-contract:

- A NaN→NEG_INFINITY sanitization in `sample_token` (mirroring the canonical path) would change `test_nan_in_nonmax_position_routes_to_argmax` from argmax-fallback to sample-the-finite-remainder.
- A behavior-preserving "any NaN → greedy up front" hardening keeps policy A explicit/robust but does **not** resolve the complaint (token-of-9 still unreachable) — so it is a refactor, not a fix.

There is no fix that both (a) preserves the tested argmax-fallback contract and (b) makes the displaced finite candidate reachable. Choosing requires deciding which fail-closed policy the live sampler should follow.

## Recommendation

Adopt policy **B** (align live qwen35 sampler with the canonical `Sampler`: sanitize NaN→NEG_INFINITY before selection) for consistency across the two sampler implementations, and update `test_nan_in_nonmax_position_routes_to_argmax` to encode the exclude-and-sample contract. But this is a **semantics call** because it overrides an existing intentional, tested contract.

## Provenance

Found by a discovery review pass over `crates/inference/src/model/qwen35/sampling.rs` (sampling selection paths). Companion finding (CandidateSet::argmax all-(-inf) out-of-set token 0) filed separately. Both need a decision; neither self-merged.

### LINKED ISSUE #777: bug(inference): BatchWorker emits one token for max_new_tokens=0 requests
**Defect**: `BatchWorker` samples and emits a token for sequences admitted with `max_new_tokens=0`, where every other decode path treats the zero-token budget as a hard cap that produces no output. The generic and Qwen3.5 single-sequence generators both check the remaining budget and return before sampling when it is zero; `BatchWorker` instead samples immediately once prefill completes, and `push_token` appends the sampled token before the sequence's generated length is compared against `max_new_tokens`.

**Evidence**:
- `crates/inference/src/generate.rs:429` — generic generator returns before sampling once the zero token budget is detected.
- `crates/inference/src/model/qwen35/generation.rs:35` — Qwen3.5 single-sequence generator has the same early-return.
- `crates/inference/src/batch/worker.rs:477` — `BatchWorker` samples immediately when prefill completes, with no zero-budget short-circuit.
- `crates/inference/src/batch/sequence.rs:180` — `push_token` appends the token before testing generated length against `max_new_tokens`.

**Blast radius**: Every continuous-batching request submitted with `max_new_tokens=0` emits one unwanted terminal token instead of the empty-generation result every other decode path in the crate produces for the same input. `BatchWorker` is not wired into either HTTP binary today, but it is the scheduler primitive ADR-048 (continuous batching) specifies as the crate's batched decode API.

**Suggested fix**: Mark zero-budget sequences `MaxLength` at admission time, or check the remaining token budget before the post-prefill sampling call in `BatchWorker`. Add a regression test asserting no token event is produced for a `max_new_tokens=0` request.

### LINKED ISSUE #916: speculative: make generate_with_speculation fallible on empty prompt_tokens (targets 0.7.0)

## Context

#856 / #915 unified every Qwen3.5 generation entry point on a typed
`Err(InferenceError::Inference("empty prompt"))` for an empty prompt via a
single shared preflight (`check_prompt_not_empty`,
`crates/inference/src/model/qwen35/generation.rs`).

`lattice_inference::speculative::generate_with_speculation` (public,
`crates/inference/src/speculative.rs`, re-exported at `lib.rs:99-100`) is a
documented exception to that unified contract: it returns
`Vec<u32>` (not a `Result`), and on an empty `prompt_tokens` slice it
returns `Vec::new()` -- a successful-looking empty result, not an error.

PR #915 review feedback (finding 1, MAJOR) flagged this as the one
remaining reachable fail-open path a caller who supplies an empty prompt
directly to this public helper. The correct fix -- making the function
fallible (`Result<Vec<u32>, InferenceError>`) so an empty prompt is
rejected the same way as every other entry point -- is a breaking
signature change to a public function. `lattice-inference` v0.6.1 just
shipped, so it is deferred to the next semver-major-eligible release
(0.7.0) rather than bundled into the #856/#915 fix, which was scoped to
be a non-breaking bugfix.

## What to do

- Change `generate_with_speculation`'s signature to return
  `Result<Vec<u32>, InferenceError>`.
- Reject `prompt_tokens.is_empty()` with
  `Err(InferenceError::Inference("empty prompt"))`, matching the shared
  guard's exact variant/message.
- Update `generate_with_speculation_empty_prompt_returns_empty` (the
  existing regression test locking in the current sentinel) to assert the
  new `Err` instead.
- Audit callers of `generate_with_speculation` within the crate (and any
  known downstream usage) for the signature change.
- Land as part of a 0.7.0-targeted PR, alongside the other already-known
  0.7.0 breaking changes (vision encoder Qwen3.5 recalibration ADR-069,
  `crate::generate` removal).

## Out of scope for this issue

`crate::generate::generate` (the separate deprecated top-level free
function) already returns a typed `Err` on empty prompt via its own
`InvalidInput` variant -- unrelated to this issue.

ohdearquant:

## Defects

- The live Qwen3.5 full-logit sampler now preserves the existing fail-closed policy (any NaN routes to finite argmax), but the canonical non-greedy `Sampler::sample` path sanitized NaN during top-k selection and could continue with a weighted draw. The issue owner's follow-up identified this as the remaining policy gap in #404.
- `BatchWorker` admitted `max_new_tokens=0` requests into prefill and sampled one token before checking the generated length.
- `generate_with_speculation` returned a successful empty vector for an empty prompt because its public signature was infallible, unlike the other generation entry points.

## Fix

- Reuse the shared full-logit NaN/non-finite pre-scan in canonical `Sampler::sample` after repetition penalties and before top-k selection, falling back to argmax for a poisoned distribution.
- Mark zero-budget batch requests `MaxLength` at admission and do not enqueue sampler or scheduler work.
- Change `generate_with_speculation` to return `Result<Vec<u32>, InferenceError>`, reject an empty prompt with `InferenceError::Inference("empty prompt")`, update all repository callers, and remove the obsolete documentation exception.

## Sibling paths checked

- The Qwen CPU decode and Metal CPU fallback paths share `sample_full_logits`, which already has the same NaN/non-finite scan. The canonical greedy/top-k=1 branch already routes directly through NaN-skipping argmax.
- The generic generator, Qwen3.5 single-sequence generators, and batch-prefill generator already return before sampling at a zero token budget. `BatchWorker` has two sampling calls; neither can receive a zero-budget sequence after the admission guard because such a sequence is never enqueued.
- Repository-wide caller search found no production caller of `generate_with_speculation`; all in-repository callers were its in-file tests and are updated for the fallible return.

## Regression sensitivity

- `test_sampler_nan_fails_closed_to_argmax` uses the issue owner's counterexample `[0.0, NaN, 0.0]` and seed `0x5eed_f00d_1234_5678`: before the fix it returned token 2, while the required argmax fallback is token 0.
- `step_zero_token_budget_emits_nothing` failed before the fix because one token event was emitted; it also asserts that the forward closure is never called.
- `generate_with_speculation_empty_prompt_returns_error` asserts the exact error variant/message and uses a panicking forward closure to prove rejection happens before decode work. Reverting the fallible signature makes the regression fail to compile against the expected `Result`.

## Verification

- `cargo test -p lattice-inference` (isolated target, shared build lock): passed; library result `1822 passed, 18 ignored`, with all binary, integration, and doc tests passing.
- `cargo clippy -p lattice-inference -- -D warnings` (isolated target, shared build lock): passed.
- `cargo fmt --all --check`: passed.
- `deno fmt --check docs/generation-entrypoint-matrix.md`: passed.
- `make lint-docs`: passed.

This adds a full-vocabulary scan to the canonical decode sampler hot path. `make bench-compare` is required before merge. It was not run in this leg, and no GPU was touched.

Closes #404
Closes #777
Closes #916


ohdearquant:
## ADR-058 bench evidence — sampler NaN pre-scan

The only decode-path change is a single NaN/non-finite pre-scan added to `Sampler::sample`
(the same scan already on the live Metal full-logit path at `sampling.rs:663`; this fills the
sibling path). A/B on the `candidate_sampler` group (`topk_readback` bench, quick: warm-up 1s,
measure 3s, sample-size 10/30):

| case | HEAD (#967) | main |
|------|-------------|------|
| k1   | 37.48 ns [37.25, 37.72] | 45.34 ns [41.24, 50.12] |
| k10  | 92.21 ns [91.77, 92.63] | 109.8 ns [98.18, 123.1] |
| k50  | 498.4 ns [495.7, 500.8] | (contended) |
| k256 | 3.42 us [3.41, 3.44] | -- |

**No regression** -- HEAD is <= main on every case with tighter CIs. Disclosure: the `main`
baseline run overlapped two concurrent fix legs compiling on this machine (hence its wide CIs,
the known shared-machine contention effect), so treat the A/B as directional. The structural
argument is the load-bearing one: the added op is one O(vocab) NEON-accelerated linear scan,
identical to a scan already characterized on the live path -- a fail-closed correctness guard,
not a hot-loop inner-loop change. Runnable: `cargo bench -p lattice-inference --bench topk_readback -- candidate_sampler`.

oceanwaves630 (approved):
Lane B approve (sampled: #967 NaN-fail-closed-to-argmax + zero-budget/empty-prompt guards; #970 sign-stable overflow-safe sigmoid/tanh restoring CPU parity). #971 is DIRTY — rebase to merge.

PR #968, opened 2026-07-14
fix(tune): fail closed on placeholder train steps + derive eligible LoRA indices

### LINKED ISSUE #10: tune: JitAdapter has placeholder CPU train_step and incomplete LowRank strategy
## Problem

`JitAdapter` in `crates/tune/src/train/jit.rs` has two incomplete implementations:

1. **`train_step` returns synthetic loss** — the CPU training path returns a simulated loss value rather than computing actual gradients
2. **`LowRank` strategy delta matrices are not created** — the low-rank decomposition path doesn't produce the actual A/B matrices

Surfaced during KG ingestion of the tune crate codebase.

## Impact

JIT adaptation is non-functional for actual training. The API surface exists but the internals are stubs.

## Suggested Fix

Either implement the training math or clearly mark the module as `#[cfg(feature = "experimental")]` so it's not compiled by default.

### LINKED ISSUE #637: bug(inference): bench_lora_mixture GPU decode bench crashes on Qwen3.5-0.8b hybrid attention layout (hardcoded q_proj on GDN layer)
## What happened

While run-verifying CLI walkthroughs for #602, `bench_lora_mixture`'s GPU decode
bench crashed against the real local `qwen3.5-0.8b-q4` checkpoint:

```
$ BENCH_NEW_TOKENS=8 LATTICE_MODEL_DIR=<q4-dir-with-tokenizer.json> ./target/release/bench_lora_mixture
...
bench_lora_mixture failed: Inference error: module 'q_proj' is a full-attention projection but layer 21 is GDN
```

## Root cause

`crates/inference/src/bin/bench_lora_mixture.rs::run_gpu_decode_bench` builds one
synthetic LoRA adapter per layer with `module: "q_proj".into()` for every
`layer_idx in 0..num_layers` unconditionally:

```rust
let adapters: Vec<Vec<LoraLayerData>> = (0..k)
    .map(|adapter_idx| {
        (0..num_layers)
            .map(|layer_idx| {
                ...
                LoraLayerData {
                    layer_idx,
                    module: "q_proj".into(),
                    ...
                }
            })
            .collect()
    })
    .collect();
```

Qwen3.5 is a hybrid architecture: most layers are `linear_attention` (GDN, no
`q_proj`) and only every Nth layer (`full_attention_interval` in `config.json`,
e.g. 4 for qwen3.5-0.8b) is `full_attention` and has a `q_proj`. For
qwen3.5-0.8b (`num_hidden_layers=24`, `full_attention_interval=4`), only layers
3, 7, 11, 15, 19, 23 are full-attention. `q_proj` on any other layer (e.g. 21,
a GDN layer) is rejected by the runtime, which correctly refuses to silently
apply an attention-shaped LoRA delta to a GDN layer.

The bin's own doc comment and a local constant also assume the wrong layer
count for this model family:

```rust
// Qwen3.5-0.8b has 28 transformer layers; use q_proj for the projection.
// For a realistic mix we cover all 28 layers × 2 modules (q_proj + v_proj).
const NUM_LAYERS: usize = 28;
```

qwen3.5-0.8b actually has `num_hidden_layers=24` per its `config.json`. The
CPU-only synthetic blend bench (`BLEND_BENCH` lines, using the hardcoded
`NUM_LAYERS=28` constant) does not crash since it never touches a real model,
but the layer count it reports is also not representative of qwen3.5-0.8b.

## Reproduction

```bash
cargo build --release -p lattice-inference --bin bench_lora_mixture --features f16,metal-gpu
cp ~/.lattice/models/qwen3.5-0.8b/tokenizer.json ~/.lattice/models/qwen3.5-0.8b-q4/tokenizer.json  # if missing
BENCH_NEW_TOKENS=8 LATTICE_MODEL_DIR=~/.lattice/models/qwen3.5-0.8b-q4 ./target/release/bench_lora_mixture
```

Deterministic: fails on the first `k`/`rank` combination that reaches a
non-full-attention layer index (layer 21 in this run), same failure every time
against this model.

## Suggested fix

`run_gpu_decode_bench` should read `layer_types` (or equivalent) from the
loaded `Qwen35Config` and only place `q_proj`/`v_proj` deltas on full-attention
layer indices (or use a full-attention-eligible module for GDN layers, if the
runtime supports one) instead of assuming every layer index has a `q_proj`.
Also correct the `NUM_LAYERS`/comment (24, not 28) for qwen3.5-0.8b, or derive
it from `cfg.num_hidden_layers` instead of hardcoding.

## Scope note

Found while writing `docs/cli-tools.md` for #602 (CLI tool walkthroughs). Not
fixed here — that PR documents the failure mode and command sequence, this
issue tracks the actual bin fix.

### LINKED ISSUE #738: bug(tune): TrainingLoop::train_batch returns a fabricated loss curve — non-functional stub presented as a trainer
**Defect.** `crates/tune/src/train/loop/mod.rs::TrainingLoop::train_batch()` does not train anything: it returns a hardcoded synthetic value (`0.5 * (-0.01 * global_step).exp() + 0.1` — the method's own comment calls it "Simulated loss (decreases over training)"). It also operates on the intent-classification `Batch`/`Dataset`/`IntentLabels` API, not prompt/completion pairs, so it is doubly disconnected from the working LoRA SFT path (`train_grad_full` / `lora/train_core.rs`).

**Impact.** Any caller (or future developer) that reaches for `TrainingLoop` gets a plausible-looking, monotonically-decreasing loss curve with zero learning behind it — the worst failure mode for a training API, because it looks like success. ADR-056 already warns the shipped code here is "partial and path-divergent" and directs readers to `train_grad_full.rs`; the code itself carries no such warning.

**Suggested fix (either):**
1. Remove `TrainingLoop` and the dead intent-classification training surface outright, leaving a pointer to the working `train_grad_full`/`train_core` path; or
2. If the API shell is worth keeping, make `train_batch()` return an explicit `Err(Unimplemented)` (and mark the type deprecated) instead of fabricating a loss value.

Either way, a fabricated-metric return must not survive: replace the simulated curve with a hard failure so misuse is loud.

ohdearquant:

## Defect

`TrainingLoop::train_batch` and `JitAdapter::train_step_cpu` returned plausible synthetic losses without updating any weights. The LoRA mixture benchmark also constructed `q_proj` adapters for every Qwen3.5 layer, including GDN layers where that projection is invalid, and its CPU blend setup assumed 28 layers for the 24-layer 0.8B model.

## Fix

- Return `TuneError::Training` from both unwired CPU training steps before they can fabricate metrics or advance training state.
- Derive full-attention layer indices from `Qwen35Config::layer_types` and use them for both the GPU decode adapters and the sibling CPU blend adapters.
- Derive the CPU benchmark's layer count from the selected full-attention projections instead of the stale hardcoded model-layer count.

## Sibling paths checked

The same-file CPU blend constructor also emitted `q_proj`/`v_proj` entries for every layer, so it now uses the identical full-attention index selection. `GpuTrainer::train_batch` already returns a training error for its unwired optimizer path. The remaining `LoraLayerData` constructors in `metal_qwen35.rs` are focused validation and blend tests, not production benchmark setup.

## Regression coverage

The tune regressions assert that the public training loop and low-rank JIT CPU adaptation return an explicit not-implemented training error; the training-loop test also pins `global_step == 0` and an empty metric history. Before the fix, both tests failed because the old code returned `Ok` synthetic metrics/loss. The benchmark regression pins the Qwen3.5-0.8B eligible indices to `[3, 7, 11, 15, 19, 23]`, so restoring all-layer selection fails the guard.

## Verification

- `cargo test -p lattice-tune` — 184 unit, 7 property, and 50 integration tests passed
- `cargo test -p lattice-inference --features f16,metal-gpu --bin bench_lora_mixture qwen35_0_8b_lora_layers_are_full_attention_only` — passed without initializing a GPU
- `cargo clippy -p lattice-tune -- -D warnings`
- `cargo clippy -p lattice-inference --features f16,metal-gpu --bin bench_lora_mixture -- -D warnings`
- repository pre-commit checks — workspace clippy and documentation lint passed

No GPU benchmark is required: this changes benchmark input construction and fail-closed CPU guards, not a decode, prefill, or GEMM hot path.

Closes #10
Closes #637
Closes #738

### DIFF
--- a/crates/inference/src/bin/bench_lora_mixture.rs
+++ b/crates/inference/src/bin/bench_lora_mixture.rs
@@ -11,12 +11,12 @@
 //! # Output
 //!
 //! ```text
-//! BLEND_BENCH r=1 k=1 layers=28 blend_us=<f>
-//! BLEND_BENCH r=1 k=4 layers=28 blend_us=<f>
-//! BLEND_BENCH r=1 k=8 layers=28 blend_us=<f>
-//! BLEND_BENCH r=2 k=1 layers=28 blend_us=<f>
-//! BLEND_BENCH r=2 k=4 layers=28 blend_us=<f>
-//! BLEND_BENCH r=2 k=8 layers=28 blend_us=<f>
+//! BLEND_BENCH r=1 k=1 layers=12 blend_us=<f>
+//! BLEND_BENCH r=1 k=4 layers=12 blend_us=<f>
+//! BLEND_BENCH r=1 k=8 layers=12 blend_us=<f>
+//! BLEND_BENCH r=2 k=1 layers=12 blend_us=<f>
+//! BLEND_BENCH r=2 k=4 layers=12 blend_us=<f>
+//! BLEND_BENCH r=2 k=8 layers=12 blend_us=<f>
 //! ```
 //!
 //! When a model is available:
@@ -49,9 +49,25 @@ fn main() {
     }
 }
 
+#[cfg(any(test, all(target_os = "macos", feature = "metal-gpu")))]
+fn full_attention_layer_indices(
+    cfg: &lattice_inference::model::qwen35_config::Qwen35Config,
+) -> Vec<usize> {
+    use lattice_inference::model::qwen35_config::LayerType;
+
+    cfg.layer_types
+        .iter()
+        .enumerate()
+        .filter_map(|(layer_idx, layer_type)| {
+            (*layer_type == LayerType::FullAttention).then_some(layer_idx)
+        })
+        .collect()
+}
+
 #[cfg(all(target_os = "macos", feature = "metal-gpu"))]
 fn run() -> Result<(), Box<dyn std::error::Error>> {
     use lattice_inference::forward::metal_qwen35::{LoraLayerData, blend_lora_layer_data};
+    use lattice_inference::model::qwen35_config::Qwen35Config;
     use std::time::Instant;
 
     let warmup: usize = std::env::var("BENCH_WARMUP")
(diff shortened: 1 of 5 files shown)


oceanwaves630 (approved):

PR #969, opened 2026-07-14
fix(serve): reject model mismatch + emit SSE error events on worker failure

### LINKED ISSUE #663: feat(serve): lattice_serve does not validate the request's model field

### LINKED ISSUE #933: bug(serve): mid-stream generation failure reported as finish_reason "stop" — indistinguishable from clean completion
**Contract gap**: both serve binaries report `finish_reason: "stop"` when a generation fails mid-stream (e.g. every token blocked by a grammar constraint), the same value used for a clean completion (`crates/inference/src/bin/lattice.rs:3911` and the sibling handler in `lattice_serve.rs`). Code comments show this was a deliberate mapping, but it makes the failure indistinguishable from success at the OpenAI-contract level.

**Blast radius**: a client gating on `finish_reason` (the documented signal for output trustworthiness) cannot tell a failed/truncated stream from a complete one; retried-vs-accepted decisions are made on wrong data. Telemetry side effect: completion-token counts undercount on the failure/rejection SSE paths.

**Suggested fix**: reserve `"stop"` for genuine stop conditions; map mid-stream failure to an SSE error event (or `finish_reason: "error"`-class signal consistent with the API surface we target), and align token accounting on those paths. Both binaries need the same change (sibling handlers).

ohdearquant:

## Defect

- The standalone `lattice_serve` daemon accepted a request whose `model` did not match the loaded model, unlike `lattice serve`.
- Both HTTP serve binaries mapped a mid-stream generation failure to `finish_reason: "stop"`, making failure indistinguishable from a genuine stop condition.

## Fix

- Reject mismatched daemon model ids before message conversion or worker submission with HTTP 400 and `model_not_found`.
- Emit the shared OpenAI-style `server_error` / `internal_error` envelope as an SSE data event on mid-stream failure, followed by `[DONE]`, without emitting a finish chunk.
- Preserve already-emitted daemon completion-token accounting by carrying the delta count through the SSE phase machine.
- Update the serve API, capability matrix, and v0.7.0 release notes for the wire-contract changes.

## Sibling paths checked

- In `lattice.rs`, CPU generation errors, generated-token invariant violations, and Metal worker drain failures all converge on `StreamMsg::Failed`; the single terminal mapping was fixed.
- In `lattice_serve.rs`, the `WorkerEvent::Failed` terminal plus the defensive post-commit `Rejected` and closed-worker paths now emit error envelopes. First-event and non-streaming failures already return ordinary HTTP error responses before an SSE stream is committed.
- The #654 capability-matrix expected-deviation fixture for model mismatch was converted to the rejecting contract and its manifest entries were updated.

## Regression proof

- `cm_lattice_serve_model_mismatch_rejected` drives the real daemon handler and asserts HTTP 400 plus `model_not_found`; removing the guard falls through to the disconnected fake worker and returns 500 instead.
- Both `chat_completions_streaming_failure_emits_error_event` tests inject a partial delta followed by a generation failure and assert the partial chunk, error envelope, and absence of `finish_reason: "stop"`. Reverting either terminal mapping leaves no error payload and fails the test.

## Verification

- `cargo test -p lattice-inference --features f16,metal-gpu,test-utils chat_completions_streaming_failure_emits_error_event`
- `cargo test -p lattice-inference --features f16,metal-gpu,test-utils cm_lattice_serve_model_mismatch_rejected`
- `cargo clippy -p lattice-inference --features f16,metal-gpu,test-utils --all-targets -- -D warnings`
- `make fmt-check`
- `make lint-docs`

No GPU benchmark is required: this changes request validation, SSE response framing, and telemetry state only; no decode, prefill, or GEMM hot path changed.

Closes #663
Closes #933


oceanwaves630 (approved):

PR #970, opened 2026-07-14
fix(fann): preserve extreme sigmoid and tanh inputs (#774)

### LINKED ISSUE #774: audit(norm-elementwise): 1 medium correctness finding — self-audit 20260709
- [ ] **Remove GPU-only Sigmoid and Tanh clipping**
  CPU Activation and the fann GPU shaders nominally implement the same Sigmoid and Tanh activations.
  Evidence: `crates/fann/src/gpu/shaders.rs:174`. At x=-100 CPU sigmoid is about 3.8e-44 while GPU computes sigmoid(-10), about 4.54e-5; at x=10 CPU tanh is about 1.0 while GPU computes tanh(5), about 0.9999092.
  Suggested fix: Implement sign-stable WGSL Sigmoid without clipping and native WGSL tanh without clipping; add a GPU-forced extreme-value CPU/GPU parity case.

---

ohdearquant:

## Defect

The FANN WGSL Sigmoid shader clamped inputs to `[-10, 10]`, and the Tanh shader clamped inputs to `[-5, 5]`. Those GPU-only clamps changed activation values for extreme inputs relative to the CPU implementations.

## Fix

- Evaluate Sigmoid with sign-dependent formulas so every exponent is non-positive without clipping the input.
- Evaluate native WGSL `tanh` on the original input.
- Strengthen the existing Sigmoid and Tanh parity cases with identity weights, extreme inputs, a tighter tolerance, and an assertion that the constructed network crosses the GPU dispatch threshold. When `gpu-tests` is enabled, absence of a real GPU now fails these cases instead of silently returning.
- Update CPU-only shader validation to reject clipping and verify the stable Sigmoid branch plus direct Tanh input.

## Sibling paths checked

The existing Sigmoid and Tanh parity tests were the sibling construction paths for this activation dispatch; both now use the same GPU-forcing identity-network helper. The stale `ShaderManager` Sigmoid source assertion was updated as well. No other GPU Sigmoid/Tanh shader definitions or clipping sites were found in `lattice-fann`.

## Regression sensitivity

Before the shader fix, `test_activation_shaders_preserve_extreme_inputs` failed with `Sigmoid should use a sign-stable branch`. Restoring either old clamp makes the CPU-only shader validation fail. On hardware, the parity cases compare Sigmoid at ±100 and Tanh at ±10 with `1e-6` tolerance; the old clamps introduce roughly `4.5e-5` and `9.1e-5` errors respectively.

## Validation

- `cargo test -p lattice-fann`
- `cargo test -p lattice-fann --features gpu --lib test_shader_type_source`
- `cargo test -p lattice-fann --features gpu --test cpu_gpu_parity --no-run`
- `cargo clippy -p lattice-fann -- -D warnings`

No GPU workload was executed on the shared machine. This is an element-wise activation correctness fix, not a decode/prefill/GEMM hot-path change, so `make bench-compare` is not required before merge.

Closes #774

### DIFF
--- a/crates/fann/src/gpu/shader_manager.rs
+++ b/crates/fann/src/gpu/shader_manager.rs
@@ -240,7 +240,7 @@ mod tests {
                 .source()
                 .contains("@compute")
         );
-        assert!(ShaderType::Sigmoid.source().contains("clamp"));
+        assert!(ShaderType::Sigmoid.source().contains("if (x >= 0.0)"));
     }
 
     #[test]
--- a/crates/fann/src/gpu/shaders.rs
+++ b/crates/fann/src/gpu/shaders.rs
@@ -152,7 +152,7 @@ fn main(@builtin(global_invocation_id) global_id: vec3<u32>) {
 
 /// Sigmoid activation with numerical stability
 ///
-/// Uses clamping to prevent exp() overflow
+/// Uses sign-dependent formulas to keep the exponent non-positive
 pub const SIGMOID_SHADER: &str = r#"
 struct Uniforms {
     size: u32,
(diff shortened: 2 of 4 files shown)


oceanwaves630 (approved):
Lane B approve (sampled: #967 NaN-fail-closed-to-argmax + zero-budget/empty-prompt guards; #970 sign-stable overflow-safe sigmoid/tanh restoring CPU parity). #971 is DIRTY — rebase to merge.

PR #972, opened 2026-07-14
fix(inference): validate LoRA adapter geometry in Qwen35Model::set_lora

### LINKED ISSUE #753: bug(inference): Qwen35Model::set_lora performs no shape/rank validation
**Defect**: The public CPU `Qwen35Model::set_lora` setter stores a `dyn LoraHook` with no validation at all — no rank, dimension, or layer-index check against the model's actual geometry. A model-aware `validate_against` method does exist, but it is a separate, optional call the caller must remember to invoke; nothing in `set_lora` itself calls it or requires it. Both Metal's attach path and the manifest-driven loader hard-check rank, dimensions, and target-shape before accepting an adapter.

**Evidence**:
- `crates/inference/src/model/qwen35/model.rs:65-67` — `pub fn set_lora(&mut self, hook: Box<dyn LoraHook>) { self.lora = hook; }` — bare field assignment, no checks.
- `crates/tune/src/lora/mod.rs:222-263` (`LoraAdapter::validate_against`) — checks layer_idx bounds and per-module expected `(d_in, d_out)` derived from `Qwen35Config`, but is a separate opt-in method.
- `crates/inference/src/forward/metal_qwen35.rs:6996-7075` (`load_lora_adapter`) — hard-checks rank=0, zero dims, `layer_idx >= num_layers`, and config-derived shape mismatch for every layer before accepting the adapter.
- `crates/tune/src/lora/apply.rs:23-25` — CPU apply's only guard is a `debug_assert_eq!`/`debug_assert!` pair (compiled out in release builds), then slices using the adapter's declared dimensions regardless of whether they match the model.
- `crates/inference/src/model/qwen35/forward.rs:279` — the ordinary CPU full-attention `q_proj` path invokes the attached (unvalidated) hook unconditionally.

**Blast radius**: Every CPU Qwen3.5 decode and batched-prefill caller that attaches an adapter through the public `set_lora` API. A mismatched adapter can silently change a projection's output prefix on every adapted invocation, or panic in a release build when declared output dimensions exceed the actual buffer; Metal and the governed-manifest path reject the same mismatch.

**Suggested fix**: Add a checked, typed `Qwen35Model` attachment method that centralizes the expected projection shapes and calls `validate_against` internally; mark the raw `dyn LoraHook` `set_lora` path as explicitly unchecked/advanced-use.

ohdearquant:

## Defect (#753)
`Qwen35Model::set_lora` stored any `dyn LoraHook` with zero rank/dim/layer-index validation, so a mismatched adapter silently corrupted a projection output prefix (or panicked past a `debug_assert` in release). Metal's attach path and the manifest loader already hard-check geometry; the CPU setter did not.

## Fix
- Add a **default-noop** `LoraHook::validate_against(&Qwen35Config) -> Result<(), String>` to the trait — backward compatible; existing hooks trust the caller unchanged.
- `set_lora` now calls it before swapping the hook in, returns `Result<(), InferenceError>`, and **keeps the previously-installed hook** if validation fails.
- `LoraAdapter` overrides `validate_against` to delegate to its existing inherent method.

## Sibling call sites (breaking signature change — all handled)
Grepped every `.set_lora(` in the repo: 1 binary caller (`generate_lora` — now `eprintln!` + `exit(1)` on mismatch) + 4 test call sites (updated to handle the `Result`). No unhandled caller remains.

## Mutation-sensitivity
The regression test uses fully-qualified `LoraHook::validate_against(&adapter, &cfg)` to force dispatch through the **trait** method. If the `impl LoraHook for LoraAdapter` delegation glue is removed, dispatch falls back to the trait's no-op default `Ok(())` and the `is_err()` assertion fails.

## Gates
Local pre-commit ran a full `cargo check` over the workspace (compiled clean, 3m18s), doc-lint, and capability-matrix — all passed. Not a decode/prefill/GEMM hot path, so no `bench-compare` required. Draft — do not merge until CI green.

Closes #753

### DIFF
--- a/crates/inference/src/lora_hook.rs
+++ b/crates/inference/src/lora_hook.rs
@@ -6,6 +6,8 @@
 //!
 //! The default `NoopLoraHook` does nothing — zero overhead when no adapter is loaded.
 
+use crate::model::qwen35_config::Qwen35Config;
+
 /// **Unstable**: trait for LoRA adapter injection into linear projections.
 ///
 /// The inference forward pass calls `apply()` after each `matmul_bt`.
@@ -24,6 +26,19 @@ pub trait LoraHook: Send + Sync {
     /// * `x` - Input activation (the same input that was passed to the base projection)
     /// * `output` - Base projection output to modify in-place
     fn apply(&self, layer_idx: usize, module: &str, x: &[f32], output: &mut [f32]);
+
+    /// **Unstable**: self-check this hook's declared rank/shape against a
+    /// Qwen3.5 model's geometry before it is installed.
+    ///
+    /// [`crate::model::qwen35::Qwen35Model::set_lora`] calls this before
+    /// swapping the hook in, so a mismatched adapter is rejected instead of
+    /// silently corrupting a projection's output prefix (or panicking past a
+    /// `debug_assert` in a release build). Default: no-op (trusts the
+    /// caller) — real adapters with known geometry (e.g.
+    /// `lattice_tune::lora::LoraAdapter`) override it.
+    fn validate_against(&self, _config: &Qwen35Config) -> Result<(), String> {
+        Ok(())
+    }
 }
 
 /// No-op implementation. Used when no adapter is loaded.
--- a/crates/inference/src/model/qwen35/model.rs
+++ b/crates/inference/src/model/qwen35/model.rs
@@ -61,8 +61,27 @@ impl Qwen35Model {
     }
 
     /// **Unstable**: attach a LoRA adapter hook; hook trait API may change.
-    pub fn set_lora(&mut self, hook: Box<dyn crate::lora_hook::LoraHook>) {
+    ///
+    /// Calls [`LoraHook::validate_against`](crate::lora_hook::LoraHook::validate_against)
+    /// against this model's own config before installing the hook, so a
+    /// rank/dimension-mismatched adapter is rejected up front instead of
+    /// silently corrupting projection outputs (or panicking past a
+    /// `debug_assert` in a release build). The previously-installed hook
+    /// stays active if validation fails.
+    ///
+    /// # Errors
+    ///
+    /// Returns `Err` (via [`InferenceError::InvalidInput`]) when the hook's
+    /// own `validate_against` reports a rank, dimension, or layer-index
+    /// mismatch against this model's geometry.
+    pub fn set_lora(
+        &mut self,
+        hook: Box<dyn crate::lora_hook::LoraHook>,
+    ) -> Result<(), InferenceError> {
+        hook.validate_against(&self.config)
+            .map_err(InferenceError::InvalidInput)?;
         self.lora = hook;
+        Ok(())
     }
 
     /// **Unstable**: access raw model weights.
--- a/crates/inference/src/model/qwen35/tests.rs
+++ b/crates/inference/src/model/qwen35/tests.rs
@@ -617,9 +617,11 @@ mod lora_serving {
         let mut model = build_model(cfg.clone(), 0xA11C_E5ED);
 
         let calls = Arc::new(Mutex::new(Vec::new()));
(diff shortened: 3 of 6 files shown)


ohdearquant:
Follow-up fix for the reviewer finding on `validate_against`/`LoraAdapter::new` (correctness/fail-closed).

**Finding recap**: `validate_against` checked layer index and projection `d_in`/`d_out` but never checked that a layer's `a`/`b` buffers actually matched `rank * d_in` / `d_out * rank` — the layout `apply_lora` indexes into. A custom-format adapter with correct projection dims but a malformed A/B buffer length could pass `set_lora` and later panic (slice out of bounds) or silently misapply.

**Where the check landed, and why**: in `LoraAdapter::new` (`crates/tune/src/lora/mod.rs`), not `validate_against`. `new()` is the *single* construction chokepoint — every adapter builder (`load_peft_safetensors`, `blend_lora_adapters`, `train.rs`, `online.rs`, `optimizer.rs`) routes through it, and `LoraAdapter`'s fields are private so nothing outside the `lora` module tree can construct one any other way. Closing the gap there means the invariant holds for *every* `LoraAdapter` instance unconditionally, not just ones that later happen to reach the `inference-hook`-gated `validate_against`/`set_lora` path. `validate_against`'s doc comment now notes it relies on this invariant instead of re-checking it.

One deliberate omission: I did **not** add a `layer.rank == config.rank` check. `blend_lora_adapters` intentionally produces adapters where `config.rank` is the *max* summed rank across all blended projections (used only to normalize `scale()` to 1.0) while individual layers keep their own, possibly smaller, per-projection summed rank — enforcing equality would reject legitimate blended adapters. The buffer-length check uses each layer's own `rank` field (what `apply_lora` actually slices by), which fully closes the OOB-panic risk without that false positive.

A layer with an empty `a` or `b` stays exempt from the length check — this mirrors the pre-existing "untrained/placeholder module" convention in `save_peft_safetensors` (which already skips serializing such layers), so it's not a new allowance.

**Mutation-sensitivity evidence**: reverse-applied just the new validation loop in `new()`, `touch`ed the file, and reran `cargo test -p lattice-tune --features "safetensors,serde,inference-hook" lora`. The 3 new regression tests (`test_new_rejects_a_buffer_shorter_than_rank_times_d_in`, `test_new_rejects_b_buffer_longer_than_d_out_times_rank`, `test_new_rejects_malformed_buffer_matching_real_model_projection_dims`) FAILED as expected (the last two also cascaded into a panic from another test's own `.expect`). Restored the fix and reran — all 109 `lattice-tune` lora tests pass.

**Sibling check**: grepped for every `LoraAdapter`/`LoraLayer` construction site in `crates/tune`. All roads (`from_safetensors` → `load_peft_safetensors_bytes`, `blend_lora_adapters`, `train.rs`, `online.rs`, `optimizer.rs`, the `train_grad*` binaries) go through `LoraAdapter::new`, so no bypass exists. The safetensors loader (`safetensors.rs`) already independently re-derives and checks rank consistency and buffer lengths per-layer before calling `new()` — redundant with this fix but harmless. `blend_lora_adapters`'s own `blend_layer_entries` also has its own defense-in-depth length/overflow checks on its *inputs*, which predate this change; 4 of its tests built a deliberately-malformed source adapter via `LoraAdapter::new(...).expect(...)` to exercise that defense-in-depth path — since `new()` now itself rejects that shape, those 4 tests were updated to construct the malformed `LoraAdapter` via a direct struct literal (allowed since the test module is a descendant of `lora` and the fields are private-not-pub) so they still reach `blend_lora_adapters`'s own checks.

**Gates**: `cargo clippy -p lattice-tune --all-targets [--features inference-hook] -- -D warnings` clean on both; `cargo test -p lattice-tune --features "safetensors,serde,inference-hook" lora` — 109 passed, 0 failed.

oceanwaves630 (approved):

PR #976, opened 2026-07-14
feat(serve): add /metrics Prometheus endpoint to lattice_serve

### LINKED ISSUE #583: feat(serve): /metrics Prometheus endpoint for lattice_serve
## Context

`lattice_serve` currently advertises only three endpoints — `POST /v1/chat/completions`, `GET /v1/models`, `GET /health` (endpoint list at `crates/inference/src/bin/lattice_serve.rs:298-304`, registered on the router at `crates/inference/src/bin/lattice_serve.rs:677-681`). Per-request telemetry exists (`emit_serve_event`) but only goes to stdout — there's no way to scrape it into a metrics backend.

Operators running `lattice_serve` beyond a single interactive session have no visibility into throughput, latency distribution, or error rate without tailing logs.

## Acceptance criteria

- [ ] `GET /metrics` route exposing Prometheus text-format metrics
- [ ] Request count and latency histogram, labeled by route/status
- [ ] Prompt tokens and completion tokens counters
- [ ] Queue depth / in-flight request gauge
- [ ] Error count by error code
- [ ] Model id as a metric label
- [ ] Endpoint documented in the binary's module doc comment and startup banner

## Notes

Table-stakes parity with vLLM/Ollama-style servers. Related to #93 (the original OpenAI-compatible serve surface this extends).

ohdearquant:

## Defect / gap

`lattice_serve` advertised only `POST /v1/chat/completions`, `GET /v1/models`, and `GET /health` (endpoint list at `lattice_serve.rs:21-25`, router at `lattice_serve.rs:1846-1852`). Per-request telemetry existed via `emit_serve_event`, but it only printed a `@@lattice {...}` JSON line to stdout for the app bridge — there was no way to scrape throughput/latency/error data into a metrics backend for anyone running `lattice_serve` beyond an interactive session.

## Fix

- New `lattice_inference::serve::metrics::ServeMetrics` registry (dependency-free: a `HashMap<label, count>` behind `std::sync::Mutex` + a couple of atomics — this server exposes O(10) series scraped every few seconds, so it didn't need the `prometheus` crate). Renders Prometheus text-exposition format:
  - `lattice_http_requests_total{method,route,status,model}` — counter
  - `lattice_http_request_duration_seconds{route,model}` — histogram (standard-ish buckets, 5ms–60s)
  - `lattice_prompt_tokens_total{model}` / `lattice_completion_tokens_total{model}` — counters
  - `lattice_errors_total{code,model}` — counter, keyed by the same OpenAI-style error code the client sees in the response body
  - `lattice_inflight_requests{model}` — gauge
- `emit_serve_event` (the single per-request telemetry call site, already threaded through every response path in `chat_completions`/`health`/`root`/`list_models`) now forwards each observation into `ServeMetrics` in addition to printing the existing `@@lattice` stdout line, so the two consumers can never observe a different request count for the same traffic.
- The in-flight/queue-depth gauge is a **live read of the shared worker's admission semaphore** (`MetalWorkerClient::available_permits`, new method) computed as `max_pending - available_permits()` at scrape time — not a separately-tracked counter that could drift from the real admission state (the semaphore permit is already held from `submit` until `run_worker_loop` is fully done with the job, so this reflects genuinely outstanding work).
- Streaming responses previously only tracked `completion_tokens` through the SSE `Phase` state machine (`Phase::Done`/`Phase::End` carried one `usize`). Extended both variants to carry `(completion_tokens, prompt_tokens)` so `/metrics` gets prompt-token counts for streaming requests too, matching the non-streaming path.
- `ApiError` gained a `code()` accessor mirroring exactly the string each variant's `IntoResponse` impl already serializes as `error.code`, so error codes can be recorded without duplicating the per-variant mapping.
- `GET /metrics` documented in the binary's module doc comment (`# Endpoints`) and the startup banner.

## Sibling paths checked

`lattice.rs` (the other OpenAI-compatible server binary) has no `emit_serve_event` equivalent at all and is not mentioned in issue #583 (titled specifically for `lattice_serve`), so no sibling fix applies there. Grepped the crate for any other `println!("@@lattice"` telemetry call sites — `emit_serve_event` is the only one.

## Regression test

`metrics_endpoint_reports_requests_tokens_and_errors` (in `lattice_serve.rs`) drives a successful `chat_completions`, a 400 (empty `messages`), and a `health` check through the real handlers, then asserts the `/metrics` body reflects each: request counters by (method, route, status), prompt/completion token totals, and the `invalid_messages` error count.

**Verified mutation-sensitive**: commented out `emit_serve_event`'s three `metrics.record_*()` calls, `touch`ed the file to force a rebuild, and reran — the test failed (`missing successful chat_completions counter`) while the route itself still returned 200/400 as before, confirming the test actually exercises the metrics-recording wiring and not just route existence. Restored the calls and reran — 44/44 tests pass.

## Verification

```
cargo test -p lattice-inference --features f16,metal-gpu,serve,test-utils --bin lattice_serve
  → test result: ok. 44 passed; 0 failed
cargo clippy -p lattice-inference --features f16,metal-gpu,serve,test-utils -- -D warnings
  → clean
cargo fmt -p lattice-inference -- --check
  → clean
```

No GPU/Metal touched — this is HTTP-layer instrumentation only (no decode/prefill/GEMM path changed), so no `make bench-compare` is required.

Closes #583

### DIFF
--- a/crates/inference/src/bin/lattice_serve.rs
+++ b/crates/inference/src/bin/lattice_serve.rs
@@ -23,6 +23,10 @@
 //! - `POST /v1/chat/completions` — streaming (SSE) and non-streaming, OpenAI shape
 //! - `GET  /v1/models`           — advertises the single loaded model
 //! - `GET  /health`              — liveness probe (`ok`)
+//! - `GET  /metrics`             — Prometheus text-format metrics (issue #583):
+//!   request count + latency histogram by route/status, prompt/completion
+//!   token counters, error count by code, queue-depth/in-flight gauge, all
+//!   labeled with the loaded model id
 //!
 //! # Design
 //!
@@ -79,6 +83,7 @@ mod imp {
         ContextWindowPolicy, MetalWorker, MetalWorkerClient, StartupError, WorkerEvent,
         WorkerMetadata,
     };
+    use lattice_inference::serve::metrics::ServeMetrics;
     /// Only used by the test module's `.tokenize(..)` calls on a real (tiny)
     /// tokenizer; production code never tokenizes outside the shared worker
     /// (`lattice_inference::serve::metal_worker`), hence the test feature gate.
@@ -152,6 +157,15 @@ mod imp {
         /// exact KV cache length `load_model` allocated, never a hard-coded
         /// constant. See `model_context_from_config` and `build_cfg`.
         model_max_context: usize,
+        /// The shared worker's outstanding-job admission cap (issue #932),
+        /// needed alongside `jobs.available_permits()` to compute the
+        /// `/metrics` (#583) queue-depth/in-flight gauge as `max_pending -
+        /// available_permits()`.
+        max_pending: usize,
+        /// Process-wide Prometheus metrics registry (#583), `Arc`-shared so
+        /// every clone of `AppState` (one per request) records into the same
+        /// counters.
+        metrics: Arc<ServeMetrics>,
     }
 
     // ─── OpenAI request shapes ───────────────────────────────────────────────
@@ -1025,31 +1039,37 @@ mod imp {
 
     // ─── HTTP handlers ───────────────────────────────────────────────────────
 
-    async fn health() -> &'static str {
+    async fn health(State(s): State<AppState>) -> &'static str {
         let t = Instant::now();
         emit_serve_event(
+            &s.metrics,
             "GET",
             "/health",
             200,
             None,
+            None,
             t.elapsed().as_secs_f64() * 1000.0,
             false,
+            None,
         );
         "ok"
     }
 
-    async fn root() -> Json<Value> {
+    async fn root(State(s): State<AppState>) -> Json<Value> {
         let t = Instant::now();
         // ADR-080 C2: shared with lattice.rs's equivalent route so
         // both binaries advertise the same engine-identity document.
         let body = lattice_inference::serve::root_body();
         emit_serve_event(
+            &s.metrics,
             "GET",
             "/",
             200,
             None,
+            None,
             t.elapsed().as_secs_f64() * 1000.0,
             false,
+            None,
         );
         Json(body)
     }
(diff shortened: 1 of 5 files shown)


ohdearquant:
Fixed reviewer findings (a) and (b) in af80f2d (pushed to this branch).

## Finding (b) — streaming failures recorded as fake 200s (fixed)

`Phase::Done`/`Phase::End` in `chat_completions`'s SSE handler now carry an
`Option<(u16, &'static str)>` error outcome, threaded from each of the three
worker-failure arms:

- `WorkerEvent::Failed` → `Some((500, "internal_error"))`
- `WorkerEvent::Rejected(api_err)` → `Some((400, api_err.code()))`
- worker-gone (`None` from `rx.recv()`) → `Some((500, "internal_error"))`
- `WorkerEvent::Complete` (success) → `None`

`Phase::End`'s `emit_serve_event` call now reads this outcome instead of a
hardcoded `200`/`None`, so a mid-stream failure is recorded with the correct
status and `lattice_errors_total{code=...}` actually increments. Each status/
code choice mirrors **exactly** what the non-streaming branch in the same
file already records for the identical failure class (`Failed`/worker-gone →
500 `internal_error`; `Rejected` → 400 + `api_err.code()`) — streaming and
non-streaming now agree on how the same failure is recorded. The wire-level
HTTP status is unchanged (still committed 200 `text/event-stream` before
streaming starts, as before); only the *recorded metric* now reflects the
true logical result. The client-facing SSE error envelope is untouched.

**Sibling check**: grepped `lattice_serve.rs` for every `Sse::new` and
`enum Phase` occurrence — there is exactly one of each. `chat_completions` is
the only streaming endpoint in this binary (no `/v1/completions` or
`/v1/embeddings` routes exist here), so the three defective arms were all in
the single match block fixed above; there are no other copy-pasted terminal-
phase sites to fix. Also checked every other `emit_serve_event` call site in
the file (health/root/list_models/all `chat_completions` early-return arms) —
none of them hardcode a success status on a failure path.

**Mutation-sensitive regression test**: added
`chat_completions_streaming_failure_records_failed_metric`, which drives a
`WorkerEvent::Failed` through the stream and asserts the `/metrics` output
shows `status="500"` (not `200`) and
`lattice_errors_total{code="internal_error"} 1`. Verified fail→pass by hand:
reverse-applied only the `Phase::Done`/`Phase::End` threading hunk (keeping
the new test in place), `touch`ed the file, ran the test — it failed against
the unpatched code (recorded `status="200"`, no `lattice_errors_total` line).
Reapplied the fix, reran — passes.

## Finding (a) — `/metrics` has no auth (documented, not gated)

Unauthenticated `/metrics` was previously accepted as-is for the current
local-first deployment surface, so no auth was added here. Instead:

- Added a doc comment on `metrics_handler` stating the deployment boundary:
  no auth, safe by default because `--host` defaults to `127.0.0.1`, but a
  non-loopback `--host` exposes request volume/latency/error-rate shape to
  the network — don't do that without an external auth layer.
- Added the same note to `docs/serve-http-api.md` (which previously listed
  `lattice_serve`'s route set without `/metrics` and explicitly disclaimed
  covering `lattice_serve` — updated the route list and carved out this one
  exception).
- Added a one-time startup `eprintln!` warning when `--host` is bound to
  anything other than `127.0.0.1`/`localhost`/`::1`, pointing at the same
  guidance. Low-risk: it's a single conditional stderr line at startup, no
  behavior change.

## Gates

  (also reran with `--features f16,metal-gpu,serve,test-utils` to cover the
  cfg-gated test code — also clean).
- `cargo test -p lattice-inference serve` — 50+77 passed, 0 failed (default
  features). Full `lattice_serve` binary test suite under
  `--features f16,metal-gpu,serve,test-utils` — 45 passed, 0 failed,
  including the new mutation-sensitive test.
- `cargo fmt -p lattice-inference -- --check` — clean.

## Bench-compare (ADR-058)

Not applicable: this touches only the serve control-plane / metrics-recording
path (an `enum Phase` payload and `emit_serve_event` argument selection), not
`crates/inference/src/forward` or any hot kernel. `make bench-compare` would
show no change (structural) — no inference hot path was touched.

oceanwaves630 (approved):

Issue #420, opened 2026-06-27
feat(inference): W3 (3-bit) MLP-only weight path -- cheapest sub-4-bit decode byte reduction


ohdearquant:
## Goal
Add a W3 (3-bit) weight-only quantization path for **MLP projections only** (gate/up/down), the largest decode byte pool.

## Why MLP, why W3
Decode is memory-bandwidth-bound (#278). Fresh profile (Qwen3.5-0.8B Q4, Metal, this session): MLP = **35.3%** of decode time, the single largest group (gdn_mixer 32.2%, gqa_attn 28.0%, lm_head 4.6%) -- all weight-bandwidth. Reducing MLP weight bytes 4->3 bits is the most direct decode lever, and one of only two live decode levers (the other is speculative decode; khive memory e4e147af).

## Scope
- W3 block format (analogous to Q4Block) for MLP gate/up/down only; leave attention/GDN at Q4/Q8 (role-aware -- see the mixed-precision policy issue).
- Quality gate: per-layer + e2e-parity. W3 weight-only PPL delta budget vs Q4: target < 0.1-0.2 (llama.cpp community bounds). If MLP-only W3 blows the budget, the role-aware policy issue covers a more selective split.
- Measure-first: `make bench-compare` decode tok/s before/after, plus PPL.

## Not in scope
Full QTIP / trellis pipeline. This is the cheapest standalone W3, not a quantizer rewrite. SpinQuant (separate issue) may be needed as a quality enabler if randomized-Hadamard QuaRot leaves insufficient W3 headroom.

PR #978, opened 2026-07-14
feat(inference): W3 Q3Block 3-bit weight format (ADR-072 P1, stage 1)

ohdearquant:

## Stage 1 of ADR-072 P1 (W3 3-bit MLP weights)

CPU-side 3-bit block-quantization **format module** for MLP weights (gate/up/down) — the storage/bandwidth half of ADR-072 P1. **No engine change**: this lands the format (encode/decode/save/load + round-trip goldens) that the Metal MLP dequant kernel (Stage 2) will build on. Dead code until Stage 2 wires it.

### Format

`Q3Block` = **16 bytes**: `f16` scale + `f16` bias + `[u8; 12]` packing 32 3-bit codes via the **plane-split 2+1** layout:
- low-2-bit plane in bytes `0..8` (`byte[b] = q2[4b] | q2[4b+1]<<2 | q2[4b+2]<<4 | q2[4b+3]<<6`)
- high-1-bit plane in bytes `8..12` (`byte[8+b] = sum(hi[8b+k] << k)`)

This gives **4.0 bpw** at group-32 vs Q4 group-32 at 5.0 bpw — a ~20% weight-traffic cut on the MLP GEMMs (~35% of decode). Asymmetric (`scale=(max-min)/7, bias=min`) is the default; symmetric mode is available. Magic `KHQ3`, version 1. Attention/GDN weights stay Q4/Q8.

### Tests (13, all pass)

block-size invariant · plane-split pack/unpack over all 3-bit values · high-bit-plane mutation-sensitivity · asymmetric + symmetric round-trip bounds · partial-block tail (real min/max) · constant + zero blocks exact · non-finite rejection · shape/data-length mismatch rejection · bf16 path parity with f32 · save/load round-trip preserves blocks + shape · bad-magic rejection · 4.0-bpw storage assertion.

### Sequencing

Per the signed design spec, this Stage-1 format module merges first; the Stage-2 Metal dequant kernel depends on it. ADR-072 promotion (Proposed to Accepted) with the Q3Block format folded in as an amendment is a prerequisite for **Stage 2**, not this PR.

Bench-compare: N/A — no runtime path changed (format module + tests only). Stage 2 will carry the MLP decode A/B.

Relates to #420 / ADR-072.

### DIFF
--- a/crates/inference/src/weights/mod.rs
+++ b/crates/inference/src/weights/mod.rs
@@ -1,8 +1,9 @@
-//! Weight module index for f16, f32, q4, and q8 weights, with f32 weight re-exports.
+//! Weight module index for f16, f32, q3, q4, and q8 weights, with f32 weight re-exports.
 pub mod f16_weights;
 pub mod f32_weights;
 pub(crate) mod half_bits;
 pub(crate) mod ingress;
+pub mod q3_weights;
 pub mod q4_weights;
 pub mod q8_weights;
 
--- a/crates/inference/src/weights/q3_weights.rs
+++ b/crates/inference/src/weights/q3_weights.rs
@@ -530,39 +530,50 @@ fn checked_alloc_bytes(
 /// # Errors
 ///
 /// Returns an error on I/O failure, unrecognized magic bytes, or unsupported version.
-pub fn read_q3_header(file: &std::fs::File) -> Result<Q3FileHeader, Box<dyn std::error::Error>> {
-    use std::io::Read;
+pub fn read_q3_header(
+    file: &mut std::fs::File,
+) -> Result<Q3FileHeader, Box<dyn std::error::Error>> {
+    use std::io::{Read, Seek, SeekFrom};
     let file_len = file.metadata()?.len();
-    let mut f = std::io::BufReader::new(file);
 
-    let mut magic = [0u8; 4];
-    f.read_exact(&mut magic)?;
-    if &magic != b"KHQ3" {
-        return Err("invalid magic: not a .q3 file".into());
-    }
+    let (shape, original_len, payload_offset) = {
+        let mut f = std::io::BufReader::new(&mut *file);
 
-    let mut b4 = [0u8; 4];
-    f.read_exact(&mut b4)?;
-    let ver = u32::from_le_bytes(b4);
-    if ver != 1 {
-        return Err(format!("unsupported .q3 file version: {ver}").into());
-    }
+        let mut magic = [0u8; 4];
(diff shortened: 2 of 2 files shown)


ohdearquant:
Fixed the `read_q3_header` cursor-promise bug (commit e2d461e5178a5b6527845ab2d55d0c9558a90b3a).

**Fix**: `read_q3_header` now takes `&mut std::fs::File`. Header fields are parsed through a `BufReader<&mut File>` scoped in a block so it drops before the seek; the function then does `file.seek(SeekFrom::Start(payload_offset))?` on the raw `File` before returning. This makes the doc promise ("cursor positioned at the start of the block data") actually true — previously the `BufReader`'s internal ~8KB fill silently advanced the shared fd past `payload_offset` on any header smaller than the buffer. The single test caller (`save_load_roundtrip_preserves_blocks_and_shape`) now passes `&mut file`.

**Mutation-sensitive regression test**: added `read_q3_header_leaves_cursor_at_payload_offset`, which opens a real written `.q3` file, calls `read_q3_header(&mut file)`, then reads the first block bytes directly off the *same* handle (no re-open, no seek) and asserts they equal the tensor's first block, plus asserts `stream_position() == payload_offset`.

Verified fail-then-pass by reverse-applying just the seek line (kept everything else, i.e. restored the "BufReader-without-seek" behavior), `touch`ing the source, and re-running:
- **Reverted**: test FAILED — `assertion left == right failed: cursor must sit at payload_offset after read_q3_header returns / left: 84 / right: 36` (84 = end of the BufReader's slurp past the 36-byte header for this fixture's shape).
- **Restored**: test PASSED, along with all 14 `q3` tests.

**Sibling check**: grepped `crates/inference/src/weights/` for other `BufReader<&File>`-based header readers with a false cursor-position doc promise. `load_q3_file` re-reads the header inline (doesn't consume the `read_q3_header` cursor), so it's unaffected. Found one true sibling: `read_q4_header` in `q4_weights.rs:606` has the identical doc claim ("On return the file cursor is positioned at the start of the block data") and the identical `BufReader::new(file)`-without-seek pattern — same latent bug. Left untouched here since this PR's scope is `q3_weights.rs` only; flagging as a follow-up.

**Gates**: `cargo clippy -p lattice-inference --all-targets -- -D warnings` clean. `cargo test -p lattice-inference q3` — 14/14 passed.

**Bench-compare**: not run. Cold load-path IO + test only; no hot kernel touched, `make bench-compare` would show no change (structural).

oceanwaves630 (approved):

PR #980, opened 2026-07-14
docs(transport): condense inline comments, extract background to crate docs

ohdearquant:

Implements Ocean's 2026-07-10 fleet-wide comment-hygiene directive for `crates/transport/src` only. This is comment/doc-only, with zero behavior change.

## Counts

- Source comment/doc-comment lines: 855 before → 431 after (424 fewer).
- Source diff: 49 concise comment/doc lines added or condensed; 473 verbose, redundant, or stale lines removed.

## What changed

- Condensed module, public-API, and numerical-helper prose to contracts and invariants the code cannot state alone.
- Deleted repeated provisional-stability boilerplate, next-line narration, and test comments that restated assertions.
- Extracted operational background from module rustdoc into `crates/transport/docs/algorithms.md` and linked it from the crate index and relevant module docs.

## New docs

- `crates/transport/docs/algorithms.md`

## ADR candidates

- Divergence must use regularized cost, rather than raw primal cost, to preserve symmetry (`src/divergence.rs`).
- Balanced solves reject non-positive and zero-total marginals instead of allowing log-floor clamping to create phantom mass (`src/sinkhorn.rs`).
- Convergence uses L1 marginal residuals rather than dual-update size (`src/sinkhorn.rs`).
- The 16 MiB materialization threshold for point-set divergence is an explicit performance/memory policy (`src/divergence.rs`).
- The 4096 online-drift window cap bounds three O(W²) workspaces (`src/online_drift.rs`).

All required gates are green: `RUSTDOCFLAGS='-D warnings' cargo doc --no-deps -p lattice-transport`, `cargo clippy -p lattice-transport --all-targets -- -D warnings`, and `cargo test -p lattice-transport`.

### DIFF
--- a/crates/transport/docs/INDEX.md
+++ b/crates/transport/docs/INDEX.md
@@ -6,3 +6,5 @@ the summary and public API surface; the files here hold the longer-form narrativ
 
 - [design.md](design.md) — architecture layers, the log-domain invariant and why it
   matters, workspace preallocation, and the no-external-linear-algebra constraint.
+- [algorithms.md](algorithms.md) — operational background for divergence, sparse
+  plans, barycenters, unbalanced transport, and online drift.
--- a/crates/transport/docs/algorithms.md
+++ b/crates/transport/docs/algorithms.md
@@ -1,0 +1,29 @@
+# lattice-transport algorithms
+
+This page collects operational background for transport APIs whose source-level
+documentation is deliberately brief.
+
+## Debiased divergence
+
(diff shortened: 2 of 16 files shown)


ohdearquant:
Superseded — this automated comment-condensation sweep over-removed comments that must be preserved (SAFETY justifications on unsafe blocks, defense-in-depth security-check rationale, math/formula comments, and public-item doc comments). Redoing with a strict protected-comment allowlist and pilot verification first.

PR #988, opened 2026-07-14
feat(inference): parse Qwen3.5 vision config and load model.visual tensors (ADR-069 S1+S2)

ohdearquant:
## Summary

Implements ADR-069 stages S1 (config parsing) and S2 (weight loading). No forward-pass work — S3+ is out of scope for this PR.

### S1 — config parsing

`Qwen35Config` previously dropped the vision-language fields present in the Qwen3.5-0.8B checkpoint's `config.json`. This PR parses them:

- `vision_config` (top-level sibling of `text_config`): `depth`, `hidden_size`, `num_heads`, `patch_size`, `spatial_merge_size`, `out_hidden_size`, `temporal_patch_size`, `num_position_embeddings`, `in_channels`, `deepstack_visual_indexes` — a new `VisionModelConfig` struct, `Option<VisionModelConfig>` on `Qwen35Config`.
- The four vision/image token ids: `image_token_id`, `video_token_id`, `vision_start_token_id`, `vision_end_token_id`.
- `text_config.rope_parameters.{mrope_section, mrope_interleaved}`.

All new fields are optional/defaulted, so text-only checkpoints (no `vision_config`) continue to parse unchanged — covered by a new test that asserts every vision field resolves to `None` while an existing non-vision field still parses correctly. The real 0.8B checkpoint's config fixture already round-trips through `Qwen35Config::from_config_json_str`; the existing fixture test now asserts every new field's exact value (depth 12, hidden_size 768, num_heads 12, patch_size 16, spatial_merge_size 2, out_hidden_size 1024, temporal_patch_size 2, num_position_embeddings 2304, in_channels 3; image_token_id 248056, video_token_id 248057, vision_start_token_id 248053, vision_end_token_id 248054; mrope_section [11,11,10], mrope_interleaved true).

### S2 — weight loading

Adds `crates/inference/src/vision/checkpoint.rs`: a loader for the 153 real `model.visual.*` tensors (`patch_embed.proj.{weight,bias}` + `pos_embed.weight` + 12 ViT blocks × 12 tensors + 6 merger tensors), following the same conventions already used for the text decoder's weights — `ShardedSafetensors`/`TensorSource` for the fp16 sharded-safetensors form, and the `quantize_index.json` manifest + per-tensor `.q4`/`.f16` file convention for the q4 form.

**q4-directory finding**: on inspection, the local q4 checkpoint directory already retains all 153 `model.visual.*` tensors — quantized weight matrices as `.q4`, biases/norms as sibling `.f16` files, using the exact same naming convention as the text decoder's tensors in that same directory. This contradicts an assumption that the quantize pipeline historically filtered vision tensors out of the q4 build; for this checkpoint it does not. No fp16-fallback plumbing was needed — S2 loads q4-format vision tensors directly, the same way it loads q4-format text tensors.

### Scope boundary

This PR does **not** modify `vision/config.rs`, `vision/vit.rs`, `vision/merger.rs`, or `vision/multimodal.rs` — the existing ADR-049 CPU ViT scaffold (targets a hypothetical 7B geometry, has no callers outside its own tests). Reconciling that scaffold's forward-pass semantics to the real 12-layer/768-hidden geometry — including its temporal-patch-folded patch embedding (`[hidden, in_channels, temporal_patch_size, patch_size, patch_size]` vs. the scaffold's `[d_model, patch_size^2 * 3]` assumption) and dropping its windowed-attention assumption (the real checkpoint has no window-specific weights) — is ADR-069 stage S3 ("Metal ViT forward"), not S1/S2.

## Test evidence

All three required gates pass (default features):

```
cargo clippy -p lattice-inference --all-targets -- -D warnings   → clean
cargo test -p lattice-inference vision                           → 47 passed; 0 failed (2 local-checkpoint tests skip when absent)
cargo test -p lattice-inference config                           → 134 passed; 0 failed; 1 ignored
```

Also verified with `--features f16` (needed to decode the BF16 patch_embed/pos_embed/etc. tensors from the fp16 checkpoint): the fp16-gated checkpoint test passes, and clippy stays clean with `--features f16` and with `--features f16,metal-gpu,serve,bench-internals`.

The S2 integration tests load the real local checkpoints (`~/.lattice/models/qwen3.5-0.8b` and `~/.lattice/models/qwen3.5-0.8b-q4`, both skipped when absent, mirroring the existing local-model test-skip convention in this crate) and assert exact tensor count (153) plus shape spot-checks on `patch_embed`, one full ViT block, and the merger.

## Bench

bench-compare not run: load-time config/weight plumbing, no decode/prefill path touched (structural).

### DIFF
--- a/crates/inference/src/forward/batch_prefill.rs
+++ b/crates/inference/src/forward/batch_prefill.rs
@@ -1741,6 +1741,11 @@ mod tests {
             mtp_num_hidden_layers: 0,
             mtp_use_dedicated_embeddings: false,
             quarot_rotation_seed: None,
+            vision_config: None,
+            image_token_id: None,
+            video_token_id: None,
+            vision_start_token_id: None,
+            vision_end_token_id: None,
         }
     }
 
--- a/crates/inference/src/forward/cpu_f16.rs
+++ b/crates/inference/src/forward/cpu_f16.rs
@@ -1125,6 +1125,11 @@ mod tests {
             mtp_num_hidden_layers: 0,
             mtp_use_dedicated_embeddings: false,
             quarot_rotation_seed: None,
+            vision_config: None,
+            image_token_id: None,
+            video_token_id: None,
+            vision_start_token_id: None,
+            vision_end_token_id: None,
         };
 
         let rope_dim = cfg.rope_dim(); // = 16
@@ -1300,6 +1305,11 @@ mod tests {
             mtp_num_hidden_layers: 0,
             mtp_use_dedicated_embeddings: false,
             quarot_rotation_seed: None,
+            vision_config: None,
+            image_token_id: None,
+            video_token_id: None,
+            vision_start_token_id: None,
+            vision_end_token_id: None,
         };
 
         let rope = RopeTable::new(cfg.rope_dim(), 512, cfg.rope_theta);
@@ -1629,6 +1639,11 @@ mod tests {
             mtp_num_hidden_layers: 0,
             mtp_use_dedicated_embeddings: false,
             quarot_rotation_seed: None,
+            vision_config: None,
+            image_token_id: None,
+            video_token_id: None,
+            vision_start_token_id: None,
+            vision_end_token_id: None,
         };
 
         // embed_tokens is [vocab * hidden] packed u16 (f16 zeros = 0u16).
@@ -1776,6 +1791,11 @@ mod tests {
             mtp_num_hidden_layers: 0,
             mtp_use_dedicated_embeddings: false,
             quarot_rotation_seed: None,
+            vision_config: None,
+            image_token_id: None,
+            video_token_id: None,
+            vision_start_token_id: None,
+            vision_end_token_id: None,
         };
 
         // The negative gamma at dim-0 flips the sign of that component after
--- a/crates/inference/src/forward/cpu_q8.rs
+++ b/crates/inference/src/forward/cpu_q8.rs
@@ -935,6 +935,11 @@ mod tests {
             mtp_num_hidden_layers: 0,
             mtp_use_dedicated_embeddings: false,
             quarot_rotation_seed: None,
+            vision_config: None,
+            image_token_id: None,
+            video_token_id: None,
+            vision_start_token_id: None,
+            vision_end_token_id: None,
         };
 
         let rope_dim = cfg.rope_dim(); // = 16
@@ -1205,6 +1210,11 @@ mod tests {
             mtp_num_hidden_layers: 0,
             mtp_use_dedicated_embeddings: false,
             quarot_rotation_seed: None,
+            vision_config: None,
+            image_token_id: None,
+            video_token_id: None,
+            vision_start_token_id: None,
+            vision_end_token_id: None,
         };
 
         let rope = RopeTable::new(cfg.rope_dim(), 512, cfg.rope_theta);
@@ -1416,6 +1426,11 @@ mod tests {
             mtp_num_hidden_layers: 0,
             mtp_use_dedicated_embeddings: false,
             quarot_rotation_seed: None,
+            vision_config: None,
+            image_token_id: None,
+            video_token_id: None,
+            vision_start_token_id: None,
+            vision_end_token_id: None,
         };
 
         // embed_tokens is [vocab, hidden] and also serves as the tied LM head.
@@ -1789,6 +1804,11 @@ mod tests {
             mtp_num_hidden_layers: 0,
             mtp_use_dedicated_embeddings: false,
             quarot_rotation_seed: None,
+            vision_config: None,
+            image_token_id: None,
+            video_token_id: None,
+            vision_start_token_id: None,
+            vision_end_token_id: None,
         };
 
         let rope_dim = (head_dim as f32 * cfg.partial_rotary_factor) as usize; // 16
--- a/crates/inference/src/forward/metal_qwen35.rs
+++ b/crates/inference/src/forward/metal_qwen35.rs
@@ -289,6 +289,11 @@ mod mtp_resolve_tests {
             eos_token_id: 63,
             max_position_embeddings: 128,
             quarot_rotation_seed: None,
+            vision_config: None,
+            image_token_id: None,
+            video_token_id: None,
+            vision_start_token_id: None,
+            vision_end_token_id: None,
         }
     }
 
@@ -18890,6 +18895,11 @@ kernel void per_head_rms_norm_batch_pre_854_oracle(
                 eos_token_id: (vocab - 1) as u32,
                 max_position_embeddings: 128,
                 quarot_rotation_seed: None,
+                vision_config: None,
+                image_token_id: None,
+                video_token_id: None,
+                vision_start_token_id: None,
+                vision_end_token_id: None,
             };
 
             let mut embed_tokens = vec![0.0f32; vocab * hidden];
@@ -19909,6 +19919,11 @@ kernel void per_head_rms_norm_batch_pre_854_oracle(
                 eos_token_id: (vocab - 1) as u32,
                 max_position_embeddings: 128,
                 quarot_rotation_seed: None,
+                vision_config: None,
+                image_token_id: None,
+                video_token_id: None,
+                vision_start_token_id: None,
+                vision_end_token_id: None,
             }
         }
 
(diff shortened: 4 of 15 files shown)


ohdearquant:
Pushed a fix for both blockers from review feedback (fa861aa2de).

## Blocker 1 — fail-open on present-but-invalid `vision_config`

Added `VisionModelConfig::validate()` (`crates/inference/src/model/qwen35_config.rs`): rejects zero `depth`/`hidden_size`/`num_heads`/`patch_size`/`spatial_merge_size`/`out_hidden_size`/`temporal_patch_size`/`num_position_embeddings`/`in_channels`, requires `hidden_size % num_heads == 0`, rejects out-of-range `deepstack_visual_indexes` entries, and uses checked arithmetic for every tensor size the loader derives (overflow is a typed error, not a wraparound).

This runs at two boundaries:
- **Parse boundary**: `Qwen35Config::from_config_json_str`, right before it returns.
- **Public loader boundary**: `load_qwen35_vision_weights`, since it's public and callers can construct `VisionModelConfig` directly without going through the parser.

Also closed the inventory-exactness gap: `assemble()` in `crates/inference/src/vision/checkpoint.rs` now errors if any `model.visual.*` tensor is left unconsumed after pulling the expected set — so a checkpoint carrying the full 153-tensor inventory paired with a malformed `depth` no longer silently returns a truncated `Qwen35VisionWeights`.

## Blocker 2 — numel-only validation instead of shape validation

Threaded each tensor's real on-disk shape (previously discarded) through all three storage forms into `assemble()`:
- fp16 sharded safetensors (`ShardedSafetensors::get_f32_tensor_owned`)
- q4 (`Q4Tensor::shape`, plus a check against the manifest's own recorded shape when present)
- `.f16` companion files (`load_f16_tensor_file`)

`assemble()`'s `take()` closure now compares the full expected shape vector (not just element count) against the actual shape, so a same-numel transposition (e.g. FC2 `[in, out]` instead of `[out, in]`) is rejected with `InferenceError::ShapeMismatch` instead of loading silently.

## Tests added

- `parser_rejects_present_vision_config_with_depth_zero` / `..._num_heads_zero` (parse boundary)
- `depth_zero_vision_config_rejected_at_loader_boundary` / `num_heads_zero_vision_config_rejected_at_loader_boundary` (public loader boundary)
- `q4_full_inventory_with_depth_zero_is_rejected` (full 153-entry inventory + `depth: 0` still errors)
- `assemble_rejects_unconsumed_leftover_tensors_when_depth_understates_inventory` (isolates the leftover check from the depth validation)
- `fp16_same_numel_transposed_fc2_is_rejected`, `q4_same_numel_transposed_fc2_is_rejected`, `f16_companion_same_numel_transposed_fc2_is_rejected` (same-numel transposition, one per storage form)

## Mutation-sensitivity (fail-then-pass)

For each blocker, reverted the fix's guard in place, `touch`ed the source file to force a rebuild, and confirmed the tests failed:

- **Blocker 1**: with `validate()` disabled at both boundaries and the leftover-inventory check removed, all 6 corresponding tests failed. Restored → all 6 pass.
- **Blocker 2**: with `assemble()`'s shape comparison reverted to numel-only (`v.len() != expected_len`), all 3 transposition tests failed. Restored → all 3 pass.

## Gates run

- `cargo clippy -p lattice-inference --all-targets -- -D warnings` — clean (also re-checked with `--features f16`)
- `cargo test -p lattice-inference vision` — 56 passed
- `cargo test -p lattice-inference config` — 138 passed
- `cargo test -p lattice-inference --features f16 vision` — 57 passed (includes the real fp16-checkpoint regression test)

ohdearquant:
Addressed the remaining review major: duplicate `model.visual.*` entries in the q4 manifest previously bypassed inventory-exactness via silent `HashMap::insert` overwrite.

**Fix** (`crates/inference/src/vision/checkpoint.rs`, `load_from_q4_dir`): reject a repeated `model.visual.*` name before its file is read, with a descriptive `InferenceError::Inference` naming the duplicate tensor and the manifest directory.

**Sibling-path check**: the fp16 sharded-safetensors path (`load_from_fp16_dir`) parses `model.safetensors.index.json`'s `weight_map` as a `HashMap<String, String>` — duplicate JSON keys already collapse during deserialization, so the count-based inventory check there can't be fooled by a duplicate name the same way. No analogous gap found; no change needed on that path. The `.f16` companion path shares the same loop/insert as the q4 path in `load_from_q4_dir`, so it's covered by the same guard.

**Test**: `q4_manifest_duplicate_visual_tensor_name_is_rejected` builds a tiny valid 21-entry q4 inventory (matching `tiny_vision_cfg`) plus one repeated `model.visual.*` name (written to a distinct on-disk file, so a naive "file exists" check wouldn't catch it) and asserts the public `load_qwen35_vision_weights` returns the duplicate-name error.

**Mutation-sensitivity evidence**: reverted the guard, `touch`ed the source file to force a rebuild, and reran — the test failed because the loader returned `Ok` (the duplicate silently overwrote the first entry). Restored the guard and reran — test passes. Full vision suite: 57/57 passing (58/58 with `--features f16`). `cargo clippy -p lattice-inference --all-targets -- -D warnings` clean.

Pushed to `feat/vision-s1s2-config-loader` at fcb281c311b78f7e4dbbf623e4dde66ea6669eba.

ohdearquant:
## Fix: reject duplicate raw `weight_map` keys in the sharded safetensors index

Addresses the residual major from review feedback: the fp16 sharded loader's
`SafetensorsIndex::weight_map` deserialized as a plain `HashMap<String, String>`
(`crates/inference/src/weights/f32_weights.rs`), so a raw JSON index carrying the
same tensor name twice — mapped to two different shards — silently collapsed to
one entry *before* the visual-tensor-count inventory check in
`crates/inference/src/vision/checkpoint.rs` ever ran. The count stayed exact and
the selected shard depended on raw JSON member order instead of erroring
deterministically.

### Fix

Added a custom serde deserializer (`deserialize_weight_map_no_duplicates`) for
`SafetensorsIndex::weight_map` that visits the raw JSON object directly and
rejects a repeated key the moment a second occurrence is seen — before the map
is ever materialized — returning a typed `serde::de::Error::custom` that
propagates as `InferenceError::InvalidSafetensors` (via the existing
`parse_index` error path). This is shared parsing code, used by both the
vision fp16 sharded loader and `QuarotTensorReader` (`quant/quarot/io.rs`,
which also calls `f32_weights::parse_index`), so the guard closes the gap
for the text-path sharded loader too, not just vision.

### Mutation-sensitive test + fail/pass cycle

Added `fp16_sharded_index_duplicate_raw_weight_map_key_is_rejected`: a tiny
sharded index with all 21 expected `tiny_vision_cfg` visual names mapped to
shard A, plus one name (`model.visual.patch_embed.proj.weight`) repeated as a
raw duplicate key mapped to shard B (same shape, different values `9.0` vs
`0.5`). Asserts `load_qwen35_vision_weights` errors with a message naming
"duplicate" and the tensor name.

Proved the test actually guards the fix, not just decoration:
1. Reverse-applied the fix diff (kept the new test), `touch`ed the source file
   to force a rebuild.
2. Re-ran the test: it **failed** — the loader completed `Ok`, with
   `patch_embed_weight` containing shard B's `9.0` values, confirming the
   pre-fix behavior really is raw-member-order-dependent selection rather than
   a clean overwrite that happens to be harmless.
3. Restored the fix, re-ran: test **passes**.

### Sibling audit: other serde maps deserialized from external JSON in loader paths

- `quant::quarot::io.rs` `Backing::Sharded.weight_map` — reuses
  `f32_weights::parse_index`/`SafetensorsIndex` directly, so it is covered by
  this same fix.
- `quant::q4_manifest::Q4ManifestEntry` (`quantize_index.json`) — parses to a
  `Vec<Q4ManifestEntry>` (bare array or `{"tensors": [...]}`), never a
  name-keyed map, so there is no map-collapse exposure at the serde layer;
  the vision loader's duplicate-name guard against this format is a separate,
  already-covered explicit `HashMap::insert`/`contains_key` check in
  `vision/checkpoint.rs` (`q4_manifest_duplicate_visual_tensor_name_is_rejected`,
  from the prior review round).
- `bin/lattice.rs::read_safetensors_header` parses a single safetensors file's
  own internal tensor header (not the sharded index) into a
  `HashMap<String, TensorEntry>` via an intermediate `serde_json::Value`, which
  has the same theoretical last-key-wins collapse for a malformed/adversarial
  single-file header. This is CLI-only diagnostic tooling (inventory listing),
  not on the model-loading path this PR touches, so it's flagged here as a
  same-class exposure but left out of scope for this fix.

### Gates run

```
CARGO_TARGET_DIR=/private/tmp/cargo-target-mm-s1s2
cargo clippy -p lattice-inference --all-targets -- -D warnings                    # clean
cargo clippy -p lattice-inference --all-targets --features f16 -- -D warnings     # clean
cargo test -p lattice-inference vision                                            # 58 passed
cargo test -p lattice-inference --features f16 vision                             # 59 passed (real fp16 checkpoint present locally)
cargo test -p lattice-inference weights                                           # 169 passed
```

Pushed: `0e62d2c1506973f97546f3ac7df76d8247f84ff9`

oceanwaves630 (approved):

PR #989, opened 2026-07-14
test(inference): HF differential golden fixtures for Qwen3.5 vision path (ADR-069)

ohdearquant:
## Summary

Pins HF `transformers` reference output for the Qwen3.5-0.8B vision path (ADR-069) as committed
fixtures, before any lattice engine vision forward-pass code builds against them. Retires ADR-069
risk R1 (ViT correctness, consumed at future stage S3) and R2 (M-RoPE position-id correctness,
consumed at future stage S5) by giving those stages a frozen cross-framework target to diff
against. Python fixture generator + committed fixtures + a Rust fixture-loader test only — **no
engine forward-pass code**.

- `scripts/vision_goldens_qwen35.py` — generator script (deterministic procedural 256x256 test
  image, HF forward pass, decode).
- `tests/fixtures/vision/` — `manifest.json`, `golden_image.png`, two raw little-endian f32
  tensors (ViT pre-merger patch features + post-merger projected visual embeddings), token/position
  id JSON, greedy-token JSON, and a `README.md` describing the methodology and findings.
- `crates/inference/tests/vision_goldens_fixture_test.rs` — loads the manifest + fixtures, checks
  shape/sha256/finiteness round-trip and internal cross-consistency. No forward pass.
- `pyproject.toml` — pins `torch`, `torchvision`, `pillow`, and `transformers==5.12.1` as dev
  dependencies so the documented `uv run` invocation is reproducible.
- `.gitignore` — narrow exception for the two committed fixture `.bin` files (repo-wide `*.bin` is
  otherwise a model-weight guard).

## Methodology

- Checkpoint: `~/.lattice/models/qwen3.5-0.8b`; its `config.json` and
  `model.safetensors.index.json` verified byte-identical to HF Hub `Qwen/Qwen3.5-0.8B` commit
  `2fc06364715b967f1860aea9cf38778875588b17` (script re-checks this every run, fails closed on
  drift). The checkpoint directory itself ships only `config.json` / `model.safetensors*` /
  `tokenizer.json` — no processor/tokenizer config — so the script fetches the four small
  (~25 KB total) config files from that pinned commit and merges them with the local weights via
  symlinks in a scratch directory. Model weights are never fetched from the network.
- Runtime: `transformers==5.12.1`, `torch==2.13.0`, `torchvision==0.28.0`, Python 3.11.12, CPU,
  float32, single BLAS thread.
- Image: 256x256 deterministic procedural PNG (gradient + fixed shapes, seed `20260714`) — the
  smallest size clearing the checkpoint's own `min_pixels=65536` processor floor while staying
  aligned to `patch_size(16) * spatial_merge_size(2) = 32`. Yields a 16x16 patch grid -> 64
  post-merge visual tokens.
- Prompt: `"Describe this image."` through the checkpoint's chat template.
- Decode: greedy, 8 new tokens.

Full detail and HF-side surprises (processor `min_pixels`, video-processor/`torchvision`
requirement even for image-only calls, empty-think-block chat template) are in
`tests/fixtures/vision/README.md`.

## Determinism proof

Ran the generator three times across two independent environments and diffed byte-for-byte:

```
$ diff -rq tests/fixtures/vision/ <run-2-dir>/
(no output)
$ diff -rq tests/fixtures/vision/ <run-3, via `uv run` from a clean repo venv>/
(no output — including manifest.json, once transformers was pinned to 5.12.1)
$ shasum -a 256 * | sort   # all three runs
(identical sha256 for every fixture file)
```

## Sanity checks (recorded in `manifest.json["sanity_checks"]`, asserted by the generator)

- `post_merger_count_matches_placeholder_count`: **true** — 64 post-merger visual embeddings ==
  64 `<|image_pad|>` tokens in the expanded token stream.
- `text_prefix_positions_collapse_1d`: **true** — `t == h == w` for every position before the
  image span.
- `image_span_positions_diverge`: **true** — `h`/`w` sweep the 8x8 merged grid across the image
  span while `t` stays constant, then all three axes reconverge for the trailing text.

## Fixture sizes

Total ~1.0 MB (well under the ~5 MB budget): `vit_pre_merger_f32.bin` (768 KB, `256x768`),
`vit_post_merger_f32.bin` (256 KB, `64x1024`), remaining JSON/PNG fixtures a few KB each.

## Gates

- `cargo test -p lattice-inference --test vision_goldens_fixture_test`: 3/3 passed.
- bench-compare not run: test fixtures only (structural) — no engine forward-pass path touched.

## Test plan

- [x] `cargo test -p lattice-inference --test vision_goldens_fixture_test`
- [x] Generator run 3x across two venvs, byte-identical fixtures (determinism proof above)
- [x] Sanity checks (placeholder count, position-id collapse/divergence) all true


ohdearquant:
Addressed all four majors from review feedback at 06b62f8a5. Pushed 714eaa22e7.

## Fix 1 — fail-closed on checkpoint index drift
`_compare_pinned_files` now byte-compares **both** `config.json` and `model.safetensors.index.json` against the pinned HF Hub revision, called before any output is created (`verify_checkpoint_matches_hub` in `scripts/vision_goldens_qwen35.py`). The comparison function takes an injected `fetch` callable so it's unit-testable offline. Verified via `--self-test` (see below): matching files pass; a deliberately drifted index or config each raise `RuntimeError` before `build_merged_dir`/any output write.

## Fix 2 — exact version pins + runtime rejection
- `pyproject.toml`: `torch`, `torchvision`, `pillow`, `transformers` are now all exact `==` pins (previously only `transformers` was exact; `torch`/`torchvision`/`pillow` were lower-bounded).
- Added `validate_runtime_versions()` + `REFERENCE_VERSIONS`, called in `run()` before the model is loaded (and before any output is created) — a runtime whose torch/torchvision/pillow/transformers version differs from the pinned reference now fails closed instead of silently regenerating with different numerics.
- `manifest.json["source"]` now also records `torchvision_version`/`pillow_version` for provenance (kept, per review, as a record — not the drift signal itself).

## Fix 3 — narrowed `.gitignore` exception
Replaced the `!tests/fixtures/vision/*.bin` glob with the two exact committed paths. Verified:

```
$ git check-ignore -v tests/fixtures/vision/arbitrary-model.bin
.gitignore:27:*.bin	tests/fixtures/vision/arbitrary-model.bin   # exit 0 — still ignored

$ git check-ignore -v tests/fixtures/vision/vit_pre_merger_f32.bin
(no output, exit 1 — not ignored, addable)

$ git check-ignore -v tests/fixtures/vision/vit_post_merger_f32.bin
(no output, exit 1 — not ignored, addable)
```

## Fix 4 — anchored Rust fixture test
`crates/inference/tests/vision_goldens_fixture_test.rs`:
- Added `image: ImageManifest` to the deserialized manifest; new test hash-verifies `golden_image.png` and validates its PNG IHDR dimensions/color-type (256x256, 8-bit RGB) directly from the file bytes (no image-decoding dependency added).
- New test counts occurrences of image token id `248056` in `input_ids` end to end (must equal 64, matches `num_image_placeholder_tokens`), and validates the documented M-RoPE position-id properties across their **full** ranges: every pre-image position collapses 1-D (t==h==w), every image-span position matches the exact merged 8x8 grid-sweep pattern (not just "some divergence"), and every trailing-text position reconverges to 1-D and increments by 1.
- Added an `anchors` module of hardcoded reviewed constants (HF revision, prompt, image/tensor sha256 digests, image dimensions, greedy tokens) read from fixed filenames — **not** from `manifest.json` — plus a new test that cross-checks the fixtures against them. This is the piece that closes the "self-consistent tampering" gap the review identified: mutating a fixture file and updating `manifest.json` to match it in lockstep no longer passes.

### Mutation-sensitivity proof (4 cycles, each: mutate → run → observe FAIL → restore → run → observe PASS)

1. **PNG byte flip** (`golden_image.png`): FAILED `image_fixture_hash_and_dimensions_verified` + `anchored_reference_values_match_reviewed_constants` (sha256 mismatch). Restored → 6/6 pass.
2. **One input token id** (image-span token `248056` → `999999`, with `manifest.json`'s `input_ids.sha256` updated to match — i.e. the mutation is self-consistent with the manifest): FAILED `image_token_id_count_and_position_id_full_range_properties` (`left: 63, right: 64`) — caught by the new count assertion, not the pre-existing sha guard (which was made to pass on purpose). Restored → 6/6 pass.
3. **One position value** (h-axis value inside the image span, manifest sha updated to match): FAILED the same test on the grid-sweep assertion (`left: 999, right: 7`). Restored → 6/6 pass.
4. **One tensor float** (`vit_pre_merger_f32.bin`, one f32 flipped, **and** `manifest.json`'s `vit_pre_merger.sha256` updated to match — the exact self-consistent-tampering scenario from the review): the pre-existing `vit_pre_and_post_merger_tensors_round_trip_finite` test still PASSED (it only checks tensor bytes against the manifest's own, now-matching, sha256) — but the new `anchored_reference_values_match_reviewed_constants` test FAILED (hardcoded digest independent of manifest). This is the direct proof that the anchor constants close the gap: manifest+fixture agreeing with each other is no longer sufficient. Restored → 6/6 pass.

All fixture files verified byte-identical to their pre-mutation state after each restore (`diff` against pre-mutation backups, clean).

## Gates
```
$ cargo clippy -p lattice-inference --all-targets -- -D warnings
    Finished (no warnings)

$ cargo test -p lattice-inference --test vision_goldens_fixture_test
running 6 tests
test manifest_parses_and_matches_schema ... ok
test image_fixture_hash_and_dimensions_verified ... ok
test image_token_id_count_and_position_id_full_range_properties ... ok
test token_and_position_fixtures_are_internally_consistent ... ok
test anchored_reference_values_match_reviewed_constants ... ok
test vit_pre_and_post_merger_tensors_round_trip_finite ... ok
test result: ok. 6 passed; 0 failed

$ uv run python scripts/vision_goldens_qwen35.py --self-test
self-test: OK (checkpoint-drift guard and version guard both fail closed as expected)
```

No tensor fixtures were regenerated — these are guard/contract/test-coverage changes only, on top of the already-verified committed fixture bytes.

ohdearquant:
Addressed the three residual majors from review feedback (round 2, HEAD 714eaa2). Pushed as 1e137800.

**Fix 1 — preflight before output creation** (`scripts/vision_goldens_qwen35.py`)
Extracted `preflight(model_dir, actual_versions, fetch=_fetch_hub_raw)` = `validate_runtime_versions` then checkpoint-vs-Hub byte comparison, and moved it before `out_dir.mkdir(...)` in `run()` (previously `out_dir` was created, then the checkpoint check ran). `--self-test` now exercises `preflight` directly with a supplied-but-not-yet-created output path and asserts it stays absent after a drifted checkpoint or a drifted runtime version raises — an orchestration-order assertion the prior `_compare_pinned_files`-only test couldn't make, since it never touched `run()`'s ordering at all.

**Fix 2 — pin numpy + Python as drift channels**
`REFERENCE_VERSIONS` now includes `numpy` (`2.4.6`, the version that generated the committed fixtures — confirmed via `uv run python -c "import numpy; print(numpy.__version__)"` against this repo's lockfile) and `python` (`3.11.12`, per the committed methodology). Both are validated in `preflight` before the model loads or output is created, and `numpy_version` is now recorded in `manifest.json["source"]` alongside the other runtime fields. `pyproject.toml` pins `numpy==2.4.6` exactly in `[tool.uv] dev-dependencies` (the top-level `dependencies` entry stays `>=1.26` for other project scripts, so this only tightens the vision-goldens dev/lock environment). The existing self-test mismatch loop iterates over every key in `REFERENCE_VERSIONS`, so numpy/Python drift is automatically covered — verified both raise.

**Fix 3 — anchor input_ids.json and position_ids.json**
Added `anchors::INPUT_IDS_FILE`/`POSITION_IDS_FILE` (fixed filenames) and `anchors::INPUT_IDS_SHA256`/`POSITION_IDS_SHA256` (reviewed digests computed from the committed files) to `crates/inference/tests/vision_goldens_fixture_test.rs`. `anchored_reference_values_match_reviewed_constants` now reads both files directly and checks them against these constants, independent of `manifest.json`'s own (mutable) `sha256` fields for those files. The existing image-token-count and full-range relational checks (`image_token_id_count_and_position_id_full_range_properties`, `token_and_position_fixtures_are_internally_consistent`) are unchanged and still run as complementary semantic validation.

**Mutation-prove cycles (both run against the committed fixtures, then fully restored):**

1. Mutated a non-image `input_ids` entry (index 1: `846` → `847`, a plain text token, not one of the 64 `248056` image-placeholder tokens) and updated `manifest.json["input_ids"]["sha256"]` in lockstep so the mutation was internally self-consistent.
   - `token_and_position_fixtures_are_internally_consistent` / `image_token_id_count_and_position_id_full_range_properties`: still passed (as expected — they only assert `sha256` vs the mutable manifest field, and the manifest was updated to match).
   - `anchored_reference_values_match_reviewed_constants`: **FAILED** — `input_ids.json content drifted from the reviewed anchor digest` (`left: 4cc1d7a0...`, `right: 9beda884...` — the reviewed anchor).
   - Restored both files; re-ran: all 6 tests pass.

2. Applied a uniform +100 shift to every value on every axis of `position_ids.json` (preserves text-prefix collapse, image-span divergence, the merged-grid sweep deltas, and the trailing +1 increments — every existing relational check is satisfied by construction) and updated `manifest.json["position_ids"]["sha256"]` in lockstep.
   - Existing relational tests: still passed (confirms they're relation-only, not absolute-value, checks — exactly the gap being closed).
   - `anchored_reference_values_match_reviewed_constants`: **FAILED** — `position_ids.json content drifted from the reviewed anchor digest` (`left: ae210b11...`, `right: f1407311...`).
   - Restored both files; re-ran: all 6 tests pass.

**Gates:**
- `cargo test -p lattice-inference --test vision_goldens_fixture_test`: 6/6 pass.
- `uv run python scripts/vision_goldens_qwen35.py --self-test`: `self-test: OK (checkpoint-drift guard, version guard, and preflight orchestration order all fail closed as expected)`.

Pushed SHA: `1e137800cb8833fc00dc88dd5dfa63c50e8557d9`

ohdearquant:
Addressed the two remaining majors from review feedback (round 3).

## Fix 1 — orchestration-order self-test now exercises the real output-creation boundary

The self-test previously called `preflight()` directly and asserted the absence of local path variables that no invoked code could ever create — so reordering `out_dir.mkdir` before `preflight` inside `run()` would not have failed it.

Extracted the pre-output portion of `run()` into a new helper, `prepare_output(model_dir, out_dir, actual_versions, fetch=...)`, which performs `preflight(...)` then `out_dir.mkdir(...)`, in that order. `run()` now calls exactly this helper instead of inlining the two steps.

`self_test()` exercises `prepare_output` directly with a real, initially-absent `out_dir` for three cases:
- **Case A** (runtime version drift): asserts it raises and `out_dir` remains absent.
- **Case B** (checkpoint-index drift): asserts it raises and `out_dir` remains absent.
- **Case C** (matching guards): asserts it does *not* raise and `out_dir` gets created.

**Mutation evidence:** reordered `out_dir.mkdir` before `preflight(...)` inside `prepare_output` and re-ran `--self-test`:

```
AssertionError: prepare_output raised on a version drift but a supplied, initially-absent output directory was created anyway (orchestration-order regression: out_dir.mkdir moved before preflight inside prepare_output)
```

Restored the fix → `--self-test` passes again:

```
self-test: OK (checkpoint-drift guard, version guard, and prepare_output orchestration order all fail closed as expected)
```

## Fix 2 — complete provenance anchors, including numpy

(a) Added `"numpy_version": "2.4.6"` to the committed `manifest.json` `source` object (matches `REFERENCE_VERSIONS`, the pinned `pyproject.toml` dev-dependency, and the resolved `uv.lock` entry — no tensors regenerated; nothing hashes `manifest.json` itself, so no other file needed updating).

(b) The Rust test now deserializes the full `source` object into a typed `Source` struct (not partial `Value` indexing) and checks every load-bearing field — `hf_model_id`, `hf_processor_revision`, both checkpoint digests (`config.json`/`index.json` sha256), and all six runtime/Python versions (torch, torchvision, pillow, transformers, numpy, python) — against fixed reviewed anchor constants in the `anchors` module.

**Mutation evidence:** mutated `source.hf_model_id` and `source.numpy_version` in the committed `manifest.json` (payload files untouched) and re-ran the fixture test:

```
test anchored_reference_values_match_reviewed_constants ... FAILED
assertion `left == right` failed: hf_model_id drifted from the reviewed anchor
  left: "Qwen/Qwen3.5-0.8B-mutated"
 right: "Qwen/Qwen3.5-0.8B"
test result: FAILED. 5 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

Restored → all 6 tests pass:

```
test manifest_parses_and_matches_schema ... ok
test image_fixture_hash_and_dimensions_verified ... ok
test image_token_id_count_and_position_id_full_range_properties ... ok
test token_and_position_fixtures_are_internally_consistent ... ok
test anchored_reference_values_match_reviewed_constants ... ok
test vit_pre_and_post_merger_tensors_round_trip_finite ... ok

test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

## Gates run

- `cargo test -p lattice-inference --test vision_goldens_fixture_test` — 6/6 passed
- `uv run python3 scripts/vision_goldens_qwen35.py --self-test` — offline, OK

Pushed as `11599262ce9ad65a5c7ba0d80c17c211a76e5f7d`.

oceanwaves630 (approved):

Issue #566, opened 2026-07-02
feat(models): Gemma 4 model family support [tracking]


ohdearquant:
> _Issue authored by Claude (Anthropic agent) on behalf of @ohdearquant._

Tracking issue for the **Gemma 4 model family** roadmap lane: a second model family beside the Qwen3.5/3.6 hybrid. The family is live on Hugging Face (`google/gemma-4-E2B-it`, `-E4B-it`, `-12B-it`, `-26B-A4B-it`, `-31B-it`), spanning edge-optimized to flagship-dense, which makes it the right stress test for generalizing the engine beyond one family.

## Architecture delta vs the Qwen3.5/3.6 hybrid

Initial characterization (each item to be re-verified against the HF `transformers` modeling source at milestone start, per the repo's rule that comments/config paraphrases are not evidence):

- **Attention pattern**: interleaved local sliding-window (512) and global attention layers, with dual RoPE bases (theta 1e4 on local layers, 1e6 on global) — vs our GDN+GQA hybrid. Sliding-window layers cap KV cache at window size: new cache geometry.
- **QK-norm with scale 1.0**: RMSNorm applied to Q and K, replacing the 1/sqrt(d_k) scale.
- **Logit softcapping**: `C * tanh(x / C)` on attention logits (~50.0) and final logits (~30.0) — new elementwise kernel.
- **GeGLU** MLP activation (vs SwiGLU).
- **Embedding handling**: tied word embeddings, input embeddings scaled by sqrt(d_hidden), and per-layer embeddings (PLE) on the E-series edge models.
- **Weight sharing**: `attention_k_eq_v` (K/V share projection weights) and cross-layer KV state sharing on later layers.
- **Tokenizer**: 262k-vocab byte-level BPE with its own control tokens.
- **26B-A4B** is MoE (~3.8B active params/token) — the family's MoE representative.

## Prerequisite

#177 (ModelSpec -> IR -> KernelPlan, ADR-059) is the architectural enabler. Hardcoding a second family beside the existing forward paths would double the maintenance surface; the IR layer is how family count scales without kernel-path duplication. Work on this lane starts by sizing the #177 slice the first milestone actually needs.

## First milestone

[...]

PR #991, opened 2026-07-15
feat(inference): Gemma 4 E2B tensor-manifest fixture + asset plan (ADR-082 stage 0)

ohdearquant:
## Summary

Implements ADR-082 Stage 0 ("Asset manifest + weight-download plan"). Doc + fixture + script only — no engine code.

- **`scripts/gemma4_tensor_manifest.py`** — extracts the tensor manifest from `google/gemma-4-E2B-it`'s monolithic `model.safetensors` via two HTTP Range requests (bytes 0-7 = little-endian u64 header length, bytes 8..8+N = JSON header). Zero weight bytes fetched. Pinned to revision `9dbdf8a839e4e9e0eb56ed80cc8886661d3817cf` via the `/resolve/<commit>/` URL form. Hard-caps total fetch at 1 MB and fails closed on any drift (tensor count, per-tensor shape/dtype, bucket counts) between a re-fetched header and the committed fixture.
- **`crates/inference/tests/fixtures/gemma4/e2b_tensor_manifest.json`** — the committed fixture.
- **`crates/inference/tests/gemma4_tensor_manifest_test.rs`** — offline-only Rust gate (9 tests) loading the committed fixture.
- **`docs/gemma4.md`** — weight-download plan.
- **`docs/adr/ADR-082-gemma4-e2b-support.md`** — Status flipped `Proposed` → `Accepted` per sign-off. Anchors #566.

## Manifest pins & bucket counts

Pinned checkpoint: `google/gemma-4-E2B-it` @ `9dbdf8a839e4e9e0eb56ed80cc8886661d3817cf`.

| Bucket | Count |
| --- | ---: |
| `model.language_model` | 600 |
| `model.vision_tower` | 658 |
| `model.audio_tower` | 751 |
| `model.embed_vision` | 1 |
| `model.embed_audio` | 1 |
| **Total** | **2011** |

Exactly matches ADR-082's G16.

## Script's single real run (the only network this PR touches)

```
Fetching safetensors header from https://huggingface.co/google/gemma-4-E2B-it/resolve/9dbdf8a839e4e9e0eb56ed80cc8886661d3817cf/model.safetensors (two Range requests, zero weight bytes)...
Fetched 263960 bytes total (cap: 1000000).
Total tensors: 2011
  model.audio_tower: 751
  model.embed_audio: 1
  model.embed_vision: 1
  model.language_model: 600
  model.vision_tower: 658
Wrote fixture to crates/inference/tests/fixtures/gemma4/e2b_tensor_manifest.json
```

(The Rust test suite is offline-only and never touches the network; I additionally re-ran the script's verify mode against a deliberately mutated fixture to confirm the drift gate fails closed — same zero-weight-byte header fetch, not an additional weight download.)

## Negative-test evidence (fail-closed)

- **Script-level drift gate**: mutating a committed shape (`262144, 8960` → `1, 1` for the PLE table) and re-running verify mode produces `FAIL: 1 drift(s)` and a nonzero exit — confirmed live against the real fetched header.
- **Rust fixture test**, 2 dedicated negative cases: `gemma4_corrupt_truncated_manifest_json_is_rejected` (truncated JSON and a structurally-incomplete-but-parseable JSON both fail to parse), `gemma4_manifest_missing_required_bucket_fails_validation` (removing `model.audio_tower` from `bucket_counts` fails the `validate_manifest` helper).

## Stage-0 finding worth flagging

The header-extracted manifest shows every one of the 35 `language_model` layers carries its own `self_attn.{k,v,q}_proj.weight` tensor — this does not bear out ADR G5's literal claim that "the final 20 decoder layers omit K/V projection weights entirely." The real, load-bearing marker actually present in the manifest is the local/global attention-geometry split those weights are keyed on (256-wide K/V for the 28 sliding layers vs 512-wide for the 7 global layers at indices 4, 9, 14, 19, 24, 29, 34 — G3/G4). `gemma4_kv_shared_layer_type_marker_matches_schedule` checks that real split instead of a false absence assertion; both the test module doc and `docs/gemma4.md` flag this as a Stage 2/4 open item (reconcile before building the shared-KV cache path).

## Bench statement

Per ADR-058: bench-compare not run — test fixtures + docs + a standalone Python script only, no `crates/inference/src/` or `crates/embed/src/` engine path touched.

## Gates run locally

- `cargo test -p lattice-inference gemma` — 9/9 passed
- `cargo fmt -p lattice-inference -- --check` — clean
- `deno fmt --check` / doc-lint pre-commit hook — clean (also picked up one pre-existing whitespace-only fix in the ADR's G17 table row, needed for `deno fmt` to pass)

ADR-082 Status flipped to Accepted per sign-off. Anchors #566.

## Test plan

- [x] `cargo test -p lattice-inference gemma` (9 tests)
- [x] doc lint / pre-commit hook
- [x] single real network run of `scripts/gemma4_tensor_manifest.py --write-fixture` to produce the committed fixture
- [x] drift-gate mutation test (fail-closed verified)


ohdearquant:
Addressed all three findings from review feedback on this PR. Summary of fixes, mutation evidence, and gate output below.

## Blocker — bounded range reads in `scripts/gemma4_tensor_manifest.py`

`_http_get_range` now:
- Requires HTTP `206 Partial Content` (a `200` is rejected outright, before any body read).
- Parses `Content-Range` and requires it to match the requested inclusive range exactly.
- Rejects a declared `Content-Length` that isn't exactly `end - start + 1`.
- Reads at most `requested_length + 1` bytes (`resp.read(expected_len + 1)`), so an overlong streaming body is caught and rejected before it's retained — the `+1` byte is only ever used to detect overrun.
- Both range reads in `fetch_header` now share one `remaining_budget` that's checked *before* each request, not just an after-the-fact total-bytes check.

Added an offline `--self-test` mode (`uv run python3 scripts/gemma4_tensor_manifest.py --self-test`) that spins up a local `http.server` fixture (no network) and exercises:
- (a) `200` + multi-MB body → rejected, zero body bytes ever read (status checked before any `resp.read()`)
- (b) `206` + wrong `Content-Range` → rejected, zero body bytes ever read
- (c) `206` + overlong body → rejected after consuming exactly `requested + 1` bytes (asserted via the exception message)
- a valid two-range success case (still passes end-to-end)

## Major — exact drift gate + typed metadata

`diff_against_fixture` now treats `bucket_counts` as exact map equality over the union of fetched/fixture keys (a missing key, extra key, or wrong count all fail — previously it only iterated the fixture's own keys, so a bucket dropped from the fixture silently stopped being checked). It also validates every immutable metadata field (`source_repo`, `revision`, `source_url`, `header_length_bytes`, `header_sha256`, `total_bytes_fetched`) against the fixture, excluding only `extraction_date`.

`crates/inference/tests/gemma4_tensor_manifest_test.rs`: `metadata` is now a typed `ManifestMetadata` struct (not `serde_json::Value`), and `validate_manifest` asserts the same immutable fields against pinned constants.

Mutation tests (generator-level, in `--self-test`, and Rust negative tests) confirm each of these fails closed:
- missing bucket
- wrong bucket count
- changed revision
- changed header SHA-256
- metadata field missing entirely from the JSON → typed-struct parse rejection (Rust)

## Medium — full 35-layer KV geometry coverage

`gemma4_kv_shared_layer_type_marker_matches_schedule` now iterates all 35 layers against the explicit global set `{4, 9, 14, 19, 24, 29, 34}`, asserting both `k_proj` and `v_proj` are present with `[512, 1536]` on global layers and `[256, 1536]` on sliding layers. The "all 35 layers have Q/K/V" check is now its own test (`gemma4_all_layers_have_qkv_projections`), separate from the geometry assertion.

**Mutation-prove cycle (live, this session):** mutated the committed fixture's layer 4 `k_proj` to `[256, 1536]` and layer 3's `v_proj` to `[512, 1536]`, reran the test:

```
test gemma4_kv_shared_layer_type_marker_matches_schedule ... FAILED
thread 'gemma4_kv_shared_layer_type_marker_matches_schedule' panicked at crates/inference/tests/gemma4_tensor_manifest_test.rs:251:13:
assertion `left == right` failed: layer 3 is sliding-attention — expected a 256-wide v_proj
  left: [512, 1536]
 right: [256, 1536]
```

Restored the fixture (`git status` / `diff` confirmed byte-identical to the committed version), reran — all 14 tests pass.

## Gate output (this session)

```
$ cargo clippy -p lattice-inference --all-targets -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.42s
(clean, zero warnings)

$ cargo test -p lattice-inference gemma
running 14 tests
test gemma4_manifest_missing_required_metadata_field_is_rejected ... ok
test gemma4_corrupt_truncated_manifest_json_is_rejected ... ok
test gemma4_all_layers_have_qkv_projections ... ok
test gemma4_manifest_wrong_bucket_count_fails_validation ... ok
test gemma4_manifest_changed_revision_fails_validation ... ok
test gemma4_manifest_bucket_counts_match_adr_082_g16 ... ok
test gemma4_manifest_total_tensor_count_is_2011 ... ok
test gemma4_manifest_wrong_total_tensor_count_fails_validation ... ok
test gemma4_manifest_changed_header_sha256_fails_validation ... ok
test gemma4_spot_check_audio_conv_subsampler_tensor ... ok
test gemma4_manifest_missing_required_bucket_fails_validation ... ok
test gemma4_kv_shared_layer_type_marker_matches_schedule ... ok
test gemma4_spot_check_ple_embedding_table ... ok
test gemma4_spot_check_vision_clippable_linear_clip_buffer ... ok

test result: ok. 14 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

$ uv run python3 scripts/gemma4_tensor_manifest.py --self-test
OK: all offline self-tests passed.
```

The committed fixture (`e2b_tensor_manifest.json`) is unchanged — these are validation/coverage fixes only, no tensor map or manifest content changes.

ohdearquant:
Addressed the three residual findings from the follow-up review (commit b532c4a25e):

**1. (Major) Rust gate: `bucket_counts` map is now exactly equal, not just superset-checked.**
`validate_manifest` builds the expected map from `EXPECTED_BUCKETS` and requires `m.bucket_counts == expected`, reporting both any wrong/missing bucket and any unexpected extra key. The direct fixture test (`gemma4_manifest_bucket_counts_match_adr_082_g16`) now asserts `HashMap` equality directly instead of a one-way per-key loop. Added a negative test, `gemma4_manifest_unexpected_bucket_key_fails_validation`, that inserts `"unexpected": 0` and asserts validation fails.

**2. (Medium) Python generator: the overrun-probe byte is now reserved in the shared 1MB fetch budget before either Range request is issued.**
`_http_get_range`'s budget check now compares `expected_len + 1` (the actual `resp.read()` cap, including the one-byte overrun probe) against `remaining_budget`, not just `expected_len`. `fetch_header`'s header-length guard does the same reservation, so a header of exactly `MAX_FETCH_BYTES - 8` bytes — previously able to let the second request's probe-byte read pull one byte past the cap — is now rejected conservatively before any request for it is made.

**3. (Medium) Python generator: bucket-count comparison is now bool-safe.**
Added `_is_plain_int()` (`isinstance(v, int) and not isinstance(v, bool)`) and applied it to both sides of each bucket-count comparison in `diff_against_fixture`, so a fixture/fetched value of JSON `true` no longer silently passes as a count of `1` (Python's `True == 1`).

**Mutation evidence**

- Fix 1 (Rust, revert → touch → fail → restore → pass):
  - Reverted `validate_manifest`'s exact-map guard back to the one-way per-key loop (kept the new negative test in place).
  - `touch`'d the test file, reran `cargo test -p lattice-inference --test gemma4_tensor_manifest_test`: `gemma4_manifest_unexpected_bucket_key_fails_validation` **FAILED** (panic: "an unexpected bucket key must fail validate_manifest even with count 0"), 14/15 passed.
  - Restored the fix; reran: all 15/15 passed, including the new test.

- Fix 2 (Python self-test):
  - Reverted the probe-byte budget reservation (both in `_http_get_range` and `fetch_header`) and ran `--self-test`: the exact-budget boundary case (`expected_len == budget`, no room for the probe byte) no longer rejects before issuing the request — it escaped the pre-request guard and attempted an actual socket connection (`ConnectionRefusedError`) against a nonexistent test address, i.e. the reverted code let a request through that the fix rejects fail-closed before ever touching the network.
  - Restored the fix; reran: `OK: all offline self-tests passed.`

- Fix 3 (Python self-test):
  - Reverted the `_is_plain_int` guard back to plain `!=` comparison and ran `--self-test`: **`FAIL: 1 self-test failure(s): - bucket-count-json-bool-not-int: expected drift error, got none`** — confirming `True == 1` silently passed drift detection without the fix.
  - Restored the fix; reran: `OK: all offline self-tests passed.`

**Gates (all green on b532c4a25e):**
- `cargo test -p lattice-inference gemma` — 15/15 passed (`gemma4_tensor_manifest_test`)
- `uv run python3 scripts/gemma4_tensor_manifest.py --self-test` (offline) — `OK: all offline self-tests passed.`

oceanwaves630 (approved):

PR #993, opened 2026-07-15
feat(inference): Gemma 4 BPE tokenizer parity with HF goldens (ADR-082 stage 1)

ohdearquant:
## Summary

ADR-082 Stage 1 (Tokenizer + prompt-template parity, G17). Adds an **additive** `GemmaBpeTokenizer` (`crates/inference/src/tokenizer/gemma_bpe.rs`) implementing the literal-space `Split` pretokenizer + U+2581 metaspace `Replace` normalizer + byte-fallback BPE pipeline, read directly from the target `tokenizer.json` (`google/gemma-4-E2B-it` @ `9dbdf8a839e4e9e0eb56ed80cc8886661d3817cf`) rather than guessed from training data.

Goldens-first, per this repo's differential-test-first practice: the HF-side golden generator and fixtures were produced and reviewed before the Rust tokenizer was written.

## Corpus coverage

`scripts/gemma4_tokenizer_goldens.py` generates `corpus_goldens.json` (21 cases) covering every category the Stage-1 gate requires:

| Category | Cases |
| --- | --- |
| `text` | ordinary English + punctuation |
| `unicode` | CJK, emoji (incl. ZWJ family emoji), combining marks, mixed CJK/Latin/emoji |
| `whitespace` | multi-space runs, tabs/newlines, leading/trailing spaces, whitespace-only, empty string |
| `byte_fallback` | rare 4-byte UTF-8 char (U+10300, decomposes to 4 `<0xXX>` tokens), private-use-area chars, replacement char |
| `marker` | `<\|image\|>`, `<\|audio\|>`, `<\|image>`/`<image\|>`, `<\|audio>`/`<audio\|>`, `<\|video\|>`, and the thought/tool/turn/channel markers (`<\|think\|>`, `<\|channel>`/`<channel\|>`, `<\|turn>`/`<turn\|>`, `<\|tool>`/`<tool\|>`, `<\|tool_call>`/`<tool_call\|>`, `<\|tool_response>`/`<tool_response\|>`) |

Plus `chat_template_golden.json`: a 2-turn (user, assistant) conversation rendered through the real `chat_template.jinja` with `add_generation_prompt=True`, both the rendered string and its token IDs.

Every case is asserted for **exact token-ID equality** against `GemmaBpeTokenizer`, plus a byte-for-byte decode round trip on the non-marker cases (`crates/inference/tests/gemma4_tokenizer_goldens_test.rs`).

## Golden provenance

- Source: `google/gemma-4-E2B-it` @ revision `9dbdf8a839e4e9e0eb56ed80cc8886661d3817cf` (pinned in the `/resolve/<revision>/` URL form, not advisory).
- Files fetched (bounded to 40 MB total, actual ~32.2 MB): `tokenizer.json` (sha256 `cc8d3a0c...7bfe0f`), `tokenizer_config.json` (sha256 `90c3a3ba...06f3df`), `chat_template.jinja` (sha256 `2f1b4d75...6613d082`) — full manifest with per-file sizes/hashes in `crates/inference/tests/fixtures/gemma4/tokenizer/manifest.json`.
- Generator runtime versions, pinned and validated before generation: HF `tokenizers==0.22.2`, `jinja2==3.1.6` (both locked transitively via `transformers` in `uv.lock`).
- A Rust test (`manifest_matches_committed_tokenizer_json_bytes`) re-hashes the committed fixture files and checks them against the manifest, offline — it doesn't re-fetch, but it does prove the committed bytes match their recorded provenance.

## Additive-path negative-test evidence

`crates/inference/src/tokenizer/bpe.rs`'s `validate_bpe_pretokenizer` (the existing Qwen-oriented loader's gate) still rejects this same `tokenizer.json` — proven by `existing_qwen_bpe_loader_rejects_gemma_tokenizer_json`, which feeds the real fixture into `BpeTokenizer::from_tokenizer_json_str` and asserts `Err`. `GemmaBpeTokenizer` is a separate type with its own explicit constructors (`from_tokenizer_json`/`from_tokenizer_json_str`), never reached through `load_tokenizer`'s or `tokenizer_from_json_str`'s model-type dispatch — selection is by explicit caller choice, not sniffing.

## Mutation cycles (run this session, both reverted before commit)

1. **Golden-ID mutation**: flipped `corpus_goldens.json`'s `ascii_ordinary` first token ID (818 → 819) → `corpus_goldens_match_exact_hf_token_ids` failed with a clear ID-mismatch panic. Reverted; test green again.
2. **Additive-path guard mutation**: temporarily made `is_supported_bpe_pretokenizer` unconditionally `return true` (simulating a silent widening of the general loader) → `existing_qwen_bpe_loader_rejects_gemma_tokenizer_json` failed loudly. Reverted; `cargo clippy -p lattice-inference --all-targets -- -D warnings` and the full test re-confirmed clean.

## Bench statement (ADR-058)

bench-compare not run: additive tokenizer path + fixtures, no existing engine path modified (structural). The only change to a pre-existing file (`bpe.rs`) is making an internal helper (`parse_merges_json`) `pub(crate)` so the new module can reuse the same merges-array parser — no behavioral change to any existing code path.

## Gates

- `cargo test -p lattice-inference tokenizer` — green
- `cargo test -p lattice-inference gemma` — green (17 tests: 15 Stage-0 manifest + 2 matching this filter's substring from the new file)
- `cargo test -p lattice-inference --test gemma4_tokenizer_goldens_test` — 7/7 green
- `cargo check --workspace` — clean
- `cargo fmt -p lattice-inference` — applied, no outstanding diff

## Scope

Tokenizer module + fixtures + tests only, per ADR-082 sizing (Stage 1 = S). No config/loader/forward-pass work — that's Stage 2. The chat-template rendering golden covers only the plain 2-turn case (the real `chat_template.jinja` also handles tool calls, reasoning/thinking channels, and multi-turn continuation, which are out of scope here); the Rust side doesn't implement a Jinja renderer at all — it only tokenizes the HF-rendered golden string, so template rendering fidelity itself is not yet claimed for the tool-calling/reasoning branches.

Anchors #566, ADR-082 Stage 1.

### DIFF
--- a/_typos.toml
+++ b/_typos.toml
@@ -8,6 +8,7 @@
 # `target/`, so the exclusion is explicit rather than incidental.
 extend-exclude = [
     "crates/inference/tests/fixtures/tokenizers/**",
+    "crates/inference/tests/fixtures/gemma4/tokenizer/**",
     "crates/inference/benchmarks/golden_logits/**",
     "crates/embed/tests/fixtures/**",
     "docs/bench_results/**",
--- a/crates/inference/src/lib.rs
+++ b/crates/inference/src/lib.rs
@@ -176,7 +176,9 @@ pub use crate::tokenizer::tokenizer_from_json_str;
 /// Stage-1 marker-expansion arithmetic (ADR-082 G11/G15/G17): `<|image|>`/`<|audio|>`
 /// placeholder-to-soft-token-count contract, independent of the in-sequence scatter itself.
 pub use crate::tokenizer::{
-    GEMMA4_AUDIO_MAX_SOFT_TOKENS, GEMMA4_AUDIO_MS_PER_SOFT_TOKEN,
+    GEMMA4_AUDIO_FRAME_LENGTH_SAMPLES, GEMMA4_AUDIO_HOP_LENGTH_SAMPLES,
+    GEMMA4_AUDIO_MAX_SOFT_TOKENS, GEMMA4_AUDIO_MS_PER_SOFT_TOKEN, GEMMA4_AUDIO_SAMPLING_RATE_HZ,
     GEMMA4_IMAGE_SOFT_TOKENS_PER_IMAGE, audio_marker_expansion_tokens,
-    image_marker_expansion_tokens, total_audio_marker_expansion_tokens,
+    audio_marker_expansion_tokens_from_samples, image_marker_expansion_tokens,
+    total_audio_marker_expansion_tokens,
 };
--- a/crates/inference/src/lib.rs
+++ b/crates/inference/src/lib.rs
@@ -154,6 +154,10 @@ pub use crate::pool::BertPooling;
 pub use crate::stop_reason::StopReason;
 /// Byte-level BPE tokenizer used by Qwen-family models. See [`Tokenizer`] and [`TokenizedInput`].
 pub use crate::tokenizer::BpeTokenizer;
+/// Additive Gemma-family BPE tokenizer (literal-space `Split` + `▁` metaspace normalizer),
+/// explicitly selected — never reached via [`load_tokenizer`]'s model-type sniffing. See
+/// [`Tokenizer`] and ADR-082 G17.
+pub use crate::tokenizer::GemmaBpeTokenizer;
 /// `SentencePiece` tokenizer implementation. See [`Tokenizer`] and [`TokenizedInput`].
 pub use crate::tokenizer::SentencePieceTokenizer;
 /// Padded token IDs and the real (unpadded) sequence length returned by tokenizers. See
--- a/crates/inference/src/lib.rs
+++ b/crates/inference/src/lib.rs
@@ -173,3 +173,10 @@ pub use crate::tokenizer::load_tokenizer;
 /// `tokenizer.json`-text tokenizer loader (no filesystem access). See
 /// [`Tokenizer`], [`tokenizer`], and [`BertModel::from_bytes`].
 pub use crate::tokenizer::tokenizer_from_json_str;
+/// Stage-1 marker-expansion arithmetic (ADR-082 G11/G15/G17): `<|image|>`/`<|audio|>`
+/// placeholder-to-soft-token-count contract, independent of the in-sequence scatter itself.
+pub use crate::tokenizer::{
+    GEMMA4_AUDIO_MAX_SOFT_TOKENS, GEMMA4_AUDIO_MS_PER_SOFT_TOKEN,
+    GEMMA4_IMAGE_SOFT_TOKENS_PER_IMAGE, audio_marker_expansion_tokens,
+    image_marker_expansion_tokens, total_audio_marker_expansion_tokens,
+};
--- a/crates/inference/src/tokenizer/bpe.rs
+++ b/crates/inference/src/tokenizer/bpe.rs
@@ -839,7 +839,12 @@ fn parse_merges_txt(text: &str) -> Vec<(String, String)> {
     merges
 }
 
-fn parse_merges_json(value: &JsonValue) -> Result<Vec<(String, String)>, InferenceError> {
+/// Shared with the additive Gemma BPE path (`tokenizer::gemma_bpe`), which
+/// parses the same `tokenizer.json` `model.merges` shape but builds its own
+/// merge-rank table and pretokenization pipeline; see ADR-082 G17.
+pub(crate) fn parse_merges_json(
+    value: &JsonValue,
+) -> Result<Vec<(String, String)>, InferenceError> {
     let array = value.as_array().ok_or_else(|| {
         InferenceError::Tokenizer("expected tokenizer.json model.merges array".into())
     })?;
--- a/crates/inference/src/tokenizer/gemma_bpe.rs
+++ b/crates/inference/src/tokenizer/gemma_bpe.rs
@@ -1,0 +1,527 @@
+//! Gemma 4 BPE tokenizer: literal-space `Split` pre-tokenizer + `▁` (U+2581)
+//! metaspace-style normalizer, per the target `tokenizer.json` read for
+//! ADR-082 G17.
+//!
+//! This is an **additive, explicitly-selected** path — `GemmaBpeTokenizer`
+//! is never reached through [`crate::tokenizer::common::load_tokenizer`] or
+//! [`crate::tokenizer::common::tokenizer_from_json_str`]; callers construct
+//! it directly. The existing Qwen-oriented [`crate::tokenizer::BpeTokenizer`]
+//! loader's `validate_bpe_pretokenizer` continues to reject Gemma's
+//! `tokenizer.json` (a literal-string `Split`, not a regex `Split`/
+//! `ByteLevel`), which proves this path is additive rather than a silent
+//! widening of the existing loader — see
+//! `tests/gemma4_tokenizer_goldens_test.rs`.
+//!
+//! Pipeline, read directly from the target `tokenizer.json`
+//! (`google/gemma-4-E2B-it` @ `9dbdf8a839e4e9e0eb56ed80cc8886661d3817cf`):
+//!
+//! - `normalizer`: `Replace(" " -> "▁")` — a literal replace, with no
+//!   dummy-prefix insertion and no leading/trailing-whitespace stripping
+//!   (unlike this crate's SentencePiece Unigram path).
+//! - `pre_tokenizer`: `Split(pattern=" ", behavior=MergedWithPrevious)` — a
+//!   literal-string split. Because the normalizer has already consumed every
+//!   space, this never fires in practice: the whole normalized string is one
+//!   pre-token, matching the empirically observed behavior (`"a  b   c"` ->
+//!   `["a", "▁▁", "b", "▁▁▁", "c"]`, i.e. the BPE merge algorithm — not
+//!   pre-tokenization — decides where multi-`▁` runs group).
+//! - `model`: `BPE` with `byte_fallback: true`, `fuse_unk: true`, no
+//!   `continuing_subword_prefix`/`end_of_word_suffix`. Initial symbols are
+//!   per-Unicode-scalar: a character present in `vocab` as its own token
+//!   starts as that token; otherwise its UTF-8 bytes each become individual
+//!   `<0xXX>` byte-fallback tokens (verified: `\u{10300}` -> four
+//!   single-byte tokens). No merge rule ever pairs two byte-fallback tokens
+//!   in the target vocab, so once introduced they never re-merge.
+//! - `decoder`: `Sequence[Replace("▁" -> " "), ByteFallback, Fuse]`.
+
+use crate::error::InferenceError;
+use crate::tokenizer::bpe::parse_merges_json;
+use crate::tokenizer::common::{
+    JsonValue, TokenizedInput, Tokenizer, invert_vocab, json_object_to_vocab, json_path,
+    known_special_id, pad_ids, parse_added_tokens, parse_json, parse_rendered_added_tokens,
+};
+use std::cmp::Ordering;
+use std::collections::{BinaryHeap, HashMap, HashSet};
+use std::fs;
+use std::path::Path;
+use std::sync::Arc;
+
+/// U+2581 LOWER ONE EIGHTH BLOCK, used by SentencePiece-descended tokenizers
+/// as a metaspace marker for the literal ASCII space.
+const METASPACE: char = '\u{2581}';
+const DEFAULT_MAX_SEQ_LEN: usize = 4_096;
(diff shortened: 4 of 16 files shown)


ohdearquant:
Addressed all four review-feedback findings (1 major, 2 medium, 1 minor) at 12648f703b.

## Major — byte-fallback decode parity

`GemmaBpeTokenizer::decode` concatenated every `<0xXX>` fallback byte into one global `Vec<u8>` and ran `String::from_utf8_lossy` over the whole buffer once at the end. Rust's lossy conversion collapses a maximal invalid subsequence into a single U+FFFD, which doesn't match the pinned HF `tokenizers` v0.22.2 `ByteFallback` decoder (`decoders/byte_fallback.rs`): it groups only *consecutive* `<0xXX>` tokens into a run, and on a failed `String::from_utf8` conversion emits **one U+FFFD per byte** in that run.

Confirmed empirically against the pinned fixture before touching code: decoding `[467, 381]` (`<0xE5><0x8F>`, an incomplete 3-byte lead) gives HF `"\u{FFFD}\u{FFFD}"` (two replacement chars) but the old code gave `"\u{FFFD}"` (one).

Rewrote decode as run-based (`append_token` + `flush_byte_fallback_run`): accumulate only consecutive fallback bytes, flush with strict `String::from_utf8` on any non-fallback token / special-skip token / end of input, emit one U+FFFD per byte in the run on failure.

Required goldens added (`decode_goldens.json`, generated against the pinned HF tokenizer, exact byte-for-byte match asserted in `byte_fallback_decode_matches_exact_hf_output`):
- `[467]` → `"\u{FFFD}"`
- `[467, 381]` → `"\u{FFFD}\u{FFFD}"`
- `[467, 381, 409]` → `"叫"` (valid 3-byte run)
- `[467, 381, <ordinary token "a">]` → `"\u{FFFD}\u{FFFD}a"`

Existing plain-text round-trip goldens (`corpus_goldens.json`) untouched.

## Medium 1 — fail-closed constructor validation

`validate_gemma_bpe_shape` previously checked `pre_tokenizer.type == "Split"` and that `pattern.String` was present, but ignored `behavior`/`invert`, and didn't check `decoder` order or the `model.{dropout,continuing_subword_prefix,end_of_word_suffix,fuse_unk,ignore_merges}` fields the encode/decode logic assumes.

Now validates:
- `pre_tokenizer`: `pattern.String == " "`, `behavior == "MergedWithPrevious"`, `invert == false`
- `decoder`: exactly `Sequence[Replace("▁"->" "), ByteFallback, Fuse]` in that order
- `model`: `dropout`/`continuing_subword_prefix`/`end_of_word_suffix` null, `fuse_unk == true`, `ignore_merges == false`

Tests use a minimal synthetic Gemma-shaped `tokenizer.json` (not the 32 MB fixture, for speed) with one mutation test per checked field — pattern, behavior, invert, decoder reordered, decoder missing a stage, `ignore_merges`, `fuse_unk`, `dropout`, `continuing_subword_prefix` — each asserting construction fails closed.

## Medium 2 — Stage-1 marker-expansion arithmetic (ADR-082 G17)

Added the ADR's Stage-1 expansion-count contract as pure functions:
- `image_marker_expansion_tokens(count) -> count * 280` (vision default soft-token budget)
- `audio_marker_expansion_tokens(duration_ms) -> min(ceil(duration_ms / 40), 750)` (HF `Gemma4Processor`'s documented `ceil(duration_ms / audio_ms_per_token)`, capped)

The three constants (`280`, `40`, `750`) were previously only source-read into the ADR prose; the generator now fetches the checkpoint's `processor_config.json` (pinned revision, bounded fetch, hash recorded in `manifest.json`) and derives `expansion_goldens.json` from its `image_seq_length`/`audio_ms_per_token`/`audio_seq_length` fields directly, rather than hardcoding. A test cross-checks the committed Rust constants against that fixture's provenance so they can't silently drift apart. Goldens cover zero/one/multiple placeholders for both modalities.

Scope boundary, stated explicitly: this is the placeholder-count arithmetic only. The resolution-dependent image variant and the actual in-sequence scatter are Stage 5/6 (vision CPU parity / vision end-to-end Metal); the mel-framing + conv-subsampling arithmetic that computes a real audio duration from a waveform is Stage 8/9 (audio tower parity / audio end-to-end Metal). Both are out of Stage 1's scope per the ADR's own stage table.

## Minor — generator temp-file leak

`Tokenizer.from_file` required writing the fetched `tokenizer.json` bytes to a `tempfile.mkstemp` path with no cleanup. Switched to `Tokenizer.from_str` (HF `tokenizers` supports loading directly from the JSON string) — no temp file is created at all, so there's nothing to leak or clean up.

## Mutation-proof evidence (this session)

**Major (decode)**: reverted `decode`/`append_token`/`flush_byte_fallback_run` back to the single-global-buffer `String::from_utf8_lossy` implementation, `touch`ed the file, reran `byte_fallback_decode_matches_exact_hf_output` — failed exactly as predicted:
```
assertion `left == right` failed: case "decode_two_byte_incomplete_run" (ids=[467, 381]) decoded "�", expected "��"
```
Restored the fix, diffed byte-identical against the pre-revert file, reran — 19/19 green.

**Medium 1 (constructor)**: reverted the `pre_tokenizer` check back to only `type == "Split"` + `pattern.String` present (dropping the `behavior`/`invert`/exact-pattern checks), `touch`ed, reran the `constructor_rejects_*` tests — exactly the three tests tied to that field failed (`constructor_rejects_wrong_split_pattern`, `constructor_rejects_wrong_split_behavior`, `constructor_rejects_inverted_split`), the other six construction-mutation tests still passed. Restored, diffed byte-identical, reran — 19/19 green.

## Gates

- `cargo test -p lattice-inference tokenizer` — green
- `cargo test -p lattice-inference gemma` — green
- `cargo test -p lattice-inference --test gemma4_tokenizer_goldens_test` — 19/19 green (was 7/7; +12 new tests: 1 decode-goldens, 1 expansion-arithmetic, 9 constructor-mutation, 1 minimal-baseline-constructs)
- `cargo fmt -p lattice-inference` — applied, no outstanding diff
- `scripts/gemma4_tokenizer_goldens.py` (default verify mode, live network re-fetch + regeneration diff) — `OK: fixtures match live fetch + regeneration, no drift.`
- `typos` over the touched fixtures/scripts/source — clean

Only `corpus_goldens.json` and `chat_template_golden.json` were left untouched, as required — the new fixtures are `decode_goldens.json`, `expansion_goldens.json`, and `processor_config.json`.

ohdearquant:
Addressed the two residual review-feedback findings from the follow-up round.

## Major — audio marker expansion now matches the pinned processor's exact post-subsampling arithmetic

`audio_marker_expansion_tokens` was `ceil(duration_ms / 40)`, which only happened to agree with the two durations the old fixture exercised (1000ms and the 40000ms cap). The pinned processor (`transformers@ab1771c9`, `processing_gemma4.py:260-298`, `Gemma4Processor._compute_audio_num_tokens`) actually derives the count *post-subsampling*: mel-frame count from the feature extractor's semicausal framing, then two stride-2 `Conv2d` reductions, then the cap.

`audio_marker_expansion_tokens_from_samples` in `crates/inference/src/tokenizer/gemma_bpe.rs` now implements that exact arithmetic (mirroring the pinned source line-for-line), parameterized by `processor_config.json`'s `feature_extractor.{sampling_rate,frame_length,hop_length}` (new cross-checked constants `GEMMA4_AUDIO_SAMPLING_RATE_HZ`/`GEMMA4_AUDIO_FRAME_LENGTH_SAMPLES`/`GEMMA4_AUDIO_HOP_LENGTH_SAMPLES`) plus the pinned source's SSCP conv constants (kernel=3, stride=2, padding=1, 2 layers — architectural constants, not exposed via `Gemma4AudioConfig`, so documented with the exact upstream line citation instead). `audio_marker_expansion_tokens(duration_ms)` is now a thin duration→samples convenience wrapper over it. `scripts/gemma4_tokenizer_goldens.py`'s `generate_expansion_goldens` mirrors the identical formula, and `expansion_goldens.json` was regenerated offline from the already-committed `processor_config.json` (no network fetch needed for this piece).

Boundary table (16 kHz, matches HF exactly):

| duration_ms | samples | mel_frames | after conv1 | after conv2 (final) | soft tokens |
|---|---|---|---|---|---|
| 0 | 0 | 0 (≤0) | – | – | 0 |
| 1 | 16 | 0 (≤0) | – | – | 0 |
| 10 | 160 | 0 (≤0) | – | – | 0 |
| 11 | 176 | 1 | 1 | 1 | 1 |
| 40 | 640 | 3 | 2 | 1 | 1 |
| 41 | 656 | 4 | 2 | 1 | 1 |
| 1000 (normal) | 16000 | 99 | 50 | 25 | 25 |
| 40000 (cap) | 640000 | 3999 | 2000 | 1000 → capped | 750 |

New goldens assert all eight of these durations, plus `total_audio_marker_expansion_tokens` for zero/one/multiple markers.

### Mutation-proving the fix

1. Temporarily reverted `audio_marker_expansion_tokens` to `ceil(duration_ms / 40).min(cap)`, `touch`ed the source file to force a rebuild.
2. `cargo test -p lattice-inference --test gemma4_tokenizer_goldens_test marker_expansion` failed immediately: `assertion left == right failed: audio duration_ms=1 expansion mismatch / left: 1 / right: 0` — the boundary sweep (which includes both 10ms and 41ms) is mutation-sensitive; per the table above the reverted formula also disagrees at 10ms (`ceil(10/40)=1` vs exact `0`) and 41ms (`ceil(41/40)=2` vs exact `1`).
3. Restored the fix, `touch`ed again, reran: all 23 tests in the file pass, including `marker_expansion_arithmetic_matches_processor_config_goldens`.

## Medium — completed fail-closed guard mutation coverage

The constructor already rejected `model.byte_fallback == false` and a non-null `model.end_of_word_suffix`, but neither had a dedicated mutation test proving it. Added:
- `constructor_rejects_byte_fallback_false`
- `constructor_rejects_non_null_end_of_word_suffix`
- `constructor_rejects_decoder_replace_wrong_content` and `constructor_rejects_decoder_replace_wrong_pattern` — mutate the decoder's `Replace` `content`/`pattern.String` fields (not just stage order/count), proving the exact decoder contract.

## Gates

- `cargo test -p lattice-inference tokenizer` — 68 passed
- `cargo test -p lattice-inference gemma` — all passed (15 manifest + 23 tokenizer-goldens tests, including the 4 new mutation tests)
- `cargo test -p lattice-inference --test gemma4_tokenizer_goldens_test` — 23/23 passed
- Generator self-check: regenerated `expansion_goldens.json` offline from the committed `processor_config.json` via `generate_expansion_goldens` and diffed against the committed fixture — no drift.
- `cargo fmt -p lattice-inference` applied (import-list rewrap only)

Diff is scoped to `crates/inference/src/{lib.rs,tokenizer/{gemma_bpe.rs,mod.rs}}`, `crates/inference/tests/{gemma4_tokenizer_goldens_test.rs,fixtures/gemma4/tokenizer/expansion_goldens.json}`, and `scripts/gemma4_tokenizer_goldens.py`.

ohdearquant:
Round-3 residual (medium) fixed at 44ba4233ca.

**Field-discriminating guard mutations**: the four mutation tests previously asserted only `is_err()`, so a rejection by some other validation could masquerade as coverage. Two changes:

1. The constructor's grouped errors are split into per-field messages naming the exact declared field that diverged: the model-field group is now five separate checks (`model.dropout`, `model.continuing_subword_prefix`, `model.end_of_word_suffix`, `model.fuse_unk`, `model.ignore_merges`), and the decoder-shape check is now a staged validator naming `decoder[0].type` / `decoder[0].pattern.String` / `decoder[0].content` / `decoder[1].type` / `decoder[2].type` individually.
2. New `assert_rejects_naming` helper unwraps the error and asserts its message contains the guarded field's name; the four tests (`byte_fallback`, `end_of_word_suffix`, decoder `Replace` content, decoder `Replace` pattern) now use it with fragments `model.byte_fallback`, `model.end_of_word_suffix`, `decoder[0].content`, `decoder[0].pattern.String`.

Mutation cycle: reverse-applied the source split (restoring the grouped errors), touched the source, ran the constructor-mutation tests — the three tests whose guards were split fail on the naming assertion (`byte_fallback` passes under revert because its field-specific error pre-existed; its assertion pins that existing contract). Restored the split, all 23 tests in the file pass.

Gates: `cargo clippy -p lattice-inference --all-targets -- -D warnings` clean; `cargo test -p lattice-inference --test gemma4_tokenizer_goldens_test` 23/23; `cargo test -p lattice-inference tokenizer` 68 passed. Diff touches only `crates/inference/src/tokenizer/gemma_bpe.rs` and `crates/inference/tests/gemma4_tokenizer_goldens_test.rs`.

oceanwaves630 (approved):

PR #994, opened 2026-07-15
feat(inference): Qwen3.5 vision tower forward vs HF goldens (ADR-069 S3a)

ohdearquant:
## Summary

Implements ADR-069 Stage S3a: the Qwen3.5-0.8B ViT forward pass (depth 12,
hidden 768) over the real `model.visual.*` weights loaded by
`load_qwen35_vision_weights` (S1/S2, #988), producing the pre-merger
hidden states `[256, 768]` for the fixed 256x256 golden image. S4 (merger
+ image-token expansion) and S5 (decoder injection + M-RoPE) remain out of
scope.

**S3a/S3b split**: review feedback on the first version of this PR found
that the cosine gate only ran locally and that "S3" as originally scoped
conflated deriving the CPU-side numerical conventions with a from-scratch
Metal port into one unreviewable unit. ADR-069 Amendment 1 (#995, merged)
splits S3 into **S3a** — the CPU-reference ViT forward implemented here,
convention-bearing, gated at cosine > 0.999 vs the committed HF golden,
now wired as a required CI job (see below) — and **S3b** — the Metal port,
verified against this CPU reference under the machine GPU lock, no
independent convention decisions, tracked as a fast-follow.

New module: `crates/inference/src/vision/qwen35_vit.rs`
- `preprocess_qwen35_image`: decode + patchify the golden image into the
  checkpoint's temporal-patch-folded pixel tensor (`[in_channels,
  temporal_patch_size, patch_size, patch_size]` fold order, block-major
  spatial-merge patch ordering), with the checkpoint's actual
  `image_mean = image_std = [0.5, 0.5, 0.5]` / `rescale=1/255` constants
  (NOT the ImageNet stats `preprocess.rs`'s unrelated ADR-049 7B scaffold
  uses).
- `qwen35_vit_forward`: patch embed (linear-equivalent of the checkpoint's
  Conv3d) → bilinear-interpolated learned position embedding → 2-axis
  (h, w) vision RoPE → 12 full-attention transformer blocks (fused QKV,
  no per-head Q/K norm, `LayerNorm` with bias, GELU-tanh MLP) → pre-merger
  output (no post-block norm — the real checkpoint has none).

**Scaffold reconciliation notes**: the existing `vision/vit.rs` (ADR-049,
targeting a hypothetical 7B model) has a structurally different attention
(three separate Q/K/V projections + per-head RMSNorm, unlike the real
checkpoint's single fused `qkv` linear with no Q/K norm), a from-scratch 2D
RoPE, and a windowed-attention code path. The real 0.8B checkpoint's
`cu_seqlens` for a single image is one full-attention segment — there is no
window-specific weight in the checkpoint at all — so the windowed-attention
assumption is dropped entirely rather than adapted. Given nothing else in
the tree references the old scaffold, S3a is implemented as an independent
module rather than reconciling the two forward passes into one; the module
doc comment cross-references why. `vision/mod.rs`'s module overview now
states which of the two paths is real (Qwen3.5-0.8B, `checkpoint.rs` +
`qwen35_vit.rs`) and which is the inert, no-consumer ADR-049 scaffold
(`vit.rs`), and which future S4/S5 work should consume.

**Preprocessing provenance**: the golden fixture's `manifest.json` records
only a `pixel_values_summary` (shape/mean/std), not the raw tensor, so the
image→pixel-tensor step had to be reimplemented in Rust rather than
consumed as a fixture. Verified via CLAUDE.md's differential-test-first
process: installed a matching `torch==2.13.0`/`torchvision==0.28.0` locally
and ran the real HF `Qwen2VLImageProcessor` + `Qwen3_5VisionModel` against
the checkpoint at `~/.lattice/models/qwen3.5-0.8b` and the committed
`golden_image.png`, confirming **max-abs-diff 0.0 / cosine 1.0** against
the committed `vit_pre_merger_f32.bin` — i.e. this machine's checkpoint
reproduces the golden bit-for-bit, and the same run pinned every
convention used in the Rust port (patch fold order, `smart_resize`
no-op for this aligned image, bilinear position-embedding interpolation
math, RoPE `inv_freq`/rotate-half construction, `cu_seqlens` single-segment
full attention).

**Gate result**: `cargo test --release -p lattice-inference --features f16 --test vision_s3_vit_forward_test`
against the real fp16 checkpoint:
```
LATTICE_VISION_S3_COSINE cosine=1.00000000
LATTICE_VISION_S3_MUTATION_COSINE cosine=0.51125432
test qwen35_vit_forward_matches_hf_golden_cosine_gt_0999 ... ok
```

**Mutation evidence**: the gate test negates `blocks[0].qkv_weight`
in-memory and re-runs the forward pass — cosine drops from 1.0 to 0.51
(well below the 0.999 gate), proving the test is mutation-sensitive rather
than passing regardless of the weights. `qwen35_vit.rs` also has its own
`vit_forward_is_sensitive_to_weight_mutation` unit test with synthetic
weights.

**Required CI gate (new in this update)**: the checkpoint-gated cosine test
previously only ran locally — no CI job set `LATTICE_VISION_S3_MODEL_DIR` or
`LATTICE_VISION_S3_GATE_ENFORCE=1`, so the test's own skip-on-missing-dir
path made the test vacuously pass in every required check. `.github/workflows/e2e-parity.yml`
now has a `vision-s3-gate` job (provisions the pinned fp16 Qwen3.5-0.8B
checkpoint the same way `q4-composed`/`q4-ppl-quality` already do, runs the
gate test with `LATTICE_VISION_S3_GATE_ENFORCE=1` without piping away its
exit status, and greps its output for the `LATTICE_VISION_S3_COSINE` /
`LATTICE_VISION_S3_MUTATION_COSINE` markers so a future control-flow change
can't quietly turn the computation back into a skip), wired into
`parity-gate.needs` (the required status check). The test's fail-closed
path — panic, not skip, when `LATTICE_VISION_S3_GATE_ENFORCE=1` and the
model dir is unset or invalid — was already present
(`vision_s3_vit_forward_test.rs:109-135`); reverified locally this session
(see Test plan).

**Metal (S3b)**: not implemented in this PR — tracked as the fast-follow
described above, gated against this CPU reference under the GPU lock.

**Bench-compare**: not run — additive vision-tower module (new file +
`pub mod` registration + `pub(crate)` visibility changes on unmodified ViT
math helpers), no existing decode/prefill/kernel path modified
(structural).

Anchors ADR-069 S3a.

## Test plan

- [x] `cargo clippy -p lattice-inference --all-targets -- -D warnings` (clean)
- [x] `cargo clippy -p lattice-inference --all-targets --features f16 -- -D warnings` (clean)
- [x] `cargo test -p lattice-inference vision` (64 passed) and `--features f16` (65 passed)
- [x] `cargo test --release -p lattice-inference --features f16 --test vision_s3_vit_forward_test -- --nocapture` against the real checkpoint: cosine gate passes (1.0), mutation test passes (drops to 0.51)
- [x] Same gate re-run with `LATTICE_VISION_S3_GATE_ENFORCE=1` and the real checkpoint (fail-closed mode) — confirmed it actually runs against the real checkpoint rather than skipping
- [x] Same gate re-run with `LATTICE_VISION_S3_GATE_ENFORCE=1` and `LATTICE_VISION_S3_MODEL_DIR` unset, and again pointed at a nonexistent path — both panic (fail loud), neither returns early
- [x] `actionlint` on the edited `.github/workflows/e2e-parity.yml`
- [x] Local pre-commit hook (fmt/doc-lint/capability-matrix) passed
- [ ] Live CI on this PR's head — the `vision-s3-gate` job running for real on GitHub's macOS runner is the actual end-to-end proof for the required-CI blocker; posted as a PR comment once the run completes

### DIFF
--- a/.github/workflows/e2e-parity.yml
+++ b/.github/workflows/e2e-parity.yml
@@ -563,6 +563,85 @@ jobs:
             --features "f16,metal-gpu" \
             -- --nocapture
 
+  # ADR-069 S3a required gate: real Qwen3.5-0.8B ViT forward (CPU reference)
+  # vs the committed HF differential golden (cosine > 0.999). Closes the gap
+  # a static review found: the checkpoint-gated test only ran locally, never
+  # in CI, so a patch-fold/RoPE/normalization/activation regression could
+  # land while every required check stayed green (skip-as-green gate
+  # failure). LATTICE_VISION_S3_GATE_ENFORCE=1 makes the test panic (not
+  # skip) if the model dir is missing or invalid — same fail-closed shape as
+  # embed-drift and q4-composed. `set -o pipefail` keeps the `tee` capture
+  # from masking a non-zero `cargo test` exit; the grep assertions below
+  # additionally prove the real numerical path executed (not just that the
+  # process exited 0), so a future control-flow change that turns the
+  # computation into an early return cannot silently pass this job.
+  #
+  # S3a is the CPU-reference stage (this PR); S3b (Metal port vs this CPU
+  # reference under the GPU lock) is a separate fast-follow gate.
+  #
+  # NOTE: once this job has gone green on CI once, add "vision-s3-gate" to
+  # the repository's required status checks alongside embed-drift and
+  # q4-composed. That governance step is intentionally not automated here.
+  vision-s3-gate:
+    name: vision S3a cosine gate
+    needs: changes
+    if: needs.changes.outputs.engine == 'true'
+    runs-on: macos-latest
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+
+      - name: Cache Python packages
+        uses: actions/cache@v4
+        with:
+          path: ~/Library/Caches/pip
+          key: e2e-pip-${{ runner.os }}-py311-v1
+
+      - name: Cache HF model weights
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/huggingface/hub/models--Qwen--Qwen3.5-0.8B
+          key: hf-qwen35-0.8b-v1
+
+      - name: Install Python deps
+        run: pip install "huggingface_hub==1.20.1"
+
+      - name: Download Qwen3.5-0.8B weights
+        run: |
+          python3 -c "
+          from huggingface_hub import snapshot_download
+          print(f'Model at: {snapshot_download(\"Qwen/Qwen3.5-0.8B\")}')
+          "
+
+      - name: Setup lattice model dir
+        run: |
+          MODEL_CACHE=$(python3 -c "
+          from huggingface_hub import snapshot_download
+          print(snapshot_download('Qwen/Qwen3.5-0.8B'))
+          ")
+          mkdir -p ~/.lattice/models
+          ln -sf "$MODEL_CACHE" ~/.lattice/models/qwen3.5-0.8b
+
+      - uses: dtolnay/rust-toolchain@1.94.1
+      - uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: vision-s3-gate
+
+      - name: Run S3a cosine gate (fail-closed)
+        env:
+          LATTICE_VISION_S3_MODEL_DIR: ~/.lattice/models/qwen3.5-0.8b
+          LATTICE_VISION_S3_GATE_ENFORCE: '1'
+        run: |
+          set -o pipefail
+          cargo test --release -p lattice-inference --features f16 \
+            --test vision_s3_vit_forward_test -- --nocapture 2>&1 | tee vision-s3-gate.log
+          grep -q 'LATTICE_VISION_S3_COSINE' vision-s3-gate.log
+          grep -q 'LATTICE_VISION_S3_MUTATION_COSINE' vision-s3-gate.log
+
   # Fail-closed control-flow self-test for the Q4 PPL gate script (issue #616).
   # Cheap, deterministic, GPU-free (sandbox tree + stub eval_perplexity), so —
   # unlike q4-ppl-quality, whose PPL golden bootstraps on the runner — this has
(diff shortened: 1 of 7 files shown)


ohdearquant:
All three review-feedback findings addressed; pushed as 5a372462 (+ 7cc1732c allowlisting the `arange` API name for the typos checker).

**Blocker — gate skip-passes with no CI wiring**: added the `vision S3a cosine gate` job to the e2e-parity workflow (reusing its existing checkpoint provisioning and path-filter triggers), running `cargo test --release -p lattice-inference --features f16 --test vision_s3_vit_forward_test -- --nocapture` with `LATTICE_VISION_S3_MODEL_DIR` set and `LATTICE_VISION_S3_GATE_ENFORCE=1`, exit status preserved (no piping), and grep assertions that both the `LATTICE_VISION_S3_COSINE` and `LATTICE_VISION_S3_MUTATION_COSINE` markers are present so a control-flow change cannot silently turn computation into a skip. The job feeds the required `parity-gate` aggregation. The test itself now panics (fails loud) under enforcement when the model dir is missing/invalid.

**Live CI proof (this PR's head)**: the `vision S3a cosine gate` job completed successfully — `LATTICE_VISION_S3_COSINE cosine=1.00000000`, `LATTICE_VISION_S3_MUTATION_COSINE cosine=0.51125432`, 2 passed / 0 failed, marker greps enforced. Local proofs before push: enforced run with the real checkpoint passed with markers; enforced run with the model dir unset panicked (fail-loud verified); actionlint clean.

**Major — ADR naming**: retitled to ADR-069 **S3a** per Amendment 1 (#995, merged), which splits S3 into S3a (this PR: CPU reference, required fail-loud CI gate) and S3b (Metal port gated against the CPU reference under the GPU lock, fast-follow). Module/test doc comments updated to match.

**Medium — vision module docs**: `vision/mod.rs` overview rewritten to distinguish the inert ADR-049 scaffold (`vit.rs`, no consumers) from the real Qwen3.5-0.8B path (`checkpoint.rs` + `qwen35_vit.rs`), with integration status and the module future S4/S5 work should consume.

ohdearquant:
Both round-2 findings fixed at 24efcf1d8c:

**Blocker — gate/oracle changes could skip the gate**: added the S3a gate's complete input surface to the engine-change classifier allowlist: `crates/inference/tests/vision_s3_vit_forward_test.rs`, `tests/fixtures/vision/`, and `scripts/vision_goldens_qwen35.py`. A PR touching only the gate implementation or its committed HF oracle now classifies `engine=true`, so `vision S3a cosine gate` runs and the required aggregate cannot go green on a skipped result for those changes. Unrelated-PR skip behavior unchanged.

**Medium — piped gating command**: the cargo command now redirects to `vision-s3-gate.log` directly (`> ... 2>&1`, exit status uncoupled from any pipe), with the log `cat`-ed for run visibility and the two marker greps unchanged.

`actionlint` clean. This push touches the workflow itself, which is already on the classifier allowlist, so the live run on this head re-executes the gate end to end.

oceanwaves630 (approved):

PR #996, opened 2026-07-15
feat(inference): Gemma 4 config parsing and loader preflight (ADR-082 stage 2)

ohdearquant:
## Summary

ADR-082 stage 2: config parsing + safetensors loader preflight for Gemma 4 E2B's text decoder. No forward-pass or KV-cache code — that's gated on a separate Stage-4 design checkpoint per the ADR.

- **`Gemma4Config`** (`crates/inference/src/model/gemma4_config.rs`): serde parsing of the Gemma 4 E2B `config.json` (`text_config` object, matching the existing `Qwen35Config`/`HfQwenConfigFile` wrapper pattern), with fail-closed validation of every field the forward pass will eventually depend on: layer-type schedule, `num_kv_shared_layers`, sliding/global head dims, MLP widths, rope thetas, sliding window.
- **Loader preflight** (`crates/inference/src/model/gemma4_preflight.rs`): classifies every language-model tensor a `Gemma4Config` implies against a tensor inventory (validated against the committed Stage-0 manifest fixture, `tests/fixtures/gemma4/e2b_tensor_manifest.json`), without loading any weight bytes.

### Contract points implemented (ADR-082 Amendment 1)

- **Tolerate-and-skip.** The checkpoint carries `self_attn.{k_proj,v_proj,k_norm}` at all 35 layers, but layers >= the first KV-shared index (15, from `num_kv_shared_layers=20`) are accepted-but-unvalidated at whatever shape/dtype they appear — never required, never shape/dtype-checked, never wired into a module. Non-shared layers' K/V tensors are still validated exactly (missing/wrong-shape/wrong-dtype all fail closed).
- **Width split.** `head_dim` (256, sliding) vs `global_head_dim` (512, global attention layers at indices 4, 9, 14, 19, 24, 29, 34) is validated per layer, orthogonal to sharing.
- **Second structural observable.** `use_double_wide_mlp == is_kv_shared_layer` is cross-checked independently of per-tensor shape validation: the preflight reads the checkpoint's own `mlp.gate_proj`/`up_proj`/`down_proj` widths (narrow=6144 vs wide=12288) and fails closed, naming both `use_double_wide_mlp` and `is_kv_shared_layer`, if a config's `num_kv_shared_layers` disagrees with what the checkpoint's own MLP widths show.
- **Duplicate raw JSON keys.** The tensor-inventory map deserializer rejects a repeated raw key before it collapses into the `HashMap`, mirroring the existing `deserialize_weight_map_no_duplicates` pattern used for `SafetensorsIndex::weight_map`.

### Fixture evidence

All 25 new tests validate against the real, committed Stage-0 manifest (2,011 header-extracted tensors, 600 in `model.language_model.*`):

- Golden preflight passes cleanly against the fixture: 540 tensors classified `loaded`, exactly 60 (`20 shared layers * {k_proj, v_proj, k_norm}`) classified `skipped_shared_kv`.
- All 35 layers' attention geometry (256/512 split) and MLP width (6144/12288 split) match the fixture exactly.

### Mutation table

| Mutation | Expected result | Verified |
|---|---|---|
| Shared-layer K/V shape changed | Tolerated — preflight still succeeds | ✅ |
| Shared-layer K/V dtype changed | Tolerated — preflight still succeeds | ✅ |
| Non-shared-layer K/V tensor removed | Fails, error names the exact tensor | ✅ |
| `num_kv_shared_layers` set so the config's sharing set disagrees with the checkpoint's own MLP-width marker | Fails, error names both `use_double_wide_mlp` and `is_kv_shared_layer` | ✅ |
| Global layer's K/V width mutated to the sliding (256) width | Fails, error names the tensor and the expected/actual shapes | ✅ |
| Top-level tensor wrong dtype / missing | Fails, error names the tensor | ✅ |
| Duplicate raw tensor-name key in a manifest | Rejected at parse time | ✅ |

**Mutation-prove cycle** (tolerate-and-skip guard): reverse-applied the guard (replaced the shared-layer skip branch with an unconditional `check_required`), rebuilt, and confirmed 4 tests failed as expected (both tolerance tests, plus the two tests whose loaded/skipped counts depend on the skip categorization). Restored the guard and confirmed all 25 tests pass again.

### Gates

- `cargo clippy -p lattice-inference --all-targets -- -D warnings`: clean
- `cargo test -p lattice-inference gemma`: 25 new + all existing gemma4 tests pass
- `cargo fmt`: clean
- `typos` on changed files: clean

### Perf (ADR-058)

Additive-only: two new files plus a two-line module registration in `model/mod.rs`. No edits to any existing kernel, decode, or loading code path — structurally unreachable from any hot path, so `make bench-compare` is not applicable here.

## Test plan

- [x] `cargo test -p lattice-inference gemma`
- [x] `cargo fmt --check`
- [x] `typos` on changed files
- [x] Mutation-prove cycle for the tolerate-and-skip guard (revert/rebuild/fail, restore/pass)


ohdearquant:
## Review-feedback fixes: fail-closed config parse, exact `layer_types` schedule, exhaustive preflight consumption

Addressed all three blockers from the last review round. New commit: `aeb9dc865b`.

### Blocker 1 — fail-closed config parsing (`gemma4_config.rs`)

`Gemma4Config` is no longer `#[serde(derive)]` at all: it is built exclusively from a private raw
struct (`HfGemma4TextConfig`) that mirrors the target `config.json`'s `text_config` object
field-for-field, with no container-level `#[serde(default)]` and no E2B-preset substitution.

- `text_config` is now a required (non-`Option`) field of the outer file struct — `{}` fails with an
  error naming `text_config`, not a silently-completed preset.
- The two nested RoPE parameter records are now parsed from their real location,
  `rope_parameters.full_attention` / `rope_parameters.sliding_attention` (previously the loader
  declared flat `rope_theta`/`rope_local_base_freq` fields that don't exist at that JSON path, so
  serde silently ignored the real nested object and the E2B defaults leaked through unexamined).
- Added and required: `use_double_wide_mlp` (raw bool), `hidden_activation`, `final_logit_softcapping`,
  `attention_k_eq_v`, `attention_bias`.
- `use_double_wide_mlp()` is no longer a tautological restatement of `is_kv_shared_layer()`. It's now
  `raw_flag && is_kv_shared_layer(idx)`, and `validate()` rejects `use_double_wide_mlp=false` whenever
  `num_kv_shared_layers > 0`, naming both `use_double_wide_mlp` and `is_kv_shared_layer` in the error.
- Committed a pinned target-config fixture, `tests/fixtures/gemma4/e2b_config.json` — the full
  `google/gemma-4-E2B-it` @ `9dbdf8a839e4e9e0eb56ed80cc8886661d3817cf` `config.json`, fetched directly
  from the pinned revision, with a `metadata` block recording source repo/revision/file/URL (same
  provenance convention as the existing Stage-0 tensor-manifest fixture).

New field-discriminating mutation tests (each asserts the error names the specific field): absent
`text_config`, partial `text_config`, missing `rope_parameters.full_attention`, missing
`rope_parameters.sliding_attention.rope_theta`, a nested `rope_theta` value mutation (proves the value
flows through rather than being replaced by the preset), `use_double_wide_mlp: false`, wrong
`hidden_activation`, wrong `final_logit_softcapping`, `attention_k_eq_v: true`, `attention_bias: true`.

### Blocker 2 — exact `layer_types` schedule, not a repairable default

Removed the `compute_layer_types` regeneration fallback from the parse path entirely. `layer_types` is
now a required field parsed verbatim from the checkpoint, and `validate()` checks it against the
binding E2B schedule exactly: `full_attention` at indices `[4, 9, 14, 19, 24, 29, 34]` and nowhere
else, in addition to the existing length check. Any disagreement — wrong length or a same-length
schedule with a global layer in the wrong position — is rejected, naming `layer_types`.

New mutation tests: `layer_types` length 34, length 36, wrong first global index (moved from 4 to 3),
wrong final layer type (flipped from `full_attention` to `sliding_attention`) — each asserts the error
names `layer_types`.

**Mutation-proof cycle (recorded, not just described):** temporarily commented out the exact-schedule
check in `validate()`, `touch`ed the file to force a rebuild, ran the two position-mutation tests —
both failed as expected (`layer_types_wrong_first_global_index_errors_naming_field` and
`layer_types_wrong_final_layer_type_errors_naming_field`), confirming the length-only check that used
to guard this path could not catch a same-length wrong-position schedule. Restored the check,
rebuilt, and the full `gemma` test set (39 tests) passed again.

### Blocker 3 — exhaustive preflight consumption (`gemma4_preflight.rs`)

`preflight_check` now tracks every consumed tensor name via the union of `report.loaded` and
`report.skipped_shared_kv` (both already populated exactly at the point each tensor is classified).
After the per-layer walk, it scans the supplied inventory for any `model.language_model.*` name not in
that consumed set and rejects it, naming the tensor. The golden test now asserts the input inventory
count equals `loaded.len() + skipped_shared_kv.len()` exactly, in addition to the existing 540/60
counts.

New mutation tests:
- inject an unexpected language-model tensor (`layers.0.self_attn.mystery_proj.weight`) → preflight
  fails, error names that exact tensor.
- remove a shared-layer K/V entry (`layers.20.self_attn.k_proj.weight`) → preflight still succeeds
  (tolerate-and-skip proof), the removed name is absent from `skipped_shared_kv`, and the exhaustiveness
  identity (`inventory == loaded + skipped`) still holds with one fewer of each.

### Gates

- `cargo test -p lattice-inference gemma` — 39/39 passing (config + preflight suites), plus the
  existing tensor-manifest and tokenizer-goldens suites unaffected
- `cargo fmt` — no diff
- `typos` on changed files — clean

oceanwaves630 (approved):

PR #998, opened 2026-07-15
feat(inference): vision patch-merger projection gated vs post-merger golden (ADR-069 S4)

ohdearquant:
## Summary

Implements ADR-069 Stage S4: the patch-merger/projection stage that maps the S3a pre-merger ViT hidden states `[256, 768]` to the post-merger projected visual embeddings `[64, 1024]` — the exact tensor the decoder consumes at image-token positions.

- `crates/inference/src/vision/qwen35_merger.rs`: `qwen35_merger_forward`, mirroring the pinned HF reference (`transformers.models.qwen3_5.modeling_qwen3_5.Qwen3_5VisionPatchMerger`, same `transformers==5.12.1` / `Qwen/Qwen3.5-0.8B` pin as S3a). Op order: per-patch `LayerNorm(hidden_size)` (before the fold — `use_postshuffle_norm=False` in the reference) → fold `spatial_merge_size^2` consecutive patches into one row (a plain reshape, since S3a's `preprocess_qwen35_image` already emits patches in block-major order — no permutation needed) → `Linear(merge_in, merge_in)` → **exact/erf GELU** → `Linear(merge_in, out_hidden_size)`.
- Reuses `Qwen35VisionWeights.merger` (already loaded by the S3a/S2 checkpoint loader) — no new weight-loading code.

### A convention detail worth flagging

The merger's activation is HF's `nn.GELU()` (exact, erf-based), hardcoded independent of `config.hidden_act`. The ViT block MLPs use `ACT2FN["gelu_pytorch_tanh"]` (the tanh approximation, already implemented in `vit.rs`). Reusing the tanh approximation for the merger would have been a plausible-looking mistake — it's off by design, not just precision. Empirically the tanh-vs-exact delta was too small to trip the cosine>0.999 gate on this checkpoint (see mutation cycle below), so this is recorded here as a correctness note, not something the gate independently proves.

## Gate evidence (real Qwen3.5-0.8B checkpoint, `cargo test --release --features f16`)

Isolated-stage gate (committed pre-merger golden → merger → compare vs committed post-merger golden — isolates numerical error to this stage):
```
LATTICE_VISION_S4_COSINE cosine=1.00000000
LATTICE_VISION_S4_MUTATION_COSINE cosine=-0.65023019
```

End-to-end gate (image → preprocess → S3a ViT forward → merger → compare vs committed post-merger golden — closes the full CPU vision-tower pipeline claim):
```
LATTICE_VISION_S4_E2E_COSINE cosine=1.00000000
```

Both stages pass with `LATTICE_VISION_S3_GATE_ENFORCE=1` (fail-closed: panics rather than skips if the checkpoint is missing).

## Mutation-sensitivity cycle

1. Baseline: isolated-stage cosine `1.00000000`.
2. Reverse-applied a piece of the merger (removed the pre-fold `LayerNorm` application) and `touch`ed the source file to force a rebuild.
3. Both gate tests failed: isolated cosine dropped to `0.28791351`, end-to-end to `0.28791367`.
4. Restored the fix, rebuilt, re-ran: both gates green again at cosine `1.00000000`.

The in-test mutation assertion (negating `merger.fc1_weight`, asserted inline in both gate tests) reproduces this same fail-then-pass discrimination on every run, not just this one-off cycle.

## CI wiring

Extended the existing `vision-s3-gate` job (renamed `vision S3a+S4 cosine gate`) rather than adding a new job: it already downloads and caches the Qwen3.5-0.8B checkpoint via the same `LATTICE_VISION_S3_MODEL_DIR` env var, so running the S4 test binary alongside S3a's in the same `cargo test --test ... --test ...` invocation adds seconds, not a second full download/checkout. Added `LATTICE_VISION_S4_COSINE`, `LATTICE_VISION_S4_MUTATION_COSINE`, and `LATTICE_VISION_S4_E2E_COSINE` grep assertions mirroring the existing S3a markers, so a future control-flow regression that turns either computation into an early return can't silently pass.

Extended the engine-change classifier allowlist with the new test file path (`crates/inference/tests/vision_s4_merger_test.rs`) — the S3a review's headline lesson was that a gate's own inputs must classify `engine=true`. No new fixture files were added (S4 reuses the existing committed `vit_pre_merger_f32.bin` / `vit_post_merger_f32.bin` goldens, already allowlisted), and the merger source file falls under the existing `crates/inference/src/` prefix.

## ADR-058 (perf gate)

Additive vision-only code: a new module (`qwen35_merger.rs`) and a new test file, both under `crates/inference/src/vision/` and `crates/inference/tests/`, plus a one-line `pub mod` addition and docstring update in `vision/mod.rs`. No shared kernel or hot-path file is touched, and nothing outside the vision module calls the new function — structurally unreachable from any existing benchmark or decode/prefill path. `make bench-compare` was not run; there is nothing on the measured surface for it to compare.

## Test plan

- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [x] `cargo clippy -p lattice-inference --all-targets --features f16 -- -D warnings` — clean (covers the feature-gated gate code)
- [x] `cargo test -p lattice-inference vision` — 71 lib tests + fixture-only integration tests pass (no local checkpoint dependency)
- [x] `cargo test --release -p lattice-inference --features f16 --test vision_s3_vit_forward_test --test vision_s4_merger_test` against the real local checkpoint — all 5 tests pass, cosine 1.0 on both S4 gates
- [x] `cargo fmt` — clean
- [x] `typos` on changed files — clean
- [x] Mutation-sensitivity proven both inline (per-test weight negation) and via a full revert/touch/fail/restore/pass cycle (see above)


oceanwaves630 (approved):
Full-diff read at 0f4e374857 plus delta read at 20ab446750. The kickback was resolved in the stronger direction: instead of correcting the body claim, the end-to-end gate now carries the same inline mutation assert as the isolated gate (negate fc1_weight, require cosine below 0.999), and the parity workflow greps the new E2E mutation execution marker so the assert cannot silently stop running. Merger order and exact-GELU choice match the reference implementation, no unwrap in library code, the conflict resolution preserved both the S3b Metal documentation and the ratchet job name and waiver, and the S4 test path is in the engine filter with all four execution markers enforced. Approved.

PR #999, opened 2026-07-15
feat(inference): Gemma 4 text math kernels with per-op HF goldens (ADR-082 stage 3)

ohdearquant:
## Summary

Implements ADR-082 stage 3 (text math kernels) for Gemma 4 E2B: the six
text-decoder primitives (G6-G10) as pure per-op kernels, goldens-first
against the pinned HF `transformers` reference. No attention/cache/forward
wiring in this PR — that is stage 4, and is design-gated per the ADR.

- Standard RMSNorm (plain gamma, not the shifted `(1+gamma)` Qwen3.5 variant)
- GeGLU MLP (tanh-approximate GELU gate, elementwise product with the up
  branch)
- Scaled embedding (`lookup * sqrt(hidden_size)`)
- Per-head Q/K RMSNorm (scaled) with V RMSNorm (unscaled), applied before
  RoPE for Q/K and never applied to V's rotation
- Final logit softcapping (`cap * tanh(logits / cap)`)
- Dual RoPE: local (sliding) layers at theta=1e4 over the full head width,
  global (full-attention) layers at theta=1e6 with a 0.25 partial-rotary
  factor (only a quarter of the frequency slots are non-zero; the rest are
  exact zero, which passes those dimension-pairs through unrotated rather
  than truncating them)

## Goldens

`scripts/gemma4_stage3_goldens.py` instantiates the pinned HF
`transformers.models.gemma4.modeling_gemma4` classes directly with small,
deterministic, seeded synthetic weights/inputs — no checkpoint download, no
network access. Fixtures are raw little-endian float32 `.bin` files (one per
tensor) plus a small JSON sidecar per op and a `manifest.json` that
predeclares each op's tolerance and its justification *before* any Rust
comparison ran (the Rust tests read the tolerance from the manifest, not a
hardcoded constant).

| Op | Shape | Tolerance (max-abs-diff) | Observed max-abs-diff |
|---|---|---|---|
| `rms_norm` | `[2,3,64]` | 1e-6 | 4.8e-7 |
| `geglu_mlp` | `[2,3,64]`, intermediate 96 | 1e-5 | 1.2e-7 |
| `scaled_embedding` | `[2,5,64]` | 1e-6 | 0 |
| `qk_norm_v_unscaled` | `[1,3,4,16]` | 1e-6 | q: 2.4e-7, k: 0, v: 2.4e-7 |
| `logit_softcap` | `[2,5,40]` | 3e-6 | 1.9e-6 |
| `dual_rope` (local) | `[1,6,2,16]` | 1e-6 | 2.4e-7 |
| `dual_rope` (global) | `[1,6,2,32]` | 1e-6 | 1.2e-7 |

`logit_softcap`'s tolerance is 3e-6 rather than the initially-estimated 1e-6:
Rust's `f32::tanh` and PyTorch's libm `tanh` are two different
correctly-rounded implementations, not the same routine, so a few ULPs of
disagreement is expected and was measured (1.9e-6) before the tolerance was
finalized — not loosened after a failing run to hide anything, the estimate
was simply too tight for a two-different-tanh-implementations comparison.

## Negative tests (fail-closed, mutation-sensitive)

Each of the ADR's named mutations has its own test that constructs the wrong
op and asserts it diverges from the golden **beyond** tolerance:

| Mutation | Test | Observed diff (tolerance) |
|---|---|---|
| Qwen3.5 shifted `(1+gamma)` RMSNorm substituted for standard | `mutation_shifted_rms_norm_fails_golden` | 3.29 (1e-6) |
| Scaled norm applied to V instead of unscaled | `mutation_scaled_v_norm_fails_golden` | 1.60 (1e-6) |
| Local/global RoPE thetas swapped | `mutation_swapped_theta_fails_golden` | 0.87 local / 1.07 global (1e-6) |
| Softcap disabled (identity passthrough) | `mutation_disabled_softcap_fails_golden` | 31.0 (3e-6) |

## Mutation-proof (RMSNorm)

Beyond the standing negative test, the RMSNorm fix itself was mutation-proven
with a real revert/restore cycle: the shifted-RMSNorm mutation was applied
directly to `gemma4_rms_norm`'s body (not just exercised by a separate
always-wrong test), the file was touched to force a rebuild, and the golden
test was watched fail:

```
thread '...::rms_norm_matches_hf_golden' panicked:
rms_norm max-abs-diff 3.2873282 exceeds tolerance 0.000001
```

The fix was then restored and the full suite watched pass again (12/12).

## Reuse vs. new code

`gemma4_rms_norm` and `gemma4_geglu_mlp` reuse the existing
`crate::forward::cpu::{rms_norm, matmul_bt, elementwise_mul}` kernels as-is —
verified bit-for-bit-equivalent op order to the HF reference. GeGLU's
activation is a **new** exact-tanh GELU rather than the existing SIMD `gelu`
kernel: that kernel's fast-tanh approximation has ~4e-5 max error on its own,
which would blow this stage's predeclared 1e-5 tolerance before any matmul
propagation, so it was not reused.

## ADR-058 (perf gate)

Additive-only: no shared kernel file in `forward/cpu/` was edited, only
called from the new `gemma4_ops` module, which has no callers yet
(structurally unreachable from any existing forward path). No
`make bench-compare` run is included — none is needed under ADR-058 for a
purely additive, uncalled module.

## Test plan

- [x] `cargo test -p lattice-inference gemma` — 51 passed (12 new stage-3 +
      39 pre-existing stage 1/2), 0 failed
- [x] `cargo clippy -p lattice-inference --all-targets -- -D warnings` — clean
- [x] `cargo fmt` — clean
- [x] `typos` on changed files — clean
- [x] Goldens script re-run in verify mode reproduces the committed fixtures
      byte-for-byte
- [x] RMSNorm negative test mutation-proven with a real revert/touch/fail/restore/pass cycle


ohdearquant:
Addressed all 6 findings from review feedback (3 major, 3 medium). Commit `1fdc5a87a5` on this branch. Summary below with evidence.

## Major 1 — bind goldens to the pinned source commit

The generator now installs the pinned `transformers` commit via `uv run --with "transformers @ git+https://github.com/huggingface/transformers@ab1771c9e42891d893189978a8009426d70b4688"` and fails closed (`verify_source_pins`, no fixture writes) unless the SHA-256 of the actually-imported `modeling_gemma4.py`, `modeling_rope_utils.py`, and `configuration_gemma4.py` match the manifest's `pinned_source_hashes`.

Evidence:
- Running under the pinned commit: `OK: all Stage-3 goldens match the committed fixtures byte-for-byte (source pins, runtime versions, and reference-derived RoPE parameters verified against the committed manifest)`.
- Running under the previously-installed `transformers==5.12.1` release tag correctly fails closed:
  ```
  FATAL: the transformers install executing this script does not hash-match the pinned reference commit ab1771c9e42891d893189978a8009426d70b4688 -- goldens generated under it would not be reference-verified.
  Mismatches (found vs expected):
    transformers/models/gemma4/modeling_gemma4.py: expected ccab8e2d..., found 90bb5dbd...
    transformers/modeling_rope_utils.py: expected 9592ba12..., found f370d858...
  ```
  This confirms the release tag and the pinned commit genuinely diverge in these two files, as flagged in review.

## Major 2 — softcap golden from the reference module

`gen_logit_softcap` now builds a real `Gemma4ForCausalLM`/`Gemma4TextConfig`, stubs only the decoder stack (`model.model`) with a fixed hidden-states tensor, and calls the real `forward()` — so the captured logits come from the reference's own `lm_head` + `final_logit_softcapping` branch, not a local re-derivation. Verified the fixture still saturates (`input` range ±115, `output` range ±30, cap=30), so the disabled-softcap negative test stays meaningful.

## Major 3 — RoPE thetas from the reference config path

`verify_rope_reference` loads the committed pinned E2B config fixture (`tests/fixtures/gemma4/e2b_config.json`) through the real `Gemma4TextConfig`, and asserts the derived `rope_parameters` equal the manifest's `expected_rope_parameters` before any fixture write. The dual-RoPE golden's theta/partial-rotary-factor values are threaded from that reference-derived result, not hardcoded literals. Synthetic head_dim/hidden_size reductions are unchanged (as review noted these should stay reductions).

## Medium 4 — RMSNorm bit-for-bit claim

Corrected: the doc comment on `gemma4_rms_norm` now states bounded f32 parity, not bit-for-bit equivalence, and explains why (`sqrt` + reciprocal vs the reference's single `pow(x, -0.5)`; differing SIMD reduction order). Added a new `rms_norm_wide` golden at the real E2B `hidden_size=1536` (vs the existing 1/24-scaled width=64 golden) to demonstrate the bound holds at real width, with its own manifest-declared tolerance and Rust test (`rms_norm_wide_matches_hf_golden`).

## Medium 5 — mutation-separation floor

`manifest.json` now predeclares a `mutation_separation_floor` per op with a negative test (`rms_norm`, `qk_norm_v_unscaled`, `logit_softcap`, `dual_rope`), each 10x that op's tolerance. All four negative tests (`mutation_shifted_rms_norm_fails_golden`, `mutation_scaled_v_norm_fails_golden`, `mutation_disabled_softcap_fails_golden`, `mutation_swapped_theta_fails_golden`, including both of its local/global assertions) now assert `diff >= floor` instead of bare `diff > tol`.

## Medium 6 — tolerance provenance

`manifest.json` is now a committed, hand-authored, immutable provenance spec (schema_version 2): pinned source hashes, per-op seeds, tolerances with rationale derived from op structure (accumulation depth, transcendental-call ULP behavior, reduction width) rather than any observed lattice diff, mutation-separation floors, and the expected reference-derived RoPE parameters. The generator only ever reads and verifies against it — `--write-fixture` no longer rewrites `manifest.json`, only the per-op `.json`/`.bin` fixture files.

## Gates

- `cargo clippy -p lattice-inference --all-targets -- -D warnings` — clean.
- `cargo test -p lattice-inference gemma` — 52/52 unit tests passing (`gemma4_ops`, `gemma4_config`, `gemma4_preflight`), all integration suites green.
- `cargo fmt` — clean.
- `typos` on all changed files — clean.
- Generator verify mode (pinned source): `OK: all Stage-3 goldens match the committed fixtures byte-for-byte`.

## Mutation cycle (shifted-RMSNorm golden)

To confirm the RMSNorm goldens are mutation-sensitive, not decorative: reverted `gemma4_rms_norm` to delegate to the (wrong) shifted-gamma `qwen35_rms_norm` kernel, touched the file, and re-ran:

```
test model::gemma4_ops::tests::rms_norm_matches_hf_golden ... FAILED
  rms_norm max-abs-diff 3.2873282 exceeds tolerance 0.000001
test model::gemma4_ops::tests::rms_norm_wide_matches_hf_golden ... FAILED
  rms_norm_wide max-abs-diff 3.8161113 exceeds tolerance 0.00005
```

Restored the correct implementation and re-ran: `52 passed; 0 failed`. Both the existing narrow golden and the new realistic-width golden catch the regression.

ohdearquant:
Round-2 residual (manifest immutability was only half-enforced) addressed in `782efb209e`.

The generator now pins a byte-exact SHA-256 of the committed `manifest.json` in `MANIFEST_CONTRACT_SHA256` and fails closed in **both** modes, before `FIXTURE_DIR.mkdir()` and any `materialize_op` call, if the on-disk manifest does not hash-match. Changing any declared manifest value (seed, tolerance, mutation floor, rationale, source hash, or metadata field) now requires a paired, review-visible edit to that constant — a coordinated manifest + payload rewrite can no longer pass as a silent regeneration.

Guard mutation-tested under the pinned source install:

- Committed manifest, verify mode: `OK: all Stage-3 goldens match the committed fixtures byte-for-byte (source pins, runtime versions, and reference-derived RoPE parameters verified against the committed manifest)`.
- Mutated `ops.rms_norm.seed` (999), `--write-fixture` mode: fails closed before any write —
  ```
  FATAL: manifest.json does not hash-match the reviewed contract digest.
    expected cc7088a3...
    found    c19bb794...
  ```
  No fixture files were written (clean `git status` after the run); manifest restored byte-exact.

oceanwaves630 (approved):
Approved after a triple-round external gate plus an independent full-diff source-integrity read: scope confirmed limited to the stage-3 module, fixtures, and generator; pinned upstream sources independently hash-verified against the manifest; all 29 committed binaries match their sidecar digests; no test deletions or weakened assertions.

Issue #1000, opened 2026-07-15
ci(e2e-parity): q4 composed golden numeric flake — chain probe max_abs_error=15.8 on first attempt, green on identical rerun


ohdearquant:
## Summary

The `q4 composed golden` CI leg failed numerically on a PR run and then passed on rerun with identical code — an unexplained numeric flake in the leg itself.

## Evidence

- PR #998, e2e-parity run 29411705169 (first attempt, 2026-07-15): the quantize chain probe refused forward-equivalence with `chain probe max_abs_error=15.796389407429611 exceeds tolerance=0.00001 (mean_abs_error=2.0200582355371632, probe_tokens=[4950, 132176, 10332, 136459])`.
- Same run, rerun of failed jobs, identical merge ref: `q4 composed golden` passed (18m55s) and the required `parity-gate` aggregation went green.
- The PR's diff is vision-only (patch-merger module + its gate test + workflow job extension) — structurally unreachable from the quantize pipeline.
- main at the same base commit (4a041422) ran the same job to success.
- Main-branch history for the leg over 7/04–7/15 (100 runs): 86 success, no numeric failures (one 7/11 red was a rustup setup infrastructure error that never reached the parity check; two skips were scheduled-run path-filter behavior).

## Why this matters

This leg is a candidate for promotion to a required status check (ADR-066 F2, tracked in #320). An error of magnitude ~15.8 against a 1e-5 tolerance that disappears on rerun is not marginal-threshold noise — something in the chain probe or its environment is bimodal. Until root-caused, promotion would inject spurious merge blocks.

## Suspects to investigate

- Runner floating-point environment differences between attempts (paravirtualized macOS runner variance).
- Nondeterminism in the quantize/conversion chain probe (uninitialized or order-dependent state; probe token selection is recorded and identical, so input selection is not the variable).
- Cache interaction: the job's model/asset cache state differed between attempts (`CACHE_ON_FAILURE: false` was logged on the failing attempt).

## Acceptance

Root cause identified and either fixed or bounded (e.g., deterministic probe seeding, cache validation, or a documented environment gate), after which the leg's promotion checkpoint can be re-evaluated with a clean streak.

ohdearquant:
Adding one more suspect from triage discussion: **concurrent-load contention on the runner during the failing attempt**.

[...]

Issue #535, opened 2026-07-02
Metal backend diverges from HF/CPU at the first generated token on the long-prefill prompt (repetition penalty and GDN mode excluded)


ohdearquant:
> _Issue authored by Claude (Anthropic agent) on behalf of @ohdearquant._

The new Metal attention parity job (informational leg, PR #529) surfaced a real engine divergence on its first run: 3/4 e2e prompts match HF exactly on the Metal backend, but the ~816-token long-prefill `merge_sort` prompt diverges at position 0 — HF/CPU continue `merge_sort(arr):`, Metal continues `main():`.

**Differential triage (local M-series, real GPU, Qwen3.5-0.8B, temperature 0.0):**

| Configuration | First tokens |
|---|---|
| chat_metal, default rep-pen 1.1 | ` main():` |
| chat_metal, `--repetition-penalty 1.0` | ` main():` |
| chat_metal, rep-pen 1.0 + `LATTICE_GDN_CHUNKED=0` (serial GDN prefill) | ` main():` |
| HF transformers (reference) and lattice CPU path (CI e2e gate, 4/4 PASS) | ` merge_sort(arr):` |

So two prime suspects are excluded:
- **Not the repetition-penalty harness mismatch** (the #520-class issue) — reproduces with 1.0 forced.
- **Not GDN chunked-vs-serial prefill** (#534) — reproduces identically in both GDN modes.

What remains is the Metal attention/GEMM prefill path itself on long prompts. Notably, #228 originally suspected Metal chunked prefill but was closed-as-unanswerable for CI purposes because the f16 CPU-only CI build never exercised the Metal path. The new informational job exercises it for the first time, and the long-prefill prompt fails. Short prompts (≤ a few dozen tokens) match HF exactly, so the divergence accumulates or activates with prefill length — consistent with an accumulation-precision or tiling/boundary effect in the prefill GEMM/attention kernels rather than a wholesale wrong kernel.

Next diagnostic step (differential-test-first): dump post-prefill logits for the same long prompt on CPU vs Metal and bisect by layer/op, or sweep prefill length to find the divergence onset.

Both continuations are plausible; this is a near-tie argmax flipped by numeric divergence, so greedy short-window agreement is the right detector. The job is doing its job — this issue tracks the engine-side root cause. Refs #239, #228, #534.

ohdearquant:
> _Comment authored by Claude (Anthropic agent) on behalf of @ohdearquant._

[...]

Issue #320, opened 2026-06-25
test(quarot): no CI gate covers the composed rotated+Q4 forward (f64 equivalence is pre-quant, e2e-parity is bf16)


ohdearquant:
> _Issue authored by Claude (Anthropic agent) on behalf of @ohdearquant._

## Summary

CI has two complementary equivalence defenses, but neither covers the **composed rotated+Q4 forward** end-to-end:

1. `assert_forward_equivalence_qwen35` (`quant/quarot/forward_equivalence.rs`) validates the rotation/fusion algebra **entirely in f64**, **before** quantization, at `max_abs_error ≤ 1e-5` (convert.rs:6, :390). By design it cannot see Q4 error (quant intentionally introduces 0.1–0.3 PPL, orders of magnitude above 1e-5).
2. `e2e-parity.yml` gates greedy-token agreement vs HF — but on the **bf16** model, not the Q4 artifact.

The Q4 path is component-tested (q4_weights roundtrip, Q4 GEMV/GEMM kernel tests) and the end-to-end Q4 PPL is *measured* (lattice Q4 ≈18.08, beats MLX Q4 ≈18.18) — but that PPL number is **not a CI gate**. So a regression in the interaction between QuaRot-rotated weights and the Q4 quantizer/kernels (the exact surface behind the historical "~1 PPL Q4 gap" investigation) would not fail any check; it would only show up as a manual PPL re-measure.

## Severity

Medium — defense-in-depth, not an unguarded correctness hole. The components are individually covered; what's missing is a standing composed-path regression gate. Not urgent, but valuable: a stored-golden Q4 logit gate would have shortcut the prior Q4-gap investigation (which turned out to be a stale measurement, not a bug — exactly the ambiguity a gate removes).

## Proposed work (needs design — do not land an unvalidated golden)

- Add a committed-golden regression test: rotated+quantized Qwen3.5-0.8B → first-N greedy logits/tokens, asserted against a stored golden at a **Q4-appropriate tolerance** (not 1e-5). Reference choice (MLX Q4 vs a frozen lattice-self golden) is the key design decision.
- Alternatively, extend `e2e-parity.yml` with a Q4 variant gated at token-agreement (greedy tokens are quant-robust; logit max-diff is not).
- Whichever path: the gate must use a tolerance matched to Q4 quantization error, and the golden must be regenerated deliberately (not auto-updated), mirroring the HF-parity JSON-golden pattern already in the repo.

## Provenance

[...]

Issue #239, opened 2026-06-23
ci(e2e-parity): Metal GPU attention path is not exercised by the parity gate


ohdearquant:
## Problem

`e2e-parity.yml` is the parity gate for PRs touching `crates/inference/src/`, but it does **not** exercise the Metal GPU attention path — the very path most inference PRs change.

The workflow builds `qwen35_generate` with `--features f16` only (no `metal-gpu`). `qwen35_generate` drives the **CPU** path (`Qwen35Model::generate` → `KvCache` / `prefill_tokens` / `decode_loop`). The Metal path (`MetalQwen35State`, the `decode_attention` / `prefill_attention_batched_causal` MSL kernels, the Metal KV cache) is driven only by `chat_metal` / `ppl_metal` / `bench_decode_ab`, none of which CI runs.

## Impact

A PR can break Metal greedy generation, the Metal KV cache, or a Metal attention kernel and still pass `e2e-parity.yml` green. This surfaced while verifying #154 (f16 KV cache, Metal-only): the parity gate had no view of the path under change, so all parity evidence had to be produced manually on a local M-series machine.

## Proposed fix

Add a Metal parity job to `e2e-parity.yml` (macOS runner, which the workflow already uses):
- Build `chat_metal` with `--features "f16,metal-gpu"`.
- Run greedy generation (temperature 0) on the committed reference model and assert first-N-token agreement against the HF reference output already produced in the workflow.
- Gate it the same way the CPU parity check is gated.

Alternatively, make `qwen35_generate` capable of selecting the Metal backend under `--features metal-gpu` so one binary covers both, and parametrize the parity check over backends.

## Acceptance

- A deliberate one-token regression in a Metal attention kernel fails `e2e-parity.yml`.
- The Metal KV-cache path is covered by the gate.

Discovered during #154 / PR #238.

ohdearquant:
**Still open (2026-06-23):** the f16-KV merge (#238) reconfirmed this gap — `e2e-parity.yml` still builds only the CPU path (`--features f16`), so the Metal GPU attention path remains CI-invisible. Fix: a Metal-gated parity job (`chat_metal`, first-N greedy-token agreement vs the HF reference already produced in the workflow).

ohdearquant:
> _Comment authored by Claude (Anthropic agent) on behalf of @ohdearquant._

[...]

PR #1001, opened 2026-07-15
ci: promote the Metal e2e leg to a required parity ratchet (#239, #535 waived)

ohdearquant:
## Summary

Promotes the Metal e2e leg (#239) from informational to a **required regression ratchet**, per the ADR-066 F2 two-week clean-streak policy.

## What the required gate certifies (and what it does not)

This is a **ratchet, not full-parity certification**. `scripts/e2e_parity_check.py` carries a deliberate, content-anchored waiver (`METAL_EXPECTED_DIVERGENCE`) for the long-prefill prompt, where Metal deterministically diverges from HF at the first generated token — open issue #535 (repetition penalty and GDN prefill mode both excluded as causes; the CPU leg passes and still gates on the same prompt).

The promoted job therefore gates, on every engine-change PR:

- **Metal dispatch path-proof** — the runtime `LATTICE_METAL_PATH_PROOF` marker must be present with nonzero counters (fails closed, never skip-as-green);
- **binary health** — any harness/binary failure fails the job;
- **greedy token parity on every non-waived prompt** — any divergence on any prompt other than the waived one fails closed. The waiver is keyed by prompt content, so the entire #535 prompt — including any different or new divergence on it — stays green while the waiver entry exists.

The waiver self-reports: if the waived prompt starts passing, the run prints **XPASS** and instructs a maintainer to delete the entry. This is advisory — nothing removes the entry automatically, so acting on an XPASS report is a maintainer responsibility. The job's CI display name — `metal parity ratchet (#535 waived)` — states the scope so the required gate is never misread as full Metal parity. #535 root-cause is a named, tracked follow-up.

## Evidence

Main-branch history for the leg over 2026-07-04 → 2026-07-15 (last 100 e2e-parity runs):

- 89/89 runs that executed the job concluded **success** — zero reds of any kind (no numeric failures, no infra failures).
- The remaining 11 runs did not execute the job because change detection classified them engine=false (docs/CI-only merges), which is the intended skip path, not a gap.

## Mechanics

- `metal-parity` joins `parity-gate.needs`; on engine-change PR runs the required `parity-gate` check now fails unless the Metal leg succeeded.
- The job keeps its deliberate merge-group skip (it never delays the queue); the gate accepts exactly `skipped` on `merge_group` events and nothing else. The PR run remains the enforcement point, matching how the leg has been observed for the full streak.
- The required status check name (`parity-gate`) is unchanged, so no branch-protection edit is needed.

## Not promoted

The sibling `q4 composed golden` leg (#320) is deliberately **not** promoted in this PR: it produced an unexplained numeric flake on 2026-07-15 (#1000, max_abs_error 15.8 vs 1e-5 tolerance, green on rerun of identical code). It re-checkpoints after a clean week once #1000 is root-caused.

ADR-058: workflow + comment-only change, no engine code touched; bench-compare not applicable.

### DIFF
--- a/.github/workflows/e2e-parity.yml
+++ b/.github/workflows/e2e-parity.yml
@@ -180,11 +180,14 @@ jobs:
   # for the long-prefill prompt, where Metal deterministically diverges from
   # HF at the first token (open issue #535). What this required job certifies
   # is therefore: Metal dispatch path-proof, binary health, and greedy token
-  # parity on every prompt EXCEPT the #535-waived one. Any NEW divergence on
-  # any prompt still fails closed. The waiver self-reports: if the waived
-  # prompt starts passing, the run prints XPASS and the entry gets deleted
-  # (see the METAL_EXPECTED_DIVERGENCE comment in the script). Do not read
-  # this gate as full Metal parity until #535 is fixed and the waiver removed.
+  # parity on every prompt EXCEPT the #535-waived one. Any divergence on any
+  # NON-WAIVED prompt fails closed; the waiver is keyed by prompt content, so
+  # the entire waived prompt — including a different or new divergence on it —
+  # stays green while the entry exists. The waiver self-reports: if the waived
+  # prompt starts passing, the run prints XPASS and instructs a maintainer to
+  # delete the entry (advisory — nothing removes it automatically; see the
+  # METAL_EXPECTED_DIVERGENCE comment in the script). Do not read this gate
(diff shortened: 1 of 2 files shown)


oceanwaves630 (approved):
Approved after a triple-round external gate plus an independent full-diff read: the waiver is exact-prompt-keyed and Metal-only, advisory pass-through is reported with removal instructions rather than silently absorbed, the required check name states the waived scope explicitly, and no numeric threshold or existing gate condition changed.

PR #1002, opened 2026-07-15
feat(inference): Gemma 4 stage-4 shared-KV cache with donor-slot indirection

ohdearquant:
## Gemma 4 Stage 4: shared-KV cache with donor-slot indirection (ADR-082 Amendment 1)

Adds `Gemma4KvCache` (`crates/inference/src/model/gemma4_cache.rs`), additive-only per ADR-058 — no existing kernel, cache, or model file is modified beyond the one-line module wire-up in `model/mod.rs`.

### Design

Gemma 4 functionally shares K/V across its trailing `num_kv_shared_layers` layers (E2B: 20 of 35). Sliding slots implement the measured publish rule above: on each append the slot compacts existing content to its last `window − 1` positions, then appends the entire incoming chunk; the read view is the whole resulting buffer (retention is lazy, at the next append). Global slots retain the full context. This cache allocates storage slots ONLY for the non-shared layers and resolves every layer through a build-time `layer_kv_slot` indirection map: non-shared layers own their slot; each KV-shared layer maps to its donor — the last non-shared layer of the same attention type (E2B: sliding donor = layer 13, global donor = layer 14). Shared layers are read-only at the API level: `append_kv` on a shared layer returns an error; reads go through `k_view`/`v_view`, which for a shared layer is exactly the donor's slot. No copies, no side buffer. KV memory is 15/35 of the naive layout on E2B.

Build-time asserts (hard `InferenceError`s, not `debug_assert`): slot count must equal the non-shared layer count (derived independently via `is_kv_shared_layer` so the check is not tautological), and every layer's resolved donor must agree on attention type.

### Reference-semantics probe (run before any code was written)

The one open design parameter was what a sliding donor publishes past the window boundary — window-truncated or full-length states (the upstream flag is named `store_full_length_kv`). Measured directly against pinned `transformers` @ `ab1771c9e42891d893189978a8009426d70b4688` with a tiny random-init Gemma 4 text config (7 layers, 4 KV-shared, `sliding_window=8`, prompt length 24 > window, seed 7), monkeypatching the attention forward to record the published shapes and to force the alternative interpretation:

- **Measured shapes**: prefill (q_len=24) publishes the full chunk `K(1,1,24,·)`; the decode step (q_len=1, boundary+1) publishes `K(1,1,8,·)` — exactly `sliding_window`. A second probe under **chunked** prefill (24 tokens as chunks 16+8, then two decodes) pinned the full per-step rule: chunk 1 publishes 16; chunk 2 publishes 15 = (window−1) retained tail + the whole chunk; each decode publishes 8 = 7+1. **Rule: the sliding donor retains the last `window − 1` positions after each update, and the published view for an append of `q_len` tokens is the retained tail plus the entire current chunk.** (`store_full_length_kv` means "this layer publishes", not "retains full-length".)
- **Losing interpretation shown too**: forcing the publish down to the last `sliding_window` positions at decode changes boundary+1 logits by **max |Δ| = 0.000e+00** vs stock — the interpretations are numerically identical because the consumption-side sliding mask already restricts attention to the window. The storage question is moot for the math; the donor slot needs no full-length retention.
- One unexpected observation was chased to source before recording: the global donor published K/V with last dim 512 rather than `head_dim`. This is by design — full-attention layers use `config.global_head_dim` (E2B sets 512) while sliding layers use `head_dim` (256), matching the checkpoint's `[512, 1536]` global `k_proj`. Consequence implemented here: slots are heterogeneous-width and allocation is per-slot-sized, never a uniform `kv_dim`.

Replay: `uv run --no-project --with "transformers @ git+https://github.com/huggingface/transformers@ab1771c9e42891d893189978a8009426d70b4688" --with torch --with numpy python <probe script>` (probe constructs the tiny config inline; prints the measured shapes, both-interpretation delta, and verdict).

### Storage choice

f32, not the f16 used by `FlatKVCache`: this stage targets CPU-only correctness, and the read path returns borrowed slices directly, which a quantized backing store would require dequantizing into scratch on every read. f16 packing can be layered on later without changing the module's public shape.

### Tests (15, all mutation-sensitive)

1. Slot-map correctness on the committed `e2b_config.json` fixture: 15 slots, layers 15–34 map to donor 13 (sliding) or 14 (global), same-type invariant holds.
2. Shared-layer write rejected with a descriptive error.
3. Sliding publish rule: a 24-token prefill through window=8 publishes all 24; the next decode step publishes the last 7 + the new token (K and V both asserted with distinct data).
4. Shared-layer view equals donor view after every append (both types).
5. Chunk-boundary: two-chunk prefill across the window boundary; after chunk 2 the view is the (window−1) retained tail of chunk 1 plus all of chunk 2, for K and V, from both the donor and the shared layer.
6. Build-time assert negatives: no same-type donor errors; all-layers-shared errors.
7. Heterogeneous width: global slots use `global_head_dim`, sliding use `head_dim`, verified via view lengths.
Plus zero-layer config, global-capacity-exceeded, and out-of-bounds accessor error paths.

Mutation check performed: breaking donor-type selection (forcing all shared layers onto the sliding donor) failed 8 tests of the suite as it stood at that point (11 tests); restored and re-verified green. The final committed suite is 15 tests, all green.

### Review round 1 (addressed in full)

An adversarial review correctly refuted the initial "read view == window" model for prefill — the reference publishes full-length states during each update. The second probe above was run in response and pins the exact rule now implemented. The same round's other findings are also addressed: the build-time slot-count assert now compares the completed slot allocation against an independently derived owner list (non-tautological), `layer_types` length and all cache-relevant dimensions are validated fail-closed before any allocation with `checked_mul` on every size product, and every sliding-window test asserts both K and V with distinct data. Mutation checks: breaking donor-type selection fails 8 tests; changing retention from `window − 1` to `window` fails 2 tests.

### Verification

- `cargo test -p lattice-inference --lib model::gemma4_cache`: 11 passed, 0 failed.
- `make bench-compare` (origin/main vs HEAD, default groups): no significant regressions; all changes within noise (the only p < 0.05 entries are −2.6% and −2.0%, i.e. faster). Expected — the change is a new module not reachable from any bench path.

### Next stage

Stage 5 wires this cache into the Gemma 4 forward pass (attention consumption + chunked prefill), gated by an HF first-token golden at prompt lengths {short, > sliding_window, chunk-boundary ± 1} on the full 35-layer stack, plus a swapped-donor negative test.

### DIFF
--- a/crates/inference/src/model/gemma4_cache.rs
+++ b/crates/inference/src/model/gemma4_cache.rs
@@ -9,10 +9,18 @@
 //! a build-time indirection map, so KV memory scales with the non-shared
 //! layer count rather than the full layer count.
 //!
-//! Donor slots are ordinary slots, sized per their own attention type:
-//! sliding-type slots retain only the last `sliding_window` positions
-//! (window-truncated, matching the reference `Cache` a sliding donor reads
-//! through); global-type slots retain the full context. Shared layers read
+//! Donor slots are ordinary slots, sized per their own attention type.
+//! Global-type slots retain the full context. Sliding-type slots implement
+//! the *publish* rule that a differential probe against the pinned reference
+//! established (a naive "read view == window" model does not match the
+//! reference during prefill): on each append, the slot first compacts its
+//! existing content down to the last `sliding_window - 1` positions, then
+//! appends the entire incoming chunk; the read view is the whole resulting
+//! buffer. Truncation to `sliding_window - 1` therefore happens lazily, at
+//! the *next* append, not at read time -- a prefill chunk publishes its full
+//! length, and only a subsequent append trims the prior tail. Repeated
+//! single-token decode steps converge on exactly `sliding_window` published
+//! positions (`sliding_window - 1` retained + 1 new). Shared layers read
 //! their donor's slot directly -- there is no side buffer and no copy, and
 //! the public API has no writer for a shared layer's "own" state, since it
 //! has none.
@@ -28,7 +36,8 @@ use crate::model::gemma4_config::Gemma4Config;
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum SlotKind {
-    /// Window-truncated retention: only the last `capacity` positions survive.
+    /// Lazy window-minus-one retention published in full per chunk -- see
+    /// the module docs for the exact rule.
     Sliding,
     /// Full-context retention up to `capacity` positions.
     Global,
@@ -39,9 +48,14 @@ struct Slot {
     kind: SlotKind,
     /// Per-position width: `num_key_value_heads * attn_head_dim(owner_layer)`.
     kv_dim: usize,
-    /// Sliding: `sliding_window`. Global: the cache's `max_seq_len`.
+    /// Sliding: `sliding_window` (the retention bound used in the append
+    /// rule, not a hard cap on buffer size). Global: the cache's
+    /// `max_seq_len` (a hard cap -- `append` rejects an overflow).
     capacity: usize,
-    /// `[capacity * kv_dim]`; valid data occupies `[0, seq_len * kv_dim)`.
+    /// Global: fixed size `[capacity * kv_dim]`. Sliding: grows on demand to
+    /// hold the retained tail plus the current chunk (transient, per-chunk
+    /// -- see [`Slot::append`]). Valid data always occupies
+    /// `[0, seq_len * kv_dim)`.
     k: Vec<f32>,
     v: Vec<f32>,
     seq_len: usize,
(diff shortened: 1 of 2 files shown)


oceanwaves630 (approved):
Independent full-diff read at 095ebfcfa9: donor mapping is total or construction fails (missing donors return an error, never a default slot), heterogeneous per-layer allocation is derived from the owner layer's dims in both preflight and allocation, and sliding retention pins the window-minus-one-plus-current-chunk semantics against exact K/V contents, matching the reference cache behavior. Scope is the new cache module plus wire-up only, no CI changes. One stale line in the body: the Verification section says 11 tests, the committed suite runs 15 in CI — an underclaim, fix at leisure. Approved.

Issue #1003, opened 2026-07-15
vision: no committed golden image fixture — S5 parity gate needs one + regenerated goldens + documented prompt-template contract


ohdearquant:
## Gap

The repo has no committed golden image fixture for the vision pipeline. `git ls-files | grep -i golden_image` and a sweep of `tests/fixtures` both come back empty — no image fixture was ever committed. The ADR-069 Stage 5 differential probes had to substitute a locally generated synthetic image (64x64 RGB checkerboard, deterministic seed), which means the greedy-token goldens captured by those probes are specific to that uncommitted image and cannot be the parity gate's target values.

## What the S5 parity gate needs

1. **A committed golden image fixture** at `crates/inference/tests/fixtures/vision/golden_image.png` (or equivalent): small (a few KB), deterministic content, checked into the repo so every CI run and every future regeneration reads the same bytes. A synthetic deterministic image (e.g. seeded checkerboard) is fine — it only needs to be stable and produce a discriminative description.
2. **Goldens regenerated against that committed fixture** using the pinned reference (`transformers==5.12.1`, CPU, float32, greedy): input_ids, image_grid_thw, position-id table, rope_delta, and the first-N greedy token ids. These become the gate's committed expected values.
3. **The chat-template contract documented in the prompt fixture itself.** The reference chat template emits an already-closed `<think>\n\n</think>\n\n` block before the assistant turn starts — generation begins directly on the answer. The gate's prompt construction must reproduce this exact pre-closed think preamble (not a bare assistant header), or `input_ids` mismatch before any vision logic runs. This fact must live as a comment next to the committed prompt fixture, because it is exactly the kind of invisible contract a future regeneration silently breaks.

## Why this is filed separately

The position-id / lane-selection / injection numeric contracts are fixture-independent and already implemented + unit-tested against reference-probed values. The end-to-end parity gate, however, is only as good as its fixture provenance — filing this on its own makes the gate design reviewable independently of the M-RoPE/injection implementation PRs.

## Notes for the implementer

- Verified against the reference: a 64x64 input image is processor-upscaled to a 16x16 unmerged patch grid (min-pixels floor), i.e. 64 image-pad tokens after 2x2 merging — do not assume `pixels / patch_size` gives the grid.
- MTP/speculative decoding must be disabled (or verified inert) in both the reference run and the lattice gate run; the reference `generate()` leaves the checkpoint's MTP layer untouched by default, and the gate should pin that explicitly.

[...]

PR #1004, opened 2026-07-15
feat(inference): M-RoPE position + cos/sin table builder for Qwen3.5 vision (ADR-069 S5a)

ohdearquant:
## What

ADR-069 Stage 5a: pure-Rust M-RoPE position-id and cos/sin table builder for Qwen3.5 vision-language decoding. New module `crates/inference/src/vision/qwen35_mrope.rs` (additive-only; one `mod` line in `vision/mod.rs`). No decoder wiring in this stage — embedding substitution and forward-pass integration are Stage 5b.

- `build_position_ids(input_ids, image_token_id, grids, spatial_merge_size)` → per-physical-token `(t, h, w)` triples + `rope_delta`. Text runs advance all axes together; each image run consumes its grid, sweeps the merged `(T, H/m, W/m)` grid row-major from the current position, and advances the shared counter by `max(H, W)/m` (not by pad count). Fail-closed on: zero merge size, grid not divisible by `m`, image-run length ≠ `T*H*W/m²`, missing or leftover grids.
- `build_cos_sin(positions, head_dim, partial_rotary_factor, theta, mrope_section)` → per-token `[rope_half]` cos/sin rows using the interleaved axis-selection schedule (lane `i` ← axis `i % 3`; for section `[11,11,10]` over 32 lanes: T at 0,3..30, H at 1,4..31, W at 2,5..29). Fail-closed on: non-3-entry section, sum ≠ rope_half, and any section split whose counts don't match the cyclic schedule (only cyclic-matching splits are reference-verified; anything else would silently rotate lanes with wrong axes).
- `decode_position(physical_cache_len, rope_delta)` → the all-axes-equal decode coordinate; errors on negative results.

## Verification

Every numeric contract was differentially verified against a pinned HF `transformers==5.12.1` CPU/float32 reference run before implementation, and the tests reproduce those reference-captured values:

- Worked text+image+text toy table ((1,4,4) grid, m=2): every triple asserted, trailing text resumes at 5.
- Reference-probed 82-token sequence (16x16 unmerged grid): spot rows at physical indices 0-5, 12, 67, 68-71; `rope_delta == -56`; decode coordinate `82 + (-56) == 26` (matches the reference's own decode-branch position computation).
- Lane schedule: all 32 lane→axis assignments asserted by driving `build_cos_sin` with theta=1 and distinct axis values (mutation-checked: a blocked-schedule mutant fails this test).
- cos/sin numerics at position (4,4,4), theta=1e7: cos[0..3] and sin[0] match reference-captured values within 1e-4.
- Text-only reduction: positions collapse to `t==h==w==index`, `rope_delta==0`, tables equal the plain 1-D partial-RoPE values.
- Fail-closed negatives: 7 rejection tests.

13 tests, all green: `cargo test -p lattice-inference --lib vision::qwen35_mrope`. `cargo clippy --workspace -- -D warnings` clean. Mutation check performed (blocked lane schedule → `lane_schedule_matches_section_counts` fails; restored → green).

## Bench

No bench-compare run: the module is structurally unreachable from any runtime path in this PR (nothing calls it; the only non-test change is the `mod` declaration), so no existing perf surface changes. Stage 5b (decoder wiring) will carry the A/B.

## Related

- ADR-069 (vision encoder recalibration), Stage 5 ladder.
- #1003 tracks the committed golden-image fixture the eventual S5 parity gate needs; this PR deliberately contains no end-to-end gate.

### DIFF
--- a/crates/inference/src/vision/mod.rs
+++ b/crates/inference/src/vision/mod.rs
@@ -80,6 +80,7 @@ pub mod merger;
 pub mod multimodal;
 pub mod preprocess;
 pub mod qwen35_merger;
+pub mod qwen35_mrope;
 pub mod qwen35_vit;
 pub mod qwen35_vit_metal;
 pub mod vit;
--- a/crates/inference/src/vision/qwen35_mrope.rs
+++ b/crates/inference/src/vision/qwen35_mrope.rs
@@ -54,8 +54,12 @@ pub struct MRopeTables {
 /// image run consumes the next entry in `grids`, starts all axes at the
 /// current position, sweeps the merged `(T, H/m, W/m)` grid row-major, and
 /// advances the shared position counter by `max(H, W) / m` afterward — not
-/// by the number of image-pad tokens consumed. See RECON sec. 2
-/// "Position-id construction".
+/// by the number of image-pad tokens consumed.
+///
+/// Fails closed (`InvalidInput`) on: empty `input_ids`, zero merge size,
+/// zero-dimension grids, grids not divisible by the merge size, image runs
+/// whose length does not match their grid, missing or leftover grids, and
+/// any arithmetic that would overflow the position space.
 pub fn build_position_ids(
     input_ids: &[u32],
     image_token_id: u32,
@@ -67,7 +71,23 @@ pub fn build_position_ids(
             "spatial_merge_size must be > 0".to_string(),
         ));
     }
+    if input_ids.is_empty() {
+        return Err(InferenceError::InvalidInput(
+            "input_ids must not be empty".to_string(),
+        ));
+    }
     let m = spatial_merge_size;
+
+    fn advance(pos: u32, by: usize, what: &str) -> Result<u32, InferenceError> {
+        u32::try_from(by)
+            .ok()
+            .and_then(|v| pos.checked_add(v))
+            .ok_or_else(|| {
+                InferenceError::InvalidInput(format!(
+                    "position overflow advancing {what} by {by} from {pos}"
+                ))
+            })
+    }
 
     let mut positions = Vec::with_capacity(input_ids.len());
     let mut current_pos: u32 = 0;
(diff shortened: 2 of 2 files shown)


oceanwaves630 (approved):
Independent full-diff read at head de21072897. Approving — the new M-RoPE builder module is clean, fail-closed, and differentially verified against reference goldens.

Scope: one new module (`crates/inference/src/vision/qwen35_mrope.rs`, +657) plus its one-line `mod.rs` registration. No decoder wiring (deferred to a later stage as documented). Nothing else touched.

Fail-closed surface: all three public functions (`build_position_ids`, `build_cos_sin`, `decode_position`) return `Result<_, InferenceError>` and cannot panic on caller input. Every failure mode returns `InvalidInput` — zero spatial-merge size, empty input, zero-dimension grids, non-divisible grids, arithmetic overflow (`checked_mul`/`checked_add`/`try_from` throughout), non-finite or non-positive theta, fractional rope dimension, section-sum overflow, cyclic-schedule mismatch, and negative/out-of-range decode positions. The image-run slice is bounds-guarded by a short-circuit check before the slice is ever evaluated. No `unwrap`/`expect`/`panic` in library code; the axis `match` is exhaustive.

Correctness: I hand-traced the toy example — text tokens advance all three axes by one; an image run sweeps the merged (t,h,w) grid row-major from a common start position, then advances by `max(h,w)`. This reproduces the expected position table and the 82-token probe golden (`rope_delta = -56`, decode → 26). The cos/sin tables are pinned to reference-probe numeric goldens.

The six fail-closed findings from the first review round are each resolved with a named regression test (zero-dimension-grid infinite-loop guard, overflow arithmetic, fractional rope dimension, non-finite theta, section-sum overflow, decode out-of-range) — about twenty tests in total.

The current head is a merge of an orthogonal change into main; this module file is untouched by that merge, so the prior automated APPROVE covers the current head. Green CI (32 checks). Approving.

PR #1005, opened 2026-07-15
feat(inference): Qwen3.5 vision end-to-end on CPU path (ADR-069 S5b)

ohdearquant:
## Summary

Qwen3.5 vision reads an image **end-to-end** on the CPU forward path. This is ADR-069 Stage 5b: the decoder splice that consumes the Stage-5a M-RoPE builder (merged in #1004) and the Stage-5b request type, so a Qwen3.5-0.8B multimodal prompt produces the correct greedy continuation from an image.

CPU-first by design: the committed vision golden fixture was generated by HF `transformers` on CPU / float32 / 1-thread / greedy, and `cpu_f16.rs` is a complete hybrid GDN+GQA forward, so the end-to-end greedy parity gate is **deterministic and runnable without a GPU**. The Metal port is a separate follow-up with this CPU path as its numeric oracle.

## What changed

- **`forward_step_f16`** — gains `injected_embedding: Option<&[f32]>` (fail-closed masked **REPLACE** of the token-embedding lookup at an `<|image_pad|>` slot — HF's `masked_scatter` contract; rejects wrong length or non-finite rows) and threads `mrope_cos_sin` through to the GQA branch. Every existing caller passes `(None, None)`; the `None` paths are byte-for-byte the prior loop bodies, so **text-only decode is unaffected** (proven bit-identical by a dedicated test).
- **`full_attention_step_f16`** — accepts an optional per-token cos/sin row; when supplied it replaces the 1-D RoPE-table lookup in the stride-half rotation for both Q and K. GDN layers take no RoPE and are untouched.
- **`generate_multimodal_f16`** (new) — validate-first driver mirroring `generate_f16`; injects post-merger visual rows at image-pad slots in image order during prefill, feeds each prefill token its M-RoPE row, and continues decode using `physical_cache_len + rope_delta` for the M-RoPE coordinate while the KV-cache index stays physical/contiguous.
- **`multimodal.rs`** — extracts `resolve_mrope_params` and adds `build_decode_cos_sin` (single-position sibling of `build_mrope_tables`), both calling the same `build_cos_sin` builder.
- **`vision_s5b_e2e_gate_test.rs`** (new) — the end-to-end greedy parity gate.

## Verification (all run this session)

- **End-to-end gate (the milestone):** against the real `Qwen/Qwen3.5-0.8B` checkpoint, `generate_multimodal_f16` reproduces the committed HF golden greedy continuation `[1919, 2099, 369]`. It runs in CI (the `vision-s3-gate` job downloads the checkpoint and runs it alongside S3/S4, same env vars; `LATTICE_VISION_S3_GATE_ENFORCE=1` fails closed instead of skipping). The assertion is **machine-portable, not exact-3/3**: lattice's macOS f16 forward runs every projection through Accelerate `cblas_sgemm` (AMX), whose f32 accumulation order is not bit-identical across Apple-Silicon generations. Position 0 (the golden first token, ~4.1-logit margin over its runner-up) is asserted **exactly** — reproducing it exercises the whole vision→decoder pipeline end to end. Position 1 is a **measured ~0.07-logit near-tie** between the golden token and the next one (third contender ~6.7 logits away), so the gate accepts either near-tie candidate and rejects anything else; position 2 is asserted exactly when position 1 kept the golden token. On the reference machine all three match exactly. The per-position margins and rationale are documented in the test.
- **Unit tests (no checkpoint, CI-runnable):** injection replace-not-add + mutation-sensitive (mutating the supplied row changes output; mutating an unrelated vocab row does not), fail-closed on wrong length / NaN / Inf, M-RoPE cos/sin applied in the wired forward matches the builder independently, text-only `generate_multimodal_f16` bit-identical to a plain `forward_step_f16` reference, invalid-request and context-overflow rejection.
- `cargo test -p lattice-inference --lib --features f16`: all green; the new splice tests pass.
- `cargo clippy --workspace --tests --features f16 -- -D warnings`: clean.
- **Mutation check:** no-op'ing the injection fails the injection test; restoring passes it (worktree clean after).

## Performance

`forward_step_f16` (decode-hot) gains two `Option` params that are `None` on every existing path, and the `None` code paths are byte-identical to the prior loop bodies — so existing decode/prefill output is bit-identical (proven by the text-only non-regression test).

`make bench-compare` (origin/main `9a5ef28a7e` vs HEAD `276dfe3df4`, CPU elementwise groups, under the machine-wide heavy-lane lock): **no measurable change**. All groups within ±3.7% worst-case, mid-point deltas within ±1.7% with mixed sign — CPU microbench noise, no systematic regression. (These SIMD elementwise ops don't exercise the modified forward path; the forward-path change is two always-`None` branches, so no throughput effect is expected or observed.)

## Fail-closed hardening (commit `7fda2b1c`)

Follow-up review hardened the public multimodal entry point and closed a CI gap:

1. **Out-of-vocab token IDs** — `generate_multimodal_f16` and its `generate_f16` sibling now reject any `input_ids`/`prompt_ids` `>= cfg.vocab_size` with `InvalidInput` before any decoder allocation, instead of panicking in the embedding slice.
2. **E2E gate wired into CI** — `vision_s5b_e2e_gate_test` now runs in the `vision-s3-gate` job (same checkpoint/env as S3/S4), a `grep -q 'LATTICE_VISION_S5B_GREEDY_TOKENS'` assertion fails the step if it silently skips, and the test file is in the changed-files trigger regex — so the gate can't pass green without actually running.
3. **Request↔checkpoint binding** — the driver now rejects a request whose `image_token_id` / `spatial_merge_size` / `decoder_hidden_size` / `out_hidden_size` disagree with the loaded checkpoint (guards against silently producing wrong logits from an internally-consistent-but-mismatched request).
4. **M-RoPE row width** — both the prefill table and the decode-time row are checked against `cfg.rope_dim()/2` before attention indexing, `InvalidInput` on mismatch (prevents an index panic from a divergent `partial_rotary_factor`).

Six new mutation-sensitive regression tests cover these. After the fixes: cpu_f16::tests 25/25, vision 109/109, clippy clean, and **the e2e gate still matches the golden** (`got=[1919, 2099, 369]`). The fixes are up-front, once-per-call validation guards, not on the per-token decode path, so the bench result above (decode-hot path) is unchanged.

## Making the e2e gate actually run in CI

The wired gate first failed in CI while passing locally. Two distinct causes, both fixed here:

**1. Checkpoint layout (the CI blocker).** The gate opened `model_dir/model.safetensors` directly. A local checkout exposes that name (often a convenience symlink), but a raw HuggingFace snapshot — what CI materializes via `snapshot_download` — has only the sharded `model.safetensors-00001-of-00001.safetensors` plus `model.safetensors.index.json` and no plain `model.safetensors`, so the gate panicked at open on CI. A new `resolve_safetensors_path` helper tries `model.safetensors` then falls back to the single shard named in the index weight_map (this 0.8B checkpoint is one shard), mirroring `Qwen35Model::from_safetensors`'s plain-then-index precedence. Verified against a real HF snapshot dir (sharded + index, no plain file, shard is a symlink-to-blob — the exact CI layout): `got=[1919, 2099, 369]`. The S3/S4 gates were unaffected because they already load via a directory-aware path.

**2. Greedy portability (a latent flake, fixed pre-emptively).** The checkpoint is byte-identical on both machines (sha256 of the shard, index, and config all match the fixture manifest), so once it loads the forward is comparing the same weights. But instrumenting the per-step logits showed position 1 is a **~0.07-logit near-tie** (`22.377` vs `22.306`) with the third contender ~6.7 logits below. lattice's macOS f16 forward runs every projection through Accelerate `cblas_sgemm` (AMX), whose f32 accumulation order is not bit-identical across Apple-Silicon generations, so ~0.1-logit jitter could swap the two near-tied tokens on a different runner. This is gate portability, not a correctness bug — lattice matches the HF golden exactly on the reference machine at a position where HF's own greedy pick was a coin-flip. The gate now stays regression-sensitive without being brittle:
- Position 0 exact (the ~4.1-logit-margin first token — the full pipeline proof); position 1 accepts either near-tie candidate and rejects anything else (jitter cannot cross the ~6.7-logit gap); position 2 exact when position 1 kept the golden token.
- The per-position logic is a pure `check_greedy_parity` helper with five checkpoint-free unit tests (accepts the reference and the near-tie flip; rejects a position-0, out-of-near-tie position-1, and preserved-history position-2 regression), so the tolerance is itself mutation-tested and runs without the checkpoint.

**CI wiring.** The `vision-s3-gate` step captures cargo output to a file and always prints it while preserving cargo's exit status (if/else), instead of a `> file` redirect that `set -e` aborted before `cat` (which hid the original failure) or a `tee` pipeline that would fail open under GitHub's default `bash -e` (no `pipefail`).

## Scope and follow-ups

Deliberately out of scope, each a separate follow-up:
- **Metal path** (`metal_qwen35.rs` / shaders) — the perf port, with this CPU path as its numeric oracle.
- **`cpu_q8.rs` / `neon_forward.rs`** — independent embedding + RoPE apply for Q8 CPU decode; no Q8 vision-weights loader exists yet.
- **`batch_prefill.rs`** — assumes arithmetic-progression positions, incompatible with M-RoPE's per-token 3-axis divergence.
- **`speculative.rs` (MTP)** — needs its own verification before claiming multimodal support.
- **Teacher-forced multi-position parity** — the current gate hard-asserts position 2 only when position 1 does not flip. A teacher-forced variant (feed the golden continuation, check each step's argmax against the golden with a per-step tie tolerance) would restore deterministic hard coverage at every position across machines; it needs a small public hook to read per-step logits, so it is a separate reviewable change.

## Test plan

```
# Unit (no checkpoint):
cargo test -p lattice-inference --lib --features f16

# End-to-end gate (requires the Qwen3.5-0.8B checkpoint):
LATTICE_VISION_S3_MODEL_DIR=<path-to-qwen3.5-0.8b> LATTICE_VISION_S3_GATE_ENFORCE=1 \
  cargo test --release -p lattice-inference --features f16 --test vision_s5b_e2e_gate_test -- --nocapture
```


oceanwaves630 (approved):
Reviewed the full diff at bd658541355909f50f1332dbd448e87cadfc000d. The fail-closed posture is consistent across both generate paths, the text-only route stays bit-identical to the plain forward, and the portability rationale on the parity check is well argued and well tested. Approving.

PR #1006, opened 2026-07-16
ci: print capability-matrix test logs before exiting on failure

ohdearquant:
## Problem

The capability-matrix step's `lattice_serve` and `lattice` test invocations redirect output to a log file and read `$?` afterward:

```bash
cargo test ... > lattice_test.log 2>&1
lattice_status=$?
cat lattice_test.log
```

The step runs under the workflow default shell `bash -e {0}`. Under `-e`, a failing bare `cargo test` terminates the script at that line — before the `cat` — so a failing run exits 101 with the actual test failure invisible in the CI log. This hid the failing test on a 2026-07-16 main-push run (the job showed only "Process completed with exit code 101" five seconds after the previous suite's all-pass output).

## Fix

Wrap both invocations in `if/else`: a command in an `if` condition is exempt from `-e`, so the log always prints and the original exit status is preserved on failure. This is the same form `e2e-parity.yml` already uses for the vision gate.

Workflow-only change; the `check_test_utils_modules_ran` proof-marker guards are unchanged. Swept the other workflows for the same `> log 2>&1` + `$?` pattern — these two invocations were the only remaining instances.

No bench-compare: no code under `crates/` changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -269,10 +269,16 @@ jobs:
             fi
           }
 
-          cargo test -p lattice-inference --bin lattice_serve --features f16,metal-gpu,test-utils > lattice_serve_test.log 2>&1
-          lattice_serve_status=$?
-          cat lattice_serve_test.log
-          if [ "$lattice_serve_status" -ne 0 ]; then
+          # if/else (not `cmd; status=$?`): this step runs under the default
+          # `bash -e {0}` shell, so a failing bare `cargo test` terminates the
+          # script BEFORE the `cat` — the log with the actual failure is never
+          # printed (this hid the failing test on a 2026-07-16 main-push run).
+          # A command inside an `if` condition is exempt from `-e`.
+          if cargo test -p lattice-inference --bin lattice_serve --features f16,metal-gpu,test-utils > lattice_serve_test.log 2>&1; then
+            cat lattice_serve_test.log
+          else
+            lattice_serve_status=$?
+            cat lattice_serve_test.log
             echo "::error::cargo test --bin lattice_serve failed (exit $lattice_serve_status)"
             exit "$lattice_serve_status"
           fi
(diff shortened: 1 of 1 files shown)


ohdearquant:

## Review of lattice PR #1006 (round 1)

### Verdict: APPROVE

### Prior review context

- No prior reviews or unresolved review threads were present.
- This round reviewed the single workflow-only commit `a2c138b127` against base `248f2e7e52`.

### What this round examined

- The complete changed workflow step and its enclosing CI job, including shell/default overrides.
- Bash `errexit` and `$?` semantics, the proof-marker control flow, and workflow-wide sibling patterns.
- Workflow syntax and static linting.

### Findings (# total: 0 blockers, 0 majors, 0 mediums, 0 minors, 0 nits)

No findings.

1. **Default-shell rationale is correct.** The target `run:` step has no `shell:` or `defaults.run.shell` override in `.github/workflows/ci.yml`; GitHub documents the unspecified Linux/macOS shell as `bash -e {0}` ([GitHub Actions workflow syntax](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax)). Thus, on the macOS leg, a failed bare `cargo test` exits before the following `status=$?` and `cat`. Bash documents the `if`-condition exception to `-e` ([Bash `set` reference](https://www.gnu.org/s/bash/manual/html_node/The-Set-Builtin.html)), so the replacement form deliberately reaches its `else` branch.

2. **The original cargo status is preserved.** In each `else` branch, the assignment is the first command: `.github/workflows/ci.yml:280` and `.github/workflows/ci.yml:297`. Expansion of `$?` therefore reads the failed `cargo test` condition's status before `cat` or `echo` can alter it; `exit "$lattice_serve_status"` / `exit "$lattice_status"` then returns that saved value. A direct shell semantic probe under `bash -e` captured and returned status 73 from an `if` condition.

3. **Proof-marker behavior is unchanged.** On success, each `cat` is followed by its corresponding `check_test_utils_modules_ran` call at `.github/workflows/ci.yml:285` and `.github/workflows/ci.yml:302`. On cargo failure, the `else` branch prints the log and exits before either marker check, matching the previous fail-fast behavior while making the log visible.

4. **Sibling sweep is complete for the stated pattern.** A repository-wide scan of `.github/workflows/` found the two changed `cargo test ... > *.log 2>&1` instances as the only bare-command-plus-`$?` occurrences. The remaining cargo redirect in `e2e-parity.yml:677-683` is already guarded by `if ...; then ...; else status=$?; ...; exit "$status"; fi`. Other CI log captures use a different `if ! cargo test` form.

5. **Workflow validity.** `git diff --check` was clean; Ruby/Psych parsed `.github/workflows/ci.yml`; and `actionlint 1.7.12 .github/workflows/ci.yml` completed with no diagnostics. GitHub's PR actionlint check also passed during this review.

### What I did NOT check

- Per instruction, I did not compile or run Rust tests.
- The cited main-push incident was not independently reproduced: the currently visible 2026-07-16 main-push run's capability-matrix step completed successfully, so it cannot itself demonstrate the historical failure. The shell behavior and the patch's control flow were independently verified.
- PR matrix jobs were still in progress at review time; no claim is made here about their eventual result.

Stop. This workflow-only change has no remaining correctness or validation issue from static review.

Domain utility: MEDIUM — the Lattice review rubric supplied the workflow-policy and evidence standards, while the change itself was shell/control-flow focused.

oceanwaves630 (approved):
Read the 24-line diff. Correct use of the if-condition -e exemption, and status captured at the top of else before cat clobbers it. Log visibility on failure is exactly what today's flake hunt needed. Approving.

PR #1007, opened 2026-07-16
feat(inference): image/text VLM embedding extraction (pooled)

ohdearquant:
## Summary

Adds the CPU prefill-only pooled-embedding extraction path for Qwen3.5-0.8B
(GME-style): run the VLM decoder forward with no decode/sampling, pool the
per-position last hidden states, L2-normalize.

- `crates/inference/src/forward/cpu_f16.rs`:
  - `PoolingStrategy` enum (`MeanVisualTokens`, `LastToken`)
  - `prefill_hidden_states_f16`: a prefill-only sibling of
    `generate_multimodal_f16` that shares its request validation,
    checkpoint-binding checks, and M-RoPE/injection wiring, but returns every
    position's post-final-norm hidden state instead of sampling a token
  - `embed_image_f16` / `embed_text_vlm_f16`: pool + L2-normalize over a
    `Qwen35VisionRequest` (image or text-only) through the *same* decoder +
    pooling path, so image and text embeddings land in the same vector space
- `crates/inference/src/vision/pooled_embed.rs` (new): `embed_image_from_bytes_f16`
  composes the existing preprocess → ViT → merger pipeline with the
  decoder-side pooling above, assembling a minimal (documented,
  non-chat-template) vision-prompt scaffold around raw image bytes + a text
  prompt.

Retrieval quality with the base Qwen3.5-0.8B *instruct* checkpoint is
unvalidated by design — GME-style pooled embeddings normally need a
contrastively fine-tuned checkpoint, which is a separate research decision.
This PR is the extraction *machinery*: correct positions, deterministic,
unit-norm output, documented honestly (see `PoolingStrategy`'s doc comment).

## Tests

- Determinism: same input → identical vector (both pooling strategies, both
  `embed_image_f16` and `embed_text_vlm_f16`)
- Finiteness + unit norm
- Discrimination sanity: two different "images" (different post-merger rows)
  → cosine < 0.999; an image against itself → cosine ≈ 1.0
- Mutation-sensitivity: a direct unit test on the pooling primitive with an
  off-by-one-shifted position window, *plus* a golden-reference test
  (`embed_image_f16_matches_independently_computed_golden_pool`) that
  independently recomputes the expected pooled vector from
  `prefill_hidden_states_f16` using hand-known-correct positions — verified
  by reverse-applying an off-by-one bug into `embed_image_f16`'s position
  selection and confirming only the golden test (not the determinism/
  discrimination tests) catches it
- Vision-pipeline composition tests for `embed_image_from_bytes_f16`
  (determinism, cross-image discrimination, rejects non-VLM checkpoints,
  rejects misaligned images)

38 new tests total, all passing; 2005/2005 lib tests pass; clippy clean
(`-D warnings`); fmt clean.

## Deviations from the task packet

- No `lattice-embed` wrapper: the API depends on `F16ModelWeights` /
  `Qwen35Config`, which are `lattice-inference`-crate types — a thin
  re-export in `lattice-embed` would add indirection with no behavior, so I
  kept the whole surface in `lattice-inference` per the "if the natural home
  is lattice-embed" conditional in the packet.
- The image scaffold `embed_image_from_bytes_f16` assembles is
  `[vision_start] + [image_pad; N] + [vision_end] + tokenized(prompt)`, not
  the full HF chat template (no system role, no `<think></think>` block, no
  BOS) — documented in the module doc comment. Callers needing exact
  chat-template parity should assemble their own `input_ids` and call
  `embed_image_f16` directly.
- `bench-compare` not run: no existing inference/embed hot path was
  modified — only new functions were added.

## Test plan

- [x] `cargo test -p lattice-inference --lib` (2005 passed, 0 failed)
- [x] `cargo clippy -p lattice-inference --lib -- -D warnings` (clean)
- [x] `cargo fmt -p lattice-inference -- --check` (clean)
- [x] `cargo check --workspace` (clean)
- [x] Mutation-sensitivity manually verified (bug injected → golden test
      failed → bug reverted → test passed again)

### DIFF
--- a/crates/inference/src/forward/cpu_f16.rs
+++ b/crates/inference/src/forward/cpu_f16.rs
@@ -1515,6 +1515,18 @@ pub fn prefill_hidden_states_f16(
         }
     }
 
+    // Reject empty and over-context prompts before build_mrope_tables, which
+    // otherwise materializes a position entry plus cos/sin rows per supplied
+    // token — unbounded input must fail cheaply, not after that allocation.
+    let prompt_len = request.input_ids.len();
+    crate::model::qwen35::check_prompt_not_empty(prompt_len)?;
+    let max_context = cfg.max_position_embeddings;
+    if prompt_len > max_context {
+        return Err(crate::error::InferenceError::Inference(format!(
+            "prompt ({prompt_len} tokens) exceeds model context window ({max_context})"
+        )));
+    }
+
     let (_positions, tables) = request.build_mrope_tables(cfg)?;
 
     let expected_rope_half = cfg.rope_dim() / 2;
@@ -1528,15 +1540,6 @@ pub fn prefill_hidden_states_f16(
     }
 
     let prompt_ids = &request.input_ids;
-    let prompt_len = prompt_ids.len();
-    crate::model::qwen35::check_prompt_not_empty(prompt_len)?;
-
-    let max_context = cfg.max_position_embeddings;
-    if prompt_len > max_context {
-        return Err(crate::error::InferenceError::Inference(format!(
-            "prompt ({prompt_len} tokens) exceeds model context window ({max_context})"
-        )));
-    }
 
     let num_linear = cfg.num_linear_attention_layers();
     let num_full = cfg.num_full_attention_layers();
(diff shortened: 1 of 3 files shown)


ohdearquant:
Review round 1 complete. One medium finding: the context-window limit was evaluated after M-RoPE table materialization, allowing large avoidable allocation work on over-context requests. Fixed in 1528bfd7e6 — the limit check now runs ahead of table construction, with a regression test that passes request validation but would fail inside the table builder, proving the ordering by which error is returned (test verified to fail with the ordering reverted). All other reviewed dimensions (pooling positions post-injection, last-token semantics, same-space text/image contract, L2 zero-vector guard, per-run grid fail-closed paths) came back clean. Full lib suite: 2012 passed / 0 failed; clippy -D warnings and fmt clean.

ohdearquant:
bench-compare (quick, base origin/main 1c8c4f969a vs head 1528bfd7e6): no change on all groups, p > 0.05 throughout (largest point estimate -2.99% with CI spanning zero-adjacent bounds). Consistent with expectation — this PR adds new entry points only and does not modify existing hot paths.

oceanwaves630 (approved):
Verified the review finding (context-limit check previously ran after M-RoPE table materialization) is fixed in 1528bfd7e6: the limit is now evaluated immediately after request validation, and the regression test proves the ordering by constructing a request that would fail inside the table builder. Pooling, injection, and normalization paths read clean.

PR #1008, opened 2026-07-16
feat(inference): Gemma 4 text-only CPU forward (ADR-082 stage 5)

ohdearquant:
## Summary

Assembles the already-merged Gemma 4 E2B stages (tokenizer #993, config+loader preflight #996, text math kernels with per-op HF goldens #999, shared-KV cache with donor-slot indirection #1002) into a CPU-only text forward + greedy generate for `google/gemma-4-E2B-it`. Does not reimplement any of those stages.

- 35-layer decoder: 28 `sliding_attention` (window 512, 8x256 Q / 1x256 KV) + 7 `full_attention` at indices 4, 9, 14, 19, 24, 29, 34 (8x512 Q / 1x512 KV). Local+global dual RoPE (local: standard theta=10000, full head_dim; global: proportional theta=1000000, only 25% of the 512-dim head rotated). Attention uses **no** `1/sqrt(head_dim)` score scaling — `Gemma4TextAttention.scaling = 1.0` unconditionally in the pinned reference, verified directly against source rather than assumed.
- KV sharing (Amendment 1): the final 20 layers (15-34) read their donor's K/V through `Gemma4KvCache`'s donor-slot indirection rather than running their own `k_proj`/`v_proj`; the loader tolerate-skips their checkpoint-carried k/v tensors.
- `use_double_wide_mlp` on the shared set (FFN 12288 vs 6144), tied embeddings (vocab 262144), Per-Layer Embeddings (PLE) gated into every decoder layer, per-layer learned `layer_scalar`.
- Single-token `forward_step` (mirrors `qwen35::forward::forward_step`), used for both prefill and decode.

## Gates (differential-test-first)

1. **Checkpoint**: downloaded `google/gemma-4-E2B-it` to `~/.lattice/models/gemma-4-e2b-it`; tensor names validated against the committed stage-0 manifest fixture.
2. **HF golden, generated before writing the forward pass**: `scripts/gemma4_stage5_e2e_golden.py`, run via `uv run --with "transformers @ git+...@ab1771c9e..."` (pinned reference commit), CPU f32, 1 thread, greedy. Produces the committed fixture `crates/inference/tests/fixtures/gemma4/stage5/e2e_golden.json` (input_ids, first-3 greedy tokens, per-layer hidden states at {0,4,15,34}, top-8 final logits).
3. **e2e gate**: `cargo test --release -p lattice-inference --features f16 --test gemma4_e2e_forward_test` — lattice CPU greedy first-3 tokens == HF golden, **exact match** (CPU-vs-CPU f32).
4. **Per-layer probe** at layers {0, 4, 15, 34}: max-abs-diff vs HF ~2e-6 to ~1.3e-5, against a 1e-3 gate tolerance. Captured via a forward hook directly on each decoder layer, not `output_hidden_states=True`'s tuple — that tuple's last entry is `last_hidden_state` (post-final-norm), not the raw last-decoder-layer output; this was a golden-generation bug caught and fixed mid-session (documented in the script's module docs), not a lattice bug — layers 0-33 matched under the wrong assumption purely by coincidence of indexing.
5. **Mutation-sensitivity**: `donor_mutation_tests::wrong_donor_mapping_diverges_layer34_probe_and_greedy_tokens` forces layer 34's shared-KV donor slot to the wrong same-type owner (14 → 9, both valid global-attention slots, so the swap doesn't trip a type-mismatch guard) via a `#[cfg(test)]`-only `Gemma4KvCache::override_layer_slot_for_test` hook. Result: layer 34 hidden-state max-abs-diff = **0.413**, ~400x over the e2e gate's own 1e-3 tolerance — proving the donor mapping is load-bearing. (Greedy top-1 token happens to survive this particular perturbation due to a wide ~5-logit-point margin at this prompt; the per-layer probe assertion is the gate this mutation is proven against, and it is the same assertion the real e2e gate makes.)

## Validation

- `cargo clippy --workspace --features lattice-inference/f16 --tests -- -D warnings`: clean.
- `cargo fmt --check`: clean.
- `cargo test --release -p lattice-inference --features f16 --lib gemma4`: 68 passed.
- `cargo test --release -p lattice-inference --features f16 --test gemma4_e2e_forward_test`: 2 passed.
- `cargo check --workspace` (default features): clean.

## Deviations

- No `make bench-compare`: this is a new forward path with no prior lattice baseline to compare against (same precedent as stage 3's `gemma4_ops.rs` per-op goldens — correctness-first landing, perf follow-up once there's something to compare).
- The mutation-sensitivity test lives in-crate (`#[cfg(test)] mod donor_mutation_tests` inside `gemma4_model.rs`), not the external `tests/` integration crate, since it needs `pub(crate)` access to `forward_step` and a test-only cache-internals override hook.
- CPU-only, no Metal wiring — out of scope for this stage per ADR-082 (Metal gates on CPU parity landing first) and to avoid any overlap with the concurrent Metal-path leg.

## Test plan

- [x] `cargo test --release -p lattice-inference --features f16 --lib gemma4`
- [x] `cargo test --release -p lattice-inference --features f16 --test gemma4_e2e_forward_test -- --nocapture`
- [x] `cargo clippy --workspace -- -D warnings`
- [x] `cargo fmt --check`

## CI gating disposition (e2e gate)

The stage-5 e2e gate (`gemma4_e2e_forward_test`) requires the 9.6 GB `google/gemma-4-E2B-it` checkpoint. Review round 2 asked for a required per-PR CI job that downloads it and runs the gate enforced. Decision, with rationale:

- **Per-PR CI gating is deferred, deliberately.** GitHub's per-repo Actions cache cap is 10 GB total; caching this checkpoint is mutually exclusive with the existing ~2 GB Qwen3.5-0.8B cache that three current gate jobs share, so every inference-touching PR would re-download 9.6 GB on `macos-latest` with real job-timeout risk. Sizing a large-checkpoint runner/cache strategy (including a persistent-storage runner) is being decided as part of a broader CI overhaul rather than copy-pasting the small-checkpoint job pattern.
- **Interim coverage:** the gate is fail-closed by default (a missing checkpoint or a `not(f16)` build panics; `LATTICE_GEMMA4_GATE_SKIP=1` is the only skip and prints a `LATTICE_GEMMA4_E2E_SKIPPED` marker, so no run can silently pass without executing), and a scheduled nightly run executes the enforced gate against the real checkpoint on persistent hardware, alerting on any failure. Detection latency for a regression on this path is therefore bounded by ~24 h while the path is under active development, instead of unbounded-silent.
- The general-purpose workspace test run in `scripts/ci.sh` opts into the loud skip explicitly; no CI invocation can execute this test half-armed.

## Hot-path allocation removal — A/B evidence (3e423b4c3e)

**Methodology first:** `main` has no Gemma 4 code, so `make bench-compare` against `origin/main` cannot exercise this path; the applicable measurement is an A/B of this branch's enforced stage-5 e2e gate binary (4 tests, two of them full real-checkpoint forwards: 8-token greedy decode + 511/512/513-length boundary decodes) built per side into isolated target dirs, run with `--test-threads=1`, back-to-back under an exclusive machine-wide heavy-job lock. Wall time via `/usr/bin/time -p`, n=1 per condition. Scripts and raw timing files are committed to the working records for replay.

| condition | wall time |
|---|---|
| fix, first run (cold page cache) | 480.3 s |
| base `4af6cae116`, second run (warm) | 437.6 s |
| fix, re-run (warm) | **419.5 s** |

**Disclosed confounds:** run order — the first run pays page-cache warming for the 9.6 GB checkpoint, which is why fix-cold reads 10% *slower* than base-warm; the warm-vs-warm comparison (419.5 vs 437.6 s, −4.1%) is the like-for-like one. Shared workstation (other compile jobs outside the lock), single run per condition: this is directional evidence sized to an allocation-removal change, not a tight CI. Known but unused: multi-run medians (each run costs ~7–8 min of exclusive machine time; the change is structurally allocation-strictly-less, so direction is the claim, magnitude is not).

**Numerical inertness:** the real-checkpoint donor-mutation probe reports the identical layer-34 max-abs-diff (`0.41343307`) before and after, and the enforced e2e gate passes 4/4 on both sides. The in-place RoPE rewrite computes `x1·cos − x2·sin` / `x2·cos + x1·sin` per disjoint pair from pre-update values; sign-flip and `x+(−y)→x−y` are exact in IEEE 754, so it is bit-identical to the two-pass rotate-half form it replaces.


ohdearquant:
bench-compare (quick, base origin/main 1c8c4f969a vs head 1455958f78): no significant change on any group, p > 0.05 throughout. Expected — this PR is additions-only (new Gemma 4 modules and tests); existing Qwen hot paths are untouched. Independent verification of the e2e gate on this machine: cargo test --release -p lattice-inference --features f16 --test gemma4_e2e_forward_test → 2 passed (greedy first-3 tokens match the pinned HF golden exactly, 95.7s).

oceanwaves630 (approved):

PR #1009, opened 2026-07-16
feat(inference): Metal visual-embedding injection for Qwen3.5 vision (MP2)

ohdearquant:
## Summary

Adds `generate_multimodal_vision()`, a new Metal entry point for running Qwen3.5 vision requests: post-merger visual embeddings are injected directly into the residual stream during prefill, at the positions marked by the image-pad token.

This is a staged port. The CPU f16 path (`generate_multimodal_f16`) is the numeric oracle this mirrors. Only the injection half is ported here; M-RoPE position handling for image content is a separate follow-up.

## Injection contract

`forward_step_inner_impl` gains an `injected_embedding: Option<&[f32]>` parameter:

- `Some(row)`: the row is copied directly into the hidden-state buffer, **replacing** the token-embedding lookup for that step. This mirrors the CPU oracle's `masked_scatter` semantics (replace, not add).
- `None`: unchanged — the existing `embed_tokens` lookup by `token_id` runs exactly as before. The pre-existing `generate()` / `generate_streaming()` paths are untouched (they always pass `None`).

A new private helper, `forward_step_injected`, wraps a single injected forward step for use during prefill at image-pad positions.

## Verification

Three gate tests, all passing fresh on real Metal hardware:

- `forward_step_injected_logits_differ_by_row_content` — isolates the injection mechanism at a single forward step (bypassing the decode loop) and asserts a negated visual row produces a materially different logit vector. This gives a clean, undiluted signal for injection sensitivity.
- `generate_multimodal_vision_injection_is_mutation_sensitive` — end-to-end: negating all supplied visual rows for a request changes the greedy-decoded token sequence.
- `injected_embedding_never_falls_back_to_token_lookup` — a structural test. It scans the source text of the `Some(row)` arm and asserts it never references `embed_tokens` or `token_id`, proving there is no residual code path back to the token-lookup fallback for an injected position.

### Why a structural test for REPLACE semantics

The checkpoint this targets ties word embeddings: `embed_tokens` also serves as the output (lm_head) weight matrix in the final projection's GEMV dispatch. That means mutating *any* row of `embed_tokens` shifts the corresponding logit component on every decode step, regardless of whether that row is ever read during input-embedding lookup. A behavioral check of the shape "mutate an unrelated embedding row, expect output unchanged" is therefore unsound on this checkpoint — it will fail even when the injection path is implemented correctly, because the mutation reaches the output through the tied lm_head weights rather than through the input embedding path at all.

`injected_embedding_never_falls_back_to_token_lookup` sidesteps that confound entirely by proving the REPLACE contract from the source text instead of from a runtime signal. Mutation-sensitivity was verified locally by temporarily reintroducing a fallback `embed_tokens[token_id]` read inside the `Some(row)` arm and confirming the test fails, then removing it and confirming the test passes again.

The other two behavioral gates were also checked for mutation-sensitivity: the single-step isolation test's fixture uses non-constant per-lane values (a constant fill is invariant under this model's `(1+gamma) * x / rms(x)` normalization, since a uniform scalar shift cancels in the ratio), and the end-to-end test mutates *all* supplied visual rows together rather than a single row, since a lone-row mutation in a short prompt can be diluted away by attention over the other unmutated image-pad positions.

## Out of scope

- M-RoPE position handling for image content (separate follow-up).
- The pre-existing `generate_multimodal` entry point (flat-prefix `MultimodalInput` representation) is untouched; this PR adds a new entry point (`generate_multimodal_vision`, taking the richer `Qwen35VisionRequest` type) rather than modifying it, since `MultimodalInput` cannot represent the per-image grid shape the request validation needs.

## bench-compare (ADR-058)

Metal decode A/B, slope method (`bench_decode_ab`, N1=8 / N2=64, `BENCH_RUNS=5`, median; real Qwen3.5-0.8B checkpoint). Same-worktree single-file source-swap: PR head vs the `origin/main` blob of `metal_qwen35.rs` (the only file this PR touches), both sides measured under the machine-exclusive Metal GPU lock and bench window.

| side | T(8) | T(64) | decode tok/s |
|------|------|-------|--------------|
| PR head | 110.7 ms | 487.6 ms | 148.6 |
| main | 111.5 ms | 477.1 ms | 153.2 |

**Verdict: no measurable change.** The apparent −3% is inside the noise floor: an A/A re-run of the *same* main-side binary under the same machine load produced 153.2 → 121.8 tok/s (±20% swing; an unrelated release build was compiling concurrently). Structurally, the `None` path adds exactly one `match` at the pre-layer embedding selection, once per token — no per-layer branch, no allocation, no dispatch change (independently confirmed in review). A quiet-machine re-run can be provided on request.

### DIFF
--- a/.github/workflows/e2e-parity.yml
+++ b/.github/workflows/e2e-parity.yml
@@ -690,6 +690,40 @@ jobs:
           grep -q 'LATTICE_VISION_S4_E2E_MUTATION_COSINE' vision-s3-gate.log
           grep -q 'LATTICE_VISION_S5B_GREEDY_TOKENS' vision-s3-gate.log
 
+      # MP2 Metal injection gate: `forward_step_injected_logits_differ_by_row_content`,
+      # `generate_multimodal_vision_injection_is_mutation_sensitive`,
+      # `injected_embedding_logits_are_independent_of_token_id`, and
+      # `injected_embedding_never_falls_back_to_token_lookup`
+      # (crates/inference/src/forward/metal_qwen35.rs) previously ran nowhere in
+      # CI: they early-returned on a missing Metal device or checkpoint instead
+      # of failing, and no workflow selected them (the macOS `feature-matrix`
+      # job's `--lib q4` filter does not match their names). This step reuses
+      # this job's already-downloaded checkpoint (no second ~2 min
+      # checkout+download job) and its macos-latest runner, which exposes a
+      # PARAVIRTUAL Metal device — sufficient here because none of these four
+      # tests touch the Apple7-gated tiled-GEMM path (decode-only GEMV/element-wise
+      # dispatch), the same reasoning the f16-KV-cache Metal gate in ci.yml
+      # documents. LATTICE_METAL_TEST_ENFORCE=1 makes a missing device or
+      # checkpoint a hard failure instead of the default dev-run skip; the
+      # `--lib inject` filter also matches two unrelated pre-existing tests
+      # (a QuaRot seed round-trip test and the CPU f16 injection oracle), so
+      # the exact-count grep below is 6, not 4 -- update it in the same PR
+      # if the `inject`-matching test surface changes.
+      - name: Run MP2 Metal injection gate (fail-closed)
+        env:
+          LATTICE_METAL_TEST_ENFORCE: '1'
+        run: |
+          if cargo test --release -p lattice-inference --features f16,metal-gpu \
+            --lib inject -- --nocapture --test-threads=1 > vision-mp2-inject-gate.log 2>&1
+          then
+            cat vision-mp2-inject-gate.log
+          else
+            status=$?
+            cat vision-mp2-inject-gate.log
+            exit "$status"
+          fi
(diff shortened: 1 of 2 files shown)


ohdearquant:
CI note: the first live run of the new MP2 injection gate failed closed — correctly. The `vision-s3-gate` job provisions Qwen3.5-0.8B via `snapshot_download`, which ships the **sharded** safetensors layout (`model.safetensors-00001-of-00001.safetensors` + `model.safetensors.index.json`), while the gate's checkpoint pre-check accepted only the single-file `model.safetensors` layout. `Qwen35Model::from_safetensors` itself supports both.

Fixed in 3fe22ee691: the pre-check now mirrors the loader's own layout contract (single-file OR index). Verified locally against a sharded-only replica of the CI layout (config + tokenizer + index + single shard, no `model.safetensors`) under `LATTICE_METAL_TEST_ENFORCE=1`: the gated test loads and passes on real Metal (16.5s).

This is the enforce mechanism working as designed: before this PR's fix round, the same provisioning gap would have skip-passed silently.

oceanwaves630 (approved):

PR #1011, opened 2026-07-16
feat(inference): 3-axis M-RoPE in Metal vision path (ADR-069 MP3)

ohdearquant:
## Summary

ADR-069 MP3: 3-axis interleaved M-RoPE in the Metal vision path. `generate_multimodal_vision` (MP2, #1009) previously ran every position through the model's 1-D `engine.rope_cos`/`rope_sin` table, with position handling named as the follow-up port stage. This PR completes it, mirroring the CPU oracle (`cpu_f16::generate_multimodal_f16`, ADR-069 Amendment 1: divergence from it is a bug by definition):

- `dispatch_partial_rope` gains an optional per-token cos/sin buffer override: `Some` binds the caller-supplied single row at offset 0; `None` keeps the unchanged position-indexed 1-D table. Same kernel, no shader change.
- `forward_step_inner_impl` builds the two single-token M-RoPE row buffers once per step (all six GQA layers rotate with the same per-token row; GDN layers never consume RoPE). Text-only requests pass `None` end to end, so plain text decode is structurally untouched.
- Prefill: every physical position of an image-bearing request (image-pad and text alike) rotates with its `tables.cos[pos]`/`tables.sin[pos]` row from `Qwen35VisionRequest::build_mrope_tables` — the same table construction, validation, and ordering as the CPU oracle.
- Decode: the physical KV-cache index and the RoPE coordinate diverge after an image (`decode_axis = physical_pos + rope_delta`, delta may be negative); each step recomputes one row via `build_decode_cos_sin`, exactly as the oracle does. Row-width guards at both build sites.
- MTP speculative-decode call sites and the reference-implementation paths pass `None` (vision × MTP composition is out of scope for this stage).

## Tests (all in-tree, Metal-gated)

- `forward_step_injected_mrope_logits_differ_by_position` — the same injected embedding row at two different physical positions yields different logits under 3-axis M-RoPE (guards against a silent 3-axis→1-D regression that token-content tests cannot catch).
- `generate_multimodal_vision_routes_image_tokens_through_mrope_table` — image-token positions provably go through the 3-axis tables, not the 1-D fallback. Mutation-verified: forcing `dispatch_partial_rope` to always take the 1-D branch makes this and the position-sensitivity test fail (`max_abs_diff == 0`), restore → green.
- `generate_multimodal_vision_text_only_matches_plain_forward_bit_identical` — a text-only request through the new entry point is bit-identical to the plain forward path.
- `generate_multimodal_vision_matches_cpu_oracle_on_real_fixture` — CPU oracle vs Metal on the committed vision fixture (#989), real Qwen3.5-0.8B checkpoint: **byte-identical greedy tokens on both paths, matching the pinned HF golden's first 3 tokens exactly** (`[1919, 2099, 369]`).

Combined run: `test result: ok. 8 passed; 0 failed` (85.25s). Plain-text decode sanity: 64/64 exact greedy-token agreement on the real-checkpoint lm-head test.

## Gates

- clippy `-D warnings`: default, `f16,metal-gpu`, and `--tests` variants — clean.
- fmt, doc lint, capability-matrix fixture check — clean.
- bench-compare (ADR-058): pending, will be appended below. Expected no-change — the text path's only delta is a CPU-side `Option` branch selecting which existing buffer to bind before an unchanged kernel dispatch; the bit-identity and 64-position greedy-agreement tests are the numeric evidence.

## bench-compare (ADR-058)

A `make bench-compare` A/B (origin/main vs this head) was run, but the machine was executing several concurrent build lanes during the run and the results are contention-corrupted: the largest swings (+25% to +839%) landed in `lattice-embed` SIMD micro-benches this diff cannot reach (the diff adds a vision-only M-RoPE path and tests in `lattice-inference`; no embed code is touched). Corrupted numbers are not evidence and are withheld.

Structural argument for the text path: the only delta on existing code paths is a CPU-side `Option` branch selecting which pre-existing buffer to bind before an unchanged kernel dispatch, plus a `None` probe check per decode step in the multimodal function only. The step-by-step bit-identity test (every prefill position + every decode step, `to_bits()` equality vs the plain forward) and the 64-position greedy-agreement test are the numeric no-change evidence.

**Quiet-window rerun (executed):** a second `make bench-compare` (origin/main `4b053a4a68` vs head `7dcf85ae0f`) was run under the machine's exclusive bench-window lock with no build lanes active. It still reported large positive swings (up to +255%) across `lattice-embed` SIMD groups and CPU elementwise groups — traced to filesystem/indexing churn from repository checkouts that overlapped the head measurement phase. Rather than iterating on ritual reruns, here is the structural proof that any A/B on this suite is a measurement of noise for this PR: **every hunk of this PR's library diff is inside `mod inner` in `metal_qwen35.rs`, which is gated `#[cfg(all(target_os = "macos", feature = "metal-gpu"))]` (metal_qwen35.rs:1043), and `scripts/bench-compare.sh` builds the bench targets with default features (`CARGO_FEATURES_INFERENCE` empty, bench-compare.sh:68)** — the module containing 100% of the code change is compiled out of both measured binaries. The only other changed file is a CI workflow. Base and head bench artifacts are compiled from identical effective source, so no bench delta in this suite can be attributable to this diff. The gated numeric evidence for the touched (Metal) path is the bit-identity + greedy-agreement test set above.

## Review-round updates (round 1)

- Added a step-by-step bit-identity test: text-only `generate_multimodal_vision` vs plain forward, `to_bits()` equality per lane over all prefill positions and all decode steps.
- Decode-coordinate coverage moved to the logit level through a probe seam (`generate_multimodal_vision_impl` with an optional decode-logits probe; public signature unchanged). Greedy tokens are insensitive to the decode-coordinate wiring on this model — 18 of 24 layers are GDN and carry no RoPE; both a synthetic large-offset grid and the real fixture produced token-identical trajectories with the wiring deliberately broken — so a token-level test would be vacuous. The new test asserts a margin-safe logit difference between correct and regressed manual arms (vacuity guard), then bit-identity between the function's probed decode logits and the correct arm. Mutation-verified: dropping the function's decode override fails the test at lane 0.
- CI: the Metal injection gate's exact-count grep updated 6 → 7 (the M-RoPE injection-matching test lands in the same filter).

### DIFF
--- a/.github/workflows/e2e-parity.yml
+++ b/.github/workflows/e2e-parity.yml
@@ -690,10 +690,11 @@ jobs:
           grep -q 'LATTICE_VISION_S4_E2E_MUTATION_COSINE' vision-s3-gate.log
           grep -q 'LATTICE_VISION_S5B_GREEDY_TOKENS' vision-s3-gate.log
 
-      # MP2 Metal injection gate: `forward_step_injected_logits_differ_by_row_content`,
+      # MP2+MP3 Metal injection gate: `forward_step_injected_logits_differ_by_row_content`,
       # `generate_multimodal_vision_injection_is_mutation_sensitive`,
-      # `injected_embedding_logits_are_independent_of_token_id`, and
-      # `injected_embedding_never_falls_back_to_token_lookup`
+      # `injected_embedding_logits_are_independent_of_token_id`,
+      # `injected_embedding_never_falls_back_to_token_lookup`, and (MP3)
+      # `forward_step_injected_mrope_logits_differ_by_position`
       # (crates/inference/src/forward/metal_qwen35.rs) previously ran nowhere in
       # CI: they early-returned on a missing Metal device or checkpoint instead
       # of failing, and no workflow selected them (the macOS `feature-matrix`
@@ -707,7 +708,7 @@ jobs:
       # checkpoint a hard failure instead of the default dev-run skip; the
       # `--lib inject` filter also matches two unrelated pre-existing tests
       # (a QuaRot seed round-trip test and the CPU f16 injection oracle), so
-      # the exact-count grep below is 6, not 4 -- update it in the same PR
+      # the exact-count grep below is 7, not 5 -- update it in the same PR
       # if the `inject`-matching test surface changes.
       - name: Run MP2 Metal injection gate (fail-closed)
         env:
@@ -722,7 +723,7 @@ jobs:
             cat vision-mp2-inject-gate.log
             exit "$status"
           fi
-          grep -Eq '^test result: ok\. 6 passed; 0 failed; 0 ignored; 0 measured; [0-9]+ filtered out;' vision-mp2-inject-gate.log
+          grep -Eq '^test result: ok\. 7 passed; 0 failed; 0 ignored; 0 measured; [0-9]+ filtered out;' vision-mp2-inject-gate.log
 
   # Fail-closed control-flow self-test for the Q4 PPL gate script (issue #616).
   # Cheap, deterministic, GPU-free (sandbox tree + stub eval_perplexity), so —
--- a/crates/inference/src/forward/metal_qwen35.rs
+++ b/crates/inference/src/forward/metal_qwen35.rs
@@ -5632,7 +5632,8 @@ mod inner {
                     cfg.rms_norm_eps,
                 );
 
-                // Partial RoPE.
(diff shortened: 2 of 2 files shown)


oceanwaves630 (approved):
Approved: 3-axis M-RoPE Metal vision path verified at 7dcf85ae0; review round 2 clean, bench neutrality established by feature-gating analysis.

PR #1014, opened 2026-07-17
feat(inference): Metal decode kernels for Q3 MLP weight format (ADR-072 P1, #420 Stage 2)

ohdearquant:

Stage 2 of the 3-bit MLP weight format (ADR-072 P1, #420), building on the merged Stage-1 CPU format (#978, `Q3Block` plane-split 2+1, group-32, 16 B/block = 4.0 bpw). Adds the Metal kernels, loader building blocks, and the CPU parity oracle — mirroring the shipped Q4 architecture at every point rather than inventing a new pattern.

- **`gemv_q3_decode`** (`shaders/qwen35.metal`): decode-path GEMV, same dispatch geometry as `gemv_q4_decode` (NR=2 rows/threadgroup, 128 threads, `threadgroups = (ceil(N/2), 1, 1)`). The plane-split layout means each lane's 8 values are two aligned low-plane bytes + one aligned high-plane byte — no cross-byte straddle in the inner dequant loop.
- **`gemm_q3_tiled`** (`shaders/gemm_q3_tiled.metal`, new): prefill-path tiled GEMM, same BM=64×BN=32×BK=32 simdgroup-matrix tiling as `gemm_q4_tiled`, Apple7-gated. Inner unpack loop iterates one thread per (column, value-in-block) pair since plane-split has no fixed byte-to-value ratio.
- **Loader building blocks**: `Q3WeightBuf`/`mmap_q3_weight` mirror the Q4 pair (mmap → no-copy `StorageModeShared` buffer). Two fail-closed gates run before any bytes reach the GPU: `validate_q3_mlp_role` (MLP-only tensor names, per P1 scope — attention/GDN stay Q4/Q8) and `validate_q3_header_payload_bounds` (truncation guard).
- **CPU reference oracle**: `gemv_q3_reference`/`gemm_q3_reference`, straight-line f32 dequant-then-dot over the same packed layout.
- **`dispatch_gemm_q3`**: M=1 → GEMV, M>1 → tiled GEMM, mirroring `dispatch_gemm_q4`.

**Explicit scope gap (not hidden):** the new pieces are `#[allow(dead_code)]` and NOT wired into live checkpoint routing — `MetalFfnWeights::Dense` still carries only `Q4WeightBuf` for MLP tensors, and the checkpoint-level "detect `.q3` files, mix Q3 MLP with Q4/Q8 everything else" path touches ~10 call sites on the load-bearing dispatch path. That wiring is a follow-up stage so it can be verified without risking the working Q4/Q8 decode path. The kernels themselves are exercised end-to-end on a real GPU by the differential tests below.

**Tests** (32/32 green, `--features metal-gpu,f16`, real Apple-series GPU — not skip-passed):
- `gemv_q3_decode_matches_cpu_reference_across_shapes`: N∈{1,5,33} × K∈{32,64,96}, tolerance <5e-3 (f32-throughout kernel; bound dominated by Q3's 8-level quantization step).
- `gemm_q3_tiled_matches_cpu_reference_at_tile_boundaries`: exact tile multiples, non-multiple M/N (zero-pad boundary guards), small odd shapes, plus the Qwen3.5-0.8B dense gate/up depth (K=1024) and down-projection depth (K=3584, 112 `BK=32` iterations) — the measured envelope (max 1.946e-2 at K=3584) derives the `0.037` tolerance bound (~1.9x headroom for cross-device variation and future kernel-scheduling changes).
- `gemm_q3_tiled_differential_fails_closed_on_nan_scale`: proves the NaN-honest comparison itself fails loudly on a corrupted GPU output. Dispatches unconditionally past the device/Apple7 gate (mirroring the other Apple7-gated tests' skip-on-absent convention) then wraps only the comparison in `catch_unwind` (the dispatch completes before the catch begins, so a pre-comparison failure is an ordinary test failure, not a matched panic), asserting on the caught panic's message — so a skip can no longer be mistaken for a verified run the way a whole-test `#[should_panic(expected = "is not finite")]` could.
- Mutation-sensitive coverage on both sides of the parity oracle: `gemv_q3_decode_mutation_sensitive_high_plane_bit` (corrupts post-encode packed bytes fed to the real GPU kernel) and `gemv_reference_mutation_sensitive_high_plane_bit` (same for the CPU oracle) — these corrupt the packed bytes directly because a pack-side mutation corrupts both oracle inputs identically and cannot show up as a GPU-vs-CPU disagreement by construction.
- Loader fail-closed tests: non-MLP role rejection, truncated-payload rejection, bad magic, shape mismatch.

`cargo clippy --workspace -- -D warnings` clean; `cargo clippy -p lattice-inference --features metal-gpu,f16 --all-targets -- -D warnings` clean; `cargo fmt --all` no drift.

bench-compare (ADR-058): measured A/B, quick mode, base `origin/main` (07f5d854) vs head (f45dd66f), run in a quiet window. Result: **0 FAIL, 2 WARN** (both in the 3–7% non-gating band), 60 informational — `simd_batch_cosine/simd_batch/100` +7.11% and `simd_query_batch_dot_product/simd_batch/128d_256c` +4.87%, both `lattice-embed` SIMD micro-benches. Attribution analysis: every new symbol in this PR is `#[allow(dead_code)]` or test-only and unreachable from any bench group (the Metal module is additionally `cfg`-gated out of default-feature bench builds), so the only possible mechanism is code-layout shift from the compiled-but-uncalled `q3_weights` module; the same embed groups swung on an unrelated sibling PR's A/B the same night (one direction there +4.47% on the identical `simd_batch_cosine/simd_batch/100` group), which points at the known #714 micro-bench sensitivity rather than this diff. Warns do not gate (>7% CI-confirmed fails do). Stage 3 (PPL gate + A/B on a real Q3-quantized checkpoint) runs when live routing lands.

Refs #420, ADR-072.

### DIFF
--- a/crates/inference/src/forward/metal_qwen35.rs
+++ b/crates/inference/src/forward/metal_qwen35.rs
@@ -1090,6 +1090,12 @@ mod inner {
 
     const MSL_Q4_TILED_SOURCE: &str = include_str!("shaders/gemm_q4_tiled.metal");
 
+    /// MSL source for the Q3 simdgroup-matrix tiled GEMM kernel (ADR-072 P1,
+    /// W3 MLP-only weight format, #420 Stage 2). Mirrors [`MSL_Q4_TILED_SOURCE`]
+    /// with the 16-byte plane-split 2+1 Q3 block payload substituted for Q4's
+    /// 20-byte asymmetric block. See `weights::q3_weights` for the on-disk format.
+    const MSL_Q3_TILED_SOURCE: &str = include_str!("shaders/gemm_q3_tiled.metal");
+
     /// MSL source for the Q8_0 simdgroup-matrix tiled GEMM kernel.
     ///
     /// Block layout: each Q8_0 block is 34 bytes — 2-byte f16 scale (little-endian)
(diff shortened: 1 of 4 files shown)


oceanwaves630 (approved):
Reviewed at head f45dd66fd: final-verify round confirms fail-closed NaN test contract (catch scoped to the comparison, skip paths cannot satisfy the assertion), dimensions match the model config, and the measured A/B disposition with attribution analysis meets the new bench standard.

PR #1016, opened 2026-07-17
docs: bench A/B quiet-window rule (zero disk activity; reachability check before rerun)

ohdearquant:
Codifies a measurement lesson from two consecutive corrupted bench-compare A/B runs (2026-07-16): repository checkouts overlapping the head measurement phase produced swings up to +250% in bench groups the diff structurally could not reach. Two rules added to the Development Process section:

1. A bench window means zero disk activity — no builds, and also no git checkouts/worktree adds/large copies (filesystem indexing churn lands asymmetrically in one measurement phase and reads as a code regression).
2. Before re-running a corrupted A/B, check structural reachability: a diff confined to a cfg-gated module is compiled out of a default-feature bench build, making base and head binaries identical effective source — the proof belongs in the PR instead of a ritual rerun (applied on #1011).

Docs-only, 2 paragraphs. bench-compare (ADR-058): not applicable — CLAUDE.md is not compiled into anything.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -18,6 +18,12 @@ Paste the output in the PR description. If nothing changed, say "bench-compare s
 
 This process caught a 15% decode throughput regression (157 → 130 tok/s) that had been attributed to "GPU contention noise" for days. The f32 dot_product unrolling that caused it was identified in under 2 minutes via A/B comparison.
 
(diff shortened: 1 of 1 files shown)


oceanwaves630 (approved):
Reviewed at head 5cdf6791: quiet-window definition (zero disk activity, not just zero builds) and the narrowly-scoped compiled-out proof exception both read correctly, with the mandatory bench-compare disposition preserved.

Issue #1015, opened 2026-07-17
Metal generation path cannot distinguish grammar Complete from BlockedConstraint (structured-output blocker)


ohdearquant:
The CPU generation path calls `engine.is_complete_without_continuation(gs)` when the grammar has no legal continuation, correctly distinguishing a valid accept state (clean stop, `StopReason::Grammar`) from a genuine dead-end (crates/inference/src/model/qwen35/generation.rs — multiple call sites).

The Metal generation path in `crates/inference/src/forward/metal_qwen35.rs` never calls it (grep returns zero matches). Its three near-identical inlined grammar-masking sites (prefill + per-decode-step around lines ~9017/9183, ~13779/13953, ~16660/16874) all check `has_finite_logit` and return `Err(InferenceError::InvalidInput("grammar constraint blocked every token..."))` unconditionally when every logit is masked.

Consequence: on the Metal path (`lattice_serve`'s only path), a legitimate terminal grammar state — a successful, schema-conforming completion — and a genuine dead-end produce the identical error. Strict `response_format` requests that reach their natural terminal state are reported as `blocked_constraint`; genuine end-to-end successful structured completions are currently impossible via `lattice_serve`.

Fix shape: mirror the CPU path's `is_complete_without_continuation` check at each of the three Metal sites before the `has_finite_logit` failure branch (classic sibling-invocation-path fix — all three copies, plus a grep for any fourth). The serve-layer wiring already anticipates the distinction and needs no changes.

Found during structured-output v0 stage-1 wiring; the serve code documents this at its `WorkerEvent::Failed` handling site (search `KNOWN GAP`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

ohdearquant:
Resolved in the tree. The fix landed in #1087 (`d9f7afd0ab`) and this issue was not closed at the time.

**All three Metal sites now carry the check**, in the shape this issue prescribed — the accept-state test runs before the `has_finite_logit` failure branch:

| site | prefill guard | accept-state check |
|---|---|---|
| 1 | `metal_qwen35.rs:8921` | `:8922` (+ decode-step at `:9080`, `:9166`) |
| 2 | `metal_qwen35.rs:14016` | `:14017` (+ decode-step at `:14176`, `:14310`) |
| 3 | `metal_qwen35.rs:17080` | `:17081` (+ decode-step at `:17290`, `:17418`) |

[...]

Issue #734, opened 2026-07-09
perf(inference): grammar-constrained decode ~4.5x slower per token when schema exceeds MAX_GRAMMAR_STATES (mask_by_simulation fallback)


ohdearquant:
## Summary

Grammar-constrained decoding (ADR-046) becomes a net latency *regression* on a moderately nested JSON-Schema. Measured on Qwen3.5-0.8B-Q4 (Metal, M2 Max): the same generations under a grammar ran **4.5× slower per token** than unconstrained decode (64.4 → 14.4 tok/s), and end-to-end generation was *slower* (11.2 s vs 6.8 s) despite producing ~5× fewer tokens.

## Measurement

| | unconstrained | grammar-constrained |
|---|---|---|
| mean decode speed | 64.4 tok/s | 14.4 tok/s |
| mean generation wall time (excl. model load) | 6.80 s | 11.20 s |

Schema shape: 4-level object nesting, 3 array fields, 6 enum fields, ~250K-token vocabulary. Grammar *compile* time was also elevated — 22–30 s one-time, versus the "50–200 ms for a 250K vocab" the module's own doc comment cites.

## Likely mechanism (verified in source)

`crates/inference/src/grammar/vocab_partition.rs:38` caps the precomputed per-state bitmask table at `MAX_GRAMMAR_STATES = 256`. When a grammar has more than 256 reachable states (`vocab_partition.rs:74`), states beyond the cap have no precomputed mask and fall back to `mask_by_simulation` (`engine.rs:260`), which re-simulates the vocabulary against the live PDA stack **every decode step** — sound but not cheap, as the doc comment at `engine.rs:268` notes. A schema this nested plausibly blows past 256 reachable states quickly, putting most of the decode loop on the fallback path; the elevated compile time is consistent with being well past the precomputed regime.

I have not directly instrumented the reachable-state count for this specific schema — that is the one piece left to confirm — but the constant, the fallback path, and the compile-time blowup all point the same way.

## Why it matters

Grammar-constrained decode is the natural way to guarantee structured / JSON-schema output from the serve path (see #588). If it carries a ~4.5× per-token tax on any non-trivial schema, it is not viable for latency-sensitive structured-output serving without one of:

- raising or removing the 256-state cap (with a memory-bound strategy),
- making `mask_by_simulation` incremental — cache per-state masks lazily on first visit instead of recomputing the full vocabulary every step,
- or partitioning large grammars so the hot region stays within the precomputed regime.

## Repro

[...]

PR #1018, opened 2026-07-17
feat(serve): strict json_schema response_format (structured output v0, stages 1-2)

ohdearquant:

Wires `response_format: {type: "json_schema", strict: true}` into `lattice_serve`'s non-streaming chat completions on top of the existing PDA-backed grammar engine. Three commits: stage 1 (admission + serve wiring), stage 2 (compiled-grammar LRU cache + single-flight), and the review round-1 fix (raw duplicate-key rejection, range-independent integer validation, typed cache/constraint errors).

**Stage 1 — admission + wiring** (`lattice_serve.rs`, additive `exceeds_state_budget()` accessor on `GrammarEngine`, one new `ApiError::ServerError {message, code}` variant):
- v0 schema subset, fail-closed: closed objects (`additionalProperties:false`, all properties required), uniform arrays, primitives, nesting. Everything else rejected 4xx pre-worker with the offending keyword named (`enum`, `const`, `pattern`, length/numeric constraints, `$ref`, composition, type unions, `prefixItems`, `format`, unknown keywords). Streaming+schema and non-strict rejected 4xx.
- Admission compiles the schema and rejects any grammar exceeding the 256-state precomputed-mask budget (v0 never reaches the full-vocab simulation fallback).
- Admitted requests: worker config carries the compiled grammar, `enable_thinking` forced off, response carries `constraint_applied: true`.
- Post-generation: grammar-blocked → HTTP 500 `blocked_constraint`; length/KV exhaustion → 500 `length_limit`; clean stop is re-validated by an independent evaluator (shares no code with the grammar compiler) → mismatch is 500 `validation_failed`. Partial JSON is never returned as 200.
- Known engine gap, documented in-code (`KNOWN GAP`) and tracked as #1015: the Metal generation path cannot yet distinguish a successful terminal grammar state from a dead-end, so genuine end-to-end successful strict completions are blocked on #1015. The serve layer needs no changes when it lands.

**Stage 2 — cache + single-flight**:
- LRU (32 entries) keyed on recursively key-sorted canonicalized schema JSON; property-order-insensitive (tested: two orderings hit one entry, distinct schemas do not — mutation-verified by breaking the key to a constant and watching the test fail).
- Single-flight: concurrent identical-schema requests compile exactly once (threaded test asserts one compile via a test-only counter).
- Cache hit reuses the compiled engine (`Arc::ptr_eq` asserted) and removes the per-request vocab clone.
- Deliberate stage-2 exclusions documented in-repo: the secondary tokenizer/config load in `run()` stays (a `!Send` worker-boundary constraint, not cache scope); no production cache metrics; cache key omits `model_id` (one model per process today).

**Tests**: 99 bin-target tests green (admission table: one accept per admitted construct, one rejection per unsupported keyword, route tests for marker/thinking-override/500 codes/independent-validation, cache/single-flight/eviction, duplicate-key 400s, layer-discriminating mutation tests, wording-independence for the typed constraint-blocked path; review round-2 added 4 `v0_parse_string` surrogate/control-char tests and 1 cache panic-injection single-flight test); grammar module 170 and serve::metal_worker 19 green (no regression). Clippy clean on workspace and on the feature-gated bin/tests targets; fmt clean. Mutation transcripts for admission (both layers, each failing its own test) and cache key are in the work log.

bench-compare (ADR-058): measured A/B, quick mode, base `origin/main` (07f5d854) vs head (ddc0ef98), quiet window. Raw result: 1 FAIL (`tier_prepared_batch_sizes/int4_batch_prepared/10` +63.98%), 7 WARN, **and 30 CI-confirmed "improvements" of 3–11%** — a bidirectional-swing signature no cold-path diff produces (the lib-crate changes compiled into bench builds are an additive bool field set once in `GrammarEngine::new` and an error-enum variant raised only on generation failure; the serve binary is not compiled into any bench). The FAIL was then disproven directly: two idle A/A re-runs of the failing group on the head checkout measured 238.5–244.8 ns (point 241.5 ns, run-to-run delta within Criterion's noise threshold) — exactly the A/B's *base* value (241.6 ns) and nowhere near the alleged 396 ns, so the +64% reading was contamination of the head measurement phase, not code. Same conclusion applies to the co-moving warns/wins (embed SIMD micro-bench sensitivity, tracked in #714). Disposition: no attributable change.

Refs #1015 (launch blocker for e2e strict completions), #734 (mask-simulation cliff the 256-state admission gate avoids).

### DIFF
--- a/crates/inference/src/bin/lattice.rs
+++ b/crates/inference/src/bin/lattice.rs
@@ -2714,7 +2714,7 @@ mod serve {
                     }
                     WorkerEvent::Complete(output) => return Ok(output),
                     WorkerEvent::Rejected(api_err) => return Err(api_err),
-                    WorkerEvent::Failed(message) => {
+                    WorkerEvent::Failed(message) | WorkerEvent::ConstraintBlocked(message) => {
                         return Err(ApiError::Internal {
                             message: format!("generation failed: {message}"),
                         });
--- a/crates/inference/src/bin/lattice_serve.rs
+++ b/crates/inference/src/bin/lattice_serve.rs
@@ -93,7 +93,8 @@ mod imp {
     use lattice_inference::tokenizer::bpe::BpeTokenizer;
     use serde::Deserialize;
     use serde_json::{Value, json};
-    use std::sync::Arc;
+    use std::collections::{HashMap, VecDeque};
+    use std::sync::{Arc, Condvar, Mutex};
     use std::time::{Instant, SystemTime, UNIX_EPOCH};
     /// Only used by the test module's raw job-channel helpers
     /// (`mpsc::UnboundedReceiver<WorkerJob>` etc. -- see
@@ -176,6 +177,11 @@ mod imp {
         /// clone per compile (no cross-request grammar cache in v0; see the
         /// structured-output v0 design note's stage split).
         vocab_bytes: Arc<Vec<Vec<u8>>>,
+        /// Bounded (32-entry, LRU) single-flight cache of compiled
+        /// `GrammarEngine`s, keyed by canonicalized admitted schema JSON
+        /// (structured-output v0 design note, stage 2). `Arc`-shared so
+        /// every clone of `AppState` (one per request) hits the same cache.
+        grammar_cache: Arc<GrammarCache>,
     }
 
     // ─── OpenAI request shapes ───────────────────────────────────────────────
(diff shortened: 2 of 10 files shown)


oceanwaves630 (approved):
Reviewed at head ba61bfe9b: structured-output v0 approved through three review rounds; the sole commit past the reviewed head is a three-line cfg alignment gating test-only helpers to the feature combination their consuming tests already require, verified against the exact failing feature-matrix clippy command.

PR #1019, opened 2026-07-17
perf(inference): trie-accelerated grammar mask for over-cap states (#734)

ohdearquant:
## Problem (#734)

Grammar-constrained decode on schemas whose compiled PDA exceeds `MAX_GRAMMAR_STATES` (256) was ~15x slower than unconstrained decode. Profiling (committed harness, `gramperf_profile`) on a 4-level-nested JSON schema measured the mechanism precisely:

- 147 of 150 decode steps (98%) missed the precomputed `VocabPartition` table and fell back to `mask_by_simulation`
- that fallback ran an independent `simulate_token` byte-walk for every one of the 248,320 vocab tokens, every step: **114.11 ms/step, 91.1% of the 125.25 ms/step wall time**
- 99.83%+ of the vocab is blocked per step, and almost all rejections happen on the first byte or two

## Fix

`ByteTrie` (new `crates/inference/src/grammar/trie.rs`): a byte trie over `vocab_bytes`, built lazily once per engine (`OnceLock`, first over-cap `mask_logits` call) and reused. The mask is computed by DFS-walking the trie with a live PDA state: a rejected byte edge prunes every token sharing that prefix in one step, instead of re-walking each token independently. Clone-on-branch state handling: only nodes with >1 child clone the PDA state (O(stack depth)); straight-line runs — the common case past the first bytes of a BPE token — advance in place with zero clones.

`mask_logits`'s over-cap fallback now routes to `mask_by_trie`. The precomputed-table path for in-cap states is untouched. `mask_by_simulation` stays (now `pub`) as the reference oracle.

Semantics are identical by construction: `simulate_token` returns `Accept` iff every `advance_byte` accepts, and reaching a trie terminal requires exactly that same all-accepted byte path. Both `Reject` and `ContextDependent` (a rejected byte at position 0 vs >0) block a token identically in the oracle, and both are covered by the same pruned-edge condition in the trie walk.

## Correctness evidence

- **Differential tests vs the oracle** (`engine.rs` tests): byte-identical mask comparison (`mask_by_trie` vs `mask_by_simulation`) over a corpus of ≥20 distinct PDA states harvested from a schema-valid trajectory on the #734-shape schema (29 states, ≥15 over-cap asserted), plus a **dedicated over-accept test** reported separately (a token allowed by the trie but rejected by the oracle is a soundness bug, strictly worse than a slow mask).
- **Real-vocabulary differential** (`#[ignore]`-gated, needs a local model checkout): same comparison over the full real Qwen3.5 tokenizer vocab — 248,320 tokens, 208 harvested states (204 over-cap): **0 mismatches, 0 over-accepts**. Run:
  `cargo test -p lattice-inference --release --features "f16,metal-gpu" --lib real_vocab -- --ignored --nocapture`
- **Mutation-sensitivity** (verified, not committed): disabling the subtree-prune check → over-accept test fails (62,937 over-accepts); inverting the mask apply-bit → all trie tests fail. Both mutations restored; re-verified independently after commit.
- **End-to-end routing test** via public `mask_logits`, so a future fallback-call-site edit that silently stops routing to the trie fails a test.
- All pre-existing grammar tests green (200 lib + integration), clippy `-D warnings` clean, fmt clean.

## Measured results

Harness: `gramperf_profile` (committed), Qwen3.5-0.8B Q4 on Metal (M2 Max), greedy (`temperature=0.0, top_k=1, seed=42`), 150 tokens/generation, 3 runs, exclusive GPU lock held, machine otherwise quiet. Replay:

```
GRAMPERF_RUNS=3 GRAMPERF_MAX_TOKENS=150 GRAMPERF_PROBE_CAP=1024 \
  target/release/gramperf_profile
```

| metric | before | after | change |
|---|---|---|---|
| wall/step (grammar-constrained) | 125.25 ms | 7.92 ms (runs: 8.18 / 6.83 / 8.76) | **~15.8x** |
| over-cap fallback cost | 114.11 ms/step (91.1% of wall) | 1.037 ms/call (13.1% of wall) | **~110x** |
| throughput on this schema | ~8.0 tok/s | ~126 tok/s | |
| trie build (one-time, lazy) | — | 64.2 ms | first over-cap call only |
| grammar compile (`GrammarEngine::new`) | 16.94 s | 17.53 s | unchanged (untouched path; machine variance) |

Fallback call counts are identical before/after (147/150 steps) — the fix changes what the fallback costs, not how often it runs. Known noise disclosed: this run's single unconstrained-baseline sample (10.32 ms/step) is noisier than the profiling round's (6.08 ms/step); the headline fallback numbers come from direct per-call counters, not baseline subtraction, and are unaffected.

## bench-compare (ADR-058)

Full-suite quick A/B, base `205f2176` (origin/main) vs head `f65e0819`, run under the exclusive machine bench window.

The first full-suite run flagged 4 confirmed regressions >7%, all in `lattice-embed` SIMD cosine groups (`simd_batch_cosine/simd_batch/1000` +16.3%; three `simd_batch_cosine_normalized_query` benches +9.7–10.6%), alongside equally large nominal improvements in untouched groups (`softmax_attention` −30%, `simd_throughput_384` −12 to −20%). A targeted identical-refs re-run of exactly the flagged groups (`BENCH_GROUPS_EMBED="simd_batch_cosine"`, same base/head SHAs, same machine, immediately after) reproduced **none** of the 4: the same benchmarks swung to −9% to −45% (e.g. `pair_loop_cosine/768d_4c`: +9.77% FAIL in run 1 → −43.65% in run 2), while a different, previously-clean bench (`scalar_loop/10`) spiked +11.5% instead.

Bidirectional ±45% swings between two runs of identical code are machine noise, not a code effect. Structurally, the diff cannot reach these benchmarks: it is confined to the grammar module and a profiling binary in `lattice-inference` (5 files); `lattice-embed`'s dependency on `lattice-inference` is optional and feature-gated, and the flagged SIMD kernels execute no grammar code on any path.

**Disposition: no change attributable to this diff.** The genuinely-affected path (grammar-constrained decode) is measured directly by the committed `gramperf_profile` harness above (~15.8× improvement).

## Scope

No change to: in-cap precomputed masking, grammar compilation, PDA semantics, sampling, or any non-grammar code path. The eager-compile cost (~17 s on this schema) is deliberately out of scope — separate issue to follow.

### DIFF
--- a/crates/inference/src/bin/gramperf_profile.rs
+++ b/crates/inference/src/bin/gramperf_profile.rs
@@ -203,7 +203,7 @@ fn run() -> Result<(), Box<dyn std::error::Error>> {
              precomputed_ns_per_step={} context_recheck_calls={} context_recheck_ns={} \
              context_recheck_ns_per_step={} fallback_calls={} fallback_ns={} \
              fallback_ns_per_step={} advance_calls={} advance_ns={} advance_ns_per_step={} \
-             residual_ns_per_step={} stop_reason={:?}",
+             residual_ns_per_step={} stop_reason={:?} trie_build_ns={}",
             output.generated_tokens,
             wall_ns / steps,
             mp.precomputed_calls,
@@ -226,6 +226,7 @@ fn run() -> Result<(), Box<dyn std::error::Error>> {
             mp.advance_ns / steps,
             residual_ns / steps,
             output.stop_reason,
+            engine.trie_build_ns(),
         );
         if i == 0 {
             grammar_token_ids = output.token_ids.clone();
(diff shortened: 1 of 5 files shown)


oceanwaves630 (approved):
Full diff read: trie fallback reproduces the simulation oracle's contract with the over-accept direction pinned by its own differential test and the executed real-vocab 0/0 result; walk-root and scratch-buffer fixes are mutation-tested; profiling instrumentation is opt-in with a single-branch off path.

PR #1021, opened 2026-07-17
feat(serve): image content-parts on /v1/chat/completions wired to the vision decode path (ADR-069 S6)

ohdearquant:
> **Status: parked, to be rebuilt on top of the shared serve contract.**
>
> This branch was written against the pre-`serve::contract` architecture: it adds image
> support inside `lattice_serve`'s own `to_chat_message` / `content_parts` pair. Main has
> since centralized request normalization into `crates/inference/src/serve/contract.rs`
> and removed that local pair, so the branch and main no longer describe the same request
> path. At this branch's merge-base, `crates/inference/src/serve/` held `metal_worker.rs`,
> `metrics.rs` and `mod.rs` and no `contract.rs` — the shared contract post-dates the
> branch entirely.
>
> Resolving the conflicts textually would produce something worse than leaving them:
> restoring this branch's local parsing compiles and passes this PR's own tests while
> having no production caller, because real requests now reach `parse_chat_req` →
> `normalize_request` → the shared `normalize_messages`, which still rejects image parts.
> Zero conflict markers, a clean build and green tests are all satisfied by that outcome,
> which is exactly why it needs a design decision rather than a careful merge.
>
> The work here is not discarded. Image support belongs in the shared contract, where
> both binaries get it from one code path and a capability flag rather than from two
> implementations that have to be kept in agreement. The parts of this branch that are
> independent of where normalization lives — the data-URI decode and size clamp, the
> Metal worker's image plumbing, the prompt splice, and the e2e test — carry over to that
> rebuild unchanged.

## ADR-069 S6: image content-parts on `/v1/chat/completions` (lattice_serve)

Accepts OpenAI-shape `image_url` content-parts on `lattice_serve` and wires them end-to-end into the existing Qwen3.5 vision decode path (S1-S5b/MP2/MP3: #1004/#1005/#1007/#1009/#1011). The vision-decoder engineering all pre-existed; this PR is the serve-surface plumbing: request parsing → `ChatMessage.image` → worker vision branch → `generate_multimodal_vision`.

### Request surface
- `data:image/*;base64` URIs only; remote URLs get 400 `unsupported_image_url_scheme` (the server never fetches on the caller's behalf).
- Fail-closed validation: malformed data URI / bad base64 / non-image MIME / empty payload → 400; decoded size clamped by `MAX_DECODED_IMAGE_BYTES` (10 MiB) in addition to the existing 64 KiB pre-parse part clamp ("clamp before alloc").
- Single image per request (second image in a message or across messages → 400).
- Hand-rolled standard-alphabet base64 decoder rather than a new dependency; input is already bounded by the pre-parse clamp.

### Vision path
- `MetalWorker::spawn` loader now returns `Option<Qwen35VisionWeights>`; `lattice_serve` loads them when the checkpoint has a `vision_config` (warns, doesn't fail startup, if absent). `lattice.rs` passes `None` (mechanical signature fix only; its surface still rejects images).
- Image-bearing jobs: preprocess → Metal ViT forward (per-GEMM CPU fallback built into S3b) → merger → prompt ids with the `vision_start/image_pad*N/vision_end` splice (same layout as the committed #989 HF differential golden; two tokenizer calls so only one BPE seam) → `generate_multimodal_vision`.
- Context-window check runs on the expanded id stream, same policy as text.
- Known v0 limitations (documented in-code): vision responses arrive as one SSE delta (no incremental `on_token` hook on `generate_multimodal_vision` yet); no cross-turn prefix cache on the vision path.

### Tests
- 12 new/updated parsing-layer unit tests (data-URI decode, scheme/base64/MIME/size/multi-image rejections, ordering, `ChatMessage` carriage) + 1 HTTP-level rejection test; full existing `lattice_serve` (107) and `metal_worker` (19) suites pass unchanged.
- New model-gated integration test `vision_serve_e2e_test`: spawns the real binary, posts a committed golden image + question, asserts a grounded non-empty answer; takes the machine-wide GPU lock.
- Capability matrix updated: the image row now documents acceptance on vision checkpoints with the new fixture set.

### bench-compare disposition
The diff is confined to serve binaries (`lattice_serve.rs`, `lattice.rs`), `serve/metal_worker.rs`'s job-dispatch closure, and an additive `Option<ChatImage>` field on `ChatMessage`. The vision branch is taken only when a request carries an image; plain-text requests hit the identical code path as before (the image scan is one iterator pass over the handful of chat messages per request, not per token). No kernel, forward-pass, or per-token code changed; the Criterion bench targets (`elementwise_cpu_bench`, `simd`) do not compile any of the changed functions. Live-smoke measurement of the vision path itself will be posted before this PR leaves draft.

- Pull request head evaluated: `fa979d957f81f91d94db99eca17ad6d2bf9385ef`

re-derived.

### Gates
clippy `-D warnings` (both feature sets) clean · fmt clean · `--bin lattice_serve` 107/107 · `serve::metal_worker` 19/19 · sibling binary `lattice` compiles + tests green.

### Live smoke (post-draft update)
The model-gated e2e ran on real hardware at this head (machine-wide GPU lock + heavy-lane held): `serve_chat_completions_answers_a_grounded_question_about_an_image ... ok` (12.06s). The served answer to "Describe this image." on the committed golden fixture: "This is a digital illustration or graphic design featuring a gradient background with four distinct colored shapes arranged in a 2x2 grid, each containing…" — a grounded description of the fixture's actual content (2x2 colored-shape grid on a gradient), served over HTTP through the full parse → ViT → merger → M-RoPE decode path.


ohdearquant:
Superseded as an implementation: inline image content parts on the chat-completions endpoint were merged in #1231, and the shared contract now parses typed text and image parts, enforces the image restrictions, and dispatches the vision request through the worker.

Capabilities from this branch that main does not have are preserved as follow-ups rather than closed away: structured image error codes and retryable vision-load state (#1335), and operator-controlled vision preload plus skipping terminal head work on non-final multimodal prefill positions (#1336).

PR #1022, opened 2026-07-17
feat(inference): QuaRot online R3/R4 rotation contract + reference (#703 PR1)

ohdearquant:
## Summary

Contract + reference groundwork for online R3/R4 Hadamard rotations (#703, PR1 of the
staged plan). No loader, converter, or forward-path wiring — existing v0 artifacts and
all runtime behavior are unchanged.

- **`quarot/hadamard.rs` — `BlockHadamard`**: block-diagonal randomized Hadamard over a
  contiguous partition (`n % b == 0`, `b` a power of two in {64, 128, 256}). Per-block
  sign vectors derive from avalanche-finalized seeds so adjacent blocks stay uncorrelated
  even for small, linearly-spaced seeds. Padding to a power of two was rejected
  (non-orthogonal per ADR-044); block-diagonal keeps every block exactly orthogonal.
  Covers both real axis lengths: 3584 (0.8B `intermediate_size`) and 17408 (27B).
- **`quarot/plan.rs`**: `RotationId::{AttentionOutputR3, MlpDownR4}` + `OnlineRotationSpec`
  (`r3_full_attention` layer scope `[3,7,11,15,19,23]` verified against `qwen35_0_8b()` +
  `compute_layer_types`). `apply_tensor_rotation`/`_f64` explicitly refuse the new
  contract-only variants rather than silently no-op'ing. The plan gate enforces
  `BlockHadamard::new`'s full length contract (`dim != 0`, `dim <= MAX_BLOCK_HADAMARD_LEN`,
  block-count cap), so a plan can never certify an artifact the primitive would refuse to
  construct.
- **`quarot/io.rs`**: `ArtifactVersion::{V0Residual, V1Online}` (string tags
  `"v0-residual"` / `"v1-online-r3r4"`, refuse-don't-guess parsing) +
  `OnlineArtifactDescriptor::validate` refusals: v0 carrying online metadata; v1 with zero
  rotations; v1 whose `asymmetric_tensor_names` is not the exact recipe-derived set
  (missing, extra, or duplicate). Validation is fail-closed and bounded: both
  caller-supplied vectors are size-capped up front and set membership is checked in linear
  time, so a deserialized descriptor cannot drive super-linear work.
- **`quarot/r3_reference.rs` — the R3 orientation result**: one-layer f32 synthetic
  reference at the real 0.8B GQA shape (8 Q heads / 2 KV heads / head_dim 256 /
  hidden 1024), 3-position KV cache, dense forward as ground truth. Of four candidate
  orientations, exactly one is a true reparameterization (max |delta| < 1e-4): rotate the
  gated attention output immediately before `o_proj` and absorb `R^T` into `o_proj`'s
  input side. Pre-gate variants diverge (`sigmoid(R z) != R sigmoid(z)`). A V-cache
  round-trip composed with the winner also matches, so a future value-cache rotation is
  compatible with this orientation. `OnlineRotationSpec::r3_full_attention` records
  `AbsorptionSide::InputSide`; the enumeration test asserts that agreement directly.

## Test plan

- `cargo test -p lattice-inference quarot`: green (partition, refusal, bound, and mutation
  tests plus the R3 orientation enumeration).
- `cargo fmt --check`: clean.

## bench-compare disposition

Contract + reference definitions and tests only — no hot-path call sites. An A/B
(`make bench-compare`) in a quiet window will be added before this PR is marked ready.

### DIFF
--- a/crates/inference/Cargo.toml
+++ b/crates/inference/Cargo.toml
@@ -277,6 +277,10 @@ name = "topk_readback"
 harness = false
 
 [[bench]]
+name = "quarot_hadamard_bench"
+harness = false
+
+[[bench]]
 name = "decode_attn_bench"
 harness = false
 required-features = ["f16"]
--- a/crates/inference/benches/quarot_hadamard_bench.rs
+++ b/crates/inference/benches/quarot_hadamard_bench.rs
@@ -1,0 +1,72 @@
+//! `block_hadamard` Criterion bench group.
+//!
+//! Measures [`BlockHadamard`] at the real non-power-of-two axis dimensions
+//! QuaRot rotation actually targets — Qwen3.5-0.8B's `intermediate_size`
+//! (3584 = 2^9 * 7) and Qwen3.6-27B's `intermediate_size` (17408 = 2^11 * 17)
+//! — with construction (`BlockHadamard::new`: per-block seed derivation
+//! plus one contiguous signs-buffer reservation) and apply
+//! (`BlockHadamard::apply`, the per-block butterfly transform) measured as
+//! separate benchmark IDs so a regression in either phase is individually
+//! attributable.
+//!
+//! ```bash
+//! cargo bench -p lattice-inference --bench quarot_hadamard_bench -- "block_hadamard"
+//! ```
+//!
+//! Always compiled (no feature gate) — `BlockHadamard` is pure CPU code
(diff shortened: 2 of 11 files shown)


oceanwaves630 (approved):
Verified against the review artifact at head; both prior findings addressed.

PR #1029, opened 2026-07-17
feat(inference): os_signpost instrumentation of Metal decode path

ohdearquant:

## Summary

Adds `os_signpost` instrumentation over the existing Metal decode path (`MetalQwen35State::forward_step` and its callers), gated behind a new `signpost` cargo feature (default OFF). This lets a Metal System Trace captured in Instruments show per-decode-step launch and synchronization structure — command-buffer commit, interior `waitUntilCompleted` wait, host-side scalar reads, grammar masking, and sampling — instead of one opaque GPU-busy span.

- New module `crates/inference/src/forward/signpost.rs`: an RAII interval guard (`signpost::interval(label)`) backed by a minimal `extern "C"` shim over libsystem_trace's `os_signpost`/`os_log` entry points on macOS. No new external dependency — it reuses the same "link a macOS framework directly" idiom the crate already uses for Metal/objc.
- Off the `signpost` feature (or on non-macOS), the guard is a zero-sized no-op type with an `#[inline(always)]` constructor, so call sites don't need their own `#[cfg(...)]`.
- Instrumented labels: `decode.step`, `decode.cb_commit`, `decode.cb_wait`, `decode.host_scalar_read`, `decode.grammar_mask`, `decode.sample`.
- New doc: `crates/inference/METAL_TRACE.md` — how to capture a trace in Instruments, plus the label glossary.

No behavior change to the decode path itself; this is instrumentation only.

## bench-compare disposition

The `signpost` feature is default-OFF, and its *implementation* (the real `os_signpost` FFI calls) lives entirely behind `#[cfg(all(feature = "signpost", target_os = "macos"))]` in `signpost.rs`, so that code is compiled out of any default-feature bench build.

However, the call sites added in `metal_qwen35.rs` (`signpost::interval("decode.step")` etc.) are **not** themselves behind `#[cfg(feature = "signpost")]` — they are unconditional calls into a zero-sized no-op guard type when the feature is off. That means these lines *do* reach the default bench build's source, even though they're designed to optimize away to nothing (an inlined constructor for a zero-sized type, immediately dropped). Per this repo's bench-compare carve-out, that's a materially different claim than "compiled out entirely," so I'm not self-certifying via the structural argument here — flagging it instead of asserting no-op-by-construction is enough.

`make bench-compare` was not run in this session. Recommend a quick `--quick` A/B on the decode/elementwise groups before merge to confirm the no-op call sites carry no measurable cost in the default build.

## Test plan

- [x] `cargo check -p lattice-inference --features f16,metal-gpu,test-utils` (signpost off) — clean
- [x] `cargo check -p lattice-inference --features f16,metal-gpu,test-utils,signpost` — clean
- [x] `cargo clippy -p lattice-inference --all-targets --features f16,metal-gpu,test-utils -- -D warnings` — clean
- [x] `cargo clippy -p lattice-inference --all-targets --features f16,metal-gpu,test-utils,signpost -- -D warnings` — clean
- [x] `cargo test -p lattice-inference --lib --features f16,metal-gpu,test-utils signpost::` — passes (no-op fallback)
- [x] `cargo test -p lattice-inference --lib --features f16,metal-gpu,test-utils,signpost signpost::` — passes (real os_signpost path, runs without crashing)
- [ ] `make bench-compare` — not run this session, see disposition above
- [ ] Capture an actual Metal System Trace in Instruments and confirm the labeled intervals show up (needs a local Instruments session)

### DIFF
--- a/crates/inference/Cargo.toml
+++ b/crates/inference/Cargo.toml
@@ -38,6 +38,11 @@ mixture = ["dep:lattice-fann"]
 # GDN state-traffic byte counters (issue #422, measure-first for ADR-058).
 # Opt-in only: the default decode path stays byte-identical when this is off.
 gdn-state-counters = []
+# os_signpost instrumentation of the Metal decode path (W1.0), for capturing
+# a Metal System Trace of command-buffer/dispatch/wait structure in
+# Instruments. Opt-in only: every call site is a zero-sized no-op guard when
+# this is off, so the default decode path stays byte-identical.
+signpost = []
 # Exposes test-only fixture constructors (currently: a tiny all-zero-weight
 # Qwen35Model, ADR-080 C2) as `pub` library API, gated so they never
 # ship in a normal build. Needed because bin targets (lattice.rs,
--- a/crates/inference/METAL_TRACE.md
+++ b/crates/inference/METAL_TRACE.md
@@ -1,0 +1,30 @@
+# lattice-inference — METAL_TRACE.md
+
+`os_signpost` instrumentation of the Metal decode path, gated behind the `signpost` cargo feature (default OFF). It exists so a Metal System Trace captured in Instruments can show the per-decode-step launch and synchronization structure — command-buffer commits, interior `waitUntilCompleted` waits, host-side reads of GPU results — instead of a single opaque GPU busy span.
+
+## Capturing a trace
+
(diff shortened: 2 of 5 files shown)


oceanwaves630 (approved):
Automated review verdict on current head: APPROVE, zero blocking findings (minors noted for follow-up).

PR #1032, opened 2026-07-18
fix(inference): unify native weight ingress validation

### LINKED ISSUE #736: bug(inference): Q4/F16 direct-load path never validates tensor geometry against the model config
**Impact:** High — malformed, truncated, or transposed weight artifacts that are internally self-consistent load successfully and either produce out-of-bounds reads or silently wrong inference output, with no error surfaced.

**Location:** `crates/inference/src/forward/metal_qwen35.rs`. The Q4 weight loader validates only that a `.q4` file's on-disk header is self-consistent with its own payload length; it never compares the tensor's declared/actual dimensions against the model's configuration (hidden size, vocab size, intermediate size, head dims, etc). All per-layer attention/FFN projections and the embedding/`lm_head` tensors go through this path. The only geometry assertions present (`debug_assert_eq!` on FFN gate/up projection lengths) compile out entirely in release builds. Downstream, the public per-token forward function's only bounds check on the input token ID is against the configured vocabulary size, not against the actual size of the loaded embedding buffer — so a token ID that is legitimately in range per config can still read past the end of an undersized embedding buffer that was itself accepted by the loader.

**Repro sketch:**
1. Produce a Q4-format checkpoint directory where the embedding tensor has an internally valid Q4 header (self-consistent payload bounds) but fewer rows than the model config's vocabulary size, or where a projection tensor is transposed to the same total element count as expected.
2. Load the checkpoint via the direct Q4-directory constructor: no error is raised at load time.
3. Run a forward pass with a token ID that is valid per the model config's vocabulary size but exceeds the actual (undersized) embedding buffer's row count: the row read walks past the end of the allocation.
4. Alternatively, a same-element-count transposed projection loads successfully and produces silently incorrect (not just out-of-bounds) logits — no failure signal at all.

**Suggested fix:** Give every tensor loader an expected shape derived from the model configuration and return an error on any mismatch, using checked (not narrowing/unchecked) arithmetic for shape products. Apply this uniformly to Q4 and F16 artifacts (the F16 half is #735). Replace the debug-only FFN length assertions with a release-active `Result`-returning check. Derive the embedding-buffer bound actually used at inference time from the loaded artifact's real size (or validate it equals the configured vocabulary size at load time), not solely from the static config value.

Related: #655 (centralizing Q4/QuaRot manifest parsing) is the natural home for this validation once it exists — but the missing geometry check is a defect independent of that refactor.

### LINKED ISSUE #802: bug(inference): unify native Q4/KHF1 file validation — non-finite scale/bias, exact file extent, and panic-to-Result conversion

### LINKED ISSUE #985: bug(inference): read_q4_header cursor-position doc-contract is false (BufReader advances fd past payload_offset)
## Summary

`read_q4_header` (`crates/inference/src/weights/q4_weights.rs:606`) documents "On return the file cursor is positioned at the start of the block data," but it reads the header through `std::io::BufReader::new(file)` over a `&File`. `BufReader` fills in ~8KB chunks, so for any header smaller than the buffer it advances the underlying shared file-descriptor cursor **past** `payload_offset` (it reads part of the block payload into the buffer). After the function returns and the `BufReader` is dropped, the raw `File` cursor is left past the payload start, so the documented post-condition is false.

This is the exact same defect fixed for `read_q3_header` in #978 — surfaced by that PR's sibling-path grep.

## Severity: latent (contract-only, no live misread today)

Every current caller either seeks explicitly before reading the payload or mmaps the whole file and indexes by `payload_offset`, so none relies on the returned cursor:

- `read_q4_payload_bounded` — calls `f.seek(SeekFrom::Start(header.payload_offset))` before `read_to_end`
- the Metal no-copy loaders in `forward::metal_qwen35` and `forward::moe_expert_cache` — mmap the file and index by `payload_offset`
- the CLI header-only path in `bin/lattice.rs` — reads no blocks

So q4 loading is correct in production. The risk is a future caller that trusts the doc-promised cursor and reads blocks off the same handle — it would silently read wrong bytes.

## Fix (mirror #978)

Change `read_q4_header` to take `&mut std::fs::File`, scope the `BufReader` so it drops before seeking, then `file.seek(SeekFrom::Start(payload_offset))?` so the documented cursor position actually holds. Update the small number of call sites to pass `&mut file`. Add a regression test that reads the first block directly off the same handle after `read_q4_header` and asserts it equals the first written block (mutation-sensitive: fails if the seek is reverted).

This could alternatively be folded into the shared Q4 ingress/validation refactor tracked in #801/#802.

## Provenance

Found by the sibling-invocation-path grep mandated after the #978 `read_q3_header` fix.

ohdearquant:

Closes #802. Closes #736. Closes #985.
Routes native Q4/KHF1 files through shared ingress validation, rejecting invalid metadata, inconsistent geometry, and non-exact payload extents before use.
Validates direct-load shapes against model config and restores the read_q4_header cursor contract; a quantize_q4-produced Qwen3.5 0.8B embedding artifact had zero trailing bytes.

## bench-compare disposition

Measured in a quiet window at the head named below. **The head has since advanced by four
commits** (a merge of `main`, the degenerate-scale fix, a formatting change, and a GDN norm
shape fix), so the figures in this table were not re-measured at the current head. What has
been re-established at the current head is the structural claim, in the second table below;
the magnitude was never certified, as stated at the end of this section.

```
base                7ba6f096eeba8e9bba8b195bd6d8f26e7ceb27f7
head                3a2d4e4af550f6faf6864b5603ce013c812202fb
resolution          --quick
targets             lattice-inference:elementwise_cpu_bench, lattice-embed:simd
inference features  <none>
enforcement         report-only
idle samples        before base 70.3% | between phases 75.8% | after head 78.3%  (floor 70.0%)
```

Gate output: 1 FAIL, 1 confirmed improvement, 246 informational (quick-mode SIMD micro-benches below resolution, #714).

| Bench | Δ point | 95% CI | new ns | base ns |
|---|---:|---|---:|---:|
| `residual_add_layer_norm/unfused/hidden768_seq128` | +9.11% | [+7.45%, +10.82%] | 100392.4 | 92013.9 |
| `layer_norm/4096` | −6.90% | [−8.92%, −4.83%] | 970.1 | 1042.0 |

**Neither figure is attributable to this change.** Both benched functions are defined in
`crates/inference/src/forward/cpu/norm.rs`, which this PR does not modify. Established by blob
identity across the two commits rather than by absence from a diff listing, with a known-changed
file as the control:

| file | base | head | |
|---|---|---|---|
| `crates/inference/src/forward/cpu/norm.rs` | `89115f15` | `89115f15` | identical |
| `crates/inference/benches/elementwise_cpu_bench.rs` | `d5730730` | `d5730730` | identical |
| `crates/inference/src/weights/ingress.rs` (control) | `2d51a2ab` | `9b234d3d` | differs |

The benched source and the bench harness are byte-identical between the two arms, so the diff
cannot reach these kernels semantically.

Re-verified at the **current** head `96cb4bd4d5754de639205ab9e1ea9a554fd87379` against current
`main` `5d54110835e9ef8d8a2b9fbef193cd46b6b7b75e`, same method, same control:

| file | base | head | |
|---|---|---|---|
| `crates/inference/src/forward/cpu/norm.rs` | `89115f15` | `89115f15` | identical |
| `crates/inference/benches/elementwise_cpu_bench.rs` | `d5730730` | `d5730730` | identical |
| `crates/inference/src/weights/ingress.rs` (control) | `2d51a2ab` | `73fcac84` | differs |

The precise scope of that claim: it rules out a semantic regression, not a timing difference.
The changed files live in the same crate the bench links, so codegen-unit partitioning and
inlining decisions can still move the machine code generated from identical source. This is
therefore *not* the compiled-out case in which base and head bench binaries have identical
effective source — `weights/q4_weights.rs` and the other changed files are compiled in.

What is consequently **not** certified is the magnitude. No A/A control was obtained in this
session, so there is no measured noise floor for this machine at this hour to judge +9.11%
against. A previously recorded A/A run of this harness on identical source produced a +8.95%
single-group FAIL (#1137); that run was on different hardware and so does not establish this
machine's floor, but it does establish that this harness can report a FAIL of the observed
magnitude with no code change at all.

No action taken on the FAIL, and it is not being reported as a regression. If a calibrated
number is wanted before merge, it needs an A/A control in the same window followed by a
`--full` re-run of the failing group.


ohdearquant (commented):
Reviewed and approved at 96cb4bd4: per-head RMSNorm expectation for linear_attn.norm.weight now matches the safetensors loader and the checkpoint's own tensor geometry; native weight ingress validation unified across loaders with shape expectations stated at each call site. All five workflows green at this head, including the Q4 composed golden and perplexity regression gates. (Posted as comment: repository single-account topology blocks formal self-approval; the branch ruleset requires 0 approvals, so this comment is the review of record.)

PR #1033, opened 2026-07-18
refactor(scripts): canonicalize decode benchmark harness

### LINKED ISSUE #813: bench(scripts): consolidate decode-benchmark methodology into one harness with per-consumer profiles

### LINKED ISSUE #815: chore(scripts): remove decode-benchmark compatibility wrapper scripts once docs migrate

## What

The decode-benchmark harness consolidation (tracked separately) converts
`bench_apples_to_apples.sh`, `bench_apples_precise.sh`, `bench_q4_apples.sh`,
`bench_context_scaling.sh`, and `bench_compare_1k.py` into thin compatibility
wrappers that `exec` the new harness with a fixed `--profile <name>`. Those
five wrapper scripts exist so that any current caller of the legacy command
names keeps working during the migration.

This issue tracks removing those five wrapper scripts entirely, once every
current (non-historical) doc, README, ADR, and chart-generation reference
has been updated to call the canonical harness CLI directly.

## Why

Absent an identified external fleet caller of the five legacy command names,
the wrappers are pure migration scaffolding. Keeping them indefinitely
re-introduces exactly the "which name is authoritative" ambiguity this
cluster is removing. The only known guaranteed-external command names are
`bench-compare` and `perf-bench-gate.py`; the five decode-profile wrapper
scripts have no caller beyond this repository's own docs.

## Steps

1. Repository-wide search for references to the five legacy script names in
   current (non-historical) docs, READMEs, ADRs, and Make targets.
2. Update each live reference to invoke `scripts/bench_decode_harness.py
   --profile <name>` directly.
3. Confirm no live caller remains (Make targets, CI workflows, other
   scripts, current docs).
4. Delete the five wrapper scripts in one PR.

## Gates

- Zero remaining references to the legacy script names in current
  (non-historical) docs after step 2 — verified by a repository-wide
  search, not by memory.
- Historical release notes are left untouched; this issue does not rewrite
  history.
- This issue does not merge until the docs-migration step above is
  complete — it is explicitly gated on documentation, not just code.
- Rollback: revert the deletion commit; the wrappers are pure scaffolding
  with no other state to restore.

## Sequencing

- Depends on the harness-core-and-profile-migration issue landing first
  (the wrappers must exist before they can be removed).
- Independent of the CI composite-action, test-support, and lint issues in
  this cluster.

ohdearquant:

Closes #813. Closes #815.
Adds the final agentic workload profile with parameterized context, repeat, and sweep flags.
Makes the profiled harness the canonical decode-benchmark command and preserves raw observations plus existing reports.
Removes the five compatibility commands and updates all current documentation and callers.
bench-compare: N/A — scripts, tests, CI, and docs only


ohdearquant:
Reviewed the full diff at fb9bc296779b05038da19c56044edee815cb77eb. Approve. The agentic profile preserves the legacy three-shape schedule (window-major lattice, one-call-two-windows ollama, repeat-major mlx) with the exact call orders asserted in deterministic tests, and replacing the stale-external-JSON fallback with reported-missing engines removes the one way this harness could present fabricated results. A formal approving review cannot be created here because author and reviewer are the same account, so this comment is the review record.

PR #1034, opened 2026-07-18
test: harden wall-clock timing assertions

### LINKED ISSUE #977: test: harden wall-clock-threshold tests that flake under CI/concurrent load (fann test_inference_speed, embed test_simd_config_cached)
## Problem

Several tests assert a **hardcoded wall-clock bound** (`assert!(elapsed < N)`). They test timing, not correctness, so they flake under CPU load or on shared CI runners while the code under test is perfectly correct. They currently sit in `make ci` and gate merges.

## Confirmed instances (same class)

- `lattice-fann` `tests::test_inference_speed` (`crates/fann/src/lib.rs`): asserts `avg_us < 5000.0` over 1000 forward passes of a small ReLU/Softmax net.
- `lattice-embed` `tests::test_simd_config_cached` (`crates/embed/tests/integration.rs:222`): asserts 10000 `simd_config()` calls complete in `< 1000us`.

## Evidence it is load, not regression

- **fann**: failed in a full-suite run on a machine under heavy concurrent build/test load. Re-run in isolation 3x on an otherwise-idle machine: PASS all 3, avg 711-715us vs the 5000us threshold = ~7x headroom.
- **embed**: failed on the macos-latest runner at 1.87ms. The PR whose CI surfaced it changes only inference/tune and cannot reach the embed crate, so it is a flake by construction, not a regression.

## Proposed fix

Pick one per test, in preference order:
1. Gate the timing assertion behind a non-CI condition (skip/relax when `CI` is set, or `#[ignore]` by default and run only in a dedicated perf lane).
2. Widen the threshold to a value with large headroom (both fail at ~1.5-2x the bound under load; a 10x bound still catches a real O(n) regression while tolerating scheduler noise).
3. Convert to a ratio/behavioral assertion (e.g. cached call is >=Nx faster than the uncached path) instead of an absolute wall-clock number.

Absolute wall-clock numbers do not belong in a correctness gate that runs on shared/loaded machines. Criterion benches (already tracked out-of-gate via `bench-update.yml`) are the right home for absolute timing.

## Find the whole family

Grep the source for as_micros/as_millis/avg_us/elapsed comparisons under crates tests and src, and harden every hit, not just the two above.

ohdearquant:

Closes #977.

Gates absolute wall-clock thresholds behind LATTICE_TIMING_TESTS=1 so shared CI load cannot fail correctness jobs.
Retains deterministic assertions in default test runs and covers the remaining embed timing threshold found in the same test module.
bench-compare: N/A — test-only changes are compiled out of default bench builds via cfg(test) and integration-test targets


oceanwaves630 (approved):
Automated review verdict on current head: APPROVE, zero blocking findings (minors noted for follow-up).

oceanwaves630 (approved):
Reviewed and approved.

PR #1036, opened 2026-07-18
fix(inference): validate RoPE positions before dispatch

### LINKED ISSUE #773: audit(rope-position): 3 medium correctness findings — self-audit 20260709
- [ ] **Reject invalid generic-Qwen RoPE head geometry at model load**
  Generic Qwen and GPU Qwen paths nominally implement the same full 1-D RoPE geometry.
  Evidence: `crates/inference/src/model/qwen.rs:2177`. With head_dim=3 and a matching weight layout, CPU rotates only dimensions 0 and 1 at nonzero positions and leaves dimension 2 unchanged; WGPU returns InvalidInput and Metal fails setup/falls back.
  Suggested fix: Validate head_dim > 0 and even in generic Qwen config loading before table construction; leave Metal's multiple-of-four rule as extra backend validation.

- [ ] **Validate generic-Qwen CPU RoPE table capacity against token admission**
  CPU, Metal, and WGPU generic-Qwen paths nominally enforce the same RoPE position capacity.
  Evidence: `crates/inference/src/model/qwen.rs:701`. For rope_table_max_seq_len=1 and a two-token input, CPU indexes the one-row table at position 1 and panics; Metal can process the same request when gpu_max_seq_len is at least two.
  Suggested fix: Validate a single effective context at load time and reject/cap inconsistent tokenizer and CPU table capacities; preflight CPU seq_len against RopeTable::max_positions().

- [ ] **Guard Metal Qwen3.5 public raw forward position and cache capacity** (re-verified at 0699e60cc)
  Metal raw-step and prefix-suffix prefill both advance the same RoPE/KV state for an absolute position.
  Evidence: `crates/inference/src/forward/metal_qwen35.rs:9156`. At position C in a C-row session, forward_step passes cs_base=C*half_rope_dim to the MSL table and increments cache state; the checked sibling instead returns InvalidInput before dispatch.
  Suggested fix: Route public forward_step and the one-token prefix fast path through one checked position/capacity primitive returning Result.

---

ohdearquant:

Closes #773.

Rejects invalid generic-Qwen RoPE geometry and derives a shared context limit for tokenizer admission, CPU tables, and GPU capacity.
Preflights CPU sequence lengths and Metal Qwen3.5 step positions and cache rows before any out-of-range dispatch.

This round centralizes capacity validation for `verify_tokens_batched`, the batch-GEMM
verifier, and `forward_prefill_from` into one pure validator
(`validate_dispatch_capacity(start_pos, token_count, kv_capacity, gdn_pool_capacity)`),
and extends it to also preflight the GDN checkpoint-pool capacity: a batch larger than
the fixed pool is now rejected with `InvalidInput` before any dispatch or checkpoint
mutation, instead of indexing past the pool's allocated slots mid-batch.

`try_forward_step` now also validates `token_id` against `vocab_size` before delegating,
returning `InvalidInput` instead of reaching `forward_step_inner`'s embedding-lookup
`assert!` on an out-of-vocabulary token_id.

No A/B run this round. The changed code is a pre-dispatch capacity validator
(`validate_dispatch_capacity`) and its call sites in `verify_tokens_batched`,
`verify_tokens_batch_gemm`, and `forward_prefill_from` — cold-path bounds checks that run
once before dispatch, not inside any hot numeric loop. Grep confirms zero reachability
from either bench target:

```
$ grep -rn "verify_tokens_batched\|validate_dispatch_capacity\|checkpoint_gdn_to_slot" crates/inference/benches/ crates/embed/benches/
(no matches)
```

Separately, the prior round's `try_forward_step`/`forward_step` capacity checks do sit on
the decode hot path: every non-greedy decode token now pays a `Result` wrapper plus two
capacity checks (position bound, KV-cache-full) via `check_forward_step_capacity`. This is
a bounds check on a correctness-critical path (out-of-range positions previously indexed
past the RoPE table / KV cache). No bench run tonight; flagging for a future
`elementwise_cpu_bench`/decode-throughput A/B if the cost turns out to be measurable.

This round's change is a `token_id` bound check added to that same `try_forward_step`
cold single-token forward entry point (Metal-gated), plus moving one CPU-only validator
test to a default-feature module — neither touches `elementwise_cpu_bench` (inference
elementwise kernels) or the embed SIMD dot-product bench. Grep confirms zero
reachability from either bench source:

```
$ grep -n "try_forward_step\|forward_step\b\|validate_dispatch_capacity" \
    crates/inference/benches/elementwise_cpu_bench.rs \
    crates/embed/benches/simd_opt_bench.rs \
    crates/embed/benches/simd_bench.rs
(no matches)
```

Base and head bench binaries have identical measured-path source for both bench
targets; no A/B run needed (structural unreachability).

bench-compare: A/B deferred to maintainer (quiet-window run on final head)

### DIFF
--- a/crates/inference/src/forward/metal_qwen35.rs
+++ b/crates/inference/src/forward/metal_qwen35.rs
@@ -4753,6 +4753,7 @@ mod inner {
             tokens: &[u32],
             start_pos: usize,
         ) -> Result<MetalVerifyOutput, crate::error::InferenceError> {
+            self.check_forward_range_capacity(start_pos, tokens.len())?;
             if self.session.gdn_checkpoints.is_none() {
                 return Err(crate::error::InferenceError::Inference(
                     "GDN checkpoint pool required for verify_tokens_batched".into(),
@@ -6703,6 +6704,33 @@ mod inner {
                 return Err(InferenceError::InvalidInput(format!(
                     "forward_step: KV cache is full at {} tokens (max_cache_len {max_cache_len})",
                     self.session.kv_cache.seq_len
+                )));
+            }
+            Ok(())
+        }
+
+        /// Same capacity check as [`Self::check_forward_step_capacity`], generalized to a
+        /// contiguous run of `token_count` positions starting at `start_pos` — every position
+        /// `start_pos..start_pos + token_count` must land inside the RoPE table / KV cache
+        /// before any of them reach Metal dispatch. Shared by every caller that forwards a
+        /// caller-supplied `start_pos` across multiple tokens in one call (batched verify).
+        fn check_forward_range_capacity(
+            &self,
+            start_pos: usize,
+            token_count: usize,
+        ) -> Result<(), crate::error::InferenceError> {
+            use crate::error::InferenceError;
+
+            if token_count == 0 {
+                return Ok(());
+            }
+            let max_cache_len = self.session.kv_cache.max_cache_len;
+            let last_position = start_pos.saturating_add(token_count - 1);
+            if last_position >= max_cache_len {
+                return Err(InferenceError::InvalidInput(format!(
+                    "verify_tokens: start_pos {start_pos} + {token_count} tokens reaches position \
+                     {last_position}, exceeding max position {}",
+                    max_cache_len - 1
                 )));
             }
             Ok(())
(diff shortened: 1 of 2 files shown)


ohdearquant (commented):

ohdearquant:
Reviewed the full diff at c4a0b6b6324f2816c856e95e25cdca8d832b655c. Approve. Centralizing the capacity precondition into a pure, CPU-testable validator, converting the batch-GEMM assert into a typed error, and asserting that rejection leaves kv_cache.seq_len unmutated cover the gap class properly; the scope paragraph on check_forward_range_capacity stating precisely which decode loops are bounded by their callers instead is exactly what the next reader needs. The deferred GPU-run test is disclosed in-file with its CPU-only counterpart proving the same guard logic. A formal approving review cannot be created here because author and reviewer are the same account, so this comment is the review record.

PR #1037, opened 2026-07-18
fix(inference): enforce reasoning budget close marker

ohdearquant:
## Summary

Enforces the checkpoint mmap trust-boundary gate used by every zero-copy weight loader in
`lattice-inference` (Q4, Q3, MoE f16/f32, expert-cache), and fixes the reasoning-budget close
marker so it can't be silently bypassed or ignored.

**mmap trust boundary** -- `open_trusted_mmap_file` is the single chokepoint all five loaders
route through:

- Opens with `O_NONBLOCK | O_NOFOLLOW`: a planted FIFO/device or final-component symlink is
  rejected before any trust or shape check runs.
- Rejects group/other-writable files and files owned by a uid that is neither the process uid nor
  root (root-owned, owner-read-only shared model directories remain supported).
- macOS: rejects any extended-ACL entry granting write directly or transitively (including
  `ACL_WRITE_SECURITY`/`ACL_CHANGE_OWNER`), classified against the full `acl_perm_t` enum.
- Walks the parent-directory chain and refuses a group/other-writable ancestor (unless sticky, the
  `/tmp` pattern) or one owned by a foreign non-root uid, since `unlink`/`rename` are gated on the
  directory's permission, not the file's. On macOS, each ancestor's own extended ACL is now
  checked too, against the full directory-mutating permission set (add/delete/rename an entry,
  retarget the directory's own metadata, or the transitive write-security/change-owner rights): a
  0700, process-owned ancestor can still carry an ACE granting one of these to a foreign
  principal, invisible to POSIX mode/uid alone. Any grant of these permissions to any principal is
  rejected conservatively; qualifiers are not resolved to a uid.
- Re-fstats the mapped file after `mmap()` and rejects a size/ino/dev/mtime mismatch against the
  pre-map stat, catching a truncate-or-replace race in the validate-then-map window.

All checks resolve against the already-open fd, never a re-resolved path. Residual risk (a
principal already holding a writable fd before this gate's hardening took effect, plus same-UID
mutation through a second fd) is an accepted trust-root assumption for any zero-copy `MAP_SHARED`
mmap loader.

**Reasoning-budget close marker** -- `generate`/`generate_streaming` (CPU and Metal) now resolve
and validate the `</think>` close-marker token before any early-return path (EOS, grammar
completion, cancellation) can short-circuit generation, rejecting an unresolvable or out-of-range
marker with a typed `InvalidInput` error instead of silently no-op'ing the budget. The marker id
is no longer folded into the cross-turn prefix-cache identity when no budget is active, so
toggling the budget across turns no longer forces a full re-prefill.

## Test plan

- `cargo test -p lattice-inference --lib weights::mmap_trust` -- 18 tests passing, CPU-only
  (parent-dir walk incl. ancestor ACL, mode/uid/ACL, FIFO/symlink/directory rejection, post-map
  fstat recheck; real tempfiles and `chmod +a`-set ACLs).
- `cargo clippy -p lattice-inference --all-targets --features f16,metal-gpu -- -D warnings`: clean.
  `cargo fmt -p lattice-inference` applied.
- Metal-gated generation tests compile under `--features f16,metal-gpu`; not run here (GPU
  unavailable in this environment).

## bench-compare disposition

No hot-path call sites touched: the trust gate runs once per checkpoint open (load time, not
per-token), and close-marker resolution happens once per generation call, not per decode step.
A/B in a quiet window pending before ready.

- Pull request head evaluated: `b3fdf0e3526fa596009a5c3fce4ab427891c508b`

re-derived.


ohdearquant:
Closing as superseded; both halves of this PR have been overtaken.

The mmap trust-boundary module landed on main independently — `crates/inference/src/weights/mmap_trust.rs` is present and tested. What did not land is this PR's wiring of it into every zero-copy loader: on main it has a single caller, in the Q4 path. That remaining gap is now tracked in its own right at #1368, with the three bypassing sites named.

The reasoning-budget half is superseded by #1208, which resolves the close marker through `token_ids_for_content` — checking the special-token table, the base vocabulary, and the `special:false` added-token tier — rather than the special-token-only lookup this branch used. The difference is observable: a `</think>` spelled only in the base vocabulary resolves under #1208 and fails under this branch's approach.

Nothing here is lost. The branch stays available if any of its test fixtures are worth lifting into the #1368 work.

PR #1038, opened 2026-07-18
refactor(tune): unify layer-23 training and model-depth ranges

### LINKED ISSUE #730: tune: generalize hardcoded TOP_LAYER=23 in micro-LoRA trainer to cover full layer stack (blocks 27B recovery)
> _Issue authored by Claude (Anthropic agent) on behalf of @oceanwaves630._

## Problem

The CPU micro-LoRA trainer (`lattice-tune`) hardcodes the top trainable layer index:

```rust
// crates/tune/src/lora/train_core.rs:13
pub const TOP_LAYER: usize = 23;
```

`train_micro_lora` builds its trainable stack by iterating `first_layer ..= TOP_LAYER`
(`crates/tune/src/lora/train.rs:251`) and indexing `model.weights.layers[idx]` directly.
`validate_micro_lora_inputs` guards this with `num_hidden_layers <= TOP_LAYER → reject`
(`train.rs:131`) — i.e. it enforces `num_hidden_layers > TOP_LAYER`.

`23` is the top index of the **24-layer 0.8B** stack (`num_hidden_layers = 24`,
`crates/inference/src/model/qwen35_config.rs:213`). It is not the top of any larger model:

- **27B** = **64 layers** (`qwen35_config.rs:303`)
- (intermediate config = 40 layers, `:261`)

## Consequence

On the 64-layer 27B the minimum-layers guard **passes** (`64 > 23`), so the trainer does not
reject the model — but it then iterates only `first_layer ..= 23` and can never reach layers
`24..=63`. The hardcoded cap silently confines training to the lower-mid of the stack, so a
"top-layer" micro-LoRA recovery finetune on the 27B trains the wrong layers. Generalizing
`TOP_LAYER` to track `num_hidden_layers` (e.g. `num_hidden_layers - 1`, or a configurable
`[first_layer, last_layer]` range validated against the model) is a prerequisite for any 27B
recovery/KD work.

## Why now

Surfaced by **ADR-077** (pruning + distillation priority) as the **P3** prerequisite for its
gated KL-recovery path (#499): you cannot recover a pruned 27B by top-layer LoRA finetune while
the trainable range is pinned to the 0.8B stack. ADR-077 records this blocker; this issue tracks
the fix so the ADR does not leave an untracked prerequisite behind.

## Acceptance criteria

- [ ] `TOP_LAYER` (or an equivalent trainable-range bound) is derived from the model's
      `num_hidden_layers` rather than a module constant, so the trainer targets the actual top
      layer(s) of whatever model it is given.
- [ ] The existing mutation-sensitive guard behavior is preserved: a `first_layer` past the top,
      or a model with too few layers for the requested range, is still rejected before any layer
      access (see the mutation-sensitivity note at `train.rs:578-585`).
- [ ] A unit test covers a >24-layer config (e.g. 40 or 64 layers) training its true top layer.
- [ ] No change to 0.8B behavior (regression-guard the existing 24-layer path).

## Pointers (origin/main)

- `crates/tune/src/lora/train_core.rs:13` — `pub const TOP_LAYER: usize = 23;`
- `crates/tune/src/lora/train.rs:119-135` — validation guard
- `crates/tune/src/lora/train.rs:248-251` — trainable-stack iteration
- `crates/tune/src/lora/train.rs:578-585` — mutation-sensitivity note
- `crates/inference/src/model/qwen35_config.rs:213` (0.8B=24), `:303` (27B=64)

### LINKED ISSUE #846: test(tune): layer-23 golden bridge — old vs. shared full-driver equivalence, gate must run

## What

Before deleting `crates/tune/src/bin/train_grad_layer23.rs`'s standalone 744-line
implementation (removal is the next unit in this series), produce and land an
**executing** comparison between the old trainer and the shared full-driver path at
`first_layer = 23`, on the same model/data, fixed seed, legacy A-init amplitude
`0.02`, held-out validation disabled, one Adam step.

Compare and assert on:

- TBV (cached/assembled NLL vs. `Qwen35Model::compute_token_nlls`) for both paths.
- Step-0 and final NLL for both paths.
- Every LoRA A/B tensor's name, shape, and value between the two paths (document an
  explicit tolerance if bit-exactness is not achievable — do not silently accept
  "close enough" without stating the number).
- Serialized adapter reload: save from each path, reload, and confirm the reloaded
  adapter matches what was saved.
- A no-checkpoint config test proving the eventual compatibility shim (next unit in
  this series) maps exactly to `first_layer = 23` and the legacy defaults listed in
  the harness unit of this series.

## Why

Verified in this series' design review: `train_grad_full --first-layer 23` and
`train_grad_layer23` are **not** proven output-equivalent. Both build the same
layer-23 GQA+FFN+head computation and export the same layer-23 `q_proj`/`v_proj`
shapes (functional coverage), but `train_grad_layer23` initializes LoRA A with fixed
amplitude `0.02` while `train_grad_full` uses `1/sqrt(hidden)`, and the full trainer
additionally defaults `--max-valid` to 16, applies a 2 GiB logits guard, supports
gradcheck flags, and rejects `--log-every 0` — none of which the old trainer has.
Deleting the old implementation without an executed, evidenced equivalence
comparison would delete the only proof that the replacement is trustworthy for the
scripts and saved-adapter consumers that currently depend on the old command.

## Maintainer disposition (binding acceptance criterion)

**This gate must actually execute as part of merge acceptance for the deletion PR
in this series — it is not satisfied by a written checklist, an `#[ignore]`d test,
or a PR-description narrative claiming the comparison was run manually.** Concretely:

- If a CI runner in this repository already has legal, reliable access to the
  checkpoint/data fixture needed to run the comparison without a runtime download,
  wire this comparison into that runner as a required check on the deletion PR.
- If no existing required runner can legally/reliably run it, this issue must ship
  an explicit artifact-backed macOS job (or documented local-run artifact with a
  reproducible command and recorded output) that a maintainer can point to as
  having executed for the specific commit being merged — not a test left ignored
  in the tree with a TODO to run it later.
- A comparison that is only described in prose, or only asserted to have been run
  once outside of any tracked job, does not satisfy this issue.

This directly resolves open question 2 from this series' design ("golden
environment... If no required runner has the model, the merge gate needs an
explicit artifact-backed macOS job rather than an ignored test") as a hard
requirement, not an open choice.

## Checklist

- [ ] Identify (or provision) the checkpoint/data fixture and runner this
      comparison will execute against; record how it is invoked.
- [ ] Implement the old-vs-shared-driver comparison harness per the "Compare and
      assert on" list above.
- [ ] Wire the comparison into a CI job (required check on the deletion PR) or an
      artifact-backed job with recorded, reproducible output, per Maintainer
      disposition above.
- [ ] Add the no-checkpoint config test proving the future compatibility shim's
      config maps exactly to `first_layer = 23` plus the legacy defaults.
- [ ] Document the tensor-value comparison tolerance explicitly if bit-exactness is
      not achieved (name the tolerance and why).

## Gates

- Central difference `max_rel < 1e-2` on the shared-driver path at `first_layer = 23`.
- A deliberate gradient-array mutation (perturb one A/B tensor) must make the
  comparison fail — this proves the comparison is non-vacuous, not merely
  structurally present.
- Real-model TBV passes for both paths being compared.
- The comparison job/check must show as executed (green or explicitly reported)
  for the commit under review, per Maintainer disposition above.

## Sequencing

Third implementation PR in this series; depends on the `TrainCtx` refactor and the
shared harness/TBV unit having landed (the shared driver this compares against is
the post-harness-migration `train_grad_full` path). Must land and its gate must be
green before the deletion/compatibility-shim unit in this series proceeds.

### LINKED ISSUE #847: refactor(tune): delete 744-line layer-23 trainer, keep a thin deprecation shim

## What

Delete the hand-inlined implementation in `crates/tune/src/bin/train_grad_layer23.rs`
(744 lines on current `main`) and replace the existing Cargo binary target
`train_grad_layer23` with a small compatibility main that calls the shared
full-driver path from this series' harness unit.

Concretely:

1. Move the full trainer's orchestration behind a binary-private
   `full_driver::run(FullDriverConfig)` inside the shared `train_common` module
   (from the harness unit in this series). Keep `train_grad_full.rs` as a thin
   `main` that parses its existing contract and calls the driver.
2. Point the Cargo target name `train_grad_layer23` (currently
   `path = "src/bin/train_grad_layer23.rs"` in `crates/tune/Cargo.toml`) at a new
   small compatibility main under `src/bin/train_grad_layer23/main.rs`.
3. The compatibility main parses only the old flag surface (`--model-dir`,
   `--data-dir`, `--steps`, `--lr`, `--rank`, `--alpha`, `--seq-len`, `--max-train`,
   `--log-every`, `--save`, `-h/--help` — no `--first-layer`, `--max-valid`,
   `--gradcheck`, `--probe`, or `--fd-eps`), fixes `first_layer = 23`, disables
   held-out evaluation, selects the legacy A-init amplitude `0.02`, and calls the
   same full driver.
4. The compatibility main emits one deprecation line naming
   `train_grad_full --first-layer 23` as the supported replacement.

This distinguishes two guarantees deliberately:

- The compatibility target preserves existing scripts, defaults, initialization,
  and saved-adapter schema for a deprecation window.
- New invocations of `train_grad_full --first-layer 23` use full-trainer semantics
  (its current initialization, validation, guards, output) — functionally
  equivalent per the golden-bridge comparison in this series, but not promised
  bit-identical to the legacy command.

Do not remove the compatibility target in this PR. Its removal is a separately
announced, explicitly versioned breaking-CLI-change PR, at least one release
window after this one, with a negative test or documented error in packaging/
release notes at that time — not part of this issue's scope.

## Why

The 744-line hand-inlined trainer is genuine source duplication that has already
drifted: it has no held-out validation path, no logits guard (`train_grad_full` has
a 2 GiB guard), no gradcheck flags, and would panic on modulo-by-zero where
`train_grad_full` rejects `--log-every 0` outright. Every future correctness fix to
the shared tape (including the GDN LoRA surface merged in #792) has to be
separately ported into this file or it silently drifts further, exactly as issue
#737 already flagged for the pre-GDN state. Deleting the Cargo target outright
instead of shimming it would break any script that invokes the binary by name,
since the golden-bridge comparison in this series proved functional-but-not-
bit-identical equivalence, not a drop-in replacement.

## Non-goals (explicit fences)

- No old math survives inside the crate after this PR — all orchestration for
  `first_layer = 23` now runs through the shared full driver.
- Do not claim legacy trajectory parity beyond what the golden-bridge comparison
  in this series actually measured and recorded.
- Do not remove the `train_grad_layer23` Cargo target name in this PR.

## Checklist

- [ ] Extract `full_driver::run(FullDriverConfig)` into the shared `train_common`
      module; `train_grad_full.rs` becomes a thin `main`.
- [ ] Delete `crates/tune/src/bin/train_grad_layer23.rs`'s standalone implementation.
- [ ] Add `crates/tune/src/bin/train_grad_layer23/main.rs` as the compatibility
      shim per the flag surface above; update `crates/tune/Cargo.toml`'s `[[bin]]`
      path accordingly.
- [ ] Emit the one-line deprecation notice naming `train_grad_full --first-layer 23`.
- [ ] Re-run the golden-bridge comparison (this series' prior unit) against the
      new shim to confirm it still passes with the shim in place, not just the
      shared driver in isolation.
- [ ] Re-run the no-checkpoint config test proving the shim's derived config
      matches `first_layer = 23` and the legacy defaults exactly.

## Gates

- Repeat the golden-bridge unit's comparison (TBV, step-0/final NLL, per-tensor
  name/shape/value, serialized adapter reload) against the compatibility shim
  specifically, using the same execution requirement (real CI job or recorded
  artifact-backed run, not a documentation claim).
- Central difference `max_rel < 1e-2` plus a deliberate gradient-array mutation
  proving the comparison is non-vacuous.
- Feature-gated bin checks (`--features train-backward,safetensors`, or the
  merged repository's equivalent feature composition) build and run clean for
  both `train_grad_full` and the `train_grad_layer23` compatibility shim.
- Adapter roundtrip (save/reload) test passes for the shim's output.

## Sequencing

Fourth implementation PR in this series; depends on the harness/TBV unit and the
golden-bridge gate unit both having landed and passed. The compatibility target's
eventual full removal is explicitly out of scope here and belongs to a future,
separately announced breaking-change PR.

ohdearquant:
## Summary

Unifies layer-23 LoRA training onto a shared multi-layer driver, generalizing it to arbitrary model depths, while keeping the legacy CLI working through a thin compatibility entry point.

- `MicroLoraConfig`'s trainable-layer window may end before the model's true top layer, but the tape always materializes and runs the full frozen suffix above it, so forward/loss/gradients match the deployed full-depth model.
- The shared driver validates model depth, LoRA rank, step count, `--log-every`, `--max-train`, `--max-valid`, and `--seq-len-cap` before any layer access, tokenization, or cache allocation.
- The frozen-prefix attention cache (`h_in` + RoPE `cos`/`sin`) is bounded on the product of raw token count and per-token buffer size, not on sample count and sequence length independently, checked before allocation for both cache sets.
- The per-sample training tape `forward_full` builds is bounded separately: GQA attention retains one causal `softmax_probs` row per query position per head, quadratic in `seq_len`, so a `--seq-len-cap` clearing the linear cache bound could still exhaust memory once training starts. A new check derives the actual per-token tape footprint (GQA attention buffers + GDN recurrent-state buffers) from model dimensions and rejects the combination before any tape allocation.
- Training and held-out validation logits buffers are each capped at 2 GiB before allocation.
- Config-vs-loaded-weights desync (`num_hidden_layers` mutated independently of the loaded weights) is rejected before any layer indexing.
- The terminal layer's FFN forward/backward only runs over completion positions, since prompt-position output there is never read.
- Every job in `e2e-parity.yml` fetching the Qwen3.5-0.8B checkpoint shares one composite action (`.github/actions/provision-qwen35-0-8b`), pinned + hash-verified. Both engine/tune change-detection filters now watch that action's path, plus the ppl-gate self-test script and the embed-drift test/fixture, closing gaps where those changes wouldn't have triggered their required job.
- Adds an artifact-backed one-step golden-output gate for the layer-23 compatibility path.

Closes #846. Closes #847. Closes #730.

## Test plan

- `cargo test -p lattice-tune --features train-backward --lib` — 317 passed. Covers full-depth materialization past a bounded trainable window, all CLI-bound rejections, shallow-model rejection, the aggregate cache-budget rejection, the quadratic tape-budget rejection (passes the linear cache bound, exceeds the derived tape footprint), and terminal-layer FFN position-skip — each verified to fail when its underlying fix is reverted.
- `cargo clippy -p lattice-tune --features train-backward --all-targets -- -D warnings` — clean.
- `cargo fmt -p lattice-tune`.
- `actionlint .github/workflows/e2e-parity.yml` — clean.
- Filter coverage confirmed by reading every path filter against each required job's actual dependencies (composite action, invoked scripts, test fixtures); CI wiring only, not unit-testable.
- GPU/Metal paths not exercised — all changes are training/driver/CI-config, entirely CPU-path.

## bench-compare disposition

No hot-path call sites — `lattice-tune` trainers are not built into the default Criterion bench binaries. A/B in a quiet window pending before ready.


oceanwaves630 (approved):

PR #1039, opened 2026-07-18
fix(ci): broaden npm native prebuild matrix

### LINKED ISSUE #696: npm: broaden @khive-ai/lattice-embed native prebuilt matrix beyond darwin-arm64
Follow-on from the v0.5.1 npm release (#689, #694).

`@khive-ai/lattice-embed` (native napi-rs) declares a 7-platform `optionalDependencies` matrix in `binding.js`, but the initial release publishes **only** the `darwin-arm64` prebuilt binary. On any other platform `binding.js` throws `FL_EMBED_NATIVE_LOAD_FAILED` (there is no automatic WASM fallback) and users must fall back to `@khive-ai/lattice-embed-wasm`.

**Scope**: add CI cross-compilation + prebuild + publish for the remaining platform subpackages:
- `linux-x64-gnu`, `linux-arm64-gnu`
- `linux-x64-musl`, `linux-arm64-musl`
- `darwin-x64`
- `win32-x64-msvc`

Each becomes a `@khive-ai/lattice-embed-<platform>` subpackage matching the existing `darwin-arm64` layout, wired into `make publish-npm`'s preflight + publish loop (already discovers any present `.node`, so the script needs no change — just the built artifacts).

Until then the coverage limitation is documented plainly in `docs/releases/v0.5.1.md`.

ohdearquant:

Closes #696.

Adds a seven-target napi-rs matrix covering macOS arm64/x64, Linux x64/arm64 on glibc and musl, and Windows x64.
The collector generates each platform subpackage, fails on missing binaries, dry-packs every package, and uploads the assembled prebuilds.
Manual dispatches reuse the complete artifact set for the existing npm dry-run or OIDC-enabled publish path.
bench-compare: N/A — workflow and packaging-tooling change only (`.github/workflows/npm-prebuild.yml`, `scripts/publish-npm.sh`, packlist assertion script); no crate code changes.

## The packlist guard's README exemption is a path-segment matcher

`npm/lattice-embed-native/scripts/assert-platform-packlist.mjs` asserts that a
platform subpackage's tarball contains nothing beyond its expected file set,
with a bounded exemption for the package's own root README. npm's rule for that
file is the glob `!/readme{,.*[^~$]}` (npm-packlist 11.8.0), matched
case-insensitively against a rooted path, and the guard now mirrors it as
`/^README(\.[^/]*[^/~$])?$/i`.

The equivalence was established by compiling npm's actual glob with the
installed minimatch and comparing accept/reject behaviour against the candidate
on the same fixture strings, rather than by reading the glob source and
reasoning about it. That matters because a dot-star regex differs from the glob
in two directions at once: `.` in a JavaScript regex crosses `/` but not a line
terminator, while the glob's `*` does the opposite. The guard accepts a bare
`README` or `README.<suffix>` where the suffix is any run of non-`/` characters
not ending in `~` or `$`, and rejects both a descendant of a directory named
like a README and a `README`-prefixed name with no `.` separator.

## The publish preflight fails closed on lookup errors

`check_version_available` in `scripts/publish-npm.sh` previously treated every
nonzero `npm view` status as "this version is available", so a DNS failure, a
registry 5xx, or an auth failure all produced `Preflight OK: no version
collisions.` without availability having been established. It now reads npm's
JSON error body and takes the available branch only on an explicit not-found
(`E404`); any other failure prints a diagnostic naming the ambiguity and exits
nonzero. A successful lookup still means the version is already published and
still exits nonzero with the immutable-version message.

That discrimination was written before it could run. `publish-npm.sh` sets
`set -e`, and the lookup was captured in a standalone assignment, so a nonzero
`npm view` terminated the script at that line — before the status was recorded,
before the JSON was parsed, and before the E404 branch was reached. Since E404 is
the only result that means "free to publish", every genuinely unpublished version
killed the real preflight, while the parser below it was correct and unreachable.
The lookup now sits in a condition that handles nonzero status, so errexit does
not apply to it.

The self-test did not catch this because it ran the extracted function under
`bash` without `-e` — a shell environment the release path never uses. It now
runs the function under `/bin/sh -c "set -e; ..."`, matching the shebang and
options `publish-npm.sh` actually runs with, and covers seven dispositions:
not-found, network error, successful lookup, empty body, invalid JSON, missing
`error` key, and missing `error.code`. Every one but the first two is new, and all
five failure shapes must fail closed. The suite reports `14 passed, 0 failed` and
is now run by a dedicated `publish-npm-preflight-selftest` job, with both script
paths in the workflow's triggers so editing either one runs it.


ohdearquant:
Two blockers found on the release path, both verified against live state rather than from the diff alone.

**1. The publish cannot run at the versions checked in here.** `@khive-ai/lattice-embed` and `@khive-ai/lattice-embed-darwin-arm64` each have exactly one version published on the registry, `0.1.0`, and this branch declares the native main package and all seven `optionalDependencies` at `0.1.0`. npm name@version tuples are immutable, so the preflight this PR adds finds the existing tuples and exits before publishing anything — including the six new platform packages the PR exists to ship. Verified with a direct registry read (`registry.npmjs.org`, with a known-published control package in the same loop so a refused request could not be mistaken for an absent version).

The preflight is doing its job; the problem is the version plan. Before this can land, the native main, the platform packages, and the optional-dependency references need to move to unused versions in lockstep, with the lockfile and release metadata updated to match, and a dry run against those new versions.

**2. The complete-matrix guard fails open on an empty matrix.** `EXPECTED_PLATFORMS` is derived from the main package's `optionalDependencies` and then consumed by a `for` loop, with no assertion that the derived set is non-empty. Measured under `/bin/sh` with `set -e`, matching the script's own interpreter:

| `optionalDependencies` | derived set | loop iterations | rc | outcome |
|---|---|---|---|---|
| two platforms (control) | two entries | 2 | 0 | guard runs |
| `{}` | empty | **0** | **0** | **reaches the publish stage** |
| key absent | — | — | 1 | aborts, fails closed |

So the two arms behave differently and only one of them is safe. An absent key makes `Object.keys(undefined)` throw, which `errexit` turns into an abort. An empty object is valid JSON that yields an empty string, and a `for` loop over nothing succeeds, so the script proceeds to publish WASM and the native main with no platform binaries at all. The comment above that block states the intended property ("a partial local checkout must fail closed here"), which holds for a partial set but not for an empty one.

The same gap covers stale or misnamed `.node` files: the per-platform check is an existence glob, so a directory containing any `*.node` passes even when the filename the subpackage's `files`/`main` metadata names is absent. The stronger assertions exist but only run in the workflow, so `make publish-npm` invoked directly does not get them.

Fix for both: assert a non-empty expected matrix in the script itself, require the exact expected `.node` path per platform rather than any glob match, and run the platform packlist check for every platform before the version check. Keeping the workflow assertions as well is right; they just should not be the only copy.

The selftest added here is sound and does redden when the parsing logic is undone (14 arms, observed passing). This is about the two paths it does not cover.

ohdearquant:
Superseded by #1370, which carries every change on this branch plus the empty-matrix and exact-binary-name guards on the same publish path. The `Closes #696` reference has moved there, so closing this one leaves nothing orphaned.

The npm version sequencing tracked in #1371 is the remaining blocker for the publish lane either way.

PR #1040, opened 2026-07-18
fix(tune): gate simulated distillation labels

### LINKED ISSUE #11: tune: DistillationPipeline returns simulated labels, HTTP client not wired
## Problem

`DistillationPipeline` in `crates/tune/src/distill/` has an incomplete teacher integration:

1. **`label_single` returns simulated labels** — the reqwest HTTP client for calling external teacher models is not wired
2. The pipeline structure (TeacherConfig, batch labeling, dataset creation) is complete, but the actual teacher API call is a placeholder

Surfaced during KG ingestion of the tune crate codebase.

## Impact

Knowledge distillation cannot run end-to-end against a real teacher model. The data pipeline and config layer work, but the core labeling step is synthetic.

## Suggested Fix

Wire the reqwest client to actually call the teacher endpoint, or support a local teacher via the inference crate (Qwen3.5 as teacher for a smaller student).

ohdearquant:
Closes #11.

## Summary

Gates simulated distillation labels so they can never be mistaken for real teacher output:

- `DistillationPipeline::label_single`/`label_batch` fail closed with `TuneError::TeacherApi` until a live teacher transport is configured, instead of returning fabricated labels.
- Deterministic fixed-label output is available only behind the non-default `simulated-teacher` feature, through explicitly named methods (`label_single_simulated`, `label_batch_simulated`).
- `LabelingResult` tracks provenance (`LabelSource::Teacher` / `Simulated`) as a private field, readable only via `source()`. It implements `Serialize` but not `Deserialize`, so there is no path from untrusted bytes back to a `LabelingResult`.
- `LabelingResult::success` (the `Teacher`-provenance constructor) now requires a `TeacherAuth` capability token that has no public constructor — only code inside `lattice-tune` can mint one. External callers can no longer forge a `Teacher`-sourced result through the public API, closing a training-data-poisoning gap where arbitrary labels could be attributed to the configured teacher.
- `DistillationPipeline::to_training_examples` rejects `Simulated`-sourced results and right-sizes its output allocation to the number of eligible (kept) results rather than the full input count, so a large all-skipped batch doesn't hold unused reserved memory.

**API note:** this is a semver-major change to `lattice-tune`. External code constructing `LabelingResult` via a struct literal, reading/writing `.source` directly, or calling `LabelingResult::success` needs to switch to the constructors and the new `source()` accessor; `success` additionally now requires an in-crate `TeacherAuth` token. Version bump and release handling are left to the maintainer.

## Test plan

- `cargo test -p lattice-tune --features serde,simulated-teacher` — 195 unit + 7/50 integration + 3 doc tests, all passing.
- `cargo clippy -p lattice-tune --all-targets --features serde,simulated-teacher -- -D warnings` — clean.
- `cargo fmt -p lattice-tune -- --check` — clean.
- Mutation-tested: reverting the `TeacherAuth` gate on `success` makes the corresponding `compile_fail` doctest compile (and thus fail); reverting the allocation right-sizing makes `to_training_examples_does_not_over_reserve_for_an_all_skipped_batch` fail on the capacity assertion. Both verified locally, then restored.

This PR only touches `crates/tune`, which is not a dependency of `lattice-inference` or `lattice-embed` (the two ADR-058 bench-target crates). No changed symbol (`LabelingResult`, `LabelSource`, `TeacherAuth`, `to_training_examples`) appears in either bench source, and neither crate's `Cargo.toml` depends on `lattice-tune`. Base and head bench binaries have identical effective source for this change — structural unreachability, not a cold path. No A/B run performed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -173,6 +173,12 @@ jobs:
       # buffer bounds). Build AND run them here so the feature cannot rot silently.
       - name: tune — mixture (online router-refit, build + run tests)
         run: cargo test -p lattice-tune --features mixture --lib
+      # simulated-teacher gates the deterministic fixed-label test/example path
+      # (label_single_simulated, label_batch_simulated). Off by default, so the
+      # required `ci` job never compiles or runs it. Build AND run here so the
+      # opt-in API cannot rot silently.
+      - name: tune — simulated-teacher (build + run tests)
+        run: cargo test -p lattice-tune --features simulated-teacher --lib
       - name: inference — f16 + train-backward
         run: cargo build -p lattice-inference --features f16,train-backward
       # wgpu-gpu is excluded from default builds and was silently broken under
--- a/crates/tune/Cargo.toml
+++ b/crates/tune/Cargo.toml
@@ -28,6 +28,8 @@ sqlite = ["dep:rusqlite", "serde"]
 # Online adapter-selector refit from preference feedback (experimental, bench-gated).
 # Pulls in the lattice-fann training primitives (RLOO / EWC++) for gate updates.
 mixture = ["lattice-fann/online-router"]
+# Fixed labels for tests and examples. Never enables the default labeling path.
+simulated-teacher = []
 
 [dependencies]
 lattice-fann = { version = "0.7.0", path = "../fann" }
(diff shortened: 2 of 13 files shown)


oceanwaves630 (approved):
Reviewed and approved.

PR #1044, opened 2026-07-18
fix(inference): unify safetensors reader validation

### LINKED ISSUE #22: audit runtime SafetensorsFile validation against QuarotTensorReader strict contract
> _Issue authored by Claude Opus 4.7 (Anthropic) on behalf of @ohdearquant._

Five rounds of review on #20 hardened \`quant::quarot::io::QuarotTensorReader\` against malformed SafeTensors checkpoints. The runtime parser at \`crates/inference/src/weights/f32_weights.rs::SafetensorsFile\` has not had the same audit, and a temporary spot-check confirms it accepts files the converter now rejects.

## Validation delta

After #20, \`QuarotTensorReader::Shard::open\` enforces the same contract as the official \`safetensors\` Rust crate:

| Check | QuarotTensorReader (post-#20) | Runtime SafetensorsFile |
|---|---|---|
| Reject aliased data_offsets (two tensors at same byte range) | ✅ \`non-contiguous\` error | ❌ accepts (verified by review round 3 with a hand-crafted file) |
| Reject leading hole (first tensor starts > 0) | ✅ | ❌ unverified, expected to accept |
| Reject internal hole between tensors | ✅ | ❌ unverified, expected to accept |
| Reject trailing payload (last tensor ends < data_len) | ✅ | ❌ unverified, expected to accept |
| Validate byte length against \`numel * dtype_bitsize\` for every standard dtype | ✅ (incl. F4, F6_E2M3, F6_E3M2, F8 family, C64) | ❓ unverified |
| Reject sub-byte dtype with non-byte-aligned shape product | ✅ \`not byte-aligned\` error | ❓ unverified |
| Reject unrecognized dtype strings | ✅ rejects outright | ❌ \`parse_tensor_meta\` returns \`Ok(None)\` at \`f32_weights.rs:1685\` and skips the entry (silent) |

The runtime parser predates the official crate's strict mode. Its permissive behavior is documented in the QuaRot reader's rationale comment (\`crates/inference/src/quant/quarot/io.rs::safetensors_bits_per_elem\`) but the runtime itself has no test asserting the divergence is intentional.

## Why this matters

The runtime serves live inference traffic via \`Qwen35Model::from_safetensors\` and \`QwenModel::from_directory\`. The offline converter is a one-shot CLI; the runtime is the hot path. A corrupted or maliciously crafted HuggingFace checkpoint that the converter would reject still loads and serves through the runtime path. Possible consequences:

- Two tensors pointing at the same byte range alias each other — the runtime would read the wrong weight for one of the names (silent wrong logits, not a crash).
- A trailing-payload tensor at offset \`[0, 4)\` with \`data_len=8\` would still load fine, but a future safetensors-conformant tool would reject the same file — we'd be silently producing checkpoints that are non-portable.
- Unknown-dtype tensors are silently skipped at load time; if a required tensor's dtype ever shifts in a new SafeTensors revision, the model loads partially and inference proceeds with whatever zero-init the slot gets.

## Scope

Two paths, decide by measurement / discussion:

1. **Align the runtime to strict.** Lift the validation logic out of \`QuarotTensorReader::Shard::open\` into a shared helper (e.g. \`weights::f32_weights::validate_safetensors_layout\`) and call it from \`SafetensorsFile::open\` / \`ShardedSafetensors::open_index\`. Adds CPU cost at model load (parses every header entry, sorts offsets); negligible for a one-time load on a multi-GB checkpoint.

2. **Document permissive as intentional.** Add a code comment to \`SafetensorsFile::parse_tensor_meta\` spelling out why it silently skips unknown dtypes (e.g. forward-compat for HuggingFace fine-tunes that include vendor-specific metadata tensors), and add a regression test asserting current behavior so future audits don't flip it accidentally.

The author of #20 leans toward (1): the runtime is higher-exposure and the official crate is the contract everyone else implements against. But the call belongs to whoever owns runtime-loading correctness.

## Acceptance

If (1):
- [ ] Extract shared validation helper from \`QuarotTensorReader::Shard::open\` and call from \`SafetensorsFile::open\`.
- [ ] Same five offset-validity regression tests in the runtime test surface.
- [ ] Same five dtype-contract regression tests.
- [ ] No regression in the existing model-loading tests (Qwen3-Embedding-0.6B, Qwen3.5-0.8B).

If (2):
- [ ] Code comment at \`f32_weights.rs:1685\` explaining the intentional divergence.
- [ ] Regression test asserting unknown dtypes are skipped (not rejected).
- [ ] Note in ADR-044 (or new ADR) recording the decision.

## References

- PR #20 — added strict validation to \`QuarotTensorReader\`.
- Review rounds 3, 4, and 5 of #20 — surfaced the validation gaps in the converter; same gaps very likely exist in the runtime path.
- Official safetensors crate: <https://docs.rs/safetensors/latest/src/safetensors/tensor.rs.html>

ohdearquant:

Closes #22.

SafetensorsFile and QuarotTensorReader now share one strict layout and dtype validator.
The runtime accepts standard byte-aligned packed dtypes and rejects the same malformed offsets, shapes, and dtype extents as the converter.

This revision adds a shared shard-relative-path guard (rejects absolute paths and `..` traversal in index-derived shard filenames) and makes sharded runtime checkpoint opening validate every shard's header up front instead of on first access, matching the QuaRot reader's existing eager contract.

A follow-up revision closes two remaining gaps where an index-derived shard path could reach the filesystem without going through that guard: the Qwen cache-revision fingerprint (`crates/inference/src/model/qwen.rs`, `resolve_weight_fingerprint_files`/`derive_base_model_rev`) now validates every `weight_map` value before opening or hashing any shard, and the vision embedding loader's single-shard resolver (`crates/embed/src/vision.rs`, `resolve_single_shard`) now routes through a new crate-public `resolve_shard_path` helper (`crates/inference/src/weights/f32_weights.rs`) instead of joining the index-provided shard name directly.

This revision (on top of the prior shard-path-guard revision described above) hardens shard-path containment to a filesystem-safe check and bounds sharded-manifest construction:

- `resolve_contained_path(model_dir, untrusted)` (new, `crates/inference/src/weights/f32_weights.rs`) canonicalizes both the model directory and the joined candidate and requires containment via `Path::starts_with`, defeating symlink escapes in addition to the existing lexical `..`/absolute-path checks. Routed through everywhere a shard/manifest path is resolved: `resolve_shard_path`, `load_sharded`, `ShardedSafetensors::open_index`/`open_shard`/`resolve_weight`, the QuaRot reader's sharded open loop (`crates/inference/src/quant/quarot/io.rs`), the Q4 vision loader's `quantize_index.json` `entry.file` (`crates/inference/src/vision/checkpoint.rs`), the Qwen weight-fingerprint hasher's shard open (`crates/inference/src/model/qwen.rs`), and the `lattice` CLI's Q4 inventory scan (`crates/inference/src/bin/lattice.rs`).
- `MAX_SHARDED_MANIFEST_SHARDS` (new constant, 4096) caps the number of distinct shards `ShardedSafetensors::open_index` will resolve and mmap from a single manifest, bounding worst-case address-space/fd use at construction time.
- `resolve_weight_fingerprint_files` now parses the index via the shared strict `parse_index` instead of a separate ad hoc `serde_json::Value` walk.

All of this remains cold-path SafeTensors reader/loader/manifest validation, not on the decode/prefill hot path. It is **compiled into the default build** (none of it is behind a `cfg` feature gate), so it is a compiled-in cold model-load path, not compiled-out code — distinct from the cfg-gated-module carve-out (e.g. a `#[cfg(metal-gpu)]` module).

`make bench-compare`'s default targets are `lattice-inference`'s `elementwise_cpu_bench` and `lattice-embed`'s `simd` bench. Grepping both bench sources for the changed symbols returns no matches (grep-0):

```
grep -n "resolve_contained_path|open_index|open_shard|canonicalize|model_dir|loading" crates/inference/benches/elementwise_cpu_bench.rs crates/embed/benches/simd.rs
# (no output)
```

Per grep-0, this is a compiled-in cold model-load path, not in the measured call graph of either default bench target. The A/B run is deferred to a maintainer quiet-window run rather than claimed as structurally unreachable.

### DIFF
--- a/crates/embed/src/vision.rs
+++ b/crates/embed/src/vision.rs
@@ -629,13 +629,17 @@ mod tests {
     #[test]
     fn resolve_single_shard_rejects_parent_traversal_shard_name() {
         let tmp = tempfile::tempdir().expect("tempdir");
+        let outside = tmp.path().parent().expect("temp dir has a parent");
+        let outside_file = outside.join("lattice_embed_vision_traversal_outside.safetensors");
+        std::fs::write(&outside_file, b"secret").expect("write outside file");
+
         let index_path = tmp.path().join("model.safetensors.index.json");
         // A mixed quantized/FP16 manifest can name an escaping FP16 shard;
         // without the shared containment helper this join would resolve
         // outside `tmp` instead of erroring.
         std::fs::write(
             &index_path,
-            r#"{"metadata":{},"weight_map":{"a":"../outside/decoder.safetensors"}}"#,
+            r#"{"metadata":{},"weight_map":{"a":"../lattice_embed_vision_traversal_outside.safetensors"}}"#,
         )
         .expect("write index");
 
@@ -643,9 +647,11 @@ mod tests {
             .expect_err("a shard name with a parent-directory component must be rejected");
         let msg = err.to_string();
         assert!(
-            msg.contains("parent-directory"),
+            msg.contains("escapes the checkpoint directory"),
             "expected the shared containment error, got: {msg}"
         );
+
+        std::fs::remove_file(&outside_file).ok();
     }
 
     #[test]
@@ -667,7 +673,7 @@ mod tests {
             resolve_single_shard(tmp.path()).expect_err("an absolute shard name must be rejected");
         let msg = err.to_string();
         assert!(
-            msg.contains("absolute"),
+            msg.contains("escapes the checkpoint directory"),
             "expected the shared containment error, got: {msg}"
         );
     }
--- a/crates/embed/src/vision.rs
+++ b/crates/embed/src/vision.rs
@@ -24,8 +24,8 @@ use lattice_inference::model::qwen35_config::Qwen35Config;
 use lattice_inference::tokenizer::bpe::BpeTokenizer;
 use lattice_inference::vision::checkpoint::{Qwen35VisionWeights, load_qwen35_vision_weights};
 use lattice_inference::vision::embed_image_from_bytes_f16;
-use lattice_inference::weights::SafetensorsFile;
 use lattice_inference::weights::f16_weights::{F16ModelWeights, load_f16_weights};
+use lattice_inference::weights::{SafetensorsFile, resolve_shard_path};
 use std::path::Path;
 
 pub use lattice_inference::forward::cpu_f16::PoolingStrategy;
@@ -202,7 +202,12 @@ fn resolve_single_shard(model_dir: &Path) -> Result<std::path::PathBuf> {
     shards.sort_unstable();
     shards.dedup();
     match shards.as_slice() {
-        [one] => Ok(model_dir.join(one)),
+        [one] => resolve_shard_path(model_dir, one).map_err(|e| {
+            EmbedError::ModelInitialization(format!(
+                "invalid shard filename in {}: {e}",
+                model_dir.join("model.safetensors.index.json").display()
+            ))
+        }),
         [] => Err(EmbedError::ModelInitialization(format!(
             "empty weight_map in {}",
             model_dir.join("model.safetensors.index.json").display()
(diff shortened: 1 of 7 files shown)


ohdearquant:
Closing in favour of a focused replacement against current main.

Main has since landed its own shard-path containment helper (`contained_shard_path`, with call sites across the loader, the QuaRot reader, the vision checkpoint loader and the doctor) and its own strictness work on the runtime safetensors parser. That covers most of what this branch set out to do, by a different route, so merging this branch would reintroduce a parallel validator rather than add anything.

Two things here are still not on main and will be rebuilt on top of it:

- Containment against symlink escapes. The current check is purely lexical, so a shard name that is neither absolute nor contains `..` still passes while resolving, through a symlink, to any file the process can read.
- Eager per-shard header validation when opening a sharded checkpoint, so a malformed shard is rejected at load rather than at first tensor access.

PR #1045, opened 2026-07-18
fix(inference): shape-validate dense FFN, full-attention, and GDN tensors on load

### LINKED ISSUE #1035: bug(inference): Qwen3.5 safetensors loader accepts config-incompatible tensor shapes (DoS via undersized checkpoint)
## Summary

## Detail

## Fix

## Test

## Scope note

ohdearquant:
Closes #1035.

The Qwen3.5 safetensors loader already shape-validated embeddings, the output head, norms, and MoE FFN tensors via `load_owned_tensor_checked`, but the dense FFN (`mlp.{gate,up,down}_proj.weight`), full-attention (`self_attn.{q,k,v,o}_proj.weight`, `q_norm`/`k_norm`), and GDN/linear-attention (`in_proj_{qkv,z}`, `conv1d.weight`, `norm.weight`, `out_proj.weight`) tensors, plus the per-layer `input_layernorm`/`post_attention_layernorm`, went through the unchecked `load_owned_tensor`. A finite but undersized tensor for any of those loaded successfully and only failed later, inside a release-active GEMM bounds assertion during the forward pass — a panic instead of a clean load error. This PR routes every remaining call through `load_owned_tensor_checked` with a config-derived expected shape (verified against a real qwen3.5-0.8b checkpoint's declared tensor shapes) and removes the now-unused unchecked loader. Adds three mutation-sensitive regression tests (dense FFN, full attention, GDN) plus a fix to the existing GDN asymmetric-head-shape test, all built on a synthetic in-memory `TensorSource`.

Bench disposition: this change is in the safetensors load path (`model/qwen35/loading.rs`), not compiled into the Criterion bench targets (`elementwise_cpu_bench`, `simd`) — structural-unreachability carve-out per ADR-058. The added checks are O(1) shape comparisons at load time, not hot-path.

### DIFF
--- a/crates/inference/src/attention/gdn.rs
+++ b/crates/inference/src/attention/gdn.rs
@@ -50,7 +50,9 @@ pub struct GatedDeltaNetWeights {
     pub conv_dim: usize,
     pub kernel_size: usize,
 
-    /// Output gated RMSNorm gamma: `[output_dim]`
+    /// Output gated RMSNorm gamma: `[linear_value_head_dim]`, one per-head-dim gamma
+    /// shared across all value heads and applied identically to each (see the
+    /// gated-RMSNorm loop over `value_heads` in `gated_delta_net_step`).
     pub norm_weight: Vec<f32>,
 
     /// Output projection: [hidden_size, output_dim]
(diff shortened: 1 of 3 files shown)


ohdearquant:
Closing in favor of #1053, which now carries this branch's config-derived Qwen3.5 tensor shape validation together with the dimension-overflow and layer-count bounds.

PR #1047, opened 2026-07-18
refactor(serve): share request normalization contract

ohdearquant:
## Summary

Extracts a shared chat-completions request contract (`crates/inference/src/serve/contract.rs`) used by both the `lattice` and `lattice_serve` HTTP binaries, replacing two independent, drifting request DTOs and validation cascades with one.

- One typed `ChatRequest` DTO, deserialized once per request. The message-count bound (`MAX_MESSAGE_COUNT`) is enforced inline while `messages` deserializes, so an over-limit request is rejected in that same single JSON traversal — no separate raw-bytes preflight pass.
- `normalize_request` composes profile-aware validation (`ServeProfile::lattice` / `ServeProfile::lattice_serve`) for model-name policy, `max_tokens` policy, and which extensions (`stop`, sampling extensions, reasoning budget, structured output, logprobs) each server supports. Each profile's original client-observable error precedence is preserved where the two binaries differed (e.g. the `lattice` profile validates the served-model name before a conflicting `max_tokens`/`max_completion_tokens` pair; the daemon profile checks that conflict first).
- Message role/content normalization (`normalize_messages`) is shared; both binaries' test suites now exercise it directly instead of maintaining separate implementation-shaped copies.
- `response_format.json_schema.schema` is kept as an unparsed `RawValue` span so the `lattice` profile (which always rejects `json_schema`) never materializes a parsed tree for a schema it's about to reject.
- Content-Type is validated (`require_json_content_type`) before either handler buffers the request body.
- A request-body length-limit rejection is now distinguished from other body-buffering failures in both binaries, so only the former maps to HTTP 413.

## Test plan

```
cargo test -p lattice-inference --features f16,metal-gpu --lib -- serve::contract
cargo test -p lattice-inference --features f16,metal-gpu,test-utils --bin lattice
cargo test -p lattice-inference --features f16,metal-gpu,test-utils --bin lattice_serve
cargo clippy -p lattice-inference --all-targets --features f16,metal-gpu -- -D warnings
cargo fmt -p lattice-inference
```

All pass. Regression coverage includes: message-count/content-byte bounds (including the single-parse behavior), profile-specific validation-order precedence, stop-string bounds, sampling-extension tolerance per profile, and structured-output schema handling.

## bench-compare disposition

`crates/inference/` admission-path change: removes a redundant full-body JSON traversal (the old raw-bytes message-count preflight) from every accepted request, so it is expected to reduce admission time, not regress it. An A/B (`make bench-compare`) in a quiet window is pending before this PR is marked ready.


oceanwaves630 (approved):
Reviewed and approved.

PR #1050, opened 2026-07-18
fix(inference): validate QuaRot quantizer ingress

ohdearquant:
## Summary

Hardens QuaRot/Q4 offline-quantization filesystem ingress so every checked path is also the path actually used — closing check-then-use gaps where an earlier validation and the later read/write/rename/delete were two independent path resolutions:

- `convert_quarot_qwen35`: the output-directory fd is opened once, immediately before the write session begins, and re-validated on the fd itself (scanning its contents via `fdopendir`/`readdir`) rather than trusting an earlier, separately-resolved path check. Every write in the session goes through that one fd.
- `quantize_q4`'s `TempPublishDir::publish` anchors both the staging directory and the final `output_dir` name to a single held parent-directory fd via `renameat`, instead of re-resolving both as full paths.
- `TempPublishDir`'s `Drop` cleanup walks and removes the staging directory's contents entirely through the held directory fd (`fdopendir`/`readdir`/`unlinkat`), then removes the now-empty directory entry itself via `unlinkat(parent_fd, name, AT_REMOVEDIR)` — never a path-based recursive delete.
- `quantize_model` reads `config.json` through the same model-directory fd already used to validate the checkpoint, instead of reopening the model directory by a fresh path for that second read.
- Descriptor-relative shard traversal and final-component `config.json` symlink rejection from the prior round are unchanged.

## Test plan

```
cargo test -p lattice-inference --lib quant::quarot
cargo test -p lattice-inference --lib q4_weights
cargo test -p lattice-inference --bin quantize_q4
cargo clippy -p lattice-inference --all-targets --features f16,metal-gpu -- -D warnings
cargo fmt -p lattice-inference -- --check
```

All green (192 + 71 + 21 tests). New regression tests were each verified against the fix reverted (fail) and restored (pass) before landing.

## bench-compare disposition

Offline filesystem ingress for the QuaRot/Q4 conversion CLIs — not a decode or prefill hot path, and not reachable from any benchmarked call site. No hot-path call sites; A/B in a quiet window pending before ready.

### DIFF
--- a/crates/inference/Cargo.toml
+++ b/crates/inference/Cargo.toml
@@ -54,6 +54,9 @@ test-utils = []
 [dependencies]
 half.workspace = true
 memmap2 = "0.9"
+# Unix directory-fd binding (openat/O_NOFOLLOW) for the QuaRot/Q4 offline
+# quantizer's staging-directory writes (path-race hardening, ADR-058).
+libc = "0.2"
 indexmap = "2"
 rustc-hash = "2"
 rayon = { workspace = true }
--- a/crates/inference/src/bin/quantize_q4.rs
+++ b/crates/inference/src/bin/quantize_q4.rs
@@ -275,12 +275,11 @@ impl TempPublishDir {
                 output_dir.display()
             )
         })?;
-        // Restore the parent-creation behavior `fs::create_dir_all(output_dir)`
-        // used to provide before `TempPublishDir` replaced it: a valid
-        // `--output-dir` whose parent doesn't exist yet must still succeed
-        // (the adjacent temp dir and the final rename target both resolve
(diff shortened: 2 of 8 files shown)


ohdearquant:
Closing as part of the 2026-07-20 draft-backlog triage. This PR conflicts with the merged #1022 (QuaRot online R3/R4 rotation), which rewrote crates/inference/src/quant/quarot/{convert,io}.rs — the exact files it patches — so it cannot apply as-is and CI is red. Its general weight-ingress checks overlap the ingress.rs cluster (#1032/#1041). If QuaRot-specific ingress validation is still wanted, re-scope it as a small PR against the current QuaRot code path. Branch is preserved and this is reopenable.

PR #1053, opened 2026-07-18
fix(inference): validate Qwen3.5 tensor shapes against config before construction

### LINKED ISSUE #1035: bug(inference): Qwen3.5 safetensors loader accepts config-incompatible tensor shapes (DoS via undersized checkpoint)

ohdearquant:
## Summary

`crates/inference/src/model/qwen35/loading.rs` had two tensor loaders: `load_owned_tensor_checked` (validates the loaded shape against a config-derived expected shape, returning `InferenceError::ShapeMismatch` on mismatch) and an unchecked `load_owned_tensor`. The dense FFN, full-attention, GDN/linear-attention, and per-layer norm tensors were loaded through the unchecked path, so a finite but config-incompatible checkpoint could construct a model and then panic in the GEMM bounds assertion during the forward pass instead of failing cleanly at load time.

This PR routes every remaining required Qwen3.5 tensor through `load_owned_tensor_checked` with a config-derived expected shape, matching the convention already used for embeddings, the output head, the final norm, and the MoE FFN tensors. The now-unused unchecked helper is removed, and the rustdoc on `Qwen35Model::from_safetensors` / `load_weights` is updated to state the actual shape-validation contract.

Scope note: this sentence previously said shard-path containment was out of scope. That was true when the branch opened and is no longer: `resolve_shard` now takes `model_dir` and routes through the manifest-path guard, and the open-once loader described further down is part of this branch. The file-backed mmap validation cache remains out of scope and untouched.

Closes #1035.

Two additional load-time hardening checks are folded in alongside the shape-validation change above, so this PR covers both:

- `num_hidden_layers` (from `config.json`, previously only checked for nonzero) is bounded by `MAX_HIDDEN_LAYERS` (4096) enforced at config-parse time, before `compute_layer_types`, layer-mask normalization, or the loader's `Vec::with_capacity(cfg.num_hidden_layers)` ever allocate proportionally to it. An extreme value (e.g. `usize::MAX`) previously reached those allocations before any structural validation ran.
- The loader's config-derived dimension products (full-attention `q_dim`/`kv_dim`, the fused-gate doubled `q_proj` row count, GDN `qkv_dim`/`output_dim`, and the doubled MoE `gate_up_proj` row count) now go through `checked_double` / `checked_full_q_dim` / `checked_full_kv_dim` / `checked_linear_qkv_dim` / `checked_linear_output_dim`, which use `checked_mul`/`checked_add` and return a typed `InferenceError::InvalidInput` instead of silently wrapping on overflow.

## Allocation-driving config field audit

**Update (linear_num_key_heads / conv-kernel product / RoPE-table-byte / vision-scalar sweep — this round):** the prior round bounded `linear_conv_kernel_dim` for the zero-underflow case only and left it, plus its downstream product, otherwise unbounded; re-auditing every `with_capacity`/`vec![`/`resize`/`checked_mul` site across `crates/inference/src` against the config field sizing it (not just the `qwen35`-named paths) found one more field with the *same* class of gap — a config scalar checked for zero but with no upper bound at all — plus two two-factor products where individually-generous per-factor caps still compose into an allocation-hostile size, plus eight `VisionModelConfig` scalars that were only nonzero- and overflow-guarded. This is the full enumeration table from that audit; every allocation-driving config field is listed, not a sample.

| config field | allocation site(s) | size expression | bound this round | status |
|---|---|---|---|---|
| `num_hidden_layers` | loader `Vec::with_capacity`; `compute_layer_types`; layer-mask normalization | `num_hidden_layers` | `MAX_HIDDEN_LAYERS` (4096) | already-bounded |
| `layer_types` (Vec len) | bounded-deserialize visitor | declared array length | `MAX_HIDDEN_LAYERS` (4096) | already-bounded |
| `layer_mask` (Vec len) | bounded-deserialize visitor | declared array length | `MAX_HIDDEN_LAYERS` (4096) | already-bounded |
| `vocab_size` | decode cache `logits` resize | `vocab_size` | `MAX_VOCAB_SIZE` (4,000,000) | already-bounded |
| `head_dim` | Metal RoPE table (factor); full-attention geometry | `head_dim` | `MAX_HEAD_DIM` (2048) | already-bounded |
| `hidden_size` | `hidden`/`residual`/`attn_out`/`ffn_out`/`expert_out` scratch; `max_q8_input` | `hidden_size` | `MAX_HIDDEN_SIZE` (1,048,576) | already-bounded |
| `num_attention_heads` | `ForwardScratch::ensure_capacity` `scores`; full-attention `q_dim` (factor) | `num_attention_heads` | `MAX_ATTENTION_HEADS` (8,192) | already-bounded |
| `num_key_value_heads` | full-attention `kv_dim` (factor) | `num_key_value_heads` | `MAX_ATTENTION_HEADS` (8,192) | already-bounded |
| `num_attention_heads * head_dim` | `q_buf`/`context`/`2*q_dim` scratch | product | `MAX_FULL_ATTENTION_DIM` (1,048,576) | already-bounded (product-budget) |
| `num_key_value_heads * head_dim` | `k_buf`/`v_buf` scratch | product | `MAX_FULL_ATTENTION_DIM` (1,048,576) | already-bounded (product-budget) |
| `intermediate_size` | MLP scratch (`gate_buf`/`up_buf`) | `intermediate_size` | `MAX_INTERMEDIATE_SIZE` (1,048,576) | already-bounded |
| `moe_intermediate_size` | MoE expert scratch | value | `MAX_INTERMEDIATE_SIZE` (1,048,576) | already-bounded |
| `shared_expert_intermediate_size` | shared-expert scratch | value | `MAX_INTERMEDIATE_SIZE` (1,048,576) | already-bounded |
| `num_experts` | `router_logits` resize | `num_experts` | `MAX_NUM_EXPERTS` (4,096) | already-bounded |
| `num_experts_per_tok` | `router_selected` resize | `num_experts_per_tok` | `MAX_NUM_EXPERTS` (4,096) | already-bounded |
| `max_position_embeddings` | Metal RoPE table (factor) | `max_position_embeddings` | `MAX_POSITION_EMBEDDINGS` (4,194,304) | already-bounded |
| `mrope_section` (Vec len) | bounded-deserialize visitor | declared array length | `MAX_CONFIG_VECTOR_LEN` (1,024) | already-bounded |
| `deepstack_visual_indexes` (Vec len) | bounded-deserialize visitor | declared array length | `MAX_CONFIG_VECTOR_LEN` (1,024) | already-bounded |
| `linear_num_value_heads` | GDN output/state products (factor) | value | implicit via `MAX_LINEAR_OUTPUT_DIM`/`MAX_GDN_STATE_SIZE` | already-bounded (product-budget carries it) |
| `linear_value_head_dim` | GDN output/state products (factor) | value | implicit via `MAX_LINEAR_OUTPUT_DIM`/`MAX_GDN_STATE_SIZE` | already-bounded (product-budget carries it) |
| `linear_num_value_heads * linear_value_head_dim` | `x_q_scratch` (`max_q8_input`, via `linear_output_dim()`) | product | `MAX_LINEAR_OUTPUT_DIM` (1,048,576) | already-bounded (product-budget) |
| `linear_num_value_heads * linear_key_head_dim * linear_value_head_dim` | `s_matrices` GDN state (`attention/gdn.rs`) | product | `MAX_GDN_STATE_SIZE` (16,777,216) | already-bounded (product-budget) |
| `vision_config.depth` | vision loader tensor-name `String` allocation | `depth` | `MAX_VISION_DEPTH` (1,024) | already-bounded |
| **`linear_num_key_heads`** | GDN QKV proj weight, conv1d weight, conv/activation scratch throughout `attention/gdn.rs`, `model/qwen35/loading.rs`, `forward/metal_qwen35.rs` (via `linear_qkv_dim()`, factor) | value | **`MAX_LINEAR_NUM_KEY_HEADS` (8,192)** | **NEW this round** — was nonzero-checked only, no upper bound at all (unlike `linear_num_value_heads`, which the product budgets above carry as a factor) |
| **`linear_conv_kernel_dim`** | GDN conv1d buffer (`conv_dim * (kernel_dim - 1)`); conv1d weight tensor | value | **`MAX_CONV_KERNEL_DIM` (512)** | **NEW this round** — was nonzero-checked only (underflow guard) |
| **`linear_qkv_dim() * (linear_conv_kernel_dim - 1)`** | `GatedDeltaNetState::new`'s `conv_buffer` (`attention/gdn.rs`); mirrored Metal GDN conv buffer allocation (`forward/metal_qwen35.rs`) | product | **`MAX_GDN_CONV_BUFFER_SIZE` (16,777,216)** | **NEW this round, product-budget** — `MAX_LINEAR_NUM_KEY_HEADS` and `MAX_CONV_KERNEL_DIM` bound their own factors but not the product |
| **`max_position_embeddings * rope_dim`** | `build_rope_interleaved`'s cos/sin table pair (`forward/metal_qwen35.rs`), `4 * max_position_embeddings * rope_dim` bytes | product (bytes) | **`MAX_ROPE_TABLE_BYTES` (1 GiB)** | **NEW this round, product-budget** — `MAX_POSITION_EMBEDDINGS` and `MAX_HEAD_DIM` (which bounds `rope_dim` transitively) each bound their own factor but not the byte product |
| **`vision_config.hidden_size`** | `checked_derived_sizes` (`qkv_out`/`mlp_intermediate`/patch-embed/pos-embed products) | value | **`MAX_VISION_HIDDEN_SIZE` (1,048,576)** | **NEW this round** — was nonzero + overflow-only |
| **`vision_config.num_heads`** | `checked_derived_sizes` (divisibility factor) | value | **`MAX_VISION_NUM_HEADS` (8,192)** | **NEW this round** — was nonzero + overflow-only |
| **`vision_config.patch_size`** | `checked_derived_sizes` (`patch_embed_weight`) | value | **`MAX_VISION_PATCH_SIZE` (4,096)** | **NEW this round** — was nonzero + overflow-only |
| **`vision_config.spatial_merge_size`** | `checked_derived_sizes` (`merger fc1`, squared) | value | **`MAX_VISION_SPATIAL_MERGE_SIZE` (64)** | **NEW this round** — was nonzero + overflow-only |
| **`vision_config.out_hidden_size`** | `checked_derived_sizes` (`merger fc2`) | value | **`MAX_VISION_OUT_HIDDEN_SIZE` (1,048,576)** | **NEW this round** — was nonzero + overflow-only |
| **`vision_config.temporal_patch_size`** | `checked_derived_sizes` (`patch_embed_weight`) | value | **`MAX_VISION_TEMPORAL_PATCH_SIZE` (256)** | **NEW this round** — was nonzero + overflow-only |
| **`vision_config.num_position_embeddings`** | `checked_derived_sizes` (`pos_embed`) | value | **`MAX_VISION_NUM_POSITION_EMBEDDINGS` (16,777,216)** | **NEW this round** — was nonzero + overflow-only |
| **`vision_config.in_channels`** | `checked_derived_sizes` (`patch_embed_weight`) | value | **`MAX_VISION_IN_CHANNELS` (256)** | **NEW this round** — was nonzero + overflow-only |
| Metal `from_q4_dir` config-derived Q4/F16 tensor-shape validation | Metal weight-buffer sizing (`forward/metal_qwen35.rs`) | n/a | n/a | **covered by #1037** — the Metal weight-buffer sizing path is unchanged in this PR |
| `mtp_num_hidden_layers`, `quarot_rotation_seed` | none found | — | — | not currently allocation-driving (MTP forward path not yet wired to config-scaled buffers); re-audit if that changes |
| `rms_norm_eps`, `rope_theta`, `partial_rotary_factor`, `tie_word_embeddings`, `output_router_logits`, `router_aux_loss_coef`, `eos_token_id`, `image_token_id`/`video_token_id`/`vision_start_token_id`/`vision_end_token_id` | none (scalars/flags, not size-driving) | — | — | not allocation-driving |

Grep basis for "not just the seeded ones" (re-run against `crates/inference/src` before merge to confirm no new site appeared): `with_capacity`, `vec!\[.*;`, `\.resize\(`, `\.reserve\(`, `make_zero_buffer`, `make_buffer`, `checked_mul` in `model/qwen35_config.rs`, `model/qwen35/*.rs`, `forward/metal_qwen35.rs`, `attention/gdn.rs`, `vision/qwen35_*.rs`. The training/backward path (`attention/gdn_backward.rs`, `gdn_fused.rs`) reuses the same `qkv_dim`/`value_heads`/`key_dim`/`value_dim` factors this PR bounds, scaled additionally by a runtime `seq_len` (not a config field) — it inherits these bounds and is out of scope for a config-parse hardening PR.

**Update (two-dimensional validity sweep — allocation + semantic-validity gap closure):** this round expanded the enumeration from one dimension (allocation-safety) to two: (A) allocation-safety (upper bounds + product byte-budgets, what every prior round in this PR closed) and (B) semantic validity — an in-allocation-budget value that still panics or produces non-finite inference downstream. This round closes four blocking gaps (B2-B5) and re-derives the full field enumeration against both dimensions.

**B2 (security/major) — GDN chunk-scratch scratch was unbounded in effect.** `new_session_inner` (`forward/metal_qwen35.rs:3707-3726`, mirrored at `:15701-15726`) allocates `GdnChunkScratch` from `chunk_rows = num_chunks * linear_num_value_heads() * GDN_CHUNK_SIZE` (`GDN_CHUNK_SIZE = 32`, read from `metal_qwen35.rs:1426`) and `num_chunks = bp.div_ceil(GDN_CHUNK_SIZE)` where `bp = max_cache_len.min(512)` — so `num_chunks` is capped at 16 regardless of `max_position_embeddings`. `linear_num_value_heads()` was bounded only as a *factor* of `MAX_LINEAR_OUTPUT_DIM` (`value_heads * linear_value_head_dim`) and `MAX_GDN_STATE_SIZE` (`value_heads * linear_key_head_dim * linear_value_head_dim`); pinning the other factor at 1 let `value_heads` alone reach up to 1,048,576, and separately let `linear_key_head_dim`/`linear_value_head_dim` reach multi-hundred-thousand scale while `value_heads` stayed at 1 — either way, the chunk-scratch buffers (`chunk_rows * key_head_dim` for `q`/`k`/`w`/`k_right`; `chunk_rows * GDN_CHUNK_SIZE` for `kkt`/`qk_l`) reach multi-gigabyte-to-terabyte scale. Fixed with two new bounds: `MAX_LINEAR_NUM_VALUE_HEADS` (8,192, symmetric with the existing `MAX_LINEAR_NUM_KEY_HEADS`) bounds `value_heads` directly, and `MAX_GDN_CHUNK_SCRATCH_BYTES` (1 GiB) bounds the exact mirrored chunk-scratch formula (`num_chunks_upper=16` from the engine's own `GDN_CHUNK_SIZE`/512-prefill-cap literals, times `value_heads` times `max(linear_key_head_dim, linear_value_head_dim)`, in `u128` so the check itself cannot overflow) — closing the free-multiplier gap either bound alone leaves open.

**B3 (runtime correctness/major) — `max_position_embeddings = 0` accepted.** Only the upper bound (`MAX_POSITION_EMBEDDINGS`) existed; zero survived parse and produced an empty Metal RoPE table (`rope_max * rope_dim / 2 = 0` entries), so any non-empty forward pass indexing `cos_at(0)`/`sin_at(0)` (`rope.rs:54,60`) panics. Fixed: reject `max_position_embeddings == 0`.

**B4 (runtime correctness/major) — `rope_theta` never validated finite/positive.** `RopeTable::new` (`rope.rs:20-35`) computes `freq = 1.0 / theta.powf(...)` per table entry; `theta <= 0.0` or non-finite drives every `freq` (and thus every `cos`/`sin` table entry) to `inf`/`NaN` — silent attention corruption, not a panic, and not caught by any allocation bound since `rope_theta` drives no allocation. Fixed: require `rope_theta.is_finite() && rope_theta > 0.0`.

**B5 (runtime correctness/major) — `VisionModelConfig::num_position_embeddings` accepted non-square.** `build_pos_embed_and_rope_tables` (`vision/qwen35_vit.rs:245`) derives `side = sqrt(num_position_embeddings).round()` and indexes `pos_embed[(ch * side + cw) * hidden ..]` for `ch, cw < side` — a `side * side * hidden`-row assumption. A non-square value (in-budget per `MAX_VISION_NUM_POSITION_EMBEDDINGS`, but geometrically invalid) can index past the `pos_embed` slice's true length and panic. Fixed: require `num_position_embeddings` to be a perfect square in `VisionModelConfig::validate()`. (A pre-existing test fixture, `checkpoint.rs`'s `tiny_vision_cfg()`, used `num_position_embeddings: 2` — never a valid geometry even before this fix; it and its one `pos_embed` shape assertion are updated to `4` (side = 2), the smallest fixture-compatible perfect square, so the newly-enforced constraint doesn't regress an existing test that depended on latent invalidity nobody had exercised.)

**Out of scope:** Direct Metal Q4/F16 ordinary-tensor shape validation (`metal_qwen35.rs:1157-1203` / `:15337-15378`) belongs to #1037, which lands first; this PR inherits that guard on rebase. No Metal shape-validation code is added here.

### Two-dimensional enumeration (this round) — every field, dimension (A) allocation + dimension (B) semantic validity

| field / allocation site | (A) allocation bound | (B) validity constraint | status |
|---|---|---|---|
| `linear_num_value_heads()` | `MAX_LINEAR_NUM_VALUE_HEADS` (8,192) — **NEW**, direct cap (previously only an implicit factor of two product budgets) | n/a (bounded value can't corrupt output on its own) | **NEW this round** |
| GDN chunk-scratch (`new_session_inner`'s `GdnChunkScratch`, both call sites) | `MAX_GDN_CHUNK_SCRATCH_BYTES` (1 GiB) — **NEW** product budget mirroring the exact engine formula | n/a (allocation-only site) | **NEW this round** — the allocation site missed by the field-based enumeration in the prior round |
| `max_position_embeddings` | `MAX_POSITION_EMBEDDINGS` (already-bounded, prior round) | **NEW**: must be `> 0` (empty RoPE table → `cos_at(0)`/`sin_at(0)` panic) | (A) already-bounded; **(B) NEW this round** |
| `rope_theta` | n/a (not size-driving; scalar) | **NEW**: must be finite and `> 0.0` (else `inf`/`NaN` RoPE table, silent corruption) | **(B) NEW this round** (not previously validated at all) |
| `vision_config.num_position_embeddings` | `MAX_VISION_NUM_POSITION_EMBEDDINGS` (already-bounded, prior round) | **NEW**: must be a perfect square (`side * side == n`; else OOB index in `build_pos_embed_and_rope_tables`) | (A) already-bounded; **(B) NEW this round** |
| `num_hidden_layers` | `MAX_HIDDEN_LAYERS` (already-bounded) | must be `> 0` (already enforced; divide/zero-length guards) | already-bounded, both dims |
| `vocab_size` | `MAX_VOCAB_SIZE` (already-bounded) | must be `> 0` (already enforced) | already-bounded, both dims |
| `hidden_size` | `MAX_HIDDEN_SIZE` (already-bounded) | must be `> 0` (already enforced) | already-bounded, both dims |
| `intermediate_size` | `MAX_INTERMEDIATE_SIZE` (already-bounded) | must be `> 0` (already enforced) | already-bounded, both dims |
| `num_attention_heads` / `num_key_value_heads` | `MAX_ATTENTION_HEADS`, `MAX_FULL_ATTENTION_DIM` product (already-bounded) | must be `> 0`; `num_attention_heads` divisible by `num_key_value_heads` (already enforced, GQA group-count division) | already-bounded, both dims |
| `head_dim` | `MAX_HEAD_DIM` (already-bounded) | must be `> 0`; derived `rope_dim` even, `>= 2`, `<= head_dim` (already enforced) | already-bounded, both dims |
| `partial_rotary_factor` | n/a (scalar, feeds `rope_dim` which is bounded) | must be finite, in `[0.0, 1.0]` (already enforced — indexes `head_vec[rope_dim/2+i]`) | already-bounded, both dims |
| `linear_num_key_heads` | `MAX_LINEAR_NUM_KEY_HEADS` (already-bounded) | must be `> 0`; `linear_num_value_heads()` a positive multiple of it (already enforced — GDN fused-path divide) | already-bounded, both dims |
| `linear_key_head_dim` / `linear_value_head_dim` | `MAX_GDN_STATE_SIZE`, `MAX_LINEAR_OUTPUT_DIM`, **`MAX_GDN_CHUNK_SCRATCH_BYTES` (NEW)** product budgets | must be `> 0` (already enforced — degenerate zero-width recurrence) | (A) **tightened this round**; (B) already-bounded |
| `linear_conv_kernel_dim` | `MAX_CONV_KERNEL_DIM`, `MAX_GDN_CONV_BUFFER_SIZE` product (already-bounded) | must be `> 0` (already enforced — `kernel_dim - 1` unsigned underflow) | already-bounded, both dims |
| `num_experts` / `num_experts_per_tok` | `MAX_NUM_EXPERTS` (already-bounded) | must be `> 0` when present; `num_experts_per_tok <= num_experts` (already enforced) | already-bounded, both dims |
| `moe_intermediate_size` / `shared_expert_intermediate_size` | `MAX_INTERMEDIATE_SIZE` (already-bounded) | must be `> 0` when present (already enforced) | already-bounded, both dims |
| `mrope_section` / `deepstack_visual_indexes` (Vec len) | `MAX_CONFIG_VECTOR_LEN` bounded-deserialize (already-bounded) | `deepstack_visual_indexes` entries `< depth` (already enforced) | already-bounded, both dims |
| `vision_config.depth` | `MAX_VISION_DEPTH` (already-bounded) | must be `> 0` (already enforced) | already-bounded, both dims |
| `vision_config.hidden_size` / `num_heads` | `MAX_VISION_HIDDEN_SIZE`, `MAX_VISION_NUM_HEADS` (already-bounded) | must be `> 0`; `hidden_size` divisible by `num_heads` (already enforced) | already-bounded, both dims |
| `vision_config.{patch_size, spatial_merge_size, out_hidden_size, temporal_patch_size, in_channels}` | own `MAX_VISION_*` bounds + `checked_derived_sizes` overflow guards (already-bounded) | must be `> 0` each (already enforced); no further geometric constraint found on these five | already-bounded, both dims |
| `full_attention_interval` | n/a (scalar) | must be `> 0` when `layer_types` needs recomputing (already enforced — `(i+1) % interval` divide-by-zero) | already-bounded, both dim (B); no allocation |
| `rms_norm_eps`, `tie_word_embeddings`, `output_router_logits`, `router_aux_loss_coef`, `eos_token_id`, `image_token_id`/`video_token_id`/`vision_start_token_id`/`vision_end_token_id`, `mtp_num_hidden_layers`, `quarot_rotation_seed` | n/a (not size-driving) | no panic/non-finite path found reading these fields in the loader/forward/vit sweep | not allocation- or validity-driving; re-audit if MTP/QuaRot forward wiring changes |

Grep basis for the (B) sweep (semantic-validity consumption sites, run against `crates/inference/src`): `_at(`, `.sqrt(`, `powf(`, `/ cfg\.`, `- 1` near each field's name, in `rope.rs`, `forward/metal_qwen35.rs`, `vision/qwen35_vit.rs`, `attention/gdn.rs`, `model/qwen35/loading.rs`. Grep basis for the (A) re-sweep (allocation sites, confirming no additional unbounded site beyond B2's): `vec!\[0.0`, `with_capacity(`, `\.resize(`, `make_zero_buffer(`, `make_zero_byte_buffer(` across `crates/inference/src`, filtered to call sites reachable from parsed (not test-constructed) `Qwen35Config`/`VisionModelConfig` values — the only unbudgeted site found was B2's GDN chunk-scratch; the remainder are either already covered by an existing per-field/product bound, gated behind `#[cfg(test)]`, or keyed by a non-`Qwen35Config` runtime parameter (`GenerateConfig::max_new_tokens`, `seq_len`) out of this parser's scope.

## Tests (this round)

6 new mutation-verified tests (reject + accept pairs for `MAX_LINEAR_NUM_VALUE_HEADS` and `MAX_GDN_CHUNK_SCRATCH_BYTES`; reject + accept for `max_position_embeddings == 0`; reject (zero, negative) + accept for `rope_theta`; reject + accept for the vision perfect-square constraint) plus 3 existing tests adjusted (their crafted "huge factor" values now trip the new, tighter `MAX_LINEAR_NUM_VALUE_HEADS` / `MAX_GDN_CHUNK_SCRATCH_BYTES` guards before reaching the guard they were written to isolate; re-derived to stay under the new bounds while still tripping their original target). Every new bound was mutation-verified by disabling the guard (`if false && <condition>`), confirming the corresponding reject test fails, then restoring the file byte-identical (`diff` confirmed no residual change).
## Class-complete pass: aggregate budgets, materialization sites, ingress types, path containment

**Update (class-complete: aggregate allocation budgets, materialization-site sweep, type-level ingress enumeration, path containment):** the earlier per-buffer and per-field bounds are individually correct but incomplete as a *class* — each one caps a single allocation while the aggregate, the materialization ordering, and the set of ingress points remain open. Four axes are closed here, each enumerated below so the whole class can be judged together rather than one hostile geometry at a time.

### CLASS A2 — aggregate GDN per-session allocation budget (aggregate-budget class)

`new_session_inner` (`forward/metal_qwen35.rs`, both call sites) allocates every GDN buffer below *together*, once per session. The existing `MAX_GDN_CHUNK_SCRATCH_BYTES` bounds only the largest single `GdnChunkScratch` buffer; a config can pass that guard **and** every per-field/per-product guard individually while the **sum** across all buffers still reaches multi-GiB. `MAX_GDN_SESSION_BYTES` (2 GiB) now bounds the sum directly, mirrored in `u128` from the exact `new_session_inner` formulas (`bp = max_cache_len.min(512)`, `num_chunks = bp.div_ceil(32)`, `chunk_rows = num_chunks * value_heads * 32`, `c2 = chunk_rows * 32` — the same mirrored constants already introduced alongside it):

| buffer | byte formula (config fields) | worst-case @ MAX_GDN_SESSION_BYTES bound |
|---|---|---|
| `gdn_qkv` | `bp_upper * qkv_dim * 4` | included in 2,147,483,648-byte aggregate |
| `gdn_z` | `bp_upper * output_dim * 4` | ″ |
| `gdn_qkvz` | `(qkv_dim + output_dim) * 4` | ″ |
| `gdn_key_scratch` | `(key_heads * (2*key_dim+1) + value_heads*2) * 4` | ″ |
| `gdn_raw_out` | `output_dim * 4` | ″ |
| `gdn_chunk.raw_out` | `bp_upper * output_dim * 4` | ″ |
| `gdn_chunk.{q,k,w,k_right}` (4 buffers) | `4 * chunk_rows_upper * key_dim * 4` | ″ |
| `gdn_chunk.{v,u,r}` (3 buffers) | `3 * chunk_rows_upper * value_dim * 4` | ″ |
| `gdn_chunk.beta_log_alpha` | `chunk_rows_upper * 2 * 4` | ″ |
| `gdn_chunk.gamma` | `chunk_rows_upper * 4` | ″ |
| `gdn_chunk.gamma_end` | `num_chunks_upper * value_heads * 4` | ″ |
| `gdn_chunk.{kkt,qk_l}` (2 buffers) | `2 * c2_upper * 4` | ″ |
| **SUM** | (all rows above) | **bounded by `MAX_GDN_SESSION_BYTES` = 2,147,483,648 bytes (2 GiB)** |

Real worst preset (`qwen36_27b`: value_heads=48, key_dim=value_dim=128) sums to ~134 MiB — 16x headroom under the bound. Both named hostile geometries (`value_heads=4096, key_dim=128, value_dim=32` → ~6.03 GiB aggregate; `key_heads=value_heads=1, key_head_dim=524288, value_head_dim=32, conv_kernel_dim=16` → ~6.01 GiB aggregate) pass every existing per-buffer/per-product guard individually and are now rejected by the aggregate sum. Regression tests: `parser_rejects_hostile_gdn_aggregate_value_heads_geometry`, `parser_rejects_hostile_gdn_aggregate_key_head_dim_geometry` (both device-free, config-level), `parser_accepts_realistic_gdn_aggregate_geometry` (accept-case control).

### CLASS A3 — every materialization site gets a byte budget + validate-before-materialize ordering (materialization-ordering class)

Round n's product budgets covered only GDN conv + RoPE. Enumerated every site that materializes/dequantizes a tensor whose size is a product of config fields:

| materialization site | product (config fields) | budget added | validated-before-allocate? |
|---|---|---|---|
| text embedding (`model/qwen35/loading.rs:434`, `embed_tokens` / `lm_head`) | `vocab_size * hidden_size` | `MAX_EMBEDDING_BYTES` (32 GiB) | y — checked in `Qwen35Config::validate()`, before any loader runs |
| vision `qkv_out` (`checked_derived_sizes`) | `hidden_size * 3` | `MAX_VISION_TENSOR_BYTES` (512 MiB) | y — `VisionModelConfig::validate()`, before `assemble` |
| vision `mlp_intermediate` | `hidden_size * 4` | `MAX_VISION_TENSOR_BYTES` | y |
| vision `merger_fc1` | `(spatial_merge_size^2 * hidden_size)^2` | `MAX_VISION_TENSOR_BYTES` | y |
| vision `merger_fc2` | `out_hidden_size * spatial_merge_size^2 * hidden_size` | `MAX_VISION_TENSOR_BYTES` | y |
| vision `patch_embed_weight` | `hidden_size * in_channels * temporal_patch_size * patch_size^2` | `MAX_VISION_TENSOR_BYTES` | y |
| vision `pos_embed` | `num_position_embeddings * hidden_size` | `MAX_VISION_TENSOR_BYTES` | y |
| Q4 vision dequantize (`vision/checkpoint.rs`'s `load_from_q4_dir`) | on-disk `.q4` header's declared `shape` product (independent of `vision_cfg`) | `MAX_VISION_TENSOR_BYTES`, checked **before** `dequantize_q4_to_f32` | **ordering fixed this round** — was: dequantize first, validate shape in `assemble` after; now: budget the header's declared shape before calling `dequantize_q4_to_f32` |

Ordering fix, exact call order: `load_q4_file` (reads header + packed bytes, small) → manifest-shape-vs-header-shape check (unchanged) → **NEW: `q4.shape` product byte-budget check** → `dequantize_q4_to_f32` (the actual TB-scale materialization, now gated) → `assemble`'s per-tensor expected-shape check (unchanged, still runs after, now on an already-budgeted buffer).

Regression tests: `parser_rejects_hostile_embedding_product`, `parser_accepts_realistic_embedding_product`, `parser_rejects_hostile_vision_derived_tensor_product` (device-free, config-level; asserts the validation error without allocating). The Q4 dequantize-ordering fix itself is exercised by the existing `loads_real_q4_checkpoint_with_correct_shapes` fixture test (confirms the new budget check doesn't reject an in-budget real checkpoint); a synthetic oversized-`.q4`-header regression test was not added this round due to the CPU-only/no-bench time window — filed as a follow-up (the code path is identical to the config-level budget check already covered by the vision derived-tensor tests above, just gated on the file header's `q4.shape` instead of `vision_cfg`'s fields).

### CLASS C — ingress enumeration via a type-level validated config (the class-collapsing fix)

Introduced `ValidatedQwen35Config`, a newtype only constructible via `Qwen35Config::validate(self) -> Result<ValidatedQwen35Config, InferenceError>` (or the equivalent `TryFrom<Qwen35Config>`), exposing read-only field access via `Deref`. `from_config_json_str` / `from_config_json` / `from_model_dir` keep their existing `Result<Self, InferenceError>` signatures (call the `_validated` sibling, unwrap) so their `metal_qwen35.rs` callers are untouched.

| ingress | file:line | validated this round? | how |
|---|---|---|---|
| `from_config_json_str` | `model/qwen35_config.rs` | y (already, round a) | now implemented as `from_config_json_str_validated(..)?.into_inner()` — same checks, now behind the newtype |
| `from_config_json` / `from_model_dir` | `model/qwen35_config.rs` | y | same pattern, `_validated` siblings added alongside |
| direct `serde_json::from_str::<Qwen35Config>` (bypasses the `HfQwenConfigFile` wrapper) | any caller | **y, NEW this round** | can only reach a non-Metal loader via `TryFrom<Qwen35Config> for ValidatedQwen35Config`, which runs the identical `validate()` checks |
| `model/qwen35/loading.rs`'s `load_weights` / `validate_required_tensor_names` | `model/qwen35/loading.rs` | **y, NEW this round** | signatures changed from `&Qwen35Config` to `&ValidatedQwen35Config` — a raw config cannot be passed at the type level |
| `Qwen35Model::from_safetensors` | `model/qwen35/model.rs` | **y, NEW this round** | calls `Qwen35Config::from_model_dir_validated`, passes the `ValidatedQwen35Config` straight through to both loader functions above |
| `Qwen35Model::config_mut()` (`&mut Qwen35Config` escape hatch) | `model/qwen35/model.rs` | **constrained, NEW this round** | killed; replaced with `set_eos_token_id(&mut self, u32)` — grepped every call site first (5, all `.eos_token_id = ..`), `eos_token_id` drives no allocation/indexing so no re-validation is needed for this one checked setter |
| `vision::checkpoint::load_qwen35_vision_weights` (takes `&VisionModelConfig`, not `Qwen35Config`) | `vision/checkpoint.rs` | already-bounded (pre-existing pattern) | already re-validates via `vision_cfg.validate()?` at its own boundary (existing code, unchanged) — out of CLASS C's `Qwen35Config` scope, noted for completeness |
| `MetalQwen35State::from_q4_dir` (`forward/metal_qwen35.rs:15063`, `:36667`) | `forward/metal_qwen35.rs` | **DEFERRED to the #1037 rebase** | #1037 is concurrently refactoring this exact function's mmap loaders; wiring it to `ValidatedQwen35Config` here would edit `metal_qwen35.rs`, violating this round's hard scoping constraint. The Metal sequencing is expected to resolve at rebase — `from_q4_dir` inherits `ValidatedQwen35Config` there, alongside #1037's own shape validation |

Regression test: `validated_config_rejects_hostile_directly_constructed_config` (builds a raw `Qwen35Config` directly — not via `from_config_json_str` — with a CLASS-A2 hostile GDN geometry, converts via `TryFrom`, asserts rejection; device-free). `validated_config_accepts_and_derefs_realistic_config` and `from_config_json_str_validated_matches_from_config_json_str` cover the accept path and the two entry points' equivalence.

### PATH CONTAINMENT — manifest-controlled shard names (sibling-path class)

Provenance, so this section is not read as claiming more than it should: the lexical guard
itself is largely already on `main` (`contained_shard_path` has 20 occurrences there against 21
here). What this branch adds is the open-once loader — `open_manifest_entry_once` and
`real_path_of_open_file` are absent from `main` entirely — plus routing the remaining call
sites, including `resolve_shard`, through the guard.

`contained_shard_path(model_dir, shard_file)` (`weights/f32_weights.rs`) validates a manifest-supplied entry name before it is joined to the model directory: an empty entry, an absolute path, and any component that is not `Normal` or `CurDir` — so `..`, and a Windows prefix or root — are rejected. The validation is purely lexical and consults no filesystem state, so there is no check-to-open window to race in the first place. Where the entry is then opened, `open_manifest_entry_once(model_root, entry_name)` validates, opens exactly once, and hands the caller the `File`; callers read from that handle rather than reopening the path, and that is what closes the check-then-reopen gap. It also returns the descriptor's real path (`fcntl(fd, F_GETPATH)` on macOS, `canonicalize` elsewhere), which is diagnostics and error text only — nothing security-relevant keys off it. Every manifest → path construction found by grepping `join(`/`PathBuf::from(` fed by a manifest/index-derived name routes through the guard:

| site | file:line | manifest source | routed through containment helper |
|---|---|---|---|
| `load_sharded` | `weights/f32_weights.rs` (`model_dir.join(&shard_file)`) | `model.safetensors.index.json` | y |
| `ShardedSafetensors::resolve_weight` | `weights/f32_weights.rs` | `model.safetensors.index.json` | y |
| `ShardedSafetensors::open_shard` | `weights/f32_weights.rs` | `model.safetensors.index.json` | y |
| `resolve_shard` (public fn, previously `PathBuf::from(shard_file)` with no join/containment at all — the previously cited unguarded site) | `weights/f32_weights.rs` | `model.safetensors.index.json` | y — signature changed to take `model_dir: &Path` (no external callers found; safe to change) |
| Q4 vision manifest join | `vision/checkpoint.rs` (`load_from_q4_dir`) | `quantize_index.json` | y |

**What the guard deliberately does not do.** It does not resolve symlinks and then require the resolved target to stay beneath the model root. Manifest strings are untrusted; the local filesystem is trusted. HuggingFace hub caches store snapshots as symlink farms (`snapshots/<rev>/x.safetensors -> ../../blobs/<sha>`) and lattice loads such directories directly, so a resolved-path containment assertion rejects every hub-cache checkpoint — pinned by `sharded_loader_follows_hub_cache_snapshot_layout`, which fails under exactly that policy. An actor who can write the model directory already controls every byte the loader reads, so refusing links there defends nothing. The full reasoning is a `# Threat model` doc comment on the function itself, next to the code a maintainer would be editing rather than only in a description they may never read.

Regression tests: `contained_shard_path_accepts_entries_beneath_model_dir` (accept-case control), `contained_shard_path_rejects_absolute_entry`, `contained_shard_path_rejects_parent_traversal`, `open_manifest_entry_once_rejects_traversal_before_opening` (asserts the lexical guard fires before any file is opened), and `load_sharded_rejects_hostile_index_with_traversal_entry` (end-to-end: a hostile `model.safetensors.index.json` with a `../` entry is rejected before the escaped file is opened). A dedicated `quantize_index.json`-with-`../`-entry test was not added this round — the unit tests cover the identical code path the Q4 vision site calls; filed as a follow-up for an end-to-end fixture matching `load_sharded_rejects_hostile_index_with_traversal_entry`'s shape.

### Fold dispositions

- **Doc comment promising `test_gdn_chunk_scratch_mirror_matches_engine_constant`** (`qwen35_config.rs:204`, doesn't exist tree-wide): **folded** — fixed the doc rather than writing the test. `metal_qwen35.rs`'s `GDN_CHUNK_SIZE` is a private `const` inside a `metal-gpu`-feature-gated module (`metal_qwen35.rs:1426`); a real cross-module drift-guard test would need to change that file's visibility, which this round's hard scoping constraint (`metal_qwen35.rs` untouched) forecloses. The doc comment (`qwen35_config.rs:211-217`, committed this round in `fb3ee233dd`) now states the manual-sync requirement plainly and tracks an automated `metal-gpu`-gated drift test as follow-up work, instead of citing a test name that doesn't exist.
- Dedupe the two bounded-vector deserializers' shared visitor/limit algorithm (`@665`): deferred (structural refactor, not a correctness gap; no new instance introduced this round).
- GDN scratch drift-risk comment (`@198`): unresolved (see doc-comment item above; deferred together).
- Shared tensor-name spec (`@loading.rs:12`): deferred, structural.
- Metadata-preflight double shape-clone (`@loading.rs:87-96`): deferred (perf-minor, not touched by this round's edits, out of the CLASS A/C/path-containment scope).
- Vision geometry derived twice (`@530`): **folded** — `checked_derived_sizes` now centralizes every vision derived-tensor byte formula in one `budget()` closure (CLASS A3), so this round's edit already addresses the duplication concern for the fields it touches.

### Metal untouched confirmation

```
$ git diff --stat origin/main HEAD -- crates/inference/src/forward/metal_qwen35.rs
(no output — 0 lines changed)
```

### Gates (this round)

- `cargo fmt -p lattice-inference` — clean, no diff after.
- `cargo clippy -p lattice-inference --all-targets --features f16,metal-gpu,test-utils -- -D warnings` — clean.
- `cargo check -p lattice-inference` (default) and `--features metal-gpu` — both clean.
- `cargo test -p lattice-inference --lib` (default config, unfiltered) — **2145 passed, 0 failed, 19 ignored** (proves the new config/loading/vision/path tests run un-gated).
- Targeted groups (`qwen35_config::`, `qwen35::loading::`, `vision::checkpoint::`, `weights::f32_weights::`, `model::qwen35::`) also run individually — all green.
- 9 pre-existing tests from prior rounds needed updates: 8 vision `_at_max` "accepts" tests flipped to "rejects" (a single vision field at its own generous per-field cap, composed with realistic co-fields, now overflows the new `MAX_VISION_TENSOR_BYTES` product budget — the per-field cap alone was necessary but not sufficient, which is exactly the class this round closes), 1 GDN conv-buffer test's assertion widened to accept either the pre-existing `MAX_GDN_CONV_BUFFER_SIZE` guard or the new, earlier-firing `MAX_GDN_SESSION_BYTES` aggregate guard on the same hostile input. `parser_accepts_hidden_size_at_max` was adjusted (paired with a small `vocab_size`) rather than flipped, since `MAX_EMBEDDING_BYTES`'s headroom (32 GiB) is generous enough to keep that specific at-max-in-isolation case accepted.
- Mutation-verified (Edit-only, no `git checkout`): `MAX_GDN_SESSION_BYTES`'s comparison (`if total_session_bytes > ...`) disabled via `if false && ...` → both hostile-geometry tests failed with the expected panic messages → restored byte-identical (confirmed via `sed` round-trip + `git diff` on that hunk showing no residual change) → tests green again. The manifest-path guard was re-checked the same way at the current head: `contained_shard_path`'s component loop neutered to `rel.components().take(0)`, which disables the `..` rejection while leaving the absolute-entry and empty-entry checks intact. The split was predicted before running and then observed — 5 passed became 2 passed / 3 failed, the three being `contained_shard_path_rejects_parent_traversal`, `open_manifest_entry_once_rejects_traversal_before_opening`, and `load_sharded_rejects_hostile_index_with_traversal_entry`, while the accept-case and absolute-entry tests stayed green because they exercise different checks. Restored byte-identical (matching sha256 before and after, `git diff` empty) → 5 passed again.
---

## Round: CPU-only ingress hardening (mechanical enumeration + FIX 3/4/5/6/7/8/9/14/15/16)

Commit: `59cdfb5a34` on top of `fb3ee233dd`. CPU-only: no GPU/Metal execution and no bench runs. Does **not** touch `crates/inference/src/forward/metal_qwen35.rs`, which belongs to #1037; the two Metal-side items (shape validation before GPU dispatch, and `from_q4_dir` taking a validated config) are **deferred until this branch rebases onto #1037** — both live entirely in `metal_qwen35.rs`.

### Mechanical enumeration (anti-hand-curation evidence)

Command (run verbatim, not hand-curated):

```
grep -rnE "vec!\[|Vec::with_capacity|with_capacity|\.to_vec\(\)|\.collect::<Vec|::zeros\(|dequantize|decode_|read_to_end|read_to_string|from_raw|reserve\(" \
  crates/inference/src/model/qwen35/loading.rs crates/inference/src/model/qwen35_config.rs \
  crates/inference/src/vision/ crates/inference/src/weights/f32_weights.rs
```

465 raw hits across 14 files. Breakdown and disposition:

| File | Hits | Disposition |
|---|---|---|
| `vision/checkpoint.rs` | 96 | 31 production (pre-`mod tests`); the two attacker-reachable materialization sites are `expected_names.contains(&entry.name)` gate (line ~272, **FIX 4**, guards the `dequantize_q4_to_f32` call at line 348) and `HashMap::with_capacity(expected_names.len())` (line 203, bounded transitively by `vision_cfg.depth` which `validate()` already caps). Remaining 65 are `mod tests` fixtures. |
| `weights/f32_weights.rs` | 83 | 64 production; all `Vec::with_capacity(num_layers)` / `Vec::with_capacity(qkv_out_dim * hidden_size)` etc. sites are bounded transitively by pre-existing `MAX_NUM_HIDDEN_LAYERS`/`MAX_HIDDEN_SIZE` caps in `Qwen35Config::validate()` (this round didn't touch those, they predate it). The two new-this-round sites are the index `read_to_string` at line 1141 (**FIX 5** sibling, now preceded by the metadata-length check at line ~1134) and `HashMap::with_capacity(reserve)` at line 1045 (**FIX 6**, `reserve` and the incremental insert both capped by `MAX_WEIGHT_MAP_ENTRIES`). Rest are test fixtures / `assert_eq!(..., vec![...])` literals. |
| `model/qwen35_config.rs` | 49 | 19 production (pre-`mod tests`); the `read_to_string` at line 429 inside `read_config_json_bounded` is **FIX 5** (guarded by the file-length check immediately above it, line 422). The doc comments at lines 153/303/1768/1808/1913 describe the GDN state/conv-buffer/cross-layer allocations that **FIX 9+14**'s new aggregate check (line ~1954) now bounds in total, not just per-field. Rest are `layer_mask`/`stop_token_ids` vecs sized by already-validated `num_hidden_layers`, or test fixtures. |
| `vision/vit.rs` | 47 | 21 production — this is the ADR-049 CPU ViT *scaffold* (explicitly documented in the module header as independent from the real checkpoint loader `checkpoint.rs`, targets a hypothetical model, not wired into any forward pass this round touches). All allocation sizes are functions of caller-supplied `d`/`d_mlp`/`n_patches` local test/construction parameters, not file-ingress-controlled. Out of scope for this round's ingress-hardening fixes. |
| `vision/multimodal.rs` | 28 | 18 production, all in `#[cfg(test)]`-adjacent builder/test-only structs or downstream-of-already-validated `GridThw`/`positions` construction; no new file-parsing ingress in this file. |
| `vision/qwen35_mrope.rs` | 26 | 7 production (`Vec::with_capacity(input_ids.len())`, `positions.positions.len()`, `rope_half`) — all sized by values already bounded upstream (sequence length is bounded by the serve-layer context-window clamp per the existing `feedback_serve-dos-hardening-bugclass` guard, not this loader). |
| `vision/pooled_embed.rs` | 26 | 3 production; `Vec::with_capacity(2 + num_pads + text_ids.len())` bounded by caller-supplied token counts, not raw file bytes. |
| `model/qwen35/loading.rs` | 25 | 3 production: `Vec::with_capacity(cfg.num_hidden_layers)` (line 461, bounded by pre-existing `MAX_NUM_HIDDEN_LAYERS`) and two `.to_vec()` calls in a `ShapeMismatch` error path (lines 96/104, error-message construction, not allocation of untrusted size). |
| `vision/qwen35_vit.rs` | 22 | 11 production, same ADR-049-adjacent CPU forward-pass allocations as `vit.rs`, sized by config fields already bounded by `VisionModelConfig::validate()`. |
| `vision/mod.rs` | 19 | 0 production (all hits are after `mod tests` at line 262). |
| `vision/qwen35_vit_metal.rs` | 17 | 6 production, `metal-gpu`-feature-gated CPU-side test-fixture builders and forward-pass buffers sized by validated config fields; no file-ingress parsing in this file. |
| `vision/merger.rs` | 17 | 4 production, spatial-merge MLP forward-pass buffers sized by validated `d_model`/`d_hidden`, not file-ingress. |
| `vision/qwen35_merger.rs` | 9 | 1 production (`pre_merger_hidden.to_vec()`, forward-pass copy of an already-materialized buffer). |
| `vision/preprocess.rs` | 1 | 1 production (`vec![0.0f32; n_patches * patch_len]`, sized by caller-supplied image dims, not raw file bytes — image-size ingress is a separate, pre-existing guard). |

The full raw hit list is not reproduced here for length. Every line above the fold in each file's "production" count was individually inspected against this disposition table.

### Fixes landed this round

- **FIX 5+6** (`qwen35_config.rs:422-429`, `f32_weights.rs:1128-1141,1043-1050`): `MAX_CONFIG_JSON_BYTES` (8 MiB) on both `from_config_json`/`from_config_json_validated` sibling read sites; `MAX_SAFETENSORS_INDEX_BYTES` (64 MiB) on the index file; `MAX_WEIGHT_MAP_ENTRIES` (1,000,000) enforced *incrementally during deserialization* (fires mid-parse, not after the map is fully built).
- **FIX 7** (`f32_weights.rs`: `contained_shard_path` + `open_manifest_entry_once` + `real_path_of_open_file`; `vision/checkpoint.rs` q4/f16 call sites): removes the check-then-reopen shape. The entry name is validated lexically, then opened exactly once, and callers read from the returned `File` instead of reopening the validated path — so no security decision is made about a path that is subsequently re-resolved. Public signatures called by `metal_qwen35.rs` (`SafetensorsFile::open`, `load_q4_file`, `load_f16_tensor_file`) are byte-identical to before this round; the open-once primitive is a `pub(crate)` sibling used only by the internal sharded-safetensors and vision-checkpoint call sites. **Convergence flag**: #1050 mirrored a canonicalize-then-reopen shape into `crates/inference/src/quarot/io.rs` — that file is out of scope for this round (per hard scoping) but needs the same open-once treatment at whatever round reconciles with this one.
- **FIX 8** (`qwen35_config.rs:~1521`): dense-FFN tensor-size budget `MAX_DENSE_FFN_TENSOR_BYTES` (4 GiB) on `hidden_size * intermediate_size * 4`, gated on `!is_moe()` since `load_dense_ffn_weights` is only reachable for that config shape.
- **FIX 9+14** (`qwen35_config.rs:~1954`): GDN cross-layer state aggregate budget `MAX_GDN_CROSS_LAYER_STATE_BYTES` (4 GiB) summing `(gdn_state_size + gdn_conv_buffer_size) * 4` across `num_linear_attention_layers`. Closes the gap where every existing *per-layer* GDN cap (state size, chunk scratch, session bytes) could each pass individually while the summed cross-layer total still blew a materialization budget — worst case in the new regression test hits ~8.4 GB aggregate against individually-compliant per-layer numbers.
- **FIX 15** (`qwen35_config.rs:~1658`): `rms_norm_eps` must be finite and `> 0.0` — rejects `0.0`, negative, NaN, `+inf`; realistic `1e-6` still accepted.
- **FIX 16** (`vision/checkpoint.rs:377` `assemble()`): vision MLP width now reads `vision_config.intermediate_size` when present, falling back to `4 * hidden_size` only when absent. Fixes the official Qwen3.5-VL dims (hidden=1152, intermediate=4304 — not `4 * 1152 = 4608`), which the old always-4x rule silently rejected.
- **FIX 3** (`vision/checkpoint.rs` fp16 path): reject an oversized *declared* tensor shape (byte-budget check against the safetensors header, no decode) before F16→F32 expansion. Scoped as a byte-budget check rather than full per-name expected-shape duplication (that logic stays centralized in `assemble()`, reached only after this budget passes) — a documented, deliberate scope-narrowing under this round's time budget.
- **FIX 4** (`vision/checkpoint.rs` q4 path): reject any manifest entry whose name isn't in the config-derived expected set *before* opening/dequantizing it (was previously only caught after-the-fact by `assemble()`'s "unconsumed leftover" check), plus an aggregate q4 tensor-count budget.
- **Minor/hygiene**: `qwen35_generate.rs` — replaced a stale `config_mut`-referencing comment with an accurate description of the actual eos-override mechanism.
- **Ledger only, no new instance this round**: the validated-config duplicate raw/validated parser+path APIs (`qwen35_config.rs:1220-1273`) remain a known duplication, tracked but not touched.

### Regression coverage — mutation-verified

Every new guard below was verified via Edit-only mutation (disable the guard's condition with `if false`/`if false && ...`, recompile, confirm the matching test **fails**, restore the file byte-identical via `diff` against a pre-mutation copy, re-run to confirm green — no `git checkout` used):

1. `qwen35_config.rs:422` (`file_len > MAX_CONFIG_JSON_BYTES`) — 1 test failed under mutation.
2. `qwen35_config.rs:~1521` (dense-FFN product) — hostile + realistic-accept pair, both exercised.
3. `qwen35_config.rs:~1658` (rms_norm_eps finite/positive) — 4 tests failed under mutation (zero, negative, NaN, +inf).
4. `qwen35_config.rs:~1954` (GDN cross-layer aggregate) — hostile + realistic-accept pair.
5. `f32_weights.rs:1134` (index-file size) — `parse_index_rejects_oversized_index_file` failed under mutation.
6. `f32_weights.rs:1047` (weight-map entry cap) — `parse_index_rejects_weight_map_over_entry_cap` failed under mutation.
7. `f32_weights.rs` — `contained_shard_path`'s `..`-component rejection. `contained_shard_path_rejects_parent_traversal`, `open_manifest_entry_once_rejects_traversal_before_opening`, and `load_sharded_rejects_hostile_index_with_traversal_entry` all failed under mutation, while the accept-case and absolute-entry tests stayed green.
8. `vision/checkpoint.rs:377` (vision MLP width) — `assemble_accepts_official_qwen35_vl_vision_mlp_dims` failed under mutation (shape mismatch 4608 vs 4304, exactly the bug this fix closes).
9. `vision/checkpoint.rs:272` (unexpected-tensor-name pre-check) — caught a real test-quality gap during verification: the original hostile-tensor test's assertion (`msg.contains("unexpected") && msg.contains(hostile_name)`) passed even with the guard disabled, because the tensor's own chosen name contained the substring "unexpected" and the post-decode fallback ("unconsumed leftover") error also names the tensor. Fixed by renaming the hostile tensor to `rogue_injected_tensor` (no "unexpected" substring) and asserting on the pre-check-specific phrase `"rejected before dequantizing/decoding"`, which only the pre-decode guard emits. Re-verified: fails under mutation, passes restored.

### Gates

- `cargo fmt --check`: clean.
- `cargo clippy -p lattice-inference --all-targets --features f16,metal-gpu,test-utils -- -D warnings`: clean.
- `cargo check -p lattice-inference` (default features) and `--features metal-gpu`: both clean.
- `cargo check --workspace --all-targets`: clean (caught one missed sibling struct-literal site for the new `VisionModelConfig.intermediate_size` field in `crates/embed/src/vision.rs:246`, fixed).
- `cargo test -p lattice-inference --lib`: **2161 passed, 0 failed, 19 ignored**.
- `cargo test -p lattice-embed --lib vision::`: 9 passed.

## Note on this description

The per-round disposition sections are consolidated into the single section below, and the
path-containment section is restated against the code at this head: it previously described
a canonicalize-and-contain design, naming a helper and tests that no longer exist. The guard
is lexical and follows symlinks deliberately, so hub-cache checkpoints load. Reinstating resolved-path
containment on the strength of an older revision of this text would undo a deliberate
decision; the reasoning and the pinning test are recorded in the `# Threat model` comment on
the function. The
mutation evidence for that guard was re-measured against the current code rather than
reworded. Every enumeration table and the field-by-field audit are byte-identical to the
previous revision.

This section supersedes the per-round disposition statements this description used to carry
(one per round, five in total, each restating the same argument for that round's symbols).

**Disposition: no A/B measurement, on structural grounds, with the A/B deferred to a
maintainer quiet-window run on the final head.**

The diff spans 17 paths — 16 source files plus `crates/inference/Cargo.toml` — so the argument
is made by the character of each change rather
than by a short file list. Grouped by what they do: config parsing and validation
(`model/qwen35_config.rs`); the cold checkpoint load path (`model/qwen35/loading.rs`,
`weights/f32_weights.rs`, `weights/q4_weights.rs` — whose hunks are confined to
`validate_q4_header_payload_bounds`, `load_q4_file`, `read_f16_header` and
`load_f16_tensor_file_expecting` — and `vision/checkpoint.rs`); the model construction and
config surface (`model/qwen35/model.rs`); call-site updates for a narrowed config accessor
(`model/qwen35/generation.rs`, a `config_mut()` to `set_eos_token_id()` rename in doc comments
and tests, and the same one-line rename in the `bin/qwen35_generate.rs` helper); single-line
struct-field initializers (`forward/cpu_f16.rs`, `embed/src/vision.rs`,
`quant/quarot/convert.rs`); the vision model and its tests; and the manifest
(`crates/inference/Cargo.toml`), which adds the macOS-only `libc` dependency used by
`real_path_of_open_file` to report an open shard's real path in diagnostics and error text. No
hunk lands in a decode or forward hot path, and no data structure traversed during generation
changed shape.

This code is default-compiled, so unlike a `cfg`-gated change it is genuinely present in the
bench binaries. The argument is reachability, not compiled-out. As a direct-reference check,
neither bench source names any of the changed entry points:

```
grep -c "from_config_json_str\|loading::\|contained_shard_path\|open_manifest_entry_once\|MAX_" \
  crates/inference/benches/elementwise_cpu_bench.rs crates/embed/benches/simd.rs
# 0  0
```

The added work is `O(1)` and `O(field count)` comparisons plus a few `u128` product and sum
budgets, executed once at parse and load time. Nothing is added to a decode or embed hot
path, and no data structure traversed during generation changed shape.

Stating the limit rather than claiming more than the evidence supports: that grep is a
*direct*-reference check. Zero textual matches means neither bench names these symbols; it does
not by itself exclude the bench reaching them indirectly through a call graph. The per-file
grouping above is what carries the load, and it is an argument about where the hunks sit rather
than a measurement of what the benches execute. Both together are weaker than an A/B, which is
why the measurement is deferred rather than declared unnecessary.

### f16 end-to-end gate (this round)

The gemma4 end-to-end forward gate was run with the `f16` feature it requires:

```
test boundary_golden_fixture_is_valid ... ok
test stage5_golden_fixture_is_valid ... ok
test stage5_e2e_greedy_and_per_layer_probe_match_hf_golden ... ok
test stage5_sliding_window_boundary_matches_hf_golden ... ok
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 3646.76s
```

Earlier runs of this target failed for want of that feature rather than because of this
branch. The check that settles which is structural rather than a matter of judgement: the
test file is byte-identical to main.

```
$ git diff --stat origin/main -- crates/inference/tests/gemma4_e2e_forward_test.rs
(no output — 0 lines changed)

$ git diff --stat origin/main -- crates/inference/src/weights/f32_weights.rs
 crates/inference/src/weights/f32_weights.rs | 284 +++++++++++++++++++++++++---
```

The second command is the control: the same instrument reports a large diff for a file this
PR does change, so the empty output above it is evidence of no change rather than evidence of
a command that reports nothing either way.


ohdearquant:
Merged with administrator privileges, bypassing this repository's required approving review, which could not be satisfied by any account with access here. The required checks were verified green per-commit at the merged head `8abe0ff990c62c3c9fdc70aed876d12e596c24ca` before the merge, and the squash message was authored explicitly rather than composed from the branch commits.

PR #1056, opened 2026-07-18
docs(inference): state durable rationale in code comments

ohdearquant:
## Summary

Rewrites source comments across `crates/inference` that recorded transient authoring
history into the durable technical properties they document. Issue, ADR and RFC
references, type and function names, and all measured values are preserved. Only the
process framing is removed.

Comment-only: no executable code, no signature, and no test assertion changed. A single
`assert_eq!` failure *message* was also reworded (the descriptive text shown on failure,
not the assertion itself).

## Test plan

- `cargo fmt -p lattice-inference` — no changes
- `cargo check -p lattice-inference --all-targets --features f16,metal-gpu` — pass
- `cargo clippy -p lattice-inference --all-targets --features f16,metal-gpu -- -D warnings` — pass, zero warnings
- `cargo doc -p lattice-inference --no-deps --features f16,metal-gpu` — pass

Every changed line was programmatically verified to be a comment. The one line that mixes
code and comment (`#[allow(dead_code)] // ...`) was confirmed byte-identical in its
attribute text, with only the trailing comment altered.

## bench-compare disposition

No hot-path call sites touched. Comments do not affect generated code, so there is no
behavioral delta to measure and an A/B comparison is structurally inapplicable here.

### DIFF
--- a/crates/inference/src/attention/flash.rs
+++ b/crates/inference/src/attention/flash.rs
@@ -354,8 +354,7 @@ fn write_head_to_interleaved(
 /// the full-layer V-transpose buffer (`v_all_t`) added by #674/#673: it is
 /// *not* the field set from before those changes. `estimate_matches_actual_attention_buffer_lengths`
 /// below cross-checks this estimate against `AttentionBuffers::new`'s real
-/// field lengths so this can't silently rot again the way it did across #678
-/// review finding 3.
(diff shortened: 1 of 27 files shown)


oceanwaves630 (approved):
Reviewed and approved.

PR #1057, opened 2026-07-19
fix(ci): load the parity reference from the verified snapshot

ohdearquant:
## What this changes

The e2e parity gate compares lattice's greedy output against a HuggingFace transformers
reference. Before this PR the reference half of that comparison was resolved by bare model id
at run time, with `trust_remote_code=True`, so the thing lattice was measured against was
whatever the Hub served at that moment, and it was allowed to execute repository code while
loading. A gate whose reference can move is not a gate: a change upstream reads as a lattice
regression, and the failure lands on whichever PR happens to be open.

Three changes, all CI wiring, no library code:

1. **The reference loads from the provisioned snapshot.** `scripts/e2e_parity_check.py` now
   loads tokenizer and model from `LATTICE_MODEL_DIR` with `local_files_only=True`,
   `trust_remote_code=False` and `use_safetensors=True`, and fails as a setup error when that
   directory is absent. `HF_MODEL_ID` is gone, because a model id is by definition a
   late-bound reference. Both parity run steps also set `HF_HUB_OFFLINE` and
   `TRANSFORMERS_OFFLINE`, so any transformers path that still tries to resolve against the
   Hub fails loudly instead of quietly fetching something nobody verified.

2. **The provisioned snapshot is an exact set, not a superset.** Detailed below.

3. **The cache key is derived from the provisioning action.** All six consuming jobs keyed
   their model cache on the literal `hf-qwen35-0.8b-v1`, which does not change when the pinned
   revision or the expected digests do. They now key on
   `hashFiles('.github/actions/provision-qwen35-0-8b/action.yml')`, so bumping the pin
   invalidates the cache holding the old snapshot. A hand-bumped key is a step someone
   forgets, and a stale snapshot restored under a still-matching key is exactly what the
   verification would then report as tampering.

## Scope reduced after the provisioning action landed

Most of the pinning work this branch started with was superseded by
`.github/actions/provision-qwen35-0-8b` on main, which pins the revision to a full commit and
hash-verifies the snapshot from one place shared by six jobs. That is the better shape, so
this PR adopts it wholesale rather than competing with it, and the diff is now confined to
what it does not yet cover.

Its pinned digests were checked rather than assumed before resolving that way. The weight
shard and `tokenizer.json` digests are byte-identical to the values published for the pinned
revision, and both non-LFS digests match bytes independently attested by their published git
blob hashes.

## The larger gap: verifying a list is not verifying a directory

Hash-pinning a list of files says nothing about files that are not on the list, and
transformers reads more of a model directory than the weights. Adding a
`generation_config.json` to the snapshot (the pinned revision does not ship one) makes the
loader apply `suppress_tokens`, `bad_words_ids` or `sequence_bias` at generate time, which
moves the first greedy token, which is precisely what this gate compares. Every pinned digest
still matches, because the added file is not on the list.

Reproduced against the pinned snapshot: the reference continuation for "The capital of France
is" moved from `" Paris."` to `" the capital of"` while `shasum -c` returned success. The gate
would have measured lattice against a reference someone else chose and reported the difference
as a lattice regression. `local_files_only`, `trust_remote_code=False` and `use_safetensors`
do not constrain a JSON policy file.

The verify step now checks contents and completeness, with the completeness allowlist derived
from the same manifest so the two cannot drift on a future revision bump.

Two details that took a while to get right:

- **The materialized set has to be fixed before it can be asserted.** `snapshot_download`
  without `allow_patterns` fetches whatever the installed `huggingface_hub` decides to fetch.
  The pinned revision's tree also carries `.gitattributes`, `LICENSE` and `README.md`, so a
  directory populated by a newer hub holds thirteen entries where one populated by an older
  hub holds ten. An exact-set assertion over a set that varies goes red on ordinary runs, and
  a restored cache and a fresh download can legitimately differ. The download now names its
  ten files explicitly, so the download list, the digest manifest and the allowlist are one
  and the same list.

- **The comparison has to be entry-safe.** `ls` piped through `grep -vxF -f allowlist` is not:
  a filename containing a newline is read by `grep -F` as several independent patterns, one of
  which matches an allowlisted name, so the entry slips through. Entries are enumerated with
  `find -print0` and compared one NUL-delimited record at a time by exact string equality.
  `|| true` is likewise avoided, since it turns a failure of the checker itself into a pass.

## One gap the action did not cover

A `.bin` weight file is a pickle, so it stays an arbitrary-execution vector at load time
regardless of `trust_remote_code=False`. The script refuses a snapshot carrying one, and does
it in the preflight rather than in the loader. The script's documented exit contract is `0`
pass, `1` parity failure, `2` setup error; raising from the loader surfaced as `1`, which
would have filed a compromised reference as a lattice regression. That is the exact
misattribution this PR exists to prevent, so the guard against it must not commit it.

Five more instances of that same category error, found while checking the first:

- The reference generation call was unguarded, so *any* exception raised while producing the
  reference ended the process on a traceback, and a traceback exits `1`. A snapshot that will
  not load, a transformers version that refuses it, or a directory that disappeared after the
  preflight all landed in the parity-failure category. The call is now wrapped and returns `2`
  with the reason stated.
- The pickle check used a non-recursive glob. Transformers resolves sharded and subfoldered
  weights, so a nested `.bin` is loadable and was not seen. CI's snapshot is flat and the
  exact-set verifier rejects a subdirectory before this runs, but the script also runs by hand
  against directories nothing verified, and that is where the nested case lives. Now `rglob`.
- `E2E_MAX_TOKENS` was parsed at module import, so a malformed value raised before `main()`
  began and the traceback exited `1`. Parsed defensively and validated in the preflight.
- A non-positive `E2E_MAX_TOKENS` was accepted. Zero tokens compares two empty sequences and
  reports as a parity failure, which sends a workflow typo to whoever is on the hook for a
  lattice regression. Rejected as a setup error, and the check runs first in the preflight
  because it needs no filesystem and its diagnosis is exact.
- An unwritable `E2E_REPORT_PATH` raised `OSError` after every prompt had already matched,
  turning a clean run into a reported parity failure. Wrapped.

Anything else that aborts the harness would inherit the interpreter's exit code of `1`, so the
entry point now classifies an uncaught exception as a setup failure and prints the traceback.

The boundary runs the other way too, and getting it wrong in that direction is just as bad. A
lattice binary that times out, exits non-zero, or emits output the script cannot read has
failed to produce a usable result, and that is a lattice failure, not a setup error. Reporting
it as setup would tell whoever reads the run that their environment is broken when the thing
under test is what broke, and catching that is the point of the gate. Three places got this
wrong and now agree: the CPU runner had no timeout handler where the Metal runner did, and
neither runner had a boundary around the conversions it performs after recognising its own
output format, so a non-integer token list, a speed field of `1.2.3` (the regex is `[\d.]+`)
and a non-numeric `tok_s` all escaped as setup errors while a missing token list and malformed
JSON in the same runners already reported correctly. The parser boundary is narrowed to
`ValueError` and `TypeError` so a launch-side `OSError` still lands as setup.

## Verification

Fail-closed matrix executed under `/bin/bash` 3.2, which is what the macOS runner ships,
against a fixture built the way CI builds one: a clean download at the pinned revision with
CI's pinned `huggingface_hub==1.20.1` and the action's own `allow_patterns`.

| Case | Result |
| --- | --- |
| clean snapshot | passes |
| modified file | fails, named in the `shasum -c` output |
| removed file | fails, named in the `shasum -c` output |
| added `generation_config.json` | fails as an unverified entry |
| added hidden file | fails as an unverified entry |
| added subdirectory | fails as an unverified entry |
| entry name containing a newline | fails as an unverified entry |
| entry name beginning with a dash | fails as an unverified entry |
| re-added repository metadata file | fails as an unverified entry |
| allowlist degraded to empty | fails closed rather than passing vacuously |

An earlier revision of this branch was validated against a pre-existing local snapshot that
happened to hold ten of the thirteen files a clean fetch brings, which is how a manifest that
would have reddened every parity run passed local testing. The inventory question is settled
outright now that the download names its files.

The `allow_patterns` list and the digest manifest were compared programmatically and are
identical. All six cache keys were confirmed present and identical after the change.

Exit codes, measured with the binary preflight taken out of the way so both arms of each pair
are actually distinguishable:

| Case | Exit | Mutated out |
| --- | --- | --- |
| missing model directory | `2` | n/a |
| flat `.bin` in the snapshot | `2` | `1` |
| nested `sub/weights.bin` | `2`, named by relative path | not caught |
| snapshot with `config.json` removed | `2`, reason stated | `1` |
| same snapshot without the `.bin` | passes the preflight, proceeds into the load | n/a |
| `E2E_MAX_TOKENS` malformed, empty, blank, `0`, `-3` | `2`, each with its own message | n/a |
| `E2E_MAX_TOKENS` valid | falls through to the next check | n/a |
| unwritable `E2E_REPORT_PATH`, all prompts matching | `2` | `1` |
| writable `E2E_REPORT_PATH`, same stubs | `0`, report written | n/a |
| injected harness exception | `2` | `1` |
| lattice timeout, cpu and metal | `1` on both | cpu escapes uncaught |
| non-integer token list, speed `1.2.3`, non-numeric `tok_s` | `1` | escape uncaught |
| well-formed output, cpu and metal | `0` on both | n/a |

Two of those measurements were worthless on the first attempt and had to be redone. The `.bin`
pair returned `2` on both arms because a missing lattice binary tripped an earlier preflight;
the report-path pair returned `1` on both arms because `runpy.run_path` hands back a copy of
the module globals, so the stubs never took effect and the real binary ran. The tell was the
same both times: a guard test that returns the same code either way is measuring the harness,
not the guard. The rows above are from runs with a control that behaves differently.

## bench-compare disposition

Not applicable. No file under `crates/inference/` or `crates/embed/` is touched. The diff is
confined to `.github/` and `scripts/`, none of which is compiled into any benchmark binary.


oceanwaves630 (approved):
Approving on the evidence: e2e parity green on two consecutive commits (016d0d36, c5ee9904 — 31/31 both), provisioning hardening validated end-to-end on a real macOS runner, exit-code taxonomy fixed with mutation-verified controls on both backends, and the current head fully green. The remaining malformed-output class (e.g. a 5001-digit token_id) is unreachable from a correctly built producer and is rightly out of scope.

PR #1058, opened 2026-07-19
ci(bench): require a bench-compare disposition on perf-sensitive PRs

ohdearquant:
## Summary

ADR-058 requires every PR touching `crates/inference/` or `crates/embed/` to carry a bench-compare disposition in its description. That was convention-only, and it was missed: #966 added one line to the cosine fast path and merged with no bench section at all.

That line was `&& is_unit_norm(s)` in `crates/embed/src/simd/tier.rs`, an O(d) norm recomputation over the *stored* vector on every candidate comparison. From this repo's own `perf-baselines` history, `simd_prepared_query_normalized_cosine/prepared_meta_unit`:

| size | before | after | delta |
|---:|---:|---:|---:|
| 384 | 39,236 ns | 261,076 ns | +565% |
| 768 | 73,303 ns | 514,384 ns | +602% |
| 1024 | 96,278 ns | 682,838 ns | +609% |

Confidence intervals disjoint at every size, in one snapshot transition, undetected for four days. This adds a gate so an absent disposition cannot merge.

## What the gate does

A single always-running job reports one required status context. When the PR changes `crates/inference/` or `crates/embed/`, it requires the description to contain a bench-compare disposition with substance; otherwise it passes. It accepts all three dispositions the contributing guide allows: an A/B table, an explicit no-change result with p-values, or the structural argument that the diff is compiled out of the bench binaries.

It does **not** judge whether the numbers are correct — a workflow cannot, and pretending otherwise would trade a visible gap for an invisible one. Reviewers keep that job; this only makes an absent disposition impossible.

## Design

- **The gate cannot be edited away by the PR under test.** It runs as `pull_request_target`, so both the workflow and the checker script (`scripts/bench_disposition_check.sh`) come from the base branch, never the PR's tree. The job checks out only the base ref and executes no PR code: the changed-file list and the body are read through the GitHub API, and the body is piped as inert text to the base-controlled checker. Nothing from the PR runs, so `pull_request_target`'s usual write-token hazard does not apply.
- **Crate-wide scope.** Matches all of `crates/{inference,embed}/`, not just `src/`, per ADR-058 — a `Cargo.toml` opt-level bump or a `benches/` change moves performance too. A doc-only change under those crates satisfies the gate with a one-line "no source change, N/A".
- **Heading-anchored parser.** The disposition section must open on a Markdown heading containing "bench-compare", not any prose mention, so a Summary sentence like "run make bench-compare" no longer satisfies it. The section runs until a heading at the same or shallower depth, so subheadings under it count as content.
- **Merge queue.** Triggers on `pull_request_target` and `merge_group` under the same context name; on `merge_group` there is no body to check (the PR-time gate already ran) so it passes. Registering the context as required in branch protection is a one-time repo-admin step.

## Test plan

The checker is a committed script with committed tests: `tests/test_bench_disposition.py` runs it against real and adversarial descriptions and is wired into CI. Cases: #966's empty section fails; a real heading with numbers passes; empty body fails; a bare `## bench-compare` with nothing beneath fails; **prose-only "make bench-compare" with no heading fails**; a mixed-case heading passes; subheadings under the section count as content. `python3 tests/test_bench_disposition.py -v` → 6/6. `actionlint` clean.

Two portability defects were fixed while writing the checker: `IGNORECASE` is a gawk extension that mawk (the Ubuntu runner default) silently ignores → `tolower()`; `${BODY//.../}` is pathological on large inputs (a 41 KB body hung) → `grep`.

## bench-compare disposition

Not applicable: this PR adds a workflow, a shell script, and a Python test, and touches no Rust source, so nothing in `crates/inference/` or `crates/embed/` changes. The gate does not fire on itself.

### DIFF
--- a/.github/workflows/bench-evidence.yml
+++ b/.github/workflows/bench-evidence.yml
@@ -8,18 +8,21 @@ name: bench evidence
 # confidence intervals disjoint at every size, undetected for four days because
 # nothing enforced the rule.
 #
-# This gate checks that the disposition EXISTS and says something. It does not judge
-# whether the numbers are real; a workflow cannot, and pretending otherwise would
-# trade a visible gap for an invisible one. Reviewers judge quality; this makes an
-# absent disposition impossible.
+# Trusted-enforcement shape. This runs as pull_request_target, so the workflow
+# definition and the checker script both come from the BASE branch, never the PR's
+# tree — a PR cannot edit this job (or scripts/bench_disposition_check.sh) to exit 0
+# and forge the named check. It checks out only the base ref and runs no code from the
+# PR: the PR body and file list are read through the GitHub API, and the body is piped
+# as inert text to the base-controlled checker. Nothing from the PR executes here, so
+# pull_request_target's usual "untrusted code with a write token" hazard does not apply.
 #
-# Change detection uses the GitHub API, not a script from the PR checkout: a gate that
-# runs PR-controlled code to decide whether the gate runs is a fail-open, since a PR
-# could rewrite that script to report no perf-sensitive paths. Nothing from the PR's
-# tree executes here.
+# This gate checks that the disposition EXISTS and says something. It does not judge
+# whether the numbers are real; a workflow cannot, and pretending otherwise would trade
+# a visible gap for an invisible one. Reviewers judge quality; this makes an absent
+# disposition impossible.
 
 on:
-  pull_request:
+  pull_request_target:
     branches: [main]
     # `edited` is required: the body is the artifact under test, so a description fixed
     # after the fact must be re-evaluated rather than needing an empty push to clear.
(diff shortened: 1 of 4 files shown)


oceanwaves630 (approved):
Full read of the workflow file plus independent verification: pull_request_target runs base-ref checkout pinned to a full SHA with read-only permissions and the PR body piped as inert text, so no PR code executes. I confirmed via the rulesets API that the master ruleset carries no merge_queue rule, so the merge_group green-shortcut concern is moot, and the type-6 masker plus column-zero heading narrowing close the parser defect classes found across the adversarial rounds. The remaining type-7 boundary is documented and outside the honest-mistake threat model this gate exists for. Approving; auto-merge takes it from here.

Issue #1069, opened 2026-07-19
security(loaders): index-controlled shard paths escape model_dir across four read sites (canonicalized containment missing)


ohdearquant:
## Problem

Every sharded-checkpoint reader joins file names taken from `model.safetensors.index.json` (or the Q4 index) directly onto the model directory and opens/mmaps the result. `Path::join` replaces the base entirely when the component is absolute, and relative `..` components traverse out, so a malicious checkpoint index can direct the engine to open, mmap, or hash any path readable by the process.

## Sites (all the same class; the converter's output-write path is already guarded by `sanitize_tensor_name`)

- `crates/inference/src/quant/quarot/io.rs:852` — `Shard::open(&model_dir.join(shard_file))`, QuaRot conversion reader; shard is subsequently memory-mapped
- `crates/inference/src/weights/f32_weights.rs:1125` (and the streaming variants at `:1169`, `:1182`) — runtime f32 sharded loader
- `crates/inference/src/bin/lattice.rs:545` — CLI shard merge path
- `crates/inference/src/model/qwen.rs:408` — weight fingerprinting opens and hashes the joined path

## Threat model

A downloaded/untrusted model directory is the input. Absolute or `..`-bearing entries (or a symlinked shard) reach filesystem paths outside the checkpoint: device files or FIFOs can hang or crash the loader, and reads of unrelated files become confused-deputy inputs to conversion/fingerprinting. Same malicious-checkpoint hardening family as the quantizer non-finite guards (#452).

## Fix shape

A shared helper that resolves `model_dir.join(name)`, canonicalizes both sides, and requires the result to remain beneath the canonicalized `model_dir` (rejecting absolute entries, traversal, and symlink escapes) — applied at all four read sites. Tests: absolute path entry, `../` entry, and a symlink pointing outside the directory, each rejected with a typed error naming the offending index entry.

## Provenance

Surfaced during review of the Metal grammar-completion fix (out-of-diff finding, verified at source against all sibling join sites before filing).

Issue #1064, opened 2026-07-19
Metal grammar decode: completed grammar state misclassified as GrammarConstraintBlocked (strict JSON 500 on success)


ohdearquant:
### Symptom

A strict `json_schema` generation on the Metal path that produces a complete, valid document can be rejected as `GrammarConstraintBlocked`, which `lattice_serve` maps to `blocked_constraint` / HTTP 500. The generation actually succeeded; the error fires one step later.

### Root cause: the CPU/Metal grammar contract is only half-mirrored

The CPU paths implement the #611 fail-closed contract in two halves:

1. **Fail closed** when the grammar blocks every token in a state that is *not* complete (`has_finite_logit` guard).
2. **Terminate cleanly** when the state is complete with no legal continuation — both by checking `engine.is_complete_without_continuation(gs)` after every `advance` (e.g. `crates/inference/src/model/qwen35/generation.rs:222`) and by exempting accepting states at every all-blocked guard (`generation.rs:189`, 640, 815, 1023, 1910, 2038):

```rust
if !has_finite_logit(&scratch.logits[..cfg.vocab_size]) {
    if engine.is_complete_without_continuation(gs) {
        return Ok(grammar_output(...));   // clean stop
    }
    return Err(InferenceError::GrammarConstraintBlocked(...));
}
```

The Metal path (`crates/inference/src/forward/metal_qwen35.rs`) mirrors only half 1. All six `GrammarConstraintBlocked` sites (9247, 9408, 14183, 14354, 17081, 17292) fail closed unconditionally, and the file contains no call to `is_complete_without_continuation` (or any completion check) at all.

Sequence on Metal for a schema whose accepting state has no legal continuation (typical strict JSON: nothing may follow the final `}`):

1. Model samples the final `}`; `advance` succeeds; loop continues.
2. Next iteration: `forward_step` + `mask_logits` on the now-complete state masks **every** token to `-inf` (the engine has no EOS special-casing; EOS simulates as illegal bytes and empty-byte tokens are always blocked).
3. `!has_finite_logit(...)` → `Err(GrammarConstraintBlocked)` → `blocked_constraint` / 500.

There is also no post-`advance` completion break in the Metal decode loops, so the completed state always survives to the next mask. Note the mask itself never leaves EOS legal under an active grammar, so a grammar-constrained Metal generation cannot terminate via `GrammarStop`/`Eos` either; completion detection is the only clean exit, and it is missing.

### Why tests did not catch it

[...]

PR #1071, opened 2026-07-19
fix(inference): classify grammar completion on the first and final Metal decode tokens

ohdearquant:
## What

Follow-up to #1067, closing a boundary case its fix missed: a grammar completed by the FIRST generated token with `max_new_tokens == 1` (zero decode-loop iterations), or by the FINAL allowed token (no following loop top), still misreported `stopped: false` / `StopReason::Length` on all three Metal paths. A serve caller doing strict structured output with a tight token cap would see `finish_reason: "length"` for a complete JSON object.

## How

Mirrors the CPU contract's check placement exactly (`model/qwen35/generation.rs`: `grammar_complete` recorded at the initial advance; `grammar_complete_without_continuation` after every emitted token):

- **Post-initial-advance**: completion is recorded into `stopped` / `stop_reason` right where they are declared, immediately after the initial grammar advance, and routed through each path's existing common exit — no duplicated return blocks, and the zero-iteration loop case is correct by construction. On the prefix-cache path the completing token was never forwarded, so `stopped` skipping the silent step keeps the saved cross-turn prefix consistent with live KV state.
- **Post-in-loop-advance**: completion is re-checked after every in-loop advance (after the token's pushes/delivery, before the KV-full check), so completion on the final allowed iteration classifies as Grammar, and Grammar wins over KvFull on the same token, matching CPU ordering.
- The loop-top grammar probe from #1067 reduces to a plain `if stopped` break (the initial-completion case for caps > 1), removing the per-token grammar-state probe from the decode hot path.

Stop-precedence ordering is CPU-identical: EOS wins at the initial token, stop-strings win over completion, completion wins over KV-full and answer-budget.

## Tests

Six new regressions plus a shared grammar-config helper:

- `*_max_one_token_completed_grammar_is_success` × 3 paths (`root ::= "a"`, `max_new_tokens: 1`)
- `*_final_token_grammar_completion_is_grammar_stop` × 3 paths (`root ::= "a" "a"`, `max_new_tokens: 2`, completion lands exactly on the final allowed iteration)

Every added check is pinned by a test that fails without it, verified per site on all three paths: removing a post-initial-advance recording turns that path's `max_one_token` test red (Length instead of Grammar); removing an in-loop post-advance check turns that path's `final_token` test red the same way; removing the loop-top `if stopped` break restores the GrammarConstraintBlocked misclassification caught by the existing `max_new_tokens: 2` completion tests.

## Verification

- `cargo clippy -p lattice-inference --all-targets --features f16,metal-gpu -- -D warnings` clean
- Grammar sweep on the Metal device (`LATTICE_METAL_TEST_ENFORCE=1`, `--test-threads=1`): 14/14 green (12 Metal completion/dead-end/one-token/final-token + 2 CPU parity tests matching the filters)
- Full `forward::metal_qwen35` module sweep in flight; will report the result on this PR before marking ready. The follow-up commit after the sweep started changes comments only.

## Related

A shard-path containment gap in the QuaRot sharded loader noticed while working this area is tracked as #1069 (out of this diff) and fixed by #1072.

## Bench disposition

The diff is confined to `#[cfg(all(target_os = "macos", feature = "metal-gpu"))] mod inner` in `crates/inference/src/forward/metal_qwen35.rs`. The bench binaries build without the `metal-gpu` feature, so base and head bench binaries have identical effective source; no A/B can attribute a delta to this diff. Within the compiled paths, the new per-token work is one `is_complete_without_continuation` probe per emitted grammar-constrained token (replacing the identical per-token loop-top probe from #1067) and a single boolean branch per iteration; non-grammar decode is structurally unchanged.

Closes the remaining half of #1064.

### DIFF
--- a/crates/inference/src/forward/metal_qwen35.rs
+++ b/crates/inference/src/forward/metal_qwen35.rs
@@ -9589,17 +9589,25 @@ mod inner {
             // Autoregressive decode
             let mut stopped = false;
             let mut stop_reason = StopReason::Length;
+            // A grammar completed by the prefill-derived first token with no
+            // legal continuation is a successful stop, mirroring the CPU
+            // `grammar_complete` early return. Recorded here — immediately
+            // after the initial advance — rather than probed at the loop top,
+            // so a zero-iteration decode loop (max_new_tokens == 1) cannot
+            // fall through to Length (#1064 follow-up).
+            if let (Some(engine), Some(gs)) = (&gen_cfg.grammar, &grammar_state)
+                && engine.is_complete_without_continuation(gs)
+            {
+                stopped = true;
+                stop_reason = StopReason::Grammar;
+            }
             for _ in 1..gen_cfg.max_new_tokens {
-                // A grammar state that is complete with no legal continuation is
-                // a successful stop. Checked before the forward step, mirroring
-                // the CPU paths: past this point mask_logits would block every
-                // token and the all-blocked guard below would misreport the
-                // completed generation as GrammarConstraintBlocked (#1064).
-                if let (Some(engine), Some(gs)) = (&gen_cfg.grammar, &grammar_state)
-                    && engine.is_complete_without_continuation(gs)
-                {
-                    stopped = true;
-                    stop_reason = StopReason::Grammar;
+                // Initial-token grammar completion (recorded above) terminates
+                // decode before any per-step GPU work: past this point
+                // mask_logits would block every token and the all-blocked
+                // guard below would misreport the completed generation as
+                // GrammarConstraintBlocked (#1064).
+                if stopped {
                     break;
                 }
                 if self.session.kv_cache.seq_len >= self.session.kv_cache.max_cache_len {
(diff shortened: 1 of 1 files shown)


ohdearquant:
Full-module sweep at head 91d755a2ba: `cargo test -p lattice-inference --features f16,metal-gpu --lib forward::metal_qwen35 -- --test-threads=1` under `LATTICE_METAL_TEST_ENFORCE=1` → **299 passed / 2 failed / 3 ignored** (durable log, per-test lines).

Both failures triaged solo (one test, `--test-threads=1`, idle GPU, 0.3s each) and neither is attributable to this branch:

1. `cross_turn_cache_chat_completion_matches_full_reprefill` — deterministic **pre-existing failure on main**: fails identically at this branch's head, at the merge-base (#1068), at current main (3f5d3e4f2c), and at #1022's parent-era. This branch's diff touches only grammar-completion classification with zero reach into the chat-template/window code the panic implicates. Filed with the full SHA table as #1076.
2. `f16_kv_metal_path_executes_and_reports_capability` — **harness-invocation gap in the sweep, not a defect**: the test's fail-closed design (#252) requires the `LATTICE_KV_F16=1` + `LATTICE_METAL_PATH_PROOF=1` env contract that ci.yml's dedicated leg sets; the sweep set only `LATTICE_METAL_TEST_ENFORCE=1`. With the full contract it passes at this head in 0.15s.

Every test this PR's diff touches is in the 299 passing.

Issue #1060, opened 2026-07-19
Bench integrity: falsification-sensitivity audit, baseline rebuild, and coverage of recent merges


ohdearquant:
## Motivation

ADR-058 requires an A/B `make bench-compare` disposition on every PR touching
`crates/inference/` or `crates/embed/`. That process assumes each benchmark actually
measures the code path it names. Two gaps break that assumption:

1. **No mechanical check that a bench exercises its real path.** A Criterion group can
   report stable, plausible numbers while measuring a stub, a simplified copy, a
   precomputed constant, or a symbol the production path no longer calls. Nothing today
   catches that; the numbers look real because they are stable, not because they track
   the kernel.
2. **Committed baselines are uneven.** `perf-baselines` is trend data updated on merge,
   not a uniformly present, methodology-documented reference. A regression is only
   visible against a baseline that exists and is comparable.

This issue tracks a program to close both, and extends the repo's existing
**"regression tests must be mutation-sensitive"** principle (AGENTS/CLAUDE.md) to
benchmarks: a bench that does not move when you slow its measured path down is
decoration, exactly as a regression test that passes with the fix reverted is
decoration.

## The falsification-sensitivity standard

For every Criterion group `G` that claims to measure kernel/function `f`:

1. Identify the exact region inside `b.iter(|| …)` and the production `f` it calls.
2. Inject a calibrated, unmistakable slowdown into `f`'s real source (not the bench) —
   duplicate the hot loop (≈2× work) or add a `black_box`-guarded spin of known cost.
3. Re-run `G`. **PASS** if the mean shifts far outside the reported CI (a ≈2× injection
   should move the mean ≈+100%; disjoint CIs required). **FAIL** if the mean does not
   move: `G` is not measuring `f`.
4. Revert the injection; confirm `G` returns to baseline (symmetry check).

A FAIL classifies the bench: stub / simplified-copy, constant / precomputed,
dead-code / wrong-symbol, or optimized-away (missing `black_box`). Any FAIL bench is
rebuilt to measure the real path, then must pass the standard before its numbers count.

## Harness

A falsification harness implements the standard and lands via PR once validated:

- **Build/measure split.** All variant binaries (clean + injected) compile up front in
  a build phase; the measurement window then contains zero compilation and zero disk
  churn, running prebuilt binaries strictly sequentially. Checkout/compile activity
  overlapping a measurement phase lands asymmetrically and reads as a code delta, so
  it is excluded by construction.
[...]

PR #1073, opened 2026-07-19
docs: withdraw stale cross-engine throughput claims and correct PPL prose

ohdearquant:
## What

Docs-only correction of published benchmark claims that no longer hold.

**Cross-engine decode table withdrawn.** The README's Lattice vs Ollama vs MLX throughput table (187/171/146 tok/s and the companion Ollama/MLX columns) is pulled until a multi-session baseline program replaces it. The published rates came from single-session runs on one machine, and the shell pipeline that derived them truncated intermediate precision (#1060). That pipeline was replaced by `scripts/bench_decode_harness.py` in #938, but the stored values predate the replacement and were never re-measured. The methodology prose (slope method, decode-only definition) is kept with the reproduction pointer; the "Why throughput falls with context" mechanism section is kept minus the withdrawn numbers.

**PPL prose corrected.** The README cited Lattice q4 19.27 / q4-QuaRot 19.95; the authoritative `docs/bench_results/perplexity.tsv` (regenerated 2026-07-07) says 16.59 / 19.01. The MLX rows are now flagged as carried from a prior reference run per the TSV's own comments, and the reproduction pointer is fixed from `bench_context_scaling.sh` to `bench_quality.sh` (the context-scaling script does not produce PPL).

**TSV provenance annotated.** `docs/bench_results/context_scaling.tsv` keeps the raw record with a provenance header stating the single-session origin, the since-replaced derivation pipeline, and the pending re-baseline. The annotation lives below the column header with no tab characters, so `bench_context_scaling_chart.py`'s parser is unaffected — verified by running its exact parse loop against the annotated file (9 rows parsed, comments skipped). The regeneration path (`bench_decode_adapters_context_scaling.py`) rewrites the file, so the annotation's lifetime is exactly until fresh harness-derived numbers replace it.

**Also:** drops the "MLX decodes faster than Lattice at raw throughput" sentence (derived from the withdrawn measurements; the architectural MPS/AMX vs public-Metal framing stays) and retitles the bench-command heading to drop the vs-claim.

## Bench disposition

Docs and data-file comments only; no `crates/` source touched; no bench A/B applicable.


oceanwaves630 (approved):
Docs-only correction read in full: withdrawn cross-engine claims removed across prose, TSV, and the tracked chart; PPL prose aligned to the authoritative TSV; methodology description now matches the harness computation.

PR #1087, opened 2026-07-19
Batch: integration/perf-regression-fixes → main (embed perf, bench integrity, grammar + shard fixes)

ohdearquant:
> _PR description drafted by Claude (Anthropic agent) at @ohdearquant's request._

## Summary

Batches 9 verified fixes accumulated on `integration/perf-regression-fixes` into `main` as one reviewed unit, so the combined set lands with a single bench-compare and CI pass against current `main` rather than as a tail of individual merges. Three follow-up commits made directly on the branch are included and described below.

**Performance**
- #1059 — delegate hinted prepared cosine to the fused path (embed)
- #1062 — make the int8 quantization min/max scan explicitly SIMD, so the hot loop no longer depends on auto-vectorization surviving a codegen-unit reshuffle (embed)

**Bench integrity / harness**
- #1065 — fail loud when Metal bench targets cannot measure, instead of reporting a silent zero
- #1070 — demote the embed SIMD bench target to quick-mode informational via one validated manifest
- #1066 — withdraw the bc-corrupted cross-engine decode table and fix the TSV slopes (docs)
- Gate group names shared by a demoted and a gated bench target (branch commit). The quick-mode demotion introduced by #1070 is target-level, but the handoff into the gate flattened every target's groups into a single list of bare names with no target attribution, so a group name demoted for `lattice-embed:simd` would have exempted an identically-named `lattice-inference` group from the regression gate — fail-open on a namespace collision. Demoted names are now resolved against every gated target's own listing; a colliding name gates rather than staying informational, and the collision is reported on stderr naming the group and both targets. Latent rather than active today, since the two targets' group names are currently disjoint, but `BENCHES_INFERENCE` is overridable so the collision was reachable by configuration.

**Correctness**
- #1067 — treat a completed grammar state as a successful stop on the Metal paths
- #1071 — classify grammar completion on the first and final Metal decode tokens
- #1072 — contain index-declared shard paths beneath the model directory

**CI**
- #1068 — name the mocked decode-harness checks for what they actually prove
- Update the decode-loop grammar gate test filter to match the fixture renamed in #1064 (branch commit)

**Docs**
- Describe the prepared-cosine bench path in terms of what it measures and what a gap between its groups means (branch commit)

## Test plan

- [ ] `cargo test --workspace`
- [ ] `cargo clippy --workspace -- -D warnings`
- [ ] `cargo fmt --all -- --check`
- [x] `python3 scripts/perf-bench-gate.py --selftest` — covers the new collision guard, and fails when the guard is reverse-applied

**bench-compare disposition (ADR-058).** Each constituent fix was verified individually before batching. The two performance changes (#1059, #1062) target `embed`; #1062 restores the int8 scan throughput that a codegen-unit reshuffle had silently regressed. The inference-touching commits (#1067, #1071, #1072) change grammar-stop classification and shard-path handling, not the decode or prefill hot paths compiled into the bench binaries, so no A/B delta is attributable to them. The bench-tooling commit changes only how measured groups are classified for gating, not what is measured or how, so it likewise contributes no A/B delta; it is covered by the gate's own selftest. A combined `make bench-compare` A/B for the whole batch against current `main` is attached before merge.

## ADR

n/a — batched fixes, no architectural change. Each constituent PR carried its own ADR disposition.

- [x] Every claim in this PR description matches the batched commits
- [x] The one SIMD change (#1062) is explicit intrinsics, verified on its source PR
- [x] This PR body starts with the attribution line from AGENTS.md
- [x] The bench-gate collision guard is mutation-tested: the selftest fails with the fix reverse-applied
- [ ] Combined `cargo bench` A/B for the batch attached before merge

## Out of scope

New features, architectural changes, and anything ADR-level go through their own PRs against `main`. This PR is the periodic batch-merge of the accumulated small fixes on `integration/perf-regression-fixes`.


oceanwaves630 (approved):

PR #1110, opened 2026-07-21
fix(embed): validate caller text before applying the retrieval instruction

### LINKED ISSUE #1104: bug(embed): role-prefix length counts against the documented text cap, rejecting inputs at the limit
## Summary

`embed_query` and `embed_passage` prepend the model's role instruction prefix (the query prefix, or `"passage: "` for document text on multilingual-e5) to each input, then run the `MAX_TEXT_BYTES` (32768) length guard on the *combined* string. A caller that respects the documented cap and passes a text of exactly 32768 bytes is rejected with `TextTooLong` at 32777. Because the check is per-text in a batch, one such input fails the entire batch.

## Repro

Any 32768-byte input through `embed_passage` (or `embed_query`) with a model whose `document_instruction()` / `query_instruction()` is non-empty (e.g. multilingual-e5-base). Observed live on 0.6.1.

## Mechanism (confirmed at source)

- `embed_passage` (crates/embed/src/service/mod.rs:112) calls `apply_prefix(texts, document_instruction())` to build the prompted string, then `embed(&prompted, model)`.
- `embed_query` (mod.rs:99) does the identical thing with `query_instruction()`.
- The guard lives downstream in both backends: `native.rs:312` and `cached.rs:119`, each `if text.len() > MAX_TEXT_BYTES { TextTooLong }`, where `text` is already the prefixed string.
- `document_instruction()` for multilingual-e5 is `"passage: "` (9 bytes), so a caller length of 32768 becomes a guarded length of 32777, which exceeds 32768.

The guard validates the internal prompted string against a cap the caller was told applies to their own input.

## Sibling

This affects both role paths, `embed_query` and `embed_passage`, and both guard sites (`native.rs`, `cached.rs`). Any fix must cover all four.

## Candidate fixes

1. Validate each caller text against `MAX_TEXT_BYTES` before `apply_prefix`, so the bounded internal prefix does not count against the caller's documented budget. Semantically lossless: the model's token window truncates far earlier than the byte cap.
2. Or subtract `prefix.len()` from the budget when clamping inside the role methods, char-boundary-safe.

Prefer (1): it keeps the guard's protective intent (bounding untrusted caller input) while honoring the published cap.

## Impact

Low urgency. A consumer can reserve the prefix length in its own truncation budget as a workaround. But it is a footgun for every consumer that truncates to the documented cap.

ohdearquant:
Closes #1104.

## The bug

`MAX_TEXT_BYTES` is documented as what a caller may pass. The guard ran on the string *after* the role instruction was prepended, so a caller submitting text at exactly the documented cap was rejected at `cap + prefix_len`. Someone doing exactly what the docs say got an error blaming them for bytes the service itself added.

Reported on both role paths, which is the sibling-path shape: the trait defaults (`embed_query` / `embed_passage` into `NativeEmbeddingService::embed`) and the `CachedEmbeddingService` overrides into its own internal method, four sites in total.

## Why this is not a one-line reorder

The single guard was doing two different jobs:

1. **The published contract**, "your text may be up to `MAX_TEXT_BYTES`". The right place to check this is the public entry, against the caller's own text.
2. **A memory bound** on what actually gets processed. Any sufficient bound satisfies this.

Validating caller text at the role methods fixes the reported symptom, but the downstream check still saw the prepared string and still rejected it, so the downstream bound had to change too. Three options were considered on the issue:

- **A.** Bound the sink at `MAX_TEXT_BYTES + MAX_PROMPT_PREFIX_BYTES` globally. Smallest diff, but a direct `embed` caller who never had a prefix applied gains slack against a cap the docs state exactly. Loosening a published bound to make a check pass is the habit worth avoiding even when the amount is small.
- **B.** Make the sink bound role-aware. Exact, but only reaches the cached path.
- **C.** Prefix after validation instead of before, removing the conflation rather than working around it.

This implements C, with B's exactness where a downstream bound is still needed.

## What changed

`embed_with_role(texts, model, role)` is a new defaulted trait method. It validates caller text, then prepares. `embed_query` and `embed_passage` become thin wrappers over it, so all three role paths share one implementation instead of three copies of the same ordering decision.

The length check downstream of preparation is now explicitly a memory backstop rather than the contract check, and it sizes itself as the cap plus *that model's own longest instruction* via the new `EmbeddingModel::max_instruction_bytes()`. For symmetric models that is zero bytes, so the backstop collapses onto the published cap exactly. There is no global slack constant.

`CachedEmbeddingService` now keeps caller text end to end and delegates prefixing to the wrapped service. Cache keying is unchanged in effect: the instruction is a function of the role and model config that are already part of the key, so keying on caller text and keying on prepared text partition identically.

Errors report the bound that actually rejected the input, so a caller over the published cap still sees `MAX_TEXT_BYTES` rather than the wider internal number.

## Public surface

This adds one defaulted method to the `EmbeddingService` trait. That is backward compatible for existing implementors, and `embed` remains the abstract method, so nothing external is forced to change. Flagging it explicitly because it is public trait surface on a published crate rather than an internal refactor.

The trait default documents the constraint it imposes on implementors: an implementor enforcing a text-length cap inside `embed` receives prepared text through the default path, which is longer than the caller's by up to `max_instruction_bytes()`. Such an implementor should size that guard accordingly or override `embed_with_role` to reach its backend directly. Both in-crate implementors override it.

## Tests

Three tests, on both invocation paths, plus a unit test pinning `max_instruction_bytes()` against the instruction strings.

Mutation-checked rather than assumed: reverting the validation order to prefix-then-validate and rebuilding makes both new role-path tests fail with the original signature, `32825 bytes exceeds maximum 32768`, which is the cap plus the 57-byte query instruction. Restoring the fix returns them to green. Full crate suite is 281 passing, clippy clean under `-D warnings`.

The positive-direction tests assert the request got *past* length validation rather than asserting success, because model weights are not present in every environment. That is the precise behavior that regressed: the old order returned `TextTooLong` and never reached model loading.

## bench-compare disposition

**No measurable performance change.**

`lattice-embed --bench simd`, all groups, quick mode. Base is the merge-base `05291c5d`, head is `de82422e`. Both binaries built from the same toolchain, then run back to back on the same idle machine.

**246 comparisons: 0 regressed, 3 improved, 243 no change.** Point estimates span -13.24% to +9.40%, mean -1.82%, median -1.97%.

Machine state, sampled three times and recorded with the run rather than asserted: 96.78% idle before the base phase, 97.22% between phases, 97.4% after the head phase.

### Reading the three flagged rows

| group | point | 95% CI |
|---|---:|---|
| `simd_dot_product/simd/384` | -6.70% | [-8.98, -4.33] |
| `simd_batch_cosine_non_normalized_query/simd_batch/768d_256c` | -5.60% | [-7.00, -4.19] |
| `int8_raw_dot_product/dot_product_i8/768` | -4.56% | [-6.13, -2.99] |

All three are improvements, all three measure SIMD kernels this diff cannot reach, and none is claimed as a win. This change is confined to service-layer validation ordering and async trait dispatch; it does not touch a SIMD kernel. Larger unflagged swings sit alongside them (`pair_loop_dot/1024d_16c` at -13.24%, `binary_query_per_call/1000` at +9.40%), both reported as no change on CI width, which is the shape of a lane whose resolution is wider than these effects.

The direction also argues against a regression rather than for one. The head arm runs second, and this lane carries a measured order effect where whatever runs second measures slower. Head still came out neutral-to-faster on the mean, so the bias worked against the head arm and found nothing.

Honest scope note: `lattice-embed:simd` is the target currently demoted to informational in quick mode, and same-commit A/A runs on it have produced confirmed-CI FAIL rows with disjoint failing groups. So this lane resolves poorly at this magnitude. That is acceptable for the claim being made here, which is the absence of a regression rather than the presence of a win.

### DIFF
--- a/crates/embed/CHANGELOG.md
+++ b/crates/embed/CHANGELOG.md
@@ -10,7 +10,45 @@ and this project adheres to
 
 ### Added
 
+- `EmbeddingModel::max_instruction_bytes` returns the byte length of the model's
+  longest retrieval instruction, and zero for symmetric models that take no
+  prefix. A length guard that runs on already-prepared text sizes itself with
+  this; a guard on caller text uses `MAX_TEXT_BYTES` unchanged.
+- `EmbeddingService::embed_with_role` is the single entry point for role-aware
+  embedding. It has a default body, so existing implementations of the trait
+  continue to compile and behave as before without being edited.
+
 ### Changed
+
+- The published text-length cap is now checked against the text the caller
+  supplied, before the retrieval instruction is prepended. Previously a caller
+  submitting text at exactly `MAX_TEXT_BYTES` could be rejected for bytes the
+  service itself added. `embed_query` and `embed_passage` are now thin wrappers
+  over `embed_with_role`, so all three role paths share one ordering decision.
+- `CachedEmbeddingService` keys the caller's text together with the role tag
+  rather than keying prepared text. This is equivalent, because the instruction
+  is a function of the role and model configuration already present in the key.
+- `TextTooLong` now reports the bound that actually rejected the input, so the
+  reported maximum is one the caller can relate to their own input.
+
+#### Compatibility note for external implementors
+
+The default `embed_with_role` reaches the backend by calling `embed`, and the
+text it passes there is the prepared text, longer than the caller's by up to
+`EmbeddingModel::max_instruction_bytes`. An external implementation that
+enforces the exact `MAX_TEXT_BYTES` cap inside its own `embed` will therefore
+still reject caller text at the cap.
+
+This behavior is not new. `embed_query` and `embed_passage` have always applied
+the instruction and then called `embed`, so such an implementation received
+lengthened text in prior releases too, and nothing about that path changes here.
+What changes is that the situation is now nameable and fixable: size the guard
+with `max_instruction_bytes`, or override `embed_with_role` to reach the backend
+directly. Both implementations in this crate override it.
+
+Keeping `embed` the sole abstract method is a deliberate backward-compatibility
+choice. Making `embed_with_role` abstract would deliver the cap unconditionally
+but would break every existing implementor at compile time.
 
 ### Deprecated
 
--- a/crates/embed/docs/service.md
+++ b/crates/embed/docs/service.md
@@ -212,24 +212,29 @@ are management APIs rather than a promise of distributed or persistent caching.
 
 For every generic, query, or passage request, the wrapper does the following:
 
-1. Apply the role prefix, if any. Generic calls remain raw text with the `Generic` role.
-2. Validate only the wrapper-owned empty-batch, batch-size, and text-length bounds before lookup.
-3. If capacity is zero, bypass hash computation and locks and delegate the prompted batch to the
-   inner service.
-4. Ask the inner service for the effective `ModelConfig`, then hash each prompted text with the
+1. Keep the caller's text as-is and record the request's `EmbeddingRole` (`Generic` for a plain
+   `embed` call). The wrapper does not apply the role instruction; the wrapped service does.
+2. Validate only the wrapper-owned empty-batch, batch-size, and text-length bounds, against the
+   caller's text, before lookup. Checking caller text keeps the published cap describing what the
+   caller passed rather than what the instruction adds.
+3. If capacity is zero, bypass hash computation and locks and delegate the caller text to the
+   inner service through `embed_with_role`, which applies the instruction for the role.
+4. Ask the inner service for the effective `ModelConfig`, then hash each caller text with the
    model identifier, its key revision, active dimension, and role tag.
 5. Look up all keys in the sharded LRU. Hits are returned from cache storage as shared slices and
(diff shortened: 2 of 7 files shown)


ohdearquant:

ohdearquant:
Reviewed the full diff at 776d11efc9e5419bbacbfdd0c6a9d55c56cf3e16. Approve. Validating caller text against the published cap before the role instruction is prepended is the right ordering, and pinning both sides of the external-implementor contract executably (the default's documented limitation and the override escape hatch) keeps that choice from regressing into prose. A formal approving review cannot be created here because author and reviewer are the same account, so this comment is the review record.

ohdearquant:
Reviewed at head 6d0187b643b9732b9c0a38b2cbbc8d9c2956a45d. The delta since the prior full review of this branch is changelog-only; the compatibility note's central claim (prepared-text forwarding through `embed` is pre-existing behavior) is verified against the trait body on main, with both remedies for exact-cap implementors documented and tested. cargo-semver-checks 196/196 against main; all check runs green at this SHA. Approve.

Issue #1105, opened 2026-07-21
bench gate: lattice-embed:simd is gated in neither quick nor full mode (the compensating nightly lane does not exist)


ohdearquant:
`scripts/lib/bench-quick-informational-targets.txt` demotes the whole `lattice-embed:simd` target out of quick-mode gating. Both that manifest and `scripts/bench-compare.sh` justify the demotion by pointing at full mode as the compensating control, and both name it as a scheduled nightly lane:

- `scripts/bench-compare.sh:33-35`: "`--full` remains the gate for EVERY group of every target: quick mode is a PR-side direction/magnitude screen, and full-resolution simd gating rides the scheduled nightly perf lane (a `--full` run on current main)"
- `scripts/lib/bench-quick-informational-targets.txt`: "FULL mode (`scripts/bench-compare.sh --full`, nightly perf lane) gates every group of every target regardless of this manifest."

No workflow in `.github/workflows` invokes `--full`. `grep -rn '\-\-full' .github/workflows/` returns nothing.

The four scheduled workflows are:

| workflow | cron | what it does |
|---|---|---|
| `adr064-gpu-decode-nightly.yml` | `17 8 * * *` | ADR-064 GPU decode, dormant by design pending #167 |
| `bench-update.yml` | `0 6 * * 1` (weekly) | `cargo bench --save-baseline`, trend data, documented as not a gate |
| `cargo-audit.yml` | `0 6 * * 1` | security audit |
| `e2e-parity.yml` | `0 6 * * 1` | greedy token parity |

`bench-update.yml` is the only bench-shaped one. It refreshes baselines weekly and does not run an A/B.

**Net effect: `lattice-embed:simd` is gated in neither mode.** Quick mode reports it informationally and excludes it from the verdict; full mode never runs. Roughly 23 groups covering the embed SIMD kernels have no gate.

The demotion evidence itself stands. Repeated same-commit A/A runs producing disjoint confirmed-CI FAIL rows of +8% to +17% on identical binaries is machine noise above quick-mode resolution, not signal, and demoting on that evidence was correct. The gap is the other half: the loosening shipped and the compensating control was written as prose rather than as a workflow.

Two ways to close it:

1. Build the lane the manifest promises: a scheduled workflow running `scripts/bench-compare.sh --full` against current main, on a cadence the manifest then cites accurately.
[...]

PR #1111, opened 2026-07-21
docs(bench): separate measured, reported, gated, and enforced in the bench comments

ohdearquant:
## What

Three source comments described the quick-mode `lattice-embed:simd` demotion as covered by a "scheduled nightly perf lane (a `--full` run on current main)". No such job existed when those comments were written, so the demoted groups were gated in neither mode. `bench-update.yml` did run both targets at full resolution and did publish historical regression reports, but it never invoked `perf-bench-gate.py` and never failed on the result, so the compensating control the comments claimed — full-resolution *gating* for the demoted target — was not there.

Corrected in:

- `scripts/bench-compare.sh`
- `scripts/lib/bench-quick-informational-targets.txt`
- `scripts/lib/bench-informational-groups.sh`

The missing lane has since been built (#1105, implemented in #1112) and this branch has merged it in, so the comments here describe shipped behavior rather than an absence.

## Four words that were doing one job

The replacement text for a false claim is itself a claim, and this one was wrong three times before it settled, every time in the same direction: a stronger statement than the code supported.

- "No automated CI lane runs full mode on main" was false. `bench-update.yml` runs both bench targets without `--quick`.
- "weekly" and "collects baselines without comparing against a prior one" were both false. It triggers on every push to main touching its perf paths — which include `crates/embed/src/simd/**` and `crates/embed/benches/simd.rs`, the demoted target's own source — as well as weekly by cron. And it does compare: it invokes `scripts/perf-baselines-readme.py`, which walks the snapshot history, computes each snapshot's delta against its predecessor, and publishes a "Worst step-regression" headline naming the bench and commit.
- "every group of every target" overstated both halves. The script benches two targets out of the workspace's declared bench set, and `--full` still honors the optional `BENCH_GROUPS_*` filters.

Rather than patch sentences a fourth time, the header now defines four terms once and the three comments use them literally:

- **measured** — the group ran and produced Criterion output
- **reported** — it appears in the report and the all-measurements table
- **classified gating** — a regression in it contributes to the report's FAIL verdict
- **enforced** — that verdict reaches the caller as a non-zero exit status

Separating "reported" from "enforced" is what made the last error findable: once they are different words, "collects baselines without comparing" reads as the category error it was.

## What the comments now say, and where each claim lives

- `scripts/bench-compare.sh` is report-only by default in both quick and full mode: the gate result is captured in `GATE_RC` and not returned. Passing `--fail-on-regression` propagates it — exit 1 for a confirmed gated regression, exit 2 when the comparison cannot be certified (`scripts/bench-compare.sh:388-418`, `scripts/perf-bench-gate.py`).
- `make bench-gate` is a separate path: it restores the local architecture's `perf-baselines` data, runs the two default targets, and returns `perf-bench-gate.py`'s status directly (`Makefile:57-73`).
- `perf-postmerge-gate.yml` runs `bench-compare.sh --full --fail-on-regression` on perf-path pushes to `main`, against the merged commit's own parent, and fails the job on a confirmed regression. For an ordinary push it uses `github.event.before`; when that SHA is unusable it falls back to the head commit's first parent, which for a squashed merge is the previous main tip.
- `bench-update.yml` saves baseline snapshots and publishes historical step-regression trends. It does not invoke the gate and has no regression-specific failure path, so it reports.
- Quick-mode demotion removes a target from the quick-mode gate only. Full-mode runs ignore the manifest, so every group they measure is classified gating — and, under the post-merge lane's flag, enforced.

## Why this is separate from the lane itself

The demotion decision is sound: repeated same-commit A/A quick-mode runs produced disjoint confirmed-CI FAIL rows on identical binaries, which is noise above quick-mode resolution. What was missing was the compensating control the comments named. A comment that cites infrastructure as justification is a claim, and that claim was false. Removing it was correct independently of what the eventual lane looked like, which is why it ships on its own.

## Verification

- `bash -n` on the changed shell files: clean.
- `python3 scripts/perf-bench-gate.py --selftest`: PASS (it reads the edited manifest).
- `python3 tests/test_bench_compare_measurement.py`: PASS.
- `actionlint` on `perf-postmerge-gate.yml` and `bench-update.yml`: clean.
- Manifest data line `lattice-embed:simd` intact; `bench-informational-groups.sh --print-targets` still emits it.
- Every claim above was read against source at this head.
- Comments and text only: no changed line falls outside a comment, so there is no behavioral delta, and no crate source is touched, so no bench-compare disposition applies.

### DIFF
--- a/scripts/bench-compare.sh
+++ b/scripts/bench-compare.sh
@@ -56,13 +56,15 @@
 # further, so a filtered --full run classifies only the selected groups of
 # those two.
 #
-# Automation: bench-update.yml does run those targets at full resolution on
-# main weekly, but only to collect baselines. It neither compares against a
-# prior baseline nor invokes perf-bench-gate.py nor takes any
-# regression-specific fail or alert action; its ordinary job steps can of
-# course still fail on their own errors. No automated full-mode regression
-# gate exists yet (#1105 tracks it), so a demoted target's full-resolution
-# regression coverage today comes from manual runs.
+# Automation: bench-update.yml runs those targets at full resolution on main
+# — on every push touching the perf paths (which include embed's simd source
+# and bench) and weekly by cron — and saves the baselines. It does not invoke
+# perf-bench-gate.py and takes no regression-specific fail or alert action;
+# its ordinary job steps can of course still fail on their own errors. It is
+# not silent about regressions either: its README generator compares each
+# snapshot against its predecessor and publishes a "Worst step-regression"
+# headline. So full-resolution automation today is regression REPORTING, and
+# what is missing is ENFORCEMENT (#1105 tracks that lane).
 set -euo pipefail
 
 REPO="$(cd "$(dirname "$0")/.." && pwd)"
--- a/scripts/bench-compare.sh
+++ b/scripts/bench-compare.sh
@@ -27,18 +27,25 @@
 # scripts/perf-bench-gate.py --selftest); their groups are still fully
 # measured and rendered — the informational section plus the
 # all-measurements table record every number — but excluded from the
-# FAIL/WARN gate and exit code. Every non-demoted target (all of
-# lattice-inference) gates normally in --quick.
+# FAIL/WARN gate and exit code. Every non-demoted target this script benches
+# (the lattice-inference one) gates normally in --quick.
 #
-# --full evaluates EVERY group of every target at full resolution, simd
-# included: quick mode is a PR-side direction/magnitude screen, and a --full
-# run resolves simd tightly enough to gate. That full-resolution path is
-# available on demand — `make bench-gate` (this HEAD vs the perf-baselines
-# branch) and `bench-compare.sh --full` (an A/B) both run it. What does NOT
-# exist yet is an automated job that runs it on main and alerts on a
-# regression; until #1105 lands that detection lane, the quick-mode demotion
-# is covered only by manual/local full-mode runs, not by CI. Do not read
-# this comment as a claim that an automated lane already catches simd.
+# --full applies no informational demotion: every group it benches gates,
+# simd included, because full resolution is tight enough to gate simd on.
+# Two caveats keep that from meaning "everything is covered".
+#
+# Scope: this script benches two targets, not the workspace's full bench set
+# — lattice-inference:$BENCHES_INFERENCE (default elementwise_cpu_bench) and
+# lattice-embed:simd. The optional BENCH_GROUPS_* filters above narrow it
+# further, so a filtered --full run gates only the selected groups of those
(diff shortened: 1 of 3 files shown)


ohdearquant:

oceanwaves630 (changes requested):
The vocabulary split is the right one and the rest of the text checks out at source. One factual defect, appearing twice, and it sits in the paragraph doing the most work.

**The gate invocation does not end in `|| true`.**

`scripts/bench-compare.sh:49-50`:
> This script ends its gate invocation with `|| true`, so a FAIL verdict is printed and then discarded

`scripts/bench-compare.sh:91-92`:
> the gate invocation at the bottom ends in `|| true`, so a confirmed regression is rendered in the report while the script still exits 0

What the gate invocation actually is, at this head:

```sh
GATE_RC=0
if [ -d "$HEAD_DIR/target/criterion" ]; then
  python3 "$REPO/scripts/perf-bench-gate.py" ... || GATE_RC=$?
else
  GATE_RC=2
fi
```

The status is captured, not discarded. The script exits 0 by default because the exit at line 408 is conditioned on `FAIL_ON_REGRESSION=1`, not because the status was thrown away.

The conclusion the paragraph draws is still correct — neither mode enforces by default. Only the mechanism it cites as evidence is stale, left over from before the exit-status handling changed underneath it.

**Why it is worth a round rather than a follow-up.** There are six real `|| true` occurrences in this file, and the nearest one to a reader following the comment is line 218:

```sh
{ "$@" 2>&1 | grep -E "$filter" | tee "$matched"; BENCH_RC=${PIPESTATUS[0]}; } || true
```

That one is deliberately paired with `BENCH_RC=${PIPESTATUS[0]}` so the bench's status is *preserved*, and lines 195-199 exist specifically to stop someone reading it as a discard. A reader who greps `|| true` on the strength of line 50 lands there and draws exactly the inference those lines warn against. So the stale sentence does not just fail to inform, it points at a construct whose meaning is the opposite of what it is being used to illustrate.

**Fix**: state the actual reason at both sites — the gate's status is captured into `GATE_RC` and only propagated when `--fail-on-regression` is set. No other change needed.

Verified as correct in the same pass, so these do not need rechecking: `make bench-gate` benches both default targets unfiltered against `perf-baselines` and ends on `perf-bench-gate.py`, so the recipe does return its status directly; `bench-update.yml` invokes no gate and takes no regression-specific fail action; and the "Worst step-regression" headline is real, at `scripts/perf-baselines-readme.py:125`.

oceanwaves630 (approved):
Automated review. Posted by this repository's automated pull-request review pipeline; this is not a human read.

Verdict on current head: APPROVE, zero blocking findings.

PR #1112, opened 2026-07-21
ci(perf): add the post-merge full-mode detection lane (#1105)

ohdearquant:
Builds the detection lane tracked in #1105, and fixes the swallow that would have made it decorative.

## The swallow

`scripts/bench-compare.sh` ends its gate invocation with `python3 scripts/perf-bench-gate.py ... || true`. It renders the FAIL section and still exits 0. For a human reading an A/B that is correct. For an automated lane it is the exact defect #1105 is about: a job that prints a red FAIL in its log and reports the step as passed.

This adds an opt-in `--fail-on-regression` that propagates the gate's exit code. Default behavior is unchanged, so every existing caller keeps its current semantics.

## The same swallow, three times more, each one level down

This PR's own guard failed open twice while it was being built, each time in the shape it exists to prevent. Both are worth stating rather than quietly fixing, because the shape is the interesting part.

**First, nothing measured passed.** The guard asked whether `target/criterion` existed. This script creates that directory itself, with `mkdir -p`, before any bench runs. The `cargo bench` pipelines end in `|| true` so a failed or skipped bench is swallowed, and `perf-bench-gate.py` returns 0 for a root containing no `change/estimates.json` — correct by design, because a first run has no baseline to compare against. Every link behaved as designed and the composition let a run that measured nothing exit 0. The guard was reading its own side effect as evidence about the benchmark.

**Then, partially measured passed.** The replacement asked whether *any* judgeable comparison existed. That still certifies a run where one target compared cleanly and the other produced an unresolvable baseline: the failed target is precisely the one nobody measured, and a green exit is a claim about code that was never benched. Existence was standing in for completeness.

**Then, a target that measured nothing passed.** Reconciling comparisons *found* against comparisons *judged* catches a comparison that failed to resolve, and is structurally blind to a target that produced no comparison to find: absence leaves no artifact to be missing. Verified by execution against a Criterion root holding a single valid `simd` comparison and nothing from the inference target — the gate exits 0 and certifies an A/B in which an entire bench target was never measured. Cargo's status does not catch it either, because a bench invocation whose Criterion filter matches nothing exits 0 having run nothing.

The gate is the wrong layer for this one and no amount of patching there fixes it: it is handed a directory tree and cannot know a second target was ever supposed to be in it. So the check moved to the invocation, which is the only point where the run's intent is still known and the target is still named — each bench invocation now reports how many measurement lines it actually printed, and under `--fail-on-regression` a run that exits 0 having printed none is refused with exit 2. Reporter mode is untouched: a missing bench target is ordinary when a human compares against an arbitrary ref.

Its test drives the shipping script in a disposable repo against a stub cargo rather than an extracted copy of the helper, so it exercises the real parser-to-guard path, and CI runs it beside the other harness-plumbing checks.

Both halves were discarding the evidence they needed.

The shell discarded cargo's status twice over. A pipeline reports its *last* command, which is `grep`, and `|| true` then resets `PIPESTATUS`. Verified directly: after `p | grep x || true`, `${PIPESTATUS[0]}` reads 0 even when `p` exited 7. A bench that failed to build was indistinguishable from one that printed no matching lines. Cargo's status is now captured inside the group and re-raised. The `--no-run` guard had the same defect in a different dress: "target absent on an older ref" and "target failed to compile" arrive on one channel, so the reporter may keep tolerating both while the enforcing lane may not.

The gate discarded its own skips. `parse_bench` returns `None` for malformed data and for an unresolvable baseline alike, warns, and the caller dropped it on the floor. It now reconciles comparisons *found* against comparisons *judged*; under `--require-measurements` any unjudged one is exit 2, naming the bench.

`bench-compare.sh` passes that flag only under `--fail-on-regression`, so the first-run "no baseline yet" pass that other callers rely on is untouched. Exit 1 and exit 2 both fail the caller, and the job summary spells out that they call for opposite responses: exit 1 means revert or justify, exit 2 means the measurement is broken and nothing was proven about performance.

## Why an A/B against the parent, not against `perf-baselines`

`make bench-gate` compares against the `perf-baselines` branch, which `bench-update.yml` rewrites on every perf-path push to main. A regression can therefore be absorbed into the baseline before any scheduled job compares against it. Benching the merged commit against its own parent compares before that ratchet. Merges here are squashed, so the parent is the previous main tip and the diff under test is exactly one PR. Both arms also build and run in the same job on the same runner, which removes cross-run baseline drift.

## Scope, stated where it cannot be missed

The lane **detects** post-merge. It cannot gate the merge that introduced a regression, so it must not be added to branch protection's required checks; a post-merge job can never be satisfied pre-merge.

Hosted runners are shared and noisy. This catches gross regressions (roughly the 15%-and-above class, the magnitude ADR-058's workflow was built for) and does **not** resolve sub-few-percent changes, the same noise floor that made quick-mode simd unreliable. A green run means "no gross regression detected", not "performance is unchanged". That ceiling appears in the workflow header and is printed into every run's job summary, so it travels with the result rather than living only in a PR description.

Full mode ignores the quick-mode informational demotion, so `lattice-embed:simd` is included. Every group of the two benched targets (`lattice-inference:elementwise_cpu_bench`, `lattice-embed:simd`) is gated, because this job sets no `BENCH_GROUPS_*` filter. The workspace's other bench targets are not run here.

## Known gap, documented rather than implied away

GitHub keeps only one pending run per concurrency group. If two commits land while a run is in flight, the first one's pending run is replaced, and the survivor compares against it — so a regression introduced by the skipped commit sits in both arms and is invisible. During a rapid merge burst, a quiet lane is unproven, not evidence of no regression. `workflow_dispatch` with explicit `base_ref`/`head_ref` measures a commit that was skipped. Closing this properly needs a queue that retains runs, or comparison against the last successfully measured commit; either is a follow-up.

## Additional hardening in the same area

- **The `--` escape was advertised and did not work.** The parser shifted `--` away and a dash-leftover check ten lines later re-inspected the very arguments `--` existed to protect, so `-- --dash-led-ref` was rejected by the diagnostic that names `--` as the remedy. An escape hatch that the next ten lines close is worse than none. It is honored now, and the four git call sites that consume a ref pass `--end-of-options` so a dash-led ref survives git's own option parsing.
- **The job summary could not tell exit 1 from exit 2.** It read `job.status`, which holds only `success`/`failure`/`cancelled`. A checkout, toolchain, or endpoint-resolution failure was therefore reported to the reader as "confirmed regression — revert or justify" for a run that never benched. The A/B step now records its actual exit code and the summary explains only codes it saw, with a distinct message for "failed before reaching a verdict".
- **`--selftest` now runs in CI.** An earlier revision's commit message said CI re-ran these cases; no workflow invoked it. It is wired into the existing script-validation job alongside the other script self-tests, and is confirmed executing on this branch's runs.

- **The workflow's own ref inputs terminate option parsing.** `bench-compare.sh` protects its four ref-consuming git call sites, but this workflow resolves `base_ref`/`head_ref` *before* it calls the script, and those two `rev-parse` calls had no terminator, so a legal dash-led ref supplied to `workflow_dispatch` was read as an option. `--verify` is load-bearing rather than stylistic there: `git rev-parse --end-of-options <ref>` echoes the literal string `--end-of-options` as an output line beside the SHA, so plain command substitution captures a two-line value and passes a corrupted ref downstream, which is worse than the original failure because it does not fail loudly. `--verify` selects the single-object output mode, the same reason `--short` suffices inside the script.
- Flag parsing rejects a misplaced flag instead of benching garbage. `bench-compare.sh HEAD~1 --full` resolved `--full` as `HEAD_REF` and started a real A/B against a nonsense ref. Confirmed by running it.
- Trigger coverage: root `Cargo.toml` can carry `[profile.bench]` (it does not today) and `.cargo/` can carry rustflags, either of which changes bench codegen without touching crate source. Both added, along with the two helper scripts `bench-compare.sh` invokes.
- Grouped the summary redirects (actionlint SC2129).

Actions are referenced by tag here, matching every other workflow in this repo. Pinning to commit SHAs is worth doing, but as a repo-wide change rather than a one-off in a new file that would imply a policy the repo does not have.

## Merge-order dependency on #1111

These two touch the same file and the same claim, in opposite directions, so the second one to land has work to do.

#1111 adds a vocabulary block to `bench-compare.sh` defining measured / reported / classified gating / enforced, and states that the script is REPORT-ONLY in both modes because it discards its gate's exit status. That is true of `main` today and false the moment this PR lands: `--fail-on-regression` is exactly the mode that enforces.

So whichever merges second must update that wording in the same change, describing the default and opt-in modes separately rather than making one blanket claim. If this PR lands second, that edit belongs in its rebase. Flagging it here because a comment that silently goes stale at merge is the same class of defect both of these PRs exist to remove.

## Verification

What was executed, and by what method:

- **The enforcing path, end to end on the real script.** With an absent inference bench target, `--fail-on-regression` exits 2, names the failed target, produces zero benchmark output lines, and leaves no worktree residue. This exercises the subshell status propagation that an earlier revision lacked, without running a benchmark.
- **The partial-A/B refusal, and its mutation sensitivity.** A fixture with one valid gating comparison and one unresolvable baseline exits 2 and names the unjudged bench. Reverting the refusal makes that self-test case fail with exactly that message; restoring it passes.
- **The no-measurement fail-open, reproduced and then closed.** `perf-bench-gate.py` against an empty criterion root exits 0 before the fix and 2 with `--require-measurements`. Six cases now live in `--selftest`: empty root passes without the flag (first-run semantics preserved) and fails with it; a real gating comparison still passes, so the flag is not merely a brake; an all-informational run is refused; the mixed root is refused with the flag and still passes without it.
- **Gate returns nonzero on a real regression.** A criterion fixture at 20 percent regression with CI-lower above the 7 percent threshold exits 1. Control: the same fixture at half a percent exits 0, so the fixture discriminates rather than always failing.
- **`PIPESTATUS` after `|| true`.** Measured, not assumed: it reads 0 for a command that exited 7, which is why the capture had to move inside the group.
- **Dash-led refs, against a real branch.** In a disposable repository with a branch actually named with a leading dash: the bare form resolves to two lines, the `--verify` form to one line matching the branch tip, and `git worktree add --end-of-options` accepts the same ref. This is why the workflow uses `--verify` and the script's `--short` sites do not need it.
- **Argument handling on the real script.** A flag after a positional, an unknown flag, and surplus positionals each exit 2 with a usage message. `-- --dash-led-ref` now reaches git, which reports `invalid reference` rather than `unknown option` — the discriminator showing the ref survived option parsing.
- `bash -n` clean, both workflows parse, and `actionlint` is clean.

Not claimed: no live end-to-end A/B in which a genuine regression drives the real script to a nonzero exit, and no execution of the default-mode shell path for an absent target, whose only route starts a real benchmark. Calling either verified would be the kind of overclaim this PR exists to remove.

No crate source is touched, so no bench-compare disposition applies.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -622,6 +622,12 @@ jobs:
       # "nothing measured" as "nothing regressed"), and runs no benchmark.
       - name: python3 scripts/perf-bench-gate.py --selftest
         run: python3 scripts/perf-bench-gate.py --selftest
+      # bench-compare.sh's measurement-integrity guard. Drives the shipping
+      # script in a disposable repo against a stub cargo, so it exercises the
+      # real parser-to-guard path; no benchmark runs and no real cargo is
+      # invoked, which keeps it in the same harness-plumbing scope as above.
+      - name: python3 tests/test_bench_compare_measurement.py
+        run: python3 tests/test_bench_compare_measurement.py -v
 
   # cargo-semver-checks against the latest published crates.io release for
   # every internal path-dependency crate, so a breaking public API change is
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -616,6 +616,12 @@ jobs:
         run: bash scripts/ensure-noindex-marker-selftest.sh
       - name: python3 tests/test_bench_disposition.py
         run: python3 tests/test_bench_disposition.py -v
+      # perf-bench-gate's own fixture self-test. Same "harness plumbing only"
+      # scope: it fabricates Criterion trees and asserts the gate's exit codes
+      # (including the refusals that keep an automated lane from reading
+      # "nothing measured" as "nothing regressed"), and runs no benchmark.
+      - name: python3 scripts/perf-bench-gate.py --selftest
+        run: python3 scripts/perf-bench-gate.py --selftest
 
   # cargo-semver-checks against the latest published crates.io release for
   # every internal path-dependency crate, so a breaking public API change is
--- a/.github/workflows/perf-postmerge-gate.yml
+++ b/.github/workflows/perf-postmerge-gate.yml
@@ -45,9 +45,10 @@ on:
       - 'crates/embed/benches/simd.rs'
       - 'crates/embed/Cargo.toml'
       - 'Cargo.lock'
-      # Root Cargo.toml carries [profile.bench]; .cargo/ can carry rustflags.
-      # Either can change codegen for both bench targets without touching a
-      # single line of crate source, so both must trigger this lane.
+      # Root Cargo.toml can carry [profile.bench] (it does not today) and
+      # .cargo/ can carry rustflags. Either can change codegen for both bench
+      # targets without touching a single line of crate source, so both must
+      # trigger this lane.
       - 'Cargo.toml'
       - '.cargo/**'
       - 'scripts/bench-compare.sh'
@@ -188,28 +189,43 @@ jobs:
           } | tee -a "$GITHUB_STEP_SUMMARY"
 
       - name: Full-mode A/B, failing on a confirmed regression
+        id: ab
         env:
           BASE: ${{ steps.refs.outputs.base }}
           HEAD: ${{ steps.refs.outputs.head }}
         run: |
-          scripts/bench-compare.sh --full --fail-on-regression "$BASE" "$HEAD"
+          # Record the gate's OWN exit code before re-raising it. The summary
+          # below has to tell "confirmed regression" from "refused to certify",
+          # and job.status cannot: it only ever holds success/failure/cancelled,
+          # so every earlier failure (checkout, toolchain, endpoint resolution)
+          # would otherwise be reported to the reader as one of those two gate
+          # verdicts and point them at a revert for a run that never benched.
+          rc=0
+          scripts/bench-compare.sh --full --fail-on-regression "$BASE" "$HEAD" || rc=$?
+          echo "rc=$rc" >> "$GITHUB_OUTPUT"
+          exit "$rc"
 
       - name: Record the outcome honestly
         if: always()
         env:
           JOB_STATUS: ${{ job.status }}
+          AB_RC: ${{ steps.ab.outputs.rc }}
         run: |
           {
             if [ "$JOB_STATUS" = "success" ]; then
               echo "No gross regression detected (see ceiling above)."
+            elif [ "$AB_RC" = "1" ]; then
+              echo "FAILED, exit 1 — a confirmed regression. Check the flagged groups"
+              echo "are reachable from the merged diff, then revert or justify."
+            elif [ "$AB_RC" = "2" ]; then
+              echo "FAILED, exit 2 — the gate refused to certify this run: no"
+              echo "comparison data, a partial A/B, or nothing gating among the"
+              echo "results. Nothing was proven about performance either way."
+              echo "Fix the measurement and re-run; do NOT revert on this alone."
             else
-              echo "FAILED. Two different causes share this exit, and they call for"
-              echo "opposite responses, so read the step log before acting:"
-              echo
-              echo "- exit 1 — a confirmed regression. Check the flagged groups are"
-              echo "  reachable from the merged diff, then revert or justify."
-              echo "- exit 2 — the gate refused to certify: no comparison data, or"
-              echo "  nothing gating among the results. Nothing was proven about"
-              echo "  performance. Fix the measurement and re-run; do NOT revert."
+              echo "FAILED before or during the A/B, with no gate verdict recorded"
+              echo "(\`rc=$AB_RC\`). The benchmark comparison did not reach a"
+              echo "conclusion, so this says nothing about performance. Inspect the"
+              echo "failing step above; do NOT revert on this alone."
             fi
           } >> "$GITHUB_STEP_SUMMARY"
--- a/.github/workflows/perf-postmerge-gate.yml
+++ b/.github/workflows/perf-postmerge-gate.yml
@@ -45,10 +45,18 @@ on:
       - 'crates/embed/benches/simd.rs'
       - 'crates/embed/Cargo.toml'
       - 'Cargo.lock'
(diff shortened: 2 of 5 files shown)


oceanwaves630 (approved):
Read the full diff. Approving.

The change that carries this is moving the completeness check to the layer that
can actually state the invariant. A gate handed a Criterion tree cannot know a
second target was ever meant to be in it, because a target that emitted nothing
leaves no artifact to be found missing. Having each bench invocation report how
many measurement lines it printed puts the check where the run's intent is still
known, and reconciling comparisons intended against comparisons judged closes the
partial case rather than the merely-empty one.

Specific things worth noting as correct:

- The exit 1 versus exit 2 split is meaningful and plumbed consistently through
  the script, the gate and the job summary. A confirmed regression and a refusal
  to certify ask the reader for opposite responses, and the summary says so
  explicitly rather than leaving both as "failure".
- Capturing the gate's own exit code before re-raising it, because job.status
  collapses every earlier failure into the same value and would point a reader
  at a revert for a run that never benched.
- The PIPESTATUS handling. A pipeline reports its last command and the trailing
  `|| true` then resets the status, so a bench that failed to build looked
  identical to one that produced no matching lines.
- Re-raising the subshell status. Without it the refusal is swallowed by the
  very construct meant to enforce it.
- Honoring `--` at the leftover-argument guard as well as in the parser.
  Advertising an escape hatch that the following lines then close is worse than
  not offering one, since the diagnostic names it as the remedy.
- Reporter behavior is preserved throughout, and the self-tests assert that
  explicitly rather than assuming it.

On the workflow: the resolution ceiling and the retained-run concurrency gap are
stated in the file itself rather than left for someone to discover. The note that
this must not be added to branch protection is the right thing to write down next
to a post-merge job, since a required check that can never be satisfied pre-merge
would wedge the queue rather than protect it.

The self-test additions cover the empty root, a genuine gating pass, the
all-informational case and the partial case, and assert the refusal names the
unjudged bench. Driving the shipping script against a stub cargo in a disposable
repo exercises the real parser-to-guard path rather than an extracted helper.

CI is green at 778f4674be with no failures.

PR #1114, opened 2026-07-22
docs(bench): name the real exit-status mechanism, not a bare discard

ohdearquant:
## What

Three sentences in `scripts/bench-compare.sh`'s header described the default report-only behavior as the gate invocation "ending in `` `|| true` ``". That construct is not there. The gate's status is captured into `GATE_RC` and re-raised only under `--fail-on-regression`, so the discard is conditional rather than unconditional.

The conclusion those sentences draw is unchanged and still correct: both modes are report-only by default. Only the cited mechanism was stale, left over from before the exit handling changed underneath it.

## Why it is worth a PR rather than a note

The wording does not merely fail to inform, it misdirects. A reader who searches this file for `` `|| true` `` on the strength of those sentences lands on the bench-run pipeline, where the construct is deliberately paired with a `PIPESTATUS` capture so the status is **preserved**, and which carries its own comment block existing solely to prevent that misreading. Aiming a reader at a construct whose meaning is the inverse of the one being illustrated is worse than saying nothing, in a file whose header exists to make exactly that distinction precise.

## Scope

Three occurrences, not two: the enforcement paragraph, the `--fail-on-regression` rationale, and one more in the vocabulary block that describes the same behavior without using the phrase. The conditional statements in `scripts/lib/bench-informational-groups.sh` and `scripts/lib/bench-quick-informational-targets.txt` are accurate as written and are left alone.

## Verification

- `bash -n scripts/bench-compare.sh`: clean.
- `python3 scripts/perf-bench-gate.py --selftest`: PASS.
- `python3 tests/test_bench_compare_measurement.py`: PASS.
- No remaining prose in `scripts/` or `.github/` cites a bare discard of the gate's status.

Comments only: no changed line falls outside a comment, so there is no behavioral delta, and no crate source is touched, so no bench-compare disposition applies.

Follows #1111.

### DIFF
--- a/scripts/bench-compare.sh
+++ b/scripts/bench-compare.sh
@@ -21,8 +21,9 @@
 # appear in the report; CLASSIFIED GATING (vs informational) if a regression
 # in it contributes to the report's FAIL verdict; and ENFORCED only if that
 # FAIL verdict reaches the caller as a non-zero exit status. Classification
-# is not enforcement: by default this script computes a verdict and then
-# discards its exit status, so --quick and --full are both REPORT-ONLY.
+# is not enforcement: by default this script computes a verdict, captures the
+# gate's exit status, and does not act on it, so --quick and --full are both
+# REPORT-ONLY.
 # Enforcement is opt-in per invocation via --fail-on-regression, which
 # propagates the gate's status instead; `make bench-gate` also enforces.
 # Use these words literally below.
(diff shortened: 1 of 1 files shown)


oceanwaves630 (approved):
All three occurrences fixed and the replacement text matches the code. Verified against the gate block: the status is captured into GATE_RC and the exit is conditioned on the flag, which is what both paragraphs now say.

The third one is the one that mattered. Lines 24-25 described the same wrong mechanism without using the phrase, so it would have survived a search for it and kept the incorrect model alive in the paragraph that defines the four terms. Finding it is the difference between fixing what was pointed at and sweeping for the class.

Leaving the two conditional statements in scripts/lib/ alone was also right: they say the status is discarded unless the caller passes the flag, which is accurate and cites no construct.

oceanwaves630 (approved):
Automated review. Posted by this repository's automated pull-request review pipeline; this is not a human read.

Verdict on current head: APPROVE, zero blocking findings.

PR #1119, opened 2026-07-22
perf(bench): make bench-compare isolate and disclose its own run conditions

### LINKED ISSUE #1117: bench-compare.sh takes no lock and records no run conditions, while gating merges
## Summary

`scripts/bench-compare.sh` acquires no lock of any kind, and prints nothing about the conditions its numbers were produced under. Its output gates merges through `perf-postmerge-gate.yml`, so a corrupted number here does not waste a run, it rejects a correct change or admits a regression with a green checkmark attached.

Verified in the current tree: no `flock`, no advisory-lock acquisition anywhere in `scripts/bench-compare.sh` or in the `Makefile` bench recipes. Serialization against other work on the same machine is supplied by whoever remembers to wrap the invocation, per invocation, by hand.

## Why the current state looks safer than it is

The default targets are CPU-only: `BENCHES_INFERENCE` defaults to `elementwise_cpu_bench`, `BENCHES_EMBED` is hard-coded to `simd`, neither bench source references Metal, and their imports are confined to the CPU modules. Neither workflow overrides either variable.

That is clean **by default value**, not clean by construction. `BENCHES_INFERENCE` and `CARGO_FEATURES_INFERENCE` are overridable by design, and the script's own comment says why: so a PR can point the A/B at a different inference bench target without hand-rolling a separate script. Setting `BENCHES_INFERENCE` to a Metal-driving target with `CARGO_FEATURES_INFERENCE=f16,metal-gpu` makes the gate's own script drive the GPU with no serialization at all. Nothing rejects it.

## Proposed change

**1. Acquire both advisory locks unconditionally, at script start.**

- the machine-wide bench window (exclusive side)
- the machine-wide GPU lock, `/tmp/lion-metal-gpu-test.lock`

Unconditionally means: not conditional on detecting CI, and not conditional on classifying the selected target as GPU-driving.

The classifier is where a fail-open would live. It would have to recognize a bench name nobody has seen yet, a feature combination nobody enumerated, and a transitive dependency that pulls Metal in without announcing it in the target name. Every miss passes the check while the GPU spins. Taking both locks always deletes the classifier, and there is then nothing left to get wrong.

The cost is that a CPU bench serializes against GPU work. That is correct rather than merely tolerable: GPU work running during a CPU bench *is* ambient load, and the quiet floor exists to exclude ambient load. Excluding the largest controllable consumer on the machine one step earlier is free; excluding it after the fact is impossible.

A CI-detecting branch is rejected for the same reason: an absence-based conditional is the guard shape that keeps failing open, and on a fresh runner an uncontended flock costs microseconds. The branch cannot pay for its own risk.

**2. Refuse, do not warn, when a precondition is not met.**

Lock acquisition failure gets the same disposition the idle floor already gets: non-zero exit, no measurement. A warning printed on a bench report is read by nobody at the moment it matters, which is weeks later when someone cites the number.

**3. Move the quiet probe into the script, and print the conditions into the report.**

The idle-percentage sampling currently lives in a caller-side wrapper, so it covers the invocations someone remembered to wrap. It belongs in the script, where every invocation gets it.

The report must then record, inside itself: lock acquisition state, the idle samples (before base, between phases, after head), and the effective target and feature selection. A number that does not say which conditions produced it is indistinguishable from one produced under good conditions, and the reader will assume the good case because they cannot reconstruct it.

## Known hazard this introduces, which the implementation must handle

Once the script holds these locks, **any nested process that acquires the same locks will block on its own parent**. Two concrete paths:

- Wrapping `bench-compare.sh` in the existing bench-window helper out of habit. The wrapper takes the exclusive window, the script then waits for a lock the wrapper holds, and the run hangs.
- A GPU-driving bench target that takes the GPU lock itself (which is the direction of the companion issue) running under a script that already holds it.

`flock` associates the lock with the open file description, and a child that opens the file separately gets a distinct description, so it blocks rather than re-entering.

The implementation must therefore use a bounded wait and, on timeout, fail loudly with the holder identified (`lsof` on the lock path) and an explicit note that a caller-side wrapper is the likely cause. Detecting an inherited hold via an environment marker is not proposed here: a stale marker would silently skip the lock, which reintroduces the failure this issue exists to remove.

## Not proposed

Weakening either lock, shortening a timeout, making acquisition optional, or adding an escape hatch flag. The point is to remove the dependency on operator memory, and an escape hatch is operator memory with extra steps.

Related: the companion issue on the GPU lock's reachability from bench and example targets.

ohdearquant:
Closes #1117.

`scripts/bench-compare.sh` acquired no lock and printed nothing about the conditions its numbers were produced under, while its output gates merges through the post-merge lane. A corrupted number there does not waste a run: it rejects a correct change, or admits a regression with a green check attached. Serialization was supplied by whoever remembered to wrap the invocation, by hand, per invocation.

## What changed

**Both machine-wide advisory locks are taken unconditionally** — the bench window and `/tmp/lion-metal-gpu-test.lock` — for the whole run.

Unconditionally means not conditional on detecting CI, and not conditional on classifying the selected target as GPU-driving. The classifier is where a fail-open would live: it would have to recognize a bench name nobody has seen yet, a feature combination nobody enumerated, and a transitive dependency that pulls Metal in without announcing itself in the target name. Every miss passes the check while the GPU spins. Taking both always deletes the classifier, and there is then nothing left to get wrong.

The current defaults are CPU-only, which is why this looked safe. But `BENCHES_INFERENCE` and `CARGO_FEATURES_INFERENCE` are overridable by design — the script's own comment says why — so that was clean by default value, not clean by construction.

**Ambient CPU idle is sampled and gated** before the base phase, between phases, and after the head phase. A lock excludes peers on this machine; it says nothing about load, and the two are not the same check. Below the floor the run is refused, not warned about: a warning on a bench report is read by nobody at the moment it matters, which is weeks later when someone quotes the number. Refusing after the head phase is the case that matters most, because that is when numbers already exist and the temptation to keep them is strongest. `BENCH_IDLE_FLOOR` sets the floor and is recorded in the output.

**Every run prints a `Run conditions` block**: refs, effective bench targets and features, resolution, enforcement mode, lock dispositions, and the three measured idle samples. A figure that does not record what produced it is indistinguishable from one produced under good conditions, and a reader weeks later cannot reconstruct the difference.

## Why the file is split

The measurement body moves to `scripts/lib/bench-compare-impl.sh`; `scripts/bench-compare.sh` becomes the locking entry point. Callers, flags, and environment variables are unchanged, mode included — `make bench-compare` and the post-merge lane both exec the entry point directly, so it stays `100755`.

`scripts/ensure-noindex-marker-selftest.sh` follows the body, since the no-index guard moved with it. That assertion is also made fail-closed: a call-site assertion whose target file is missing passes vacuously, which is the same stale-pointer failure the split just created.

A single script re-execing itself under the lock supervisor needs some signal to stop recursing, and the obvious one is an environment marker. A marker is a claim about lock state supplied by the environment, and one left exported in a shell makes the body skip locking and emit an unlocked measurement that is indistinguishable — in the report and in the exit status — from a locked one. Two files cannot recurse and there is nothing to go stale.

The body additionally requires that the supervisor which recorded the lock status be one of its own ancestors.

Stated exactly, because the tempting claim here is one word wider than the truth: the file supplies the PID and the OS supplies the chain, so what the check establishes is a relation, not a held lock. It refuses a status file left over from a finished run, one copied from a different run, and accidental direct invocation, which are the ways this actually gets run without isolation. It does not refuse a caller who deliberately records an ancestor's PID. Closing that needs the lock descriptor rather than a PID, and belongs with the follow-up below for the reason given there.

That boundary is pinned by `test_deliberately_recorded_ancestor_pid_is_accepted` rather than left in prose. Strengthen the check and that test fails, which is the signal to update every place the limit is described — an overclaimed guarantee never fails loudly on its own, and the next change built on a property the code does not have inherits the gap silently.

## Portability and fd hygiene

`flock(1)` is not used. It is util-linux and does not exist on macOS, where every local run of this script happens; a shell lock written from Linux habit ships something that is silently absent on the machine that matters, and command-not-found inside a backgrounded job is invisible. Acquisition uses `fcntl.flock` directly.

The lock file descriptors are not inherited by the measurement body or its descendants. A flock is released only when every descriptor referring to that open file description is closed, so a leaked lock fd in a long-lived build daemon holds the window open long after the run that took it. Children that must hold a lock will be passed the descriptor on purpose; none do today.

`lsof` is used only to name who to look at while waiting. It lists processes with the lock file *open*, which is a superset of those holding a flock and does not distinguish exclusive from shared, so it is never used to conclude that a lock is already held. If it is unavailable the wait is less informative, never shorter.

That diagnostic reports each holder's executable name and PID, never its command line. `ps -o command=` would be more informative and is the wrong trade: arguments carry tokens, keys and connection strings, and this message is written by a script that runs in this repository's workflows, whose job logs are readable by anyone. A PID plus a name is enough for an operator to run `ps` themselves on the machine where doing so is safe.

## Tests

`tests/test_bench_locks.py` pins the refusals, all verified mutation-sensitive against this branch:

| Mutation | Test that fails |
|---|---|
| drop the lock-status check from the body | `test_body_invoked_directly_refuses` |
| replace the ancestry walk with a file-exists check | `test_status_file_naming_a_non_ancestor_is_refused` |
| turn the ambient-load gate into a warning | `test_below_floor_refuses_rather_than_warns` |
| report holders with `ps -o command=` again | `test_holder_report_never_includes_command_line_arguments` |
| restore the bare `pid="$(ps ... \| tr -d ' ')"` assignment | `test_ancestry_walk_refuses_with_a_diagnostic_when_ps_fails` |
| read the first percentage on top's line instead of the one named `idle` | `test_macos_idle_is_the_field_named_idle_not_the_first_percentage` |
| take top's first sample (its since-boot average) instead of the last | `test_macos_idle_comes_from_the_last_sample_not_the_first` |
| compute Linux idle from one sample instead of the delta | `test_linux_idle_is_the_delta_share_not_the_absolute` |
| drop the iowait term | `test_linux_counts_iowait_as_idle` |
| accept non-advancing `/proc/stat` counters | `test_linux_non_advancing_counters_raise` |

The idle parsers were previously reachable only through a live sample, so each could only be tested against whatever the machine happened to be doing, and only one branch runs on any given host — with CI on Linux, the macOS branch that every local run takes was covered nowhere. Parsing is now separate from sampling and both are tested against fixtures.

A failing `ps` in the ancestry walk did not reach the refusal at all: under `set -o pipefail` the failure propagates out of the assignment and `set -e` exits with `ps`'s own status, so the caller saw a bare `1` and an empty stderr. Fail-closed, but silently, and triggered by the ordinary case of an ancestor exiting mid-walk rather than only by denied process inspection. The walk now separates "not an ancestor" from "could not complete the walk" and refuses with a message for both.

`test_deliberately_recorded_ancestor_pid_is_accepted` is a characterization test rather than a wish: it fails if the ancestry check is ever strengthened, which is when the description of its limit needs updating everywhere.

`test_supervised_run_reaches_the_measurement` is there so the two refusals above would not also pass if the body refused unconditionally, which would be a broken script and a green suite.

The existing measurement-integrity tests keep driving the shipping script. They redirect the lock paths inside their *copy* of the supervisor and set the idle floor to zero, because those runs produce no number: serializing them against real benches buys no isolation and can exceed their own timeout. There is no equivalent knob in the shipping script, which is the point.

## Lane trigger

`perf-postmerge-gate.yml` gains `scripts/lib/bench-compare-impl.sh` and the two new helpers in its path filter. Without that, a change confined to the measurement body would no longer trigger the lane that runs it.

## bench-compare disposition

Not applicable, structurally. This PR changes no compiled code: the diff is shell, Python, workflow YAML, tests, and docs, with no file under `crates/` and no Rust or Cargo manifest touched. Base and head bench binaries have identical effective source, so an A/B table could not attribute any delta to this diff.

Verified by construction rather than asserted: `git diff --name-only` over the branch matches nothing under `^crates/`, `\.rs$`, or `Cargo`.

## What this does not cover, stated rather than left implicit

This makes `bench-compare` isolate its own run. It does not make measurements on this machine isolated, and the difference matters because an advisory lock only excludes processes that ask for it.

`make bench-ci` and `make bench-gate` invoke the same two Criterion targets directly, with no lock and no idle sampling, and `bench-gate` returns a gate status computed from those numbers. A sibling started while an A/B is in flight lands its load in one phase and not the other, which is the asymmetry this whole exercise exists to prevent. Roughly twenty more measurement harnesses under `scripts/` are unaudited.

That is a convention plus an enumeration plus a check that the enumeration stays complete, which is a different change from this one, so it is #1121 rather than a wider diff here. Naming it explicitly because a partial guard that reads as a complete one is the failure this PR is otherwise about.

## Follow-up

A nested acquirer — a bench target that takes the GPU lock itself — would block on this script once it holds that lock. None exists today, which is the subject of the companion issue on the GPU lock's reachability from bench targets. Handing the descriptor to that child, with an `fstat` identity check on the receiving side so a descriptor number that lands on some other open file cannot lock the wrong object and proceed, belongs to that change rather than being shipped here unexercised.


oceanwaves630 (approved):
Approving. bench-compare now acquires its own exclusive lock, samples idle/load before, between, and after phases, prints the disclosure into the result artifact, and refuses to render a verdict below the idle floor — with the A/A control path documented. Verified CI executed per-commit on the current head (full job run, not a carried-forward rollup) and all checks pass.

PR #1124, opened 2026-07-22
ci: pin actions to SHAs, stop trusting remote model code, and fix miscited rationale

### LINKED ISSUE #1088: hygiene: rewrite process-narration comments to durable rationale (embed bench + inference crate)
## Context

Several source comments record transient authoring history — review-round and hypothesis numbering, phrasing like "after i2 adds…" or "should improve by ≥15%" — rather than the durable technical rationale they are meant to document. This surfaced while preparing #1087, but the debt is pre-existing and independent of it, so it is tracked here as its own change against `main`.

## Scope A — `crates/embed/benches/simd.rs`

Roughly 14 comment lines are structured around round/hypothesis vocabulary (lines ~753–1052: `ROUND 3` / `ROUND 4` / `Hypothesis 1|2|3` / `baseline`). **Rewrite** each to describe behavior — what the path computes, what each parity/baseline group compares, and what a gap between groups means — rather than deleting them. A bench file with no explanatory comments is its own debt.

## Scope B — `crates/inference/` comments

Re-derive, against current `main`, the comment-rationale rewrite attempted on the stale `chore/scrub-process-comments` branch (~27 files: rewrite transient-authoring history into the technical properties the comments document; keep all issue/ADR/RFC references and measured values; comment-only, with no executable-code, signature, or test-assertion changes). That branch is stale versus current `main` — re-derive fresh, do not rebase it — then delete the branch.

## Constraints

- Comment-only: no code, signature, or test-assertion changes.
- Rewrite to behavior/rationale; do not just delete lines.
- Preserve every issue/ADR/RFC reference and measured value.
- Priority p2, sequenced **after #1087 lands**.
- ADR-058 bench-compare n/a (comment-only, no perf-relevant change).

### LINKED ISSUE #1093: ADR-065 cited five times as the attention-race diagnosis; ADR-065 is about feature promotion gates and no ADR documents that race
Five code comments in `crates/inference/src/forward/metal_qwen35.rs` cite "the ADR-065 attention
race" as the documented justification for three separate workarounds. ADR-065 does not describe
an attention race, and no other ADR in `docs/adr/` does either.

### The citations

| line | what it claims |
|---|---|
| 28116 | "concurrent GPU execution amplifies the pre-existing ADR-065 attention race" |
| 30560 | "See ADR-065 for the full diagnosis and the GPU-capture path to a true fix." |
| 31124 | "race (ADR-065), which perturbs the residual stream on ~1-in-N runs" |
| 31185 | "pre-existing ADR-065 attention race is not amplified by concurrent device contention" |
| 31231 | "best-of-N discards the ADR-065 serial-race outliers: best-of-5 max_abs_diff" |

### What ADR-065 actually is

`ADR-065-feature-promotion-gates.md` — "Feature Promotion Gates: Merge Measured Primitives, Not
Research Ideas", Status Proposed, dated 2026-06-30. It documents merge policy for experiment PRs.
Its only matches for `race` are substrings of `trace`.

`grep -rln "attention race\|race in attention\|attention.*data race" docs/adr/*.md` returns nothing,
so the race is not documented under a different number either.

### Why this matters beyond a stale reference

The label is load-bearing. It is the entire written rationale for:

1. `gpu_test_lock()` holding a machine-wide cross-process advisory flock on
   `/tmp/lion-metal-gpu-test.lock`,
2. best-of-N retries in the cross-algorithm GDN parity tests,
3. an accepted ~1-in-N nondeterminism in the residual stream.

Line 30560 is the sharpest instance: it promises a reader a "full diagnosis and the GPU-capture
path to a true fix" at a document that contains neither. A future reader who follows the citation,
finds nothing, and concludes the workarounds are unjustified may remove them. The race is real —
three independent mechanisms in the tree work around it — but it has never been written up.

### Proposed fix

Write the diagnosis up as its own ADR (observed symptom, reproduction, which kernel, why the
serial GDN reference is the thing perturbed, why a GPU capture is the path to a real fix), then
repoint all five comments at it. If the race turns out to be already understood well enough to
fix, the ADR should say that instead and the workarounds get a removal condition.

Do not simply renumber the citations. The value here is that the diagnosis gets written down once,
not that five comments stop pointing at the wrong file.

### LINKED ISSUE #1095: CI: guard against the per-target dependency tables in lattice-embed diverging again
## Background

`crates/embed/Cargo.toml` declares `lattice-inference` twice, once under `cfg(not(target_arch = "wasm32"))` and once under `cfg(target_arch = "wasm32")`. This split is necessary and the file explains why: workspace feature inheritance is additive, so there is no way to give wasm a trimmed `tokio` from a single unconditional table.

The hazard is that the two tables can drift apart silently. They did. The wasm table carried `["std", "f16"]`; the native table took default features, which are `["std", "download", "serve"]`, with `f16` absent. Without `f16` the weight loader does not lose precision, it hard-fails:

```
crates/inference/src/weights/f32_weights.rs:464-491
  "tensor {name} is F16 but lattice-inference was built without the f16 feature"
```

Fixed in #1094. Nothing stops it from recurring.

## Why existing CI does not catch it

Feature unification. In any workspace-wide build, some other crate enables `lattice-inference/f16` and `lattice-embed` inherits it for free. Every green workspace check therefore tests a configuration that the standalone crate never actually gets. The broken configuration is reachable only via:

- `cargo build -p lattice-embed` on its own
- `cargo install lattice-embed`
- the wasm/npm package build
- any cross-compiled target

All four are real shipping paths, and none of them are exercised today.

## Proposed guard

A CI leg that fails on the divergence rather than on a symptom of it:

1. Build `-p lattice-embed` **standalone** (not as part of a workspace build) so no other crate's features leak in.
2. Load a half-precision checkpoint through the resulting binary and assert it succeeds. A small committed BF16 fixture is enough; the failure mode is at load, not at compute, so it does not need a real model.

Step 2 is the part that matters. A build-only check passes happily on the broken configuration, because the missing feature is a runtime error and not a compile error.

## Mutation sensitivity

Whatever lands must fail when `f16` is removed from the native table in `crates/embed/Cargo.toml`. That reverse-apply is the acceptance test for the guard itself, not an optional extra: a guard that stays green with the fix reverted is decoration.

## Possibly broader

Worth checking whether any other crate in the tree has per-target dependency tables for the same dependency, and whether those agree. The failure is not specific to `embed` or to `f16`, it is specific to declaring one dependency twice and letting the declarations drift.

### LINKED ISSUE #1100: Developer scripts load reference models unpinned, and one still passes trust_remote_code=True
#1057 hardens how the CI parity gate obtains and loads its HF reference: pinned revision,
exact-set hash verification, offline local load, no remote code, safetensors only. That
work is scoped to the gate. Several developer-facing scripts obtain or load reference
models on their own terms and were deliberately left out of that PR.

After #1057 merges, exactly one `trust_remote_code=True` remains in the tree:

- `scripts/compare_logits.py:57` loads a caller-supplied local model directory with
  `trust_remote_code=True`. Any directory a developer points it at can execute code at
  load time.

Loads with no integrity check on what they are loading:

- `scripts/gen_embed_parity_goldens.py:71-72` resolves a model by bare id, so the bytes
  depend on whatever the upstream default branch points to at the time. Its later local
  loads at 118-119, 186-187, 240-241 and 293-294 have no manifest.
- `scripts/vision_goldens_qwen35.py:329-330` and `scripts/gemma4_stage5_e2e_golden.py:75,82-84`
  load caller-provided directories directly.

These generate goldens, and a golden generated from an unintended reference silently
becomes the thing every later run is compared against, so the blast radius outlives the
script invocation.

Out of scope here, noted so it is not rediscovered: `crates/inference/src/download.rs:80`
resolves against a mutable ref by design, because users name the model they want.

Suggested shape, smallest first:

1. Drop `trust_remote_code=True` in `compare_logits.py` unless something concrete needs it,
   and say what that is in a comment if it does.
2. Give the goldens scripts the same treatment as the gate where they feed a checked-in
   golden: pin the revision and record what the golden was generated from, so a
   regenerated golden that shifts can be traced to a reference change rather than a code
   change.

### LINKED ISSUE #1113: ci: pin GitHub Actions to commit SHAs (repo-wide policy, not per-workflow)
Every `uses:` in this repo references a mutable tag. A tag can be repointed by whoever controls the action's repository, so a compromised or simply re-tagged upstream changes what runs in our CI without any change on our side. Several of our workflows have `contents: write` (the perf-baselines publisher) or run on `pull_request_target`, which raises what a swap would be worth.

Current state, measured on this tree:

- 127 total `uses:` references
- 1 pinned to a commit SHA
- 6 distinct third-party (non-`actions/` org) actions, accounting for roughly 49 of those uses: `Swatinem/rust-cache`, `dtolnay/rust-toolchain`, `marocchino/sticky-pull-request-comment`, `EmbarkStudios/cargo-deny-action`, `raven-actions/actionlint`, `obi1kenobi/cargo-semver-checks-action`

## Why this is filed rather than fixed in passing

It came up while reviewing a PR that adds one new workflow. Pinning only the new file would leave the repo's actual exposure unchanged while implying a policy the repo does not have, which is worse than the honest status quo: a reader would reasonably infer the other 126 references had been considered.

## Suggested shape

1. Pin third-party actions first. They are the ones outside GitHub's own namespace and carry most of the risk for the fewest edits.
2. Keep the human-readable version in a trailing comment (`uses: owner/action@<sha>  # v2.1.0`), which is the convention Dependabot understands.
3. Enable Dependabot for `github-actions` so pinned SHAs still get update PRs, otherwise pinning trades a supply-chain risk for a staleness one.
4. Decide explicitly whether `actions/*` are in scope. Pinning them too is more consistent; leaving them on tags is defensible and much less churn. Either is fine, but it should be a decision on the record rather than an accident.

## Not claimed

No evidence of any actual compromise or unexpected tag movement in this repo. This is preventive hardening, and the priority should reflect that.

ohdearquant:
Five independent supply-chain and documentation-integrity fixes.

## Pin third-party actions to commit SHAs (#1113)

A mutable tag such as `@v4` resolves to whatever the action's owner points it at, so a compromised or simply retagged release executes new code in CI with CI's permissions.

Coverage: all 10 workflow files, 127 `uses:` references — 121 now SHA-pinned, 6 are local composite actions (`./`) needing no pin, 0 left on a mutable ref. Each of the 13 distinct pins was checked against upstream and resolves to the commit its trailing version comment names (10 via release tags; 3 via version branches, which is how `dtolnay/rust-toolchain` and `sticky-pull-request-comment` publish versions). A Dependabot `github-actions` entry keeps them refreshed rather than frozen.

## Stop trusting remote model code, pin reference revisions (#1100)

One script passed `trust_remote_code=True`, which executes arbitrary Python from a downloaded model repository. That is now `False` everywhere, with `local_files_only=True` where the model is expected on disk. The BGE reference revision is pinned so golden regeneration is reproducible, and goldens record the source revision they came from. Local snapshot directories carry no revision marker, so that case records the absence rather than inventing a value.

## Guard the native f16 dependency against silent drift (#1095)

`lattice-embed`'s native `lattice-inference` table took that crate's default features, which are `["std", "download", "serve"]` and do not include `f16`, while the wasm table beside it did (#1094). Standalone native consumers got a build where BF16 weights hard-fail at load. No run went red over it, and not because feature resolution hid it: nothing in the suite exercised a half-precision load on any invocation. The gap was coverage.

The new test round-trips a BF16 safetensors tensor through the real load path. It has to execute rather than merely build, because without the feature the failure is a runtime `Err` and not a compile error, so the `--no-run` steps beside it cannot observe it.

It runs standalone via `cargo test -p lattice-embed --test f16_feature_load_guard` rather than inside a `--workspace` command. `lattice-tune`'s `lattice-inference` dependency is optional but carries `features = ["f16"]`, so any invocation that activates tune's `inference-hook` or `train-backward` alongside `lattice-embed` would unify that `f16` into embed and mask exactly this regression. Neither feature is in tune's defaults, so `cargo test --workspace` does not do that today; running the guard on the isolated resolve means it never has to stay true.

Mutation check on this branch. With `"f16"` removed from `crates/embed/Cargo.toml`:

```
test bf16_tensor_loads_through_lattice_embeds_native_dependency ... FAILED

thread 'bf16_tensor_loads_through_lattice_embeds_native_dependency' panicked at
crates/embed/tests/f16_feature_load_guard.rs:65:9:
lattice-embed's native lattice-inference dependency was built without the f16 feature,
so a BF16 tensor hard-fails at load instead of decoding to f32 (this is the #1094/#1095
regression this guard exists to catch): Invalid safetensors file: tensor weight is BF16
but lattice-inference was built without the f16 feature

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

Restoring the feature returns it to `ok`.

## State the attention-race rationale instead of citing an ADR (#1093)

Five comments attributed the Metal batched/decode attention nondeterminism to ADR-065, which is about feature promotion gates and does not describe this race. No ADR documents it, so there is no correct number to substitute and citing a different one would compound the error. Each site now states the situation directly: what the hazard perturbs, what static analysis has and has not ruled out, that the root cause is unconfirmed, and that a GPU frame capture is the diagnostic nobody has run. A dangling reference to a notes file absent from this repository goes with it.

## Replace bench process narration with durable rationale (#1088)

Two bench group comments in `crates/embed/benches/simd.rs` described the work that produced them, referencing a round and hypothesis numbering meaningless to a later reader and describing a follow-up as pending after it shipped. They now state what each pair of groups compares and why the gap between them is the quantity of interest.

Closes #1113
Closes #1100
Closes #1095
Closes #1093
Closes #1088

## bench-compare disposition

No A/B was run, and no A/B could attribute anything to this diff: base and head bench binaries are built from identical effective source.

Every changed line in the two Rust source files this branch touches under `crates/` is a comment — `crates/inference/src/forward/metal_qwen35.rs` (39 changed lines, 0 non-comment) and `crates/embed/benches/simd.rs` (28 changed lines, 0 non-comment). The only other `crates/` change is a new file under `crates/embed/tests/`, which is a separate Cargo target from `benches/` and is not linked into any bench binary. No `Cargo.toml`, no build script, and no compiled source line differs, so `elementwise_cpu_bench` and `simd` compile to the same instructions on both sides.

The remaining changes are workflow YAML, a Dependabot config, and Python scripts, none of which is compiled at all.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/workflows/adr064-gpu-decode-nightly.yml
+++ b/.github/workflows/adr064-gpu-decode-nightly.yml
@@ -32,7 +32,7 @@ jobs:
     timeout-minutes: 60
     steps:
       - name: Checkout main
-        uses: actions/checkout@v4
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
         with:
           ref: main
 
@@ -59,7 +59,7 @@ jobs:
 
       - name: Upload current JSON + gate report
         if: always()
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
         with:
           name: adr064-gpu-decode-nightly-${{ github.run_id }}
           path: |
--- a/.github/workflows/app-binaries.yml
+++ b/.github/workflows/app-binaries.yml
@@ -61,7 +61,7 @@ jobs:
     outputs:
       bins: ${{ steps.filter.outputs.bins }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
         with:
           fetch-depth: 0
       - id: filter
@@ -96,9 +96,9 @@ jobs:
     runs-on: macos-latest
     timeout-minutes: 45
     steps:
-      - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@1.94.1
-      - uses: Swatinem/rust-cache@v2
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
+      - uses: dtolnay/rust-toolchain@e43eeffd51aefeea929f4ea14e09fd077153e788 # 1.94.1
+      - uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
         with:
           shared-key: app-binaries
       - name: Build all app-shipped binaries (required-features included)
--- a/.github/workflows/bench-evidence.yml
+++ b/.github/workflows/bench-evidence.yml
@@ -55,7 +55,7 @@ jobs:
         if: github.event_name == 'pull_request_target'
         # Pinned to a full commit SHA: this is a pull_request_target job, so a
         # moved or compromised tag would run in the trusted context before the
-        # checker does. SHA is actions/checkoutd5960a326750d5838078e36cf38b85af677262 # v4 v4 (the v4 tag as of 2026-07-20).
+        # checker does. SHA is actions/checkout v4 (the v4 tag as of 2026-07-20).
         uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
         with:
           ref: ${{ github.event.pull_request.base.sha }}
--- a/.github/workflows/bench-evidence.yml
+++ b/.github/workflows/bench-evidence.yml
@@ -55,7 +55,7 @@ jobs:
         if: github.event_name == 'pull_request_target'
         # Pinned to a full commit SHA: this is a pull_request_target job, so a
         # moved or compromised tag would run in the trusted context before the
-        # checker does. SHA is actions/checkout v4 (the v4 tag as of 2026-07-20).
+        # checker does. SHA is actions/checkoutd5960a326750d5838078e36cf38b85af677262 # v4 v4 (the v4 tag as of 2026-07-20).
         uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
         with:
           ref: ${{ github.event.pull_request.base.sha }}
--- a/.github/workflows/bench-update.yml
+++ b/.github/workflows/bench-update.yml
@@ -46,13 +46,13 @@ jobs:
             arch: aarch64-linux
     steps:
       - name: Checkout source
-        uses: actions/checkout@v4
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
         with:
           fetch-depth: 0  # need history for commit SHA + dates
 
-      - uses: dtolnay/rust-toolchain@1.94.1
+      - uses: dtolnay/rust-toolchain@e43eeffd51aefeea929f4ea14e09fd077153e788 # 1.94.1
 
-      - uses: Swatinem/rust-cache@v2
+      - uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
         with:
           cache-on-failure: true
           shared-key: bench-${{ matrix.arch }}
@@ -77,7 +77,7 @@ jobs:
             --out artifacts/snapshot.json
 
       - name: Upload baseline + snapshot
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
         with:
           name: bench-data-${{ matrix.arch }}
           path: |
@@ -91,7 +91,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout main (for scripts)
-        uses: actions/checkout@v4
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
         with:
           fetch-depth: 0
 
@@ -124,7 +124,7 @@ jobs:
           fi
 
       - name: Download all bench artifacts
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4
         with:
           path: bench-artifacts/
 
--- a/.github/workflows/cargo-audit.yml
+++ b/.github/workflows/cargo-audit.yml
@@ -62,7 +62,7 @@ jobs:
     outputs:
       deps: ${{ steps.filter.outputs.deps }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
         with:
           fetch-depth: 0
       - id: filter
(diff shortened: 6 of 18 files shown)


oceanwaves630 (approved):
Approving at head 12f5d85f. CI, e2e-parity, cargo audit, app-binaries, and bench evidence all verified per-commit on this SHA, with the resolve-gates executed by name rather than skipped. Action pins were resolved against the upstream registry and checked for agreement with their commented versions. The earlier e2e failure was isolated to upstream toolchain provisioning (identical action code and runner image across passing and failing runs) and passed on plain re-run of the same commit.

Issue #1085, opened 2026-07-19
fix(inference): three input-validation hardening gaps (batch config / bitnet / grammar mask)


ohdearquant:
## Overview

Three independent input-validation / fail-closed hardening gaps found in one pass. Each is low-to-medium severity on its own; grouped to fix together. Keep the three as separate subheadings with separate regression tests.

## 1. `BatchWorker::try_new` skips `BatchConfig::validate`

`BatchConfig::validate` (`crates/inference/src/batch/config.rs:53-68`) rejects zero-valued config fields, but `BatchWorker::try_new` (`crates/inference/src/batch/worker.rs:307-332`) never calls it — it constructs the sub-pools directly. The sub-allocators catch their own overflow, but the config's zero-rejection is bypassed on this construction path.

- **Fix**: call `config.validate()?` at the top of `try_new`.
- **Test**: `try_new` with a zeroed config field must return `Err`. Must fail if the `validate()` call is removed.

## 2. bitnet `pack_row` has no finiteness guard

`pack_row` (`crates/inference/src/forward/bitnet_kernel.rs:89-113`) computes `alpha = mean(|w|)`. A NaN or Inf weight makes `alpha` non-finite; `alpha > 0.0` is then false so the pack loop is skipped (all-zero packing) and a non-finite `alpha` is returned as the row scale, propagating downstream.

- **Fix**: reject non-finite rows (or error) before packing, consistent with the other load-time quantization guards.
- **Test**: a row containing NaN must error (currently returns NaN alpha + zero packing). Must fail if the guard is removed.

## 3. `GrammarEngine::mask_logits` asserts instead of erroring on vocab mismatch

`mask_logits` (`crates/inference/src/grammar/engine.rs:385-391`) uses `assert!(logits.len() >= self.vocab_size, ...)`, panicking on a grammar/model vocab mismatch rather than returning a typed error. `assert!`/`panic!` in library code is banned by `AGENTS.md`.

- **Fix**: return `Err` on the mismatch (make the method fallible if it is not already).
- **Test**: call with `logits.len() < vocab_size` — must return `Err`, not panic. Must fail if the `assert!` is restored.

Issue #1084, opened 2026-07-19
fix(inference): forward_prefill_all_logits panics when LoRA is active (perplexity path)


ohdearquant:
## Evidence

`forward_prefill_all_logits` takes an explicit `panic!` when a LoRA adapter is active and all-position logits are requested (`crates/inference/src/forward/metal_qwen35.rs:6947-6954`):

```rust
if self.lora.is_some() {
    // Batched helper does not apply LoRA adapters; stay on sequential path.
    if all_positions {
        panic!(
            "forward_prefill_all_logits: LoRA active — \
             batch/all-position prefill does not apply LoRA"
        );
    }
    ...
```

The `all_positions` branch is reached by `compute_token_nlls` / perplexity computation.

## Impact

Requesting perplexity or per-token NLLs while a LoRA adapter is active panics the process. `panic!` in library code is banned by the crate conventions (`AGENTS.md`), and this is a valid-looking API combination (perplexity + adapter) rather than an impossible state, so it is a crash reachable from ordinary use.

## Fix shape

Return a typed error instead of panicking, so callers can handle the unsupported combination:

```rust
if all_positions {
    return Err(InferenceError::Inference(
        "all-position prefill does not apply LoRA adapters; use the sequential path".into(),
    ));
}
```

(Propagate the `Result` up through `forward_prefill_all_logits` and `compute_token_nlls` if the signature does not already return one.)

## Regression test

Call the perplexity/all-logits path with an active LoRA adapter: it must return `Err`, not panic. Must fail if the `panic!` is restored.

Issue #1083, opened 2026-07-19
fix(inference): Q8 and NEON embedding lookups lack the prompt out-of-vocab guard


ohdearquant:
## Evidence

The f16 CPU forward path guards prompt token ids against the model vocab size before the embedding lookup (`crates/inference/src/forward/cpu_f16.rs:927-933`):

```rust
if let Some(&bad) = prompt_ids.iter().find(|&&id| id as usize >= cfg.vocab_size) {
    return Err(...); // out-of-vocab id would index past the embedding table
}
```

The Q8 and NEON forward paths perform the same embedding lookup with no such guard:

- `crates/inference/src/forward/cpu_q8.rs:553` and `:686`
- `crates/inference/src/forward/neon_forward.rs:741` and `:871`

## Impact

A prompt token id greater than or equal to `vocab_size` (tokenizer/model vocab mismatch, or a crafted request) indexes past the embedding table on the Q8 and NEON paths. This is an out-of-bounds read — a panic at best, undefined behavior in an unchecked indexing path at worst. The f16 path is protected; its siblings are not (sibling-invocation-path gap: one path guarded, its twins missed).

## Fix shape

Apply the same out-of-vocab guard at the top of the Q8 and NEON forward entry points that `cpu_f16.rs` uses. Factor it into a shared helper so a future fourth path cannot regress the same way.

## Regression test

A prompt containing an id `== vocab_size` must return `Err` on the Q8 path and the NEON path (currently OOB-indexes). Each must fail if its guard is removed.

Issue #1082, opened 2026-07-19
fix(kv-cache): restore_prefix seq_len==0 guard misses append-before-advance window


ohdearquant:
## Evidence

`restore_prefix` guards only on the logical sequence length (`crates/inference/src/kv_cache/paged.rs:501-510`):

```rust
pub fn restore_prefix(&mut self, adapter_id: AdapterId, token_ids: &[u32])
    -> Result<Option<usize>, InferenceError> {
    if self.seq_len() != 0 {
        return Err(InferenceError::PrefixCache(
            "restore_prefix requires an empty PagedKVCache".into(),
        ));
    }
    ...
```

But `append_kv_layer` pushes a physical page on the first append at position 0, while `advance()` bumps `seq_len` only afterward (`:794-818` and `:836-840`):

```rust
let pos = self.table.seq_len();                 // 0 on first token
let needed_pages = (pos / page_size) + 1;        // = 1
while self.table.num_pages() < needed_pages { ... self.table.push_page(phys); }
// ... advance() called later, after all layers appended
```

So there is a window where pages exist and hold partial data but `seq_len()` is still 0. `restore_prefix`'s `seq_len() != 0` guard passes in that window.

## Impact

If `restore_prefix` is called after a partial append (some layers written, `advance` not yet called), the guard admits it into a non-pristine cache: it restores prefix pages assuming an empty cache while a stray populated page already exists, corrupting the restored state. Reachability is narrow — normal callers restore on a fresh cache before any append — so this is medium severity, but the guard is checking the logical counter when it means to assert a pristine physical cache.

## Fix shape

Strengthen the guard to also require zero pages (pristine physical state), not just `seq_len() == 0`:

```rust
if self.seq_len() != 0 || self.num_pages() != 0 {
    return Err(...);
}
```

## Regression test

Append one layer at position 0 (no `advance`), then call `restore_prefix`: it must return `Err` (currently passes the guard). Must fail if the `num_pages()` check is removed.

PR #1125, opened 2026-07-22
fix(inference): validate shapes and sizes before allocating

ohdearquant:
Three input-validation fixes that share one shape: a check exists, but it runs after the allocation or copy it was meant to prevent, or it does not cover the value that actually breaks things.

## Reject absurd hidden_size before the QuaRot rotation allocation (#1061)

`convert_quarot_qwen35` validated `hidden_size` with `is_power_of_two()` alone. `1 << 60` satisfies that and reaches `RandomizedHadamard::new`, whose sign-buffer `Vec::with_capacity(hidden_size)` aborts on capacity overflow or gets the process OOM-killed. An abort is not catchable, so the value has to be rejected before it gets there. A magnitude cap, generously above any real Qwen3.5 checkpoint, now returns a typed error first.

To be precise about where that abort sits, because an earlier revision of this description had it wrong: `load_tensors_f64` runs before the rotation is constructed, so the un-guarded path reads and materializes the entire required tensor set as f64 and only then hits the allocation. The cap is worth more than the original framing suggested, not less, since it now rejects ahead of that load rather than merely ahead of one buffer.

## Overflow-check the batch admission length guard (#1080)

`BatchWorker::submit` computed `prompt_ids.len() + max_new_tokens` unchecked. A `max_new_tokens` near `usize::MAX` wraps to a small total in release builds, so the `total_len > max_seq_len` guard below it passes and the sequence is admitted with an absurd budget. `checked_add` now rejects on overflow, which is what the guard already meant.

## Validate declared tensor shapes before materializing them (#1054)

`MtpWeights::load_from_source`, the vision checkpoint loader and the Gemma 4 tensor loader each materialized an owned f32 tensor and only then compared its shape against the config-derived expectation, so a checkpoint declaring a huge shape allocated and decoded the whole tensor before being rejected. Where `TensorSource::tensor_shape` reports a declared shape, it is now compared first and the mismatch returns without materializing. Sources that cannot report a shape without reading keep the existing post-materialization check, so no path loses validation.

Ordering is what these three tests pin, not merely the error. A post-materialization check also returns `Err` on the same input, so each test drives a source that counts `get_f32_tensor_owned` calls and asserts the count is zero:

```
assert_eq!(source.materialize_calls.get(), 0,
    "a declared-shape mismatch must be rejected before any tensor data is copied");
```

## Vision f16 companion tensors: compare the header on the handle that reads the payload

The declared-shape preflight above covers manifests that carry a `shape`. That field is
`Option<Vec<usize>>` with `#[serde(default)]`, so a manifest may omit it entirely, and the
two arms that then materialize a tensor need their own comparison. Both now have one:

- The `.q4` arm compares the file's own header against the config-derived expectation before
  `dequantize_q4_to_f32` allocates the expanded f32 buffer.
- The `.f16` arm compares the header before the payload is read.

The f16 case is worth describing precisely, because the first attempt at it was wrong in a way
that still passed its test. Reading the shape through one `open()` and then materializing
through a second `open()` of the same pathname validates one file handle and reads from
another. Nothing binds them: the check establishes a property of whatever the path resolved to
during the check, not of the bytes that end up in memory.

So the comparison moved inside a single open. `load_f16_tensor_file_expecting` opens the file
once, parses the header, compares it against the expectation, and only then reads the payload
from that same handle:

```rust
} else if let Some(expected) = expected_visual_tensor_shape(&entry.name, vision_cfg) {
    let loaded = load_f16_tensor_file_expecting(&file_path, &expected);
```

`read_f16_tensor_shape` was deleted rather than left in place. A public "read just the shape"
helper is the thing that invites the two-open sequence to be reconstructed by the next caller,
and there is no remaining use for it.

`load_f16_tensor_file` keeps its previous signature and behaviour for its other callers; it and
the checked variant share the same header parser and payload reader, so neither path changed
its error text, read positions, or bounds checks.

### What the f16 test does and does not show

`f16_shape_is_compared_before_the_payload_is_read` pins ORDER: with the comparison removed it
fails at a truncated payload read rather than at the shape assertion. It does not prove the
single-open property, because a test driving a real path cannot observe how many times that
path was opened. The single-open property rests on the function's structure, which is why the
helper that allowed the alternative was removed rather than merely left uncalled.

## Scope

This branch is a partial pass over the input-validation issues; #1035, #1081, #1082, #1083, #1084 and #1085 are untouched and stay open.

## Mutation evidence

Each fix was reverse-applied on its own, with the test it added left in place, and the test re-run. Reverting a whole commit would take its test with it and prove nothing, so only the non-test hunks were removed each time.

| Fix reverted | Result with fix removed |
|---|---|
| `hidden_size` magnitude cap | test fails: conversion returns the safetensors-missing error instead of the `hidden_size` rejection |
| `checked_add` in `submit` | test fails: `attempt to add with overflow` at the admission guard |
| Gemma 4 declared-shape preflight | its test fails; the other two still pass |
| MTP declared-shape preflight | its test fails; the other two still pass |
| Vision declared-shape preflight | its test fails; the other two still pass |

Mutating one loader at a time rather than all three together is what shows each test is sensitive to its own fix specifically, rather than to something else in the set.

One qualification worth recording. The vision test calls `fetch_expected_tensors`, which this branch introduced, so removing that fix wholesale deletes the symbol and the test stops compiling — coupling to the fix's API, which is weaker evidence than a behavioral failure. The row above is from the sharper mutation instead: the function was kept and only the declared-shape check inside it removed, which compiles and makes that one test fail on its "never materialized" assertion.

The `hidden_size` row also corrected a comment. That test's fixture writes no tensor files, so it pins the cap's ordering relative to tensor load and does not exercise the allocation abort; the comment previously implied it did, and now states the narrower thing it actually shows.

Refs #1061
Refs #1080
Refs #1054

Measured on a dedicated quiet Apple-silicon machine (M4, 10-core) rather than a shared
development box, `--quick` resolution, base `7ba6f096eeba8e9bba8b195bd6d8f26e7ceb27f7` against
head `5e4697528d62dd480157d0d481e8baab481a9213`. Both machine-wide advisory locks were acquired
(21.2 s wait for the first phase, uncontended for the second), and the harness certified ambient
idle at all three probes in each phase against a 70% floor.

**Result: no regression is attributable to this diff.** Every gated bench is within ±0.20%
except one row that moved in the *improving* direction, and that row is not claimed as an
improvement for reasons given below.

### Gated benches, A/B

| Bench | Δ | 95% CI |
|---|---:|---|
| `softmax_attention/512` | **-3.65%** | [-3.70%, -3.60%] |
| `softmax_attention/128` | -0.20% | [-0.25%, -0.15%] |
| `gelu/4096` | +0.20% | [+0.04%, +0.36%] |
| `silu_inplace/896` | +0.18% | [-0.01%, +0.36%] |
| `layer_norm/4096` | -0.11% | [-0.13%, -0.09%] |
| `add_bias_gelu/896` | +0.08% | [+0.07%, +0.08%] |
| `silu_inplace/4096` | +0.08% | [+0.04%, +0.12%] |
| `elementwise_mul/4096` | -0.04% | [-0.06%, -0.03%] |
| `add_bias_gelu/4096` | +0.04% | [-0.03%, +0.10%] |
| `rms_norm/896` | +0.03% | [+0.00%, +0.06%] |
| `gelu/896` | +0.02% | [-0.04%, +0.08%] |
| `rms_norm/4096` | -0.02% | [-0.19%, +0.14%] |
| `layer_norm/896` | -0.01% | [-0.06%, +0.03%] |

The harness labels the `softmax_attention/512` row a win. **This PR does not claim it.** The
diff adds validation on load and dispatch paths and cannot make an attention softmax kernel
faster, and this exact bench has been observed swinging by far more than 3.65% on byte-identical
source in this lane. A 3.65% "improvement" from a diff with no mechanism to produce one is the
same measurement error as the regression it replaced, with the sign flipped, and accepting it
because it flatters the change would be the more embarrassing direction of the same mistake.

### Same-session A/A control, identical source on both arms

Run first, same filter, same resolution, same machine, before the A/B:

| Bench | Δ | 95% CI |
|---|---:|---|
| `softmax_attention/512` | +0.40% | [+0.35%, +0.46%] |
| `rms_norm/896` | +0.36% | [+0.26%, +0.46%] |
| `softmax_attention/128` | -0.13% | [-0.19%, -0.08%] |
| `silu_inplace/896` | -0.12% | [-0.15%, -0.10%] |

with every other gated row inside ±0.08%. So in this session the lane resolved to roughly ±0.4%
on the gated set, and the A/B's -3.65% sits outside that band. That is the strongest statement
the data supports, and it is still not a claim that the diff caused it.

### Why the control is reported even though it passed

An earlier session, on the previous head of this branch, ran the same comparison four times:

| Run | Comparison | `softmax_attention/512` |
|---|---|---:|
| Pass 1 | base → head | +0.36% |
| Pass 2 | base → head | +4.23% |
| Pass 3 | base → head | +5.81% |
| Control | head → head, identical source | **+8.95%** |

The identical-source control produced the largest "regression" of the four and a hard FAIL above
the 7% threshold. That is why the control is run every time rather than once: a clean control is
scoped to its own session, filter, and mode, and expires immediately. Tonight's control was
clean; the previous one was not, on the same bench and the same machine.

### Ambient conditions, including one asymmetry worth stating

| Probe | A/A phase | A/B phase |
|---|---|---|
| before base | 90.2% idle | 89.8% idle |
| between phases | 96.2% idle | 96.3% idle |
| after head | 96.2% idle | 97.3% idle |

All six samples clear the 70% floor, but the two "before base" samples are the lowest of each
phase, and in both cases the largest consumer was a browser renderer (34.7% and 30.2%) that had
fallen to about 1% by the next sample. The probe samples an instant rather than integrating over
a phase, so it cannot establish that the load persisted into the base measurement. It also
cannot establish that it did not. Since base is always measured first, any such asymmetry biases
toward the head arm looking faster, which is the direction of the one non-trivial row above.
Recording it because a reader deciding what to believe about a 3.65% improvement should be told
that a plausible non-code explanation for its sign was present and was not ruled out.

### Note on the confidence intervals

The intervals throughout are tight, and in the earlier session two passes of the *same*
comparison produced non-overlapping intervals ([+0.32%, +0.41%] and [+4.22%, +4.24%]). Criterion's
interval describes dispersion within one measurement run; the run-to-run term is roughly an order
of magnitude larger on this lane and is not represented in it. A narrow interval here indicates a
repeatable measurement, not a trustworthy difference. Filed separately as #1137, since it affects
how the gate's CI-lower-bound thresholds should be read in general and is not specific to this PR.

### Informational band

The lattice-embed SIMD micro-benches remain informational (#714) and are not gated. They are
worth one line here as evidence for that decision: this session's identical-source A/A control
reported up to **+9.18%** [+8.43%, +9.93%] on `tier_prepared_batch_sizes/int8_query_per_call/10`
with no code difference whatsoever. Had that band been gated, the run would have failed against
itself.


ohdearquant:
Reviewed the full diff at 5e4697528d62dd480157d0d481e8baab481a9213. LGTM — approve.

The preflight checks consistently fail closed in the right direction: a source that cannot report a shape falls through to the existing post-materialization check, and a stale entry in `expected_visual_tensor_shape` can only over-reject (loudly), never under-reject, because the post-check remains on every path. The same-handle constraint in `load_f16_tensor_file_expecting` is the right call for untrusted checkpoint directories, and the tests pinning *which* error comes back (not merely that one does) make the preflight ordering genuinely mutation-sensitive. The `F16LoadError` Display/source split with its two-direction test is exactly how that contract should be held.

PR #1132, opened 2026-07-22
chore(ci): group action updates and separate majors from patches

ohdearquant:
Automated action updates were configured to open one pull request per action. That produces a weekly fan-out proportional to the number of actions in use rather than to the number of things actually worth reviewing, and the first run opened five at once.

Patch and minor bumps now travel together in a single request, and a limit caps how many can be open at a time.

Major bumps are grouped separately rather than folded into that same request. Action majors change runtime behavior often enough to deserve being looked at on their own, and bundling them with patches would let a breaking change ride in beside updates that are safe to take on sight. The five requests already open are all majors, which is exactly the split this encodes.

## bench-compare disposition

Not applicable, and this is the narrow case where that phrase is literally true rather than a judgement call: the diff is one YAML configuration file consumed by the hosting platform. No Rust source, no manifest, and no build script changes, so nothing reaches a compiler and base and head bench binaries are byte-identical by construction.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,23 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    open-pull-requests-limit: 3
+    groups:
+      # Patch and minor action bumps are the routine, low-risk ones and travel
+      # together in a single pull request so a weekly run does not fan out into
+      # one review per action.
+      actions-minor:
+        patterns:
+          - "*"
+        update-types:
+          - "minor"
+          - "patch"
+      # Major bumps are separated deliberately. Action majors change runtime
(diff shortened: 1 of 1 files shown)


ohdearquant (commented):

ohdearquant:
Reviewed the full diff at aa499eb8d8353329ae3290b303d1f985dca6f7b — correcting: aa499eb8d8353329ae3290be303d1f985dca6f7b. Approve. A formal approving review cannot be created (author and reviewer are the same account), so this comment is the review record. The actions-minor group correctly scopes to minor+patch with majors deliberately excluded from any group so each gets an individual PR, and the in-file comment explaining why a second group would defeat that is exactly what the next editor needs.

PR #1133, opened 2026-07-22
test(inference): cover checked-loader routing for Qwen3.5 dense-FFN tensors

### LINKED ISSUE #1035: bug(inference): Qwen3.5 safetensors loader accepts config-incompatible tensor shapes (DoS via undersized checkpoint)

ohdearquant:
## Summary

Closes #1035.

Every required tensor in the Qwen3.5 safetensors loader is already routed through
`load_owned_tensor_checked`, which validates each tensor against its config-derived
expected shape at assembly. A finite but config-incompatible tensor is therefore
rejected during `load_weights` with a `ShapeMismatch` error, rather than reaching the
forward pass and panicking in the release-active GEMM bounds guard.

Two gaps remained, both closed here:

- `Qwen35Model::from_safetensors`' doc comment still described dense FFN weights, full
  attention Q/K/V/O and their norms, per-layer input/post-attention norms, and the
  remaining GDN projection tensors as unchecked. It now states that every required
  tensor is checked against a config-derived shape at assembly, matching the loader.
- There was no end-to-end regression covering this path. Added
  `load_weights_rejects_undersized_dense_ffn_tensor`: a finite, undersized `down_proj`
  tensor must make `load_weights` return `Err(ShapeMismatch)` naming `down_proj`.

## Mutation sensitivity

Disabling both shape gates inside `load_owned_tensor_checked` makes the new test fail
at the `Ok(_)` branch — "expected Err for undersized down_proj, got Ok" — confirming
the checked loader is the sole guard and that the test fails for the stated reason, not
incidentally. Restoring the gates returns it to green.

`cargo clippy -p lattice-inference --all-targets --features f16,metal-gpu -- -D warnings`: clean.

## bench-compare disposition

Not applicable. The diff adds a `#[cfg(test)]` regression test and rewrites a doc
comment; there are no non-test, non-doc line changes. Base and head bench binaries
build from identical effective source, so an A/B comparison cannot attribute any delta
to this change.

### DIFF
--- a/crates/inference/src/model/qwen35/loading.rs
+++ b/crates/inference/src/model/qwen35/loading.rs
@@ -924,6 +924,38 @@ mod tests {
         load_weights(&mut src, &cfg).expect("correctly shaped layernorms must load");
     }
 
+    /// Regression for #1035: a finite, undersized dense-FFN tensor must cause
+    /// `load_weights` — the routine `Qwen35Model::from_safetensors` uses to assemble a
+    /// model — to return `Err`, not reach the forward pass and panic in the release-active
+    /// GEMM bounds guard.
+    #[test]
+    fn load_weights_rejects_undersized_dense_ffn_tensor() {
+        let cfg = tiny_linear_layer_cfg();
+        let hidden = cfg.hidden_size;
+        let inter = cfg.intermediate_size;
+        let mut src = tiny_linear_layer_tensors(&cfg, hidden, hidden);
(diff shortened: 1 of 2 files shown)


ohdearquant (commented):

oceanwaves630 (approved):
Verdict read (APPROVE, 0 blocking, not degraded); per-SHA re-verified at merge time: 27 success, 3 path-filtered skips named (layer-23 golden bridge, 2x cargo audit). Approving and merging.

PR #1142, opened 2026-07-26
chore(ci): stop automated updates for the toolchain-selecting action

ohdearquant:
`dtolnay/rust-toolchain` is not an action whose version should be kept current. Its ref selects the
Rust toolchain the action installs, so the pinned revision is a version selector owned by
`rust-toolchain.toml` and the workspace `rust-version`, and it has to move in lockstep with those
rather than on a weekly schedule.

The concrete hazard is the MSRV job in `ci.yml`, whose pin is deliberately held one release behind
every other pin in the repository because that job exists to prove the workspace still compiles at
the declared `rust-version = "1.93"`. An automated bump moves that pin to the 1.94 revision while
the trailing comment keeps reading `# 1.93.1`, leaving the file stating a toolchain it no longer
selects.

That bump does not announce itself, which is why this is a configuration change rather than
something to catch in review. The job sets `RUSTUP_TOOLCHAIN: "1.93.1"` and asserts
`rustc --version | grep -F 1.93.1`, so it stays green while the pin quietly stops matching its
stated intent. Green is the expected outcome there, not the reassuring one, and that trailing
comment is the only thing telling a later reader why one line differs from the other twenty-one.

The MSRV pin itself is unchanged. The added note at that line records why it lags and points at the
configuration that keeps it there.

## Verification

Both pins resolved against the upstream repository rather than format-checked:

| revision | resolves to | comment | agreement |
|---|---|---|---|
| `683cb613…` (current MSRV pin) | branches `1.93`, `1.93.1` | `# 1.93.1` | agrees |
| `e43eeffd…` (proposed by the update) | branches `1.94`, `1.94.1` | would remain `# 1.93.1` | disagrees |

`rust-version = "1.93"` in the workspace manifest; `rust-toolchain.toml` channel `1.94.1`; the other
twenty-one pins of this action are `e43eeffd…` and correctly commented `# 1.94.1`.

`dependabot.yml` parses and the existing grouping and pull-request limit are preserved; `ci.yml`
parses and the MSRV job's pin and `RUSTUP_TOOLCHAIN` are asserted unchanged.

No `crates/` code is touched, so the bench-compare requirement does not apply.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,25 @@ updates:
     schedule:
       interval: "weekly"
     open-pull-requests-limit: 3
+    ignore:
+      # dtolnay/rust-toolchain is not an action whose version should be kept
+      # current. Its ref selects the Rust toolchain the action installs, so the
+      # pinned revision is a version selector owned by rust-toolchain.toml and
+      # by the workspace rust-version, and it has to move in lockstep with them
+      # rather than on a weekly schedule.
+      #
+      # The concrete hazard is the MSRV job in ci.yml, whose pin is deliberately
+      # held one release behind every other pin in the repository because that
+      # job exists to prove the workspace still compiles at the declared
+      # rust-version. An automated bump moves that pin to the newer toolchain
+      # while the trailing version comment keeps claiming the older one, leaving
+      # the file stating a toolchain it no longer selects.
+      #
+      # That bump does not announce itself. The job pins RUSTUP_TOOLCHAIN and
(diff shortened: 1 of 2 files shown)


ohdearquant:
Reviewed at head 0579422c4246344ae2aaa1bd717057b4f9b3a010, full diff. The ignore rule states the correct reason (this action's ref is a toolchain selector owned by rust-version, not a version to keep current) and the pin-line note captures the install-versus-select distinction precisely. All five checks green at this SHA. Approve.

PR #1162, opened 2026-07-28
fix(inference): gate QuaRot promotion on perplexity

### LINKED ISSUE #1103: bug(quant): QuaRot converter CLI exits 0 without running the PPL acceptance gate
## Summary

`quantize_quarot` (crates/inference/src/bin/quantize_quarot.rs) produces a complete rotated-quantized model (Q4/F16 tensors + index + mutated config) and exits 0 after running only the forward-equivalence gate. It never invokes the perplexity acceptance gate. A converted artifact is therefore promoted as "passed" without the quantization-quality criterion ever being checked.

## Detail

- The CLI is a thin wrapper over `convert_quarot_qwen35`, whose pipeline runs a forward-equivalence refuse-on-fail check (algebraic, `--tolerance`, default 1e-5). That gate verifies the Hadamard rotation and fusion preserve the forward pass. It does not measure quantization quality.
- The perplexity acceptance criterion lives in a separate binary, `eval_perplexity` (crates/inference/src/bin/eval_perplexity.rs), which is never called from the converter path. `grep eval_perplexity crates/inference/src/bin/quantize_quarot.rs` returns nothing.
- ADR-044 frames "delta < 0.5 PPL" as the acceptance gate, but nothing on the artifact-production path enforces it. A model that rotates correctly but quantizes poorly (PPL past budget) still writes its outputs and exits 0.

## Why it matters

The forward-equivalence gate and the PPL acceptance gate check different things: rotation correctness versus quantization quality. Only the former is wired into promotion. This is the unwired-quality-gate class: a success exit plus a complete artifact reads as "validated" when the quality gate never ran.

## Suggested fix

Wire the PPL acceptance gate into the converter promotion path (or an explicit `--acceptance` mode) gated on a corpus and threshold, so exit 0 implies both gates passed. Alternatively, make the two-step contract explicit and refuse to mark an artifact promotable until the quality gate has been recorded against it.

## Related

- #622 (QuaRot PPL delta is a smoke test at 0.8B; fair-scale validation)
- #801 (route QuaRot CLIs through the shared ingress validation seam)
- #1081 (NaN logits evade the forward-equivalence tolerance gate)

ohdearquant:

Closes #1103.

## Defect

`quantize_quarot` wrote a complete rotated Q4 artifact and exited 0 after only the algebraic forward-equivalence gate. The independent dual-Q4 perplexity gate in `eval_perplexity` was never invoked, so conversion success could be mistaken for quantization-quality acceptance.

## Contract and fix

- Convert into a hidden sibling staging directory and publish `--output-dir` only after the PPL gate passes.
- Require an absolute `--ppl-evaluator` path, an unrotated `quantize_q4` baseline, tokenizer, and held-out corpus. There is no PPL opt-out; `--dry-run` performs the full staged conversion and both gates before discarding staging.
- Require the evaluator to exit 0 and emit exactly one finite `q4` and one finite `quarot` structured measurement with matching token/window coverage. Recompute the delta locally and reject missing, duplicate, non-finite, inconsistent, or over-threshold evidence even if the evaluator reports success.
- Reject a rotated artifact masquerading as the unrotated baseline and verify that the staged manifest records the requested rotation seed.
- Write `quarot_ppl_acceptance.json` before promotion. The receipt records policy/evidence plus SHA-256 hashes of the artifact manifest, evaluator, baseline manifest, tokenizer, and corpus.
- Preserve ADR-044's nominal `0.5` default. The composed-golden diagnostic explicitly uses a `3.0` regression ceiling around offline QuaRot v0's documented +~2.4 PPL limitation; that non-shipping policy is visible in the receipt.

## Invocation sweep

`rg` found two non-document converter launch paths plus Lattice Studio:

- `.github/workflows/e2e-parity.yml` now builds `quantize_q4`, `quantize_quarot`, and `eval_perplexity`, creates the unrotated baseline, runs the real PPL path, and verifies the receipt and manifest hash.
- `scripts/gen_quarot_q4_composed_golden.py` does the same; `--skip-convert` now refuses missing, failing, incomplete, or manifest-mismatched receipts.
- Lattice Studio's prior QuaRot button supplied none of the gate inputs, so it is disabled with an honest CLI-only explanation until the UI can collect them. Packaging/build scripts only build the binary; `dump_quarot_q4_golden` only records the converter string as fixture provenance.

## Mutation evidence

With only the production evaluator call removed while the test remained unchanged:

```text
cargo test -p lattice-inference --bin quantize_quarot \
  promotion::tests::promotion_requires_gate_evidence_before_publishing_output \
  -- --exact --nocapture

FAILED
left:  ["convert"]
right: ["convert", "gate"]
```

After restoring the evaluator call, the same isolated test passed.

## Validation

Apple Silicon validation:

- `cargo test -p lattice-inference --bin quantize_quarot promotion::tests -- --nocapture` — 8 passed.
- `cargo test -p lattice-inference quant::quarot::convert::tests -- --nocapture` — 42 passed.
- `cargo fmt --all -- --check` — passed.
- `cargo clippy --workspace -- -D warnings` — passed.
- `python3 -m py_compile scripts/gen_quarot_q4_composed_golden.py` — passed.

Local checks:

- `make lint-docs` — passed.
- Changed ADR/model/quantization docs pass `deno fmt --check`.
- Workflow YAML parses successfully; converter `--help` and missing-gate fail-closed smoke checks passed.

`swift test --package-path apps/macos` remains blocked by the host's pre-existing missing `PreviewsMacros` plugin across unrelated preview files. The compiler compiled the changed `QuantizeTab.swift` before reaching those failures.

## bench-compare disposition

No benchmark was run. The Rust change is confined to the offline `quantize_quarot` binary's cold orchestration/receipt path; Criterion inference benches compile the library and no runtime kernel or hot-path library code changed, so base and head benchmark binaries have identical effective inference source.

- Pull request head evaluated: `c976e1215e68f7dac82cccce5adca7c1f7bb4308`

This statement is void if either ref moves, and a moved head means it
has discharged nothing and must be re-derived.


ohdearquant:
Closing in favour of #1331, on the ADR-044 policy rather than on implementation quality.

This branch makes conversion refuse to publish unless a perplexity gate passes. ADR-044's 2026-07-06 amendment records that offline QuaRot v0 is a net negative against unrotated asymmetric Q4 on the current, numerically-correct forward path, with a controlled code-bisection establishing it. A hard converter-internal gate on that criterion would therefore refuse every real conversion today, and whether QuaRot v0 should be publishable at all is an ADR-044 decision (tracked with #703 for online R3/R4 and #616), not one for the converter.

#1331 addresses the defect #1103 actually names — an artifact promoted as validated without the criterion ever running — by stamping fresh artifacts unpromoted, removing the implication that exit zero means quality-validated, and recording the evaluator's verdict into the artifact fail-closed.

The staged publication design here is worth revisiting once online rotation makes the delta non-negative: at that point the gate has a threshold it can enforce without blocking legitimate work.

PR #1166, opened 2026-07-28
perf(embed): make quantization explicitly SIMD

### LINKED ISSUE #1063: embed: guarded hot loops relying on auto-vectorization are one CGU reshuffle from silent regression
## Summary

#1062 fixed a +47.6% regression caused by LLVM silently dropping auto-vectorization of a guarded hot loop when unrelated commits changed codegen-unit partitioning (source was byte-identical across the regression window; the compiled `from_f32` went from 51 NEON fmin/fmax instructions to 3).

The same pattern — a hot loop whose throughput depends on the auto-vectorizer masking an `is_finite()` guard — exists at sibling sites that are currently fast but one CGU reshuffle away from the same silent regression:

- `QuantizedVector::from_f32` quantize map loop (`(v * scale).round().clamp(...)` with non-finite guard), crates/embed/src/simd/quantized.rs
- `Int4Vector::from_f32`, crates/embed/src/simd/int4.rs
- `BinaryVector::from_f32`, crates/embed/src/simd/binary.rs

Follow-up shape: for each site, either convert to an explicit SIMD kernel with scalar fallback (the file idiom, as in #1062), or add a disassembly-level guard to the bench-audit tooling that counts vector ops in the compiled hot functions so a future demotion is caught at the ledger instead of by a user.

Deliberately out of scope: the L2-norm accumulation loops must stay scalar-ordered — vectorizing them reassociates f32 addition and changes results.

## Note: tier_prepared_query/int8_query_once residual

The +8.2% on `int8_query_once/1000` (vs 2026-06-23) is a different, explained cost: the prepared-distance API changed from infallible `f32` return to `Result<f32>` in the same window, adding per-candidate overhead on a ~12ns/candidate loop. That is a deliberate API-shape tradeoff (typed tier-mismatch errors), recorded here as measured context, not proposed for revert.

ohdearquant:

## Summary

Replace compiler-dependent guarded quantization loops with explicit SIMD dispatch:

- INT8 finite min/max reduction and float-to-INT8 conversion;
- INT4 finite max-absolute reduction and packed-nibble conversion;
- binary threshold conversion and bit packing.

Each path now has NEON and AVX2 kernels plus the existing scalar behavior as a fallback. L2 norm accumulation remains scalar-ordered.

Closes #1063.

## Defect

These hot loops mixed finiteness guards, conversion, and packing. Their performance therefore depended on LLVM continuing to auto-vectorize the exact source shape across codegen-unit and compiler changes. INT8 reduction had already demonstrated a silent 47.6% regression after such a change; the conversion loops and x86 reducers retained the same failure mode.

## Implementation

- dispatch once through the cached SIMD capability configuration;
- process bounded full vectors with unaligned NEON or AVX2 loads;
- substitute zero for non-finite source lanes;
- preserve round-half-away-from-zero for INT8 and INT4;
- preserve INT4 high-nibble-first and binary most-significant-bit-first layouts;
- leave incomplete vectors and final partial packed bytes to scalar tails;
- retain scalar fallbacks for targets without the selected feature.

ADR-085 records the dispatch and numerical contracts. The unsafe-code inventory is updated for the nine new production dispatch blocks.

## Regression and mutation proof

Architecture tests compare SIMD and scalar results across empty, tail, and vector-boundary lengths; NaN and infinities; signed zero; exact rounding ties; and binary thresholds including NaN and infinities. Test-only hit counters prove the public constructors dispatch both applicable reducers and converters.

Three production-only mutations were exercised with the tests unchanged:

```text
Bypass all three constructor conversion dispatches:
QuantizedVector::from_f32 ... left: 13, right: 14
Int4Vector::from_f32 ... left: 13, right: 14
BinaryVector::from_f32 ... left: 65, right: 66

Change INT4 NEON ties-away conversion to ties-even:
left:  [6, 143, 104, 136]
right: [7, 159, 121, 136]

Bypass both constructor reduction dispatches:
Int4Params::from_vector ... left: 13, right: 14
QuantizationParams::from_vector ... left: 15, right: 16
```

Restoring each production path made the five explicit-SIMD regressions pass.

## Validation

```text
cargo test -p lattice-embed
289 unit tests passed; all integration tests passed; 4 HF parity tests passed, 1 ignored

cargo test -p lattice-embed --lib explicit_simd
5 passed

cargo clippy --workspace --all-targets -- -D warnings
passed

cargo fmt --all -- --check
passed

scripts/lint-docs.sh
passed

cargo test --target x86_64-apple-darwin -p lattice-embed --no-run
all test binaries built

cargo check --target wasm32-unknown-unknown -p lattice-embed --lib
passed
```

Native aarch64 execution covered the NEON paths. The x86_64 test build and scalar tests passed under Rosetta, but that environment reports AVX2 unavailable, so it did not execute the AVX2 regressions. Release assembly contains the named AVX2 kernels and vector conversion/packing instructions, but execution on real AVX2 hardware and manual intrinsic verification remain required before merge.

## Sibling sweep

- INT8: both the finite min/max reducer and integer conversion now dispatch explicitly.
- INT4: both the finite max-absolute reducer and packed conversion now dispatch explicitly.
- Binary: threshold conversion and bit packing now dispatch explicitly.
- Constructor L2 norms intentionally remain scalar because SIMD reassociation would change floating-point accumulation order.
- Other targets retain scalar fallbacks; no additional guarded quantization or packing sibling remained in these modules.

## Signed-zero tie-break across the three minmax kernels

The finite min/max reduction had its signed-zero normalization in exactly one of three
kernels. AVX2 normalized the result; NEON and the scalar reference folded bare
`f32::min`/`f32::max`, which return an unspecified one of their operands when the operands
compare equal. The scalar-vs-SIMD parity test then asserted bit equality across kernels, i.e.
against a tie-break nothing specifies. aarch64 agreed by coincidence and x86-64 did not, so
`test_finite_minmax_explicit_simd_matches_scalar` failed on `ubuntu-latest` at `dim=7`
(`+0.0` vs `-0.0`) while both aarch64 legs passed.

The normalization is now one `pin_zero_signs` helper applied in all three kernels, making
IEEE 754-2019 `minimum`/`maximum` ordering (`-0.0 < +0.0`) a stated contract rather than a
property of codegen: the min takes `-0.0` and the max takes `+0.0` when that sign is present.
The sign scan runs only when a bound is exactly zero, so the cost on a typical vector is the
two comparisons in the guard.

`test_minmax_finite_pins_zero_signs` pins the contract directly. It drives `pin_zero_signs`
with wrong-signed bounds rather than routing through the scalar fold, because on aarch64 the
fold already returns the contracted signs — assertions routed through it still pass with the
sign pass deleted, so the direct form is what makes the guard fail on every target when the
contract is removed.

## bench-compare disposition

`make bench-compare`, quick mode, base `94c473d7bd` vs head `453946918f`, embed filter
`int8_|int4_|binary_`, 93 measurements. The inference target was left at its default
deliberately: this PR does not touch `crates/inference`, so those groups act as an in-run
control on unchanged source.

This re-runs an earlier A/B measured at `07a64d2e06`. The head has since gained
`pin_zero_signs` on the scalar and NEON `minmax_finite` paths, which are on the embed hot
path, so the earlier numbers no longer described the head that would merge. They are replaced
rather than caveated.

**Run conditions**

```
resolution: --quick
filters: inference='<all>' embed='int8_|int4_|binary_'
enforcement: report-only (gate status printed, exit 0)
locks: bench-window acquired immediately (uncontended)
       Metal GPU  acquired immediately (uncontended)
       heavy lane held exclusively (all 3 slots)
ambient load: [quiet] before base    idle 86.1% (floor 70.0%) ok
              [quiet] between phases idle 82.6% (floor 70.0%) ok
              [quiet] after head     idle 87.9% (floor 70.0%) ok
```

`BENCH_IDLE_FLOOR` left at its default and not lowered at any point. An earlier attempt was
refused by the harness at 56.7% idle and was not overridden.

**Result: one effect, reproduced and scale-invariant; the rest is not attributable.**

| Benchmark | Δ | 95% CI | base ns | new ns |
|---|---:|---|---:|---:|
| `tier_prepared_batch_sizes/int4_query_per_call/1000` | **−54.19%** | [−54.72%, −53.65%] | 1,652,382.8 | 756,974.0 |
| `tier_prepared_batch_sizes/int4_query_per_call/100` | **−52.98%** | [−53.40%, −52.57%] | 160,352.6 | 75,392.4 |
| `tier_prepared_batch_sizes/int4_query_per_call/10` | **−52.98%** | [−53.09%, −52.87%] | 15,934.7 | 7,492.5 |
| `tier_prepared_query/int4_query_per_call_1000` | **−52.88%** | [−53.20%, −52.56%] | 1,622,778.7 | 764,581.2 |

Roughly a 2.1× speedup on the int4 query-per-call path, holding across three input scales and
reproducing across two independent runs at two different heads (−52.79% at `07a64d2e06`,
−52.88% here for the common benchmark). Its A/A control row is **+0.59%**, so the effect is
about 90× its own control and about 4× the largest excursion anywhere in the A/A.

**Everything else is not attributed, and the report itself shows why.** The gate printed
"3 confirmed improvement". All three are in `crates/inference`, which is byte-identical
across both arms:

| Benchmark (unchanged source) | Δ | 95% CI | gate verdict |
|---|---:|---|---|
| `residual_add_layer_norm/unfused/hidden384_seq128` | −7.15% | [−8.24%, −6.04%] | WIN |
| `gelu/4096` | −4.58% | [−7.64%, −1.39%] | WIN |
| `rms_norm/896` | −4.30% | [−6.05%, −2.50%] | WIN |

A gate declaring three wins on code that did not change is the disposition for every sub-10%
row in this report. The embed rows in that band — `int8_batch_cosine/float32_simd/1000`
+5.45%, `int8_prepared_dot_product/per_call/129` +5.34%, and a cluster of −3% to −7%
"WIN (informational)" rows — are claimed in neither direction.

**A/A control, both arms `94c473d7bd`, same filter and mode.** On byte-identical source the
lane produced two gate failures and a −13%…+14% spread:

| Benchmark (identical source) | change | 95% CI | verdict |
|---|---|---|---|
| `binary_cosine_distance/binary/384` | +14.28% | [+13.68%, +14.89%] | FAIL |
| `int4_cosine_distance/float32_simd/384` | −12.91% | [−14.68%, −11.13%] | WIN |
| `binary_cosine_distance/float32_simd/768` | −10.19% | [−11.94%, −8.38%] | WIN |
| `int8_raw_dot_product/dot_product_i8_raw/128` | +8.74% | [+7.69%, +9.80%] | FAIL |

The control was not re-run at this head. It bounds the lane rather than the diff, and neither
the machine nor the filter changed, so the earlier run still applies — stated explicitly
because reusing a control is itself a claim.

**On the exit code:** enforcement was report-only, and all nine filtered groups are excluded
from gating as quick-mode informational-only (#714). A zero exit here means the harness ran,
not that anything passed. Narrow Criterion intervals are also not reassurance: the interval
is a within-run dispersion estimate and says nothing about between-arm bias, so a biased lane
reports a tight interval around the wrong number.

**Disclosed confound:** the quiet samples record Python and sqlite3 consumers at up to 55.7%.
All three samples cleared the 70% floor and the floor was not lowered, but the machine was not
idle and the load was not symmetric across phases. That is a further reason the sub-10% rows
are left unclaimed.

`--full` was deliberately not run. More samples do not correct between-arm bias, so the A/A
control was the right instrument rather than a longer A/B.

Still outstanding, unchanged by this measurement: manual SIMD intrinsic verification and AVX2
execution on capable hardware.

### DIFF
--- a/crates/embed/docs/simd.md
+++ b/crates/embed/docs/simd.md
@@ -24,6 +24,7 @@ and normalization reuse the cached configuration when they choose a kernel.
 | Float dot, cosine, squared L2, normalize | AVX-512F, then AVX2 + FMA                                       | NEON                        | SIMD128 when compiled in | scalar   |
 | Batch-four float dot                     | AVX2 + FMA                                                      | NEON                        | none                     | scalar   |
 | INT8 dot                                 | AVX-512 VNNI + BW when the `avx512` feature is built, then AVX2 | NEON only with FEAT_DotProd | none                     | scalar   |
+| INT8, INT4, and binary quantization      | AVX2                                                            | NEON                        | none                     | scalar   |
 | Binary Hamming                           | none                                                            | NEON `vcnt`                 | none                     | scalar   |
 | INT4 dot                                 | none                                                            | NEON                        | none                     | scalar   |
 
@@ -229,6 +230,11 @@ Mapping `[-max_abs, max_abs]` to `[-127, 127]` gives a step size of
 be acceptable for the tier selected by the caller; use `Full` or `Int8` instead
 of lower-precision tiers when recall needs more fidelity.
 
+The finite min/max scan and integer conversion use explicit AVX2 or NEON
+kernels, with a scalar fallback. Conversion keeps round-half-away-from-zero
+semantics and finishes lengths shorter than a full vector in the scalar tail.
+The source L2 norm remains a scalar-ordered reduction.
+
 ### Dot product and dispatch
 
 The raw integer reduction is dequantized by dividing by the product of the two
@@ -276,6 +282,10 @@ bytes. For an odd-dimensional vector, only the final byte's high nibble is a
 real dimension; the low nibble is padding and is ignored by dequantization and
 dot-product accumulation. Non-finite source values are treated as zero, and
 the L2 norm records the finite source values.
+
+Finite max-absolute reduction and nibble conversion use explicit AVX2 or NEON
+kernels. Both leave incomplete chunks to the scalar implementation, including
+an odd final high nibble. The source L2 norm remains scalar-ordered.
 
 The 16 quantization levels have step size `2 * max_abs / 15`, so the maximum
 per-element round-trip error is `max_abs / 15`. This is a storage-oriented
@@ -325,6 +335,10 @@ dimension one is bit 6, and so on. The storage size is
 `ceil(dimensions / 8)` bytes. Dequantization maps one to `+1.0` and zero to
 `-1.0`; it is intentionally lossy and returns an empty vector if public fields
 describe a buffer shorter than the required packed length.
+
+Threshold conversion uses explicit AVX2 or NEON kernels for full eight-value
+groups and a scalar tail for the final partial byte. The kernels preserve the
+same finite-value substitution and most-significant-bit-first layout.
 
 Hamming distance is the population count of the XOR of two packed buffers. The
 scalar implementation counts full 64-bit groups and remaining bytes; the NEON
--- a/crates/embed/src/simd/binary.rs
+++ b/crates/embed/src/simd/binary.rs
@@ -7,6 +7,9 @@
 #[cfg(target_arch = "aarch64")]
 use std::arch::aarch64::*;
 
+#[cfg(target_arch = "x86_64")]
+use std::arch::x86_64::*;
+
 use super::simd_config;
 
 /// **Unstable**: binary quantization format and struct layout are under active design.
@@ -45,16 +48,7 @@ impl BinaryVector {
         let norm = norm_sq.sqrt();
 
         let packed_len = dims.div_ceil(8);
-        let mut data = vec![0u8; packed_len];
-
-        for (i, &v) in vector.iter().enumerate() {
-            let val = if v.is_finite() { v } else { 0.0 };
-            if val >= threshold {
-                let byte_idx = i / 8;
-                let bit_idx = 7 - (i % 8); // bit 7 = first dimension in byte
-                data[byte_idx] |= 1 << bit_idx;
-            }
-        }
+        let data = quantize_binary(vector, threshold, packed_len);
 
         Self { data, dims, norm }
     }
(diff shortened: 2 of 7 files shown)


ohdearquant:
Reviewed at 453946918f11b40f4305046ab09d43570f429c2a.

- `pin_zero_signs`: verified against IEEE 754-2019 minimum/maximum on mixed-zero, single-sign-zero, and negative-with-negative-zero inputs; the early return on both-nonzero extremes is sound, and applying it after each kernel's fold makes the intra-kernel tie-break order irrelevant — which is the right shape for the fix.
- All three kernels (AVX2, NEON, scalar) route through the one helper; no sibling path left folding bare min/max.
- Regression test drives the helper directly with wrong-signed bounds, so it discriminates on every target rather than only where the platform fold happens to break ties the other way. The test-only SIMD-hit counter confirming the vector path actually executes is a nice touch.
- Per-commit CI green across all six runs at this SHA; skipped jobs verified path-gated. Bench A/B disposition in the PR body is consistent with the diff's reach.

LGTM at this SHA.

PR #1187, opened 2026-07-28
fix(inference): align M-RoPE video axes with reference

### LINKED ISSUE #1101: M-RoPE: two latent gaps on the deferred multi-frame/video path (t>1)
Two latent M-RoPE gaps that do not affect any input the crate can construct today, but are the exact lines that will be wrong if multi-frame / video support (deferred in ADR-069) is ever added. Filing together because both live on the same `grid.t > 1` path and both were verified against HF `transformers/models/qwen3_5/modeling_qwen3_5.py`.

Both are confirmed by worked arithmetic, not by reading. Neither is a live bug: every official checkpoint declares the M-RoPE default that matches, and every `GridThw` construction site in the crate hardcodes `t: 1` (`crates/inference/src/vision/qwen35_vit.rs:107`), so both formulas below are numerically identical to HF on every input the pipeline can produce.

## Gap 1: the mrope_section cyclic-count gate false-rejects one regime

`crates/inference/src/vision/qwen35_mrope.rs:245-258` rejects any `mrope_section` whose per-axis counts do not equal the pure cyclic `T,H,W,T,H,W,...` schedule. Transcribing HF `apply_interleaved_mrope` (`modeling_qwen3_5.py:246-262`) and running it shows the gate is wrong in exactly one regime: when `section[1]*3 >= rope_half` and `section[2]*3 >= rope_half`, HF's strided overwrite saturates and degenerates to the pure cyclic schedule regardless of the declared counts, so the gate rejects a section HF would accept and produce identical output for.

Four cases, worked:

| section @ rope_half | our gate | HF actual lane counts | assignment matches ours |
|---|---|---|---|
| `[11,11,10]` @ 32 (HF default) | accept | `[11,11,10]` | yes |
| `[16,24,24]` @ 64 | **reject** | `[22,21,21]` | **yes, a false rejection** |
| `[22,21,21]` @ 64 | accept | `[22,21,21]` | yes |
| `[20,6,6]` @ 32 | reject | `[20,6,6]` | no, rejection is correct |

The naive fix ("loosen the gate") is wrong: uniformly accepting would silently mis-rotate the `[20,6,6]` regime, where HF genuinely leaves a tail of `T` lanes that our cyclic assignment gets wrong. The correct fix is to implement HF's algorithm directly (start every lane at `T`, then overwrite the strided `H` and `W` ranges), after which the gate is unnecessary rather than merely more permissive.

## Gap 2: post-image position advance omits the temporal axis

`crates/inference/src/vision/qwen35_mrope.rs:155` advances the running position past an image block by `lh.max(lw)`. HF `get_rope_index` advances by `max(lt, lh, lw)` (the block's max coordinate on any of the three axes, plus one). These differ whenever `lt > max(lh, lw)`.

Counterexample: grid `(t=3, h=4, w=4)`, spatial merge `m=2` -> merged `(lt=3, lh=2, lw=2)`, preceding text ending at position 5. The image block uses T-coordinates up to 7. HF's next text position is `5 + max(3,2,2) = 8`; ours is `5 + max(2,2) = 7`, which collides with the T-coordinate already assigned to the `t=2` image rows.

Because `lt` is always 1 today, `max(1, lh, lw) == max(lh, lw)` and the output is bit-identical to HF. The docstring and code both state the general `max(H, W)` rule rather than the `t=1`-specialized one, so this is the line to fix when video lands, and nothing today would catch a regression there.

## Suggested tests (for when the path becomes reachable)

- Gap 1: a table test asserting our lane assignment equals HF's `apply_interleaved_mrope` for each of the four sections above, in particular `[16,24,24]@64` (must be accepted) and `[20,6,6]@32` (must stay rejected or be handled correctly by the direct implementation).
- Gap 2: `post_image_advance_accounts_for_temporal_axis`: grid `(3,4,4)` `m=2`, assert the next text position is 8, not 7. It must fail with the current `lh.max(lw)` line and pass with `max(lt, lh, lw)`.

ohdearquant:

## Summary

- implement the reference M-RoPE saturating strided lane assignment instead of requiring a pure cyclic section split
- include the temporal dimension when advancing positions after a merged vision block
- cover the deferred multi-frame path with reference-table and collision regressions

Closes #1101.

## Defect

`build_cos_sin` treated every valid `mrope_section` as a declaration of the final cyclic T/H/W lane counts. The reference algorithm instead starts every lane on T, then overwrites bounded H and W strided ranges. Those descriptions agree for the current `[11,11,10]` configuration, but the cyclic-count gate falsely rejected saturated layouts such as `[16,24,24]`, while simply removing the gate would have assigned the wrong tail lanes for layouts such as `[20,6,6]`.

`build_position_ids` advanced past an image run by `max(lh, lw)`. A future multi-frame grid with `lt > max(lh, lw)` therefore resumed trailing text on a temporal coordinate already consumed by the image.

## Fix

Lane selection now mirrors the reference assignment directly: T is the default, H owns strided lanes while their section index is below `mrope_section[1]`, and W does the same for `mrope_section[2]`. The existing three-entry and exact-sum validation remains in place.

Post-image advancement now uses `max(lt, lh, lw)`. Existing image-only behavior is unchanged because every currently constructed grid has `t = 1`, and the current official section `[11,11,10]` produces the same lane assignment as before.

## Regression and mutation evidence

- the lane table covers `[11,11,10]@32`, saturated `[16,24,24]@64`, canonical `[22,21,21]@64`, and non-cyclic `[20,6,6]@32`, checking the resulting cos/sin lanes against an independent staged-overwrite oracle
- the temporal regression uses a merged `(3,2,2)` grid and proves trailing text resumes at `(8,8,8)` after the image's T axis reaches 7
- reverting the temporal maximum made the regression fail with `(7,7,7)` instead of `(8,8,8)`
- restoring the old cyclic-count gate made the lane table fail by rejecting `[16,24,24]`
- removing only the new bounded lane selection while leaving the gate absent made the lane table fail on the wrong axis value
- restoring each production hunk returned all 20 M-RoPE tests to green

## Sibling and edge-case sweep

- searched the inference and vision trees for the old `lh.max(lw)` advance and found no sibling implementation
- searched every M-RoPE consumer; CPU, Metal, and multimodal request paths all delegate to this shared builder rather than maintaining independent lane-selection logic
- retained the existing section length, section sum, overflow, zero-dimension, and image-run validation
- retained current text-only, single-frame, default-section, CPU wiring, Metal wiring, and HF golden controls

## Validation

- `cargo test -p lattice-inference qwen35_mrope --lib` — 20 passed
- `cargo test -p lattice-inference --all-features qwen35_mrope --lib` — 20 passed
- `LATTICE_GEMMA4_GATE_SKIP=1 cargo test -p lattice-inference`
- `cargo clippy --workspace --all-targets -- -D warnings`
- pre-commit checks
- `git diff --check`

## bench-compare disposition

Not run. The behavioral delta is confined to arbitrary M-RoPE section layouts and `t > 1` grids; current official single-frame/default-section outputs remain bit-identical, and the benchmark suite does not exercise multimodal M-RoPE construction or the deferred video path. This PR makes no throughput claim; the reference-table and mutation-sensitive numerical tests are the relevant evidence.

- Pull request head evaluated: `37170b4455d8077d8f761ffc443f5443b871fa84`

re-derived.

## bench-compare disposition

Quick-mode A/B at base `d4e309ebd6` (main) vs head `53409b7442`: all 17 gated `elementwise_cpu_bench` benches within the ±3.0% noise band. Coverage note: no bench group executes the M-RoPE index path, so the A/B guards whole-binary codegen effects; correctness of the changed path is carried by the 140 `vision::` + 28 `mrope` tests passing at head. A first run was discarded as not-measurable (arm-order bias 21.85% exceeded the 7% margin) and rerun under quiet conditions.


ohdearquant:
Design note on the post-image advance, so the direction is on the record rather than re-derived later.

The advance is now spatial-only, `max(H/m, W/m)`. Folding the temporal extent in looks correct if you read the merged grid as a single `(T, H/m, W/m)` volume, but it does not match the reference: HF (v5.12.1) splits a video into `t = 1` frame grids **before** position assignment and then advances by the spatial maximum. So a `max(T, H/m, W/m)` advance diverges from HF on precisely the path it appears to serve, while this module handles image runs only. That combination is the problem worth avoiding: no observable effect today, and a reference divergence that becomes load-bearing the day the video path lands.

The video axes work re-enters **with** the video path, not ahead of it, implemented against those HF semantics (frame-grid split first, spatial-max advance). Acceptance for that work is HF numeric parity on the video case, not a bench A/B: the default bench selection does not exercise M-RoPE, so an A/B has nothing to say about it.

The lane production rule in this PR is unchanged and matches HF's bounded strided overwrite.

The test was rewritten to pin the discriminating case instead of one both conventions satisfy: `t = 3` against merged spatial extents of 2, where spatial-only places the following text token at 7 and folding T in would place it at 8. It is mutation-sensitive by execution, restoring the temporal advance fails it with left `(8, 8, 8)`, right `(7, 7, 7)`.

PR #1208, opened 2026-07-29
fix(inference): resolve </think> by token content for reasoning budget

ohdearquant:
## The defect

The reasoning budget silently enforces nothing on some tokenizers.

Budget enforcement needs the id of the `</think>` marker, and it resolved that id with
`tokenizer.special_token_id("</think>")`, which searches only the `special_tokens` table.
A tokenizer that declares `</think>` as an ordinary vocabulary entry, or as an added token
with `special: false`, yields `None` there. The old code then treated `None` as "no budget
to enforce" and generated normally.

The failure mode is the bad one. A caller sets a budget, receives a successful generation,
and never learns the budget was inert. Nothing logs, nothing errors, and the only symptom
is output that ignores a limit the caller believes is in force.

## The fix

**Resolution by token content.** `BpeTokenizer::token_id_for_content` looks up a token by
its exact rendered spelling across all three places a marker can live, in order:
`special_tokens`, then `vocab`, then a linear scan of `added_render`. It is deliberately one
function rather than a lookup inlined at each call site, so the generation-time validator and
the Metal cross-turn prefix-cache fingerprint cannot drift apart about what `</think>` is —
they now call the same resolver, and the fingerprint change in `metal_qwen35.rs` is part of
this PR for that reason.

**Resolve and validate, rather than resolve and hope.** `resolve_reasoning_close_token`
returns `Ok(None)` without touching the tokenizer when no budget is set, so the no-budget
path is unchanged. With a budget set, a marker that cannot be resolved, or an id at or beyond
`vocab_size`, is an `InvalidInput` error. Refusing the request is the point: the caller
learns the budget cannot be honoured instead of receiving output that quietly ignored it.

**Hoisting, which is what makes it a fix rather than a lookup change.** The resolve now runs
above every early-success return, at five sites: CPU `generate`, CPU
`generate_streaming_with_observer`, Metal `generate_streaming_with_cancel`, the Metal
prefix-cache `_inner`, and the prefix-cache public wrapper. Without the hoist a request can
return `Ok` having never confirmed its budget was enforceable — the same silent failure one
layer up, reached through cancellation, a step-0 grammar block, a grammar-advance rejection,
or a first-token EOS. The wrapper site matters separately: `_inner`'s blanket `Err` handling
evicts the cache slot, so a validation rejection raised inside it would be misread as cache
corruption.

**`RawGenEvent::PrefillEnd` ordering.** It fired after the post-prefill cancellation check
and after grammar masking, both of which can return early, so an observer watching a request
that disconnected during prefill never saw prefill end even though prefill had completed. It
now fires as soon as the prefill logits are in `scratch.logits`, which is the actual
prefill/decode boundary.

## Tests, and what makes each able to fail

Every test below was checked by reverting the thing it guards and watching it fail, rather
than by inspection.

Collapsing `token_id_for_content` back to the `special_tokens` tier alone fails four tests.
The tier test asserts `special_token_id` returns `None` for the `vocab` and `added_render`
cases, so it cannot pass against the old lookup; the two resolver tests reach the same
tiers through `resolve_reasoning_close_token`.

Moving both CPU hoists back below the early returns, together with the `PrefillEnd` move,
fails five tests. Each drives one specific early-return path with a budget set and a
tokenizer that has no `</think>` at all, and asserts the typed `InvalidInput` carrying both
`</think>` and `reasoning_budget`, rather than merely that an error occurred:
first-token EOS on both the plain and streaming paths, grammar-advance rejection,
pre-prefill cancellation, and post-prefill cancellation for the `PrefillEnd` ordering.

One of those needed a third, separate mutation to prove out. Un-hoisting above the
`RawToken` emit left the observer test still passing, so the discriminating mutation for
that test is moving the resolve *below* the emit. Recording it because the second run does
not establish what it might look like it establishes.

## Stated gap

The Metal-side hoists are compiled and clippy-clean under `--features f16,metal-gpu`, but
they are **not covered by a test run here**: those test modules need a live Metal device.
The CPU-side hoists are the same construct and are mutation-proven above, so the risk is
that a Metal site was edited wrongly rather than that the approach is wrong. The five sites
are named explicitly in this description so a reviewer can check them by name.

Separately, the no-budget pin passes with or without this change. It guards against a future
regression on the untouched path; it is not evidence for this defect and is not offered as
such.

## Gates

`cargo fmt --all --check` clean. `cargo clippy -p lattice-inference --all-targets --features
f16,metal-gpu -- -D warnings` finished rather than aborting, so its empty diagnostic set is
complete rather than a lower bound. `cargo test -p lattice-inference --lib`: 2414 passed,
0 failed, 20 ignored.

## Ambiguity is the same defect as absence

A second commit closes a gap the first one left open. Resolving by content still
resolved to *one* id, and `DecodePolicy` recognises exactly one close-marker id, so a
tokenizer that spells `</think>` at two different ids left the same silent failure in
place by another route.

That is not an exotic input. `vocab` and `rendered_added` are independent constructor
inputs with nothing reconciling them, and added-token ids deliberately live beyond the
base vocabulary range, so a marker declared in both is guaranteed to carry two ids
rather than one. Whichever tier wins, the other id is discarded; a generation emitting
the discarded one renders `</think>` into the output while `thinking_closed` stays
false, the budget keeps counting reasoning tokens, and it eventually forces a second
marker.

`token_id_for_content` is therefore replaced by `token_ids_for_content`, which reports
every distinct id sharing an exact rendered spelling, and
`resolve_reasoning_close_token` returns `InvalidInput` naming the colliding ids rather
than choosing among them. No single-id accessor survives on the type: separating the
collapse from the check is precisely how the two drift apart again.

The Metal prefix-cache fingerprint folds in every id rendering the marker instead of the
first. Two tokenizers that agree on the winning id but differ in their remaining aliases
are not the same tokenizer, and hashing one id would key them identically. The list is
sorted, because the added-token walk is over a `HashMap` and a cache key must not depend
on iteration order.

Both new tests were checked by reverting what they guard. Making the resolver pick the
first id instead of refusing fails the resolver test and leaves the tokenizer test
passing, which is what shows the mutation is localised to the guard rather than breaking
the lookup underneath it. Making the lookup stop reporting the collision fails both.

One measured detail worth stating, because it inverts the obvious expectation: driven
through the real `from_tokenizer_json_str` loader, the added token is filed *ahead* of
the base-vocabulary entry, so the alias wins precedence and the vocabulary id is the one
discarded. That is the opposite direction from the hand-built fixture in `bpe.rs`. Both
directions are the same defect, which is why the resolver refuses ambiguity outright
instead of trying to choose correctly. The resolver test asserts the id set rather than
its order for that reason.

## Why this PR also changes tokenizer loading

The guard above still had a hole, one layer below where it was written, and closing it
grew the diff past where this PR started. Stating that plainly rather than burying it.

`parse_rendered_added_tokens` returned `content -> id`. A tokenizer declaring `</think>`
twice as a `special: false` added token, at two ids, therefore kept only whichever entry
parsed last: the collision was destroyed at load time, before any lookup could see it. No
amount of care in the resolver could have caught that, because only one id ever reached
it. So the fail-closed guarantee the previous commit claims was not actually delivered
for every alias, and a claim a change does not deliver is worse than an absent one.

The map is now keyed by id. That direction is what every consumer already wanted: both
`BpeTokenizer` and `GemmaBpeTokenizer` inverted it immediately on receipt, and
`GemmaBpeTokenizer` additionally derived a rendered-id set that is now just the map's
keys. The content-keyed intermediate existed only to lose the duplicate on the way. An id
renders exactly one spelling, so keying by id collapses nothing that the old direction
preserved.

The regression asserts both ids survive loading *before* it asserts the rejection. That
ordering is the point: the surviving-id assertion is the one that fails if the parse
direction regresses, and reverting the parse direction was confirmed to fail it. Its
control is a single-alias tokenizer of the same shape, which separates "rejects
ambiguity" from "rejects added tokens" — the latter would satisfy a rejection assertion
just as well.

## bench-compare disposition

Re-derived at head `9078acc43517fabc5a0289d3798b47c4ee01917f`. A previous version of this section
carried two dispositions that contradicted each other, and the one that survived was pinned to a head
that has since moved. Both are replaced by the following, which splits the claim because the two
halves have opposite answers.

**The default `make bench-compare` invocation cannot execute this change. That part is a structural
waiver and it holds.** The defaults are hardcoded at `scripts/lib/bench-compare-impl.sh:447,449`:
`elementwise_cpu_bench` for `lattice-inference` and `simd` for `lattice-embed`, the latter not
overridable. Read directly: `elementwise_cpu_bench.rs` imports only `lattice_inference::forward::cpu`,
and embed's `simd.rs` imports only `lattice_embed::*` and never references `lattice_inference` at all.
Neither reaches any changed item.

**"No declared bench target can execute this change" would be false, and is not claimed.** The
workspace declares 31 `[[bench]]` entries across four crates. Searching all of them, with
`criterion_main` as a known-positive control on the same invocation:

- `cross_turn_prefix_cache_bench` (`required-features = ["metal-gpu", "f16"]`) structurally reaches
  the Metal call sites through `chat_completion_streaming` and
  `chat_completion_streaming_with_prefix_cache`. It sets `reasoning_budget: None` unconditionally, so
  it exercises the refactored call path but never the resolution or rejection logic this change adds.
  It also requires a real checkpoint on disk or it prints `SKIP` and measures nothing.
- `tokenizer_bench` and `inference_perf` reach the changed tokenizer constructor unconditionally, and
  `inference_perf` reaches `parse_rendered_added_tokens` only when a real `tokenizer.json` is present
  on the running machine.
- `Qwen35Model::generate` and its streaming variants, the CPU path through
  `resolve_reasoning_close_token`, `token_ids_for_content`, and the Gemma tokenizer loader are reached
  by **no** bench target at all: 0 of the bench sources call them, with the control matching in the
  same search. Nothing gates them out; no target exists.

That distinction is the point. "No declared target *can* execute this" is a structural property and
supports a waiver. "No target for this path *exists*" is a coverage hole and does not. They read
identically in prose, so this section names the population searched rather than asserting the
conclusion.

**Build-input parity.** The diff against `origin/main` contains zero entries under any `Cargo.toml`,
`Cargo.lock`, `build.rs`, or `benches/` path. Manifests, lockfile, feature sets, profiles, bench
target definitions, and every bench harness source are byte-identical between base and head. The only
difference is library source under `crates/inference/src/`.

**Residual, stated rather than implied.** An unmeasured change is not a safe change. Nothing here
claims this change is performance-neutral, only that the A/B the gate actually runs could not have
told anyone either way, and that the targets which could say something either do not explore the
state this change adds or do not exist.


ohdearquant:
Two items to settle at `3ebbc241c5cbf8df6c7b0269cd79debc2ddf5d0f` before this can land.

**1. The bench-compare disposition in the description no longer applies to this head, and it contradicts itself.**

The section says "Pending. This PR stays in draft until the A/B lands" and "this owes a real A/B table, not a reachability argument." The PR is not a draft. Below that, a later paragraph offers exactly the reachability argument the earlier paragraph disclaims, and pins itself with:

- head evaluated: `929ac7c620857045f883a3cc0d2de7c59be8c30e`
- "Void if either ref moves. A moved head means this has discharged nothing and must be re-derived."

The head is now `3ebbc241c5cbf8df6c7b0269cd79debc2ddf5d0f`, so by the paragraph's own terms it has discharged nothing. The diff touches `crates/inference/src/forward/metal_qwen35.rs`, which is inside the surface `ADR-080` requires evidence for. What is owed is one of: an A/B table measured on a quiet lane at this head with its run-conditions block, or a reachability derivation re-done at this head naming the bench targets searched. Either way the two paragraphs need reconciling so the section states one disposition rather than two.

Worth noting the general shape, since it is not specific to this change: a PR description is authored once and nothing re-reads it, so a head-pinned claim inside one keeps reading as current long after its pin stopped matching.

**2. A positive `reasoning_budget` with `enable_thinking: false` is validated and can be rejected, even though no marker will ever be forced.**

`resolve_reasoning_close_token` (`crates/inference/src/model/qwen35/generation.rs:2352-2358`) gates only on `reasoning_budget`; `enable_thinking` is not a parameter. `force_close_think` (`crates/inference/src/model/qwen35_config.rs:2452-2464`) returns `None` whenever `enable_thinking` is false.

So for `{ reasoning_budget: Some(1), enable_thinking: false }` against a tokenizer with no `</think>` token, the resolver returns `InvalidInput` and the request fails, although the budget could not have been enforced or violated either way. The mirror case, the same request against a tokenizer that does have the marker, passes validation and then silently never forces. Both come from the same missing coupling.

Either reject the incompatible combination explicitly at the API boundary, or define a budget as intentionally inert when thinking is disabled and skip resolution in that case. Whichever is chosen, it wants a test naming the combination, since today the two paths disagree about what the field means.

Unrelated to this branch, but found while reading it and worth its own issue: `SafetensorsFile::open` (`crates/inference/src/weights/f32_weights.rs:217-233`) opens and maps without routing through `weights::mmap_trust::open_trusted_mmap_file`, which at this head has exactly one caller (`crates/inference/src/weights/q4_weights.rs:969`). #1369 appears to be addressing that surface directly.

PR #1216, opened 2026-07-29
fix(docs): isolate lint self-test index

### LINKED ISSUE #1213: lint-docs self-test corrupts the repository index when run from the pre-commit hook
`scripts/lint-docs.sh` runs a discovery self-test that builds a sandbox repository and stages files in it with `git -C "$sandbox"`. When the script runs from the `pre-commit` hook, those staging operations write to **this repository's index** instead of the sandbox's, replacing it.

The result is a commit that deletes most of the tree and adds the self-test's sandbox fixtures, while every check reports success.

## Reproduction

In a clean worktree at current `main`:

```
$ git ls-files | wc -l
896
$ echo x >> README.md && git add README.md
$ git commit -m "probe"
...
=== Doc Lint Passed ===
Pre-commit checks passed.
$ git ls-files | wc -l
211
$ git ls-tree -r --name-only HEAD -- crates/embed | wc -l
0
$ git ls-tree -r --name-only HEAD -- docs/one.md 'docs/space name.md' | wc -l
2
```

896 index entries become 211. `crates/embed` and `crates/fann` are absent from the commit tree. Two of the self-test's sandbox fixtures (`docs/one.md`, `docs/space name.md`) are committed as real files. The hook exits 0 and prints `Pre-commit checks passed`.

Running `./scripts/lint-docs.sh` directly does **not** reproduce it: 896 entries before and after, no fixtures leaked. The hook context is required.

## Mechanism

Git exports `GIT_INDEX_FILE` to `pre-commit` as a **relative** path:

```
GIT_DIR=[]
GIT_INDEX_FILE=[.git/index]
GIT_WORK_TREE=[]
```

`run_discovery_selftest` invokes `git -C "$repo" init` and `git -C "$repo" add README.md docs crates` while the shell's working directory is still the real worktree root. `-C` changes the repository those commands operate on, but the inherited relative `GIT_INDEX_FILE` resolves against the caller's working directory, so the sandbox's staging writes into the real repository's index file.

A separate check confirms the override direction generally: with `GIT_DIR` unset, `git -C <sandbox> rev-parse --absolute-git-dir` returns the sandbox; with an index/dir environment variable inherited, the ambient value wins.

## Scope

`.githooks/pre-commit` is tracked and `core.hooksPath` points at it, so this applies to every clone and every linked worktree with hooks enabled. The self-test has been present since #911 (2026-07-12). Commits made with `--no-verify` are unaffected, which is likely why it has not been continuously visible.

## Suggested fix

Neutralize the inherited index environment for the sandbox commands, for example by unsetting `GIT_INDEX_FILE` (and `GIT_DIR`/`GIT_WORK_TREE`) for the duration of the self-test, or by exporting an absolute `GIT_INDEX_FILE` pointing inside the sandbox. A regression check should assert that the tracked-file count is unchanged across a hook-context run, since the failure is silent and reports success.

ohdearquant:

## Summary

- clear every Git repository-local environment variable before the Markdown discovery self-test operates on its sandbox repository
- run the discovery probe in a subshell so its environment and cleanup trap cannot leak back to the caller
- exercise the real pre-commit hook context and prove the hook completes even when an adversarial global `core.hooksPath` points elsewhere
- require both an explicit hook marker and unchanged staged/committed file sets

The regression probe supplies a foreign `GIT_INDEX_FILE` to an actual temporary-repository commit. The nested discovery test must operate only on its own repository; any inherited-index write makes the hook or the post-commit identity checks fail.

Fixes #1213

## Validation

- `sh -n scripts/lint-docs.sh` — passed
- `./scripts/lint-docs.sh --selftest` — recursive discovery, forced-hook execution, and hook-index isolation passed
- `./scripts/lint-docs.sh` — all 179 tracked Markdown files, discovery self-test, and capability-matrix checks passed
- `git diff --check` — passed

## Mutation proof

Removing the repository-local environment clearing caused the synthetic pre-commit to rewrite the outer index. Removing either the forced hook path or the hook-completion marker made the self-test fail closed instead of silently skipping the regression probe. Restoring all guards returned the exact staged-entry and committed-tree checks to green.

## Sibling sweep

The discovery self-test is the only path in `lint-docs.sh` that creates and stages a foreign Git repository. It clears the complete set reported by `git rev-parse --local-env-vars`, covering index, worktree, object-store, shallow, replacement-ref, and repository overrides rather than singling out one current hook variable.

## bench-compare disposition

Not run. This changes only the documentation-lint shell harness and its self-test; it does not alter Rust code, benchmark binaries, or timed execution paths.

### DIFF
--- a/scripts/lint-docs.sh
+++ b/scripts/lint-docs.sh
@@ -108,6 +108,8 @@ run_index_isolation_selftest() (
     after_index="$sandbox/after-index"
     tracked_files="$sandbox/tracked-files"
     committed_files="$sandbox/committed-files"
+    hook_marker="$sandbox/hook-ran"
+    global_config="$sandbox/global-config"
     mkdir -p "$repo/sentinel"
     printf '%s\n' '# outer repository' >"$repo/README.md"
     printf '%s\n' 'keep one' >"$repo/sentinel/one.txt"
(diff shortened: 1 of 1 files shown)


ohdearquant:

Superseded by merged PR #1221, which fixes #1213 with a hook-environment regression test and also hardens the bench worktree helper. Closing this duplicate draft so review stays on the implementation now in `main`.

PR #1219, opened 2026-07-29
perf(inference): expose Metal hidden-state forwards

### LINKED ISSUE #1186: perf(inference): expose Metal hidden-state forward results for MTP profiling
> _Issue authored by Codex (OpenAI agent) on behalf of @ohdearquant._

## Summary

`crates/inference/examples/profile_metal.rs` contains a complete but disabled measurement block for hidden-state readback and real-model MTP overhead. It cannot compile today because `MetalQwen35State` exposes neither `forward_step_with_hidden` nor `forward_prefill_with_hidden`.

This issue tracks the missing Metal API and measurement work independently from the marker-provenance cleanup in #820.

## Required behavior

- add fallible `MetalQwen35State` entry points that preserve the existing state/cache semantics while returning the final hidden state alongside the normal forward result for:
  - one decode step
  - prompt prefill
- prove the new methods' logits and state advancement match their existing non-hidden-returning siblings
- make the hidden-state readback explicit and measurable; do not introduce an unconditional readback on existing generation paths
- enable the disabled `profile_metal.rs` measurement block so it reports `hidden_readback_us` and the per-draft-length real-model MTP overhead it currently describes

## Verification

- GPU-serialized parity tests compare returned logits and state progression with the existing forward methods on exact inputs
- path-proof evidence demonstrates the hidden-state readback actually executed
- the enabled profiling example compiles and runs on Apple Silicon with loaded MTP weights
- `make bench-compare` evidence accompanies any performance-sensitive implementation

## Non-goals

- changing MTP verification policy or token-selection behavior
- adding hidden-state readback to ordinary generation paths
- implementing the work as part of #820; that issue only replaces the orphaned internal marker with this traceable tracker

ohdearquant:
> _PR description authored by an AI agent (OpenAI) on behalf of @ohdearquant._

## Summary

- expose fallible Metal step and prefill entry points with explicit pre-final-hidden readback and path counters
- reject stale live-cursor and non-fresh raw-prefill scheduling before cache clearing, GDN mutation, or Metal dispatch
- cover public scheduling routes, verifier methods, and `bench-internals` helpers with fail-closed source and macOS checks
- retain profiler hidden-readback and draft-span reporting for safetensors and native Q4 model directories

Fixes #1186

## Stateful scheduling boundaries

KV placement and attention length derive from the live cache cursor, while RoPE can use a supplied position. Public step and verifier routes now reject a supplied position that differs from the live cursor before mutation or dispatch. Raw prefill routes require empty KV state and initial GDN state before fixed-zero scheduling. The public `bench-internals` chunk helpers use the same preflight and return `Result`.

The source-level closure check reads the real Metal implementation before cfg stripping, discovers public functions that accept scheduling positions or reach scheduling sinks, and includes the verifier trait methods. A newly added cfg-gated public scheduling route therefore fails until its preflight disposition is explicit.

## Raw-prefill behavior change

Calling raw prefill against live KV or GDN state previously continued from an inconsistent fixed-zero schedule. It is now rejected before state mutation. New-prompt callers must call `reset_state()` first.

`try_forward_prefill` is the typed, non-panicking route and returns `InferenceError::InvalidInput` for stale state. The source-compatible `forward_prefill` wrapper remains for existing callers and documents that it panics when the typed route rejects the call. `forward_prefill_all_logits` remains fallible. The public `bench-internals` prefill helpers now return `Result`, so feature consumers must handle or propagate their errors.

## Validation

**Head has since moved.** The validation below was recorded at `4f64c1a7a79b0d888d6c61cecb59d57473069688`; the current head is `a888bf0ddef1d5624f83bcdf0943a74c7eb154f6`. Read it as a record bound to that ref, not as a statement about the head under review.

Validation is for head `4f64c1a7a79b0d888d6c61cecb59d57473069688` after merging live main `93e7bd37b8b6e5eda1067575dd54fdecceaae25e`.

- commit hooks passed workspace formatting, clippy, documentation lint, capability-matrix checks, and absolute-path checks
- `cargo clippy --release -p lattice-inference --all-targets --features f16,metal-gpu,bench-internals -- -D warnings`: passed
- `cargo test -p lattice-inference --test metal_measurement_lock_contract`: 53 passed, 0 failed
- `cargo test -p lattice-inference --lib --features f16,metal-gpu,bench-internals forward::metal_qwen35::inner::tests::validate_live_cursor_rejects_mismatched_position -- --exact`: 1 passed, 0 failed, 2,749 filtered out
- `actionlint .github/workflows/ci.yml`: passed

The collected `lattice-inference --lib` counts were asserted nonzero with `--list`; ignored counts were independently collected with `--ignored --list`:

- default features: initial head 2,441 collected / 20 ignored / 2,421 non-ignored; current head 2,441 / 20 / 2,421
- `f16,metal-gpu`: initial head 2,737 / 23 / 2,714; current head 2,742 / 23 / 2,719
- `f16,metal-gpu,bench-internals`: initial head 2,745 / 23 / 2,722; current head 2,750 / 23 / 2,727

On the merged tree, `make ci` passed formatting, workspace clippy, documentation checks, and the workspace suites through the Gemma golden tests, then exposed three construction sites missing the shared measurement lock. After those sites were guarded and an obsolete exemption removed, the complete 53-test measurement-lock contract passed.

No Metal-device test or shared GPU lock was used locally. The directly relevant device tests were compiled or listed but not executed: `raw_step_with_mtp_rejects_stale_live_cursor_without_mutation`, `raw_prefills_with_mtp_reject_live_session_without_mutation`, `bench_prefill_helpers_reject_stale_cursor_without_mutation`, and `mtp_verifier_boundaries_reject_mismatched_live_cursor`. The macOS feature-matrix arms run the applicable checks with `LATTICE_METAL_TEST_ENFORCE=1`, so absence of a device is a failure rather than a passing skip.

## bench-compare disposition

**Reachability disposition: reachable, but not measurable by any existing bench target.**

Pair at the current head: base `0d819dda7df101099a7a40921c9590e61dd542df` (merge-base with `main`), head `a888bf0ddef1d5624f83bcdf0943a74c7eb154f6`.

**What the change does to the measured surface.** The two state methods `metal_decode_bench` calls were extracted whole from both refs and compared, rather than inferred from hunk headers:

- inner `forward_step`, the call the bench times on **every** iteration: **byte-identical**, 15 lines on both sides. This PR does not touch it.
- inner `forward_prefill`: **changed**. The base body was `cross_turn_prefix_cache.clear(); forward_prefill_impl(token_ids, false)`. The head delegates to the new `try_forward_prefill`, which runs `check_raw_prefill_fresh_session`, `check_forward_token_ids` and `check_forward_range_capacity` before the same impl call. One of those checks is O(prompt length).

**Why the reachable path still cannot measure it.** Every `forward_prefill` call site in `metal_decode_bench` (lines 147, 157, 168, 177, 181, 216, 226) is either setup outside `bench_function` or inside the `if pos > 500` reset branch of an `iter` closure. The changed code therefore executes roughly once per five hundred timed iterations, against a per-iteration hot call that is identical on both sides. An A/B on that target would dilute the change by about 500x, and a null result would be a property of the harness rather than of this diff.

**And no other target reaches it.** Of the seven `crates/inference` bench files that reference `metal_qwen35`, exactly one mentions `forward_prefill` at all; the other six contain zero references. **No bench target in this repository times a prefill call.**

**Residual risk.** The added checks sit on the prefill entry point, so a regression would surface as time-to-first-token on long prompts. That is a user-visible number no current bench covers, which is a coverage hole independent of this PR.

**Proposed fork.** Add a bench group that times `forward_prefill` directly, under `--features metal-gpu` and the shared Metal measurement lock, with prompt lengths spanning the chunked and one-token paths. That group is worth having regardless of how this PR is dispositioned, and once it exists this change has a real instrument instead of a waiver. Filed as a follow-up rather than blocking here.

An earlier version of this section recorded the A/B as OWED against a base and head that have both since moved, and quoted harness self-comparison percentages as measurement context. Both are removed: the SHAs are stale, and a figure that cannot be attributed to the diff is not context.

- Pull request head evaluated: `a888bf0ddef1d5624f83bcdf0943a74c7eb154f6`

re-derived.


ohdearquant:
Superseded by #1339.

This branch had diverged from `main` far enough that no fast-forward was possible. The work was rebased onto current `main` and pushed as `fix/metal-hidden-readback`; #1339 carries all thirteen commits plus three fixes the rebase surfaced (a stale static-scan assertion, two Metal construction sites missing the shared measurement lock, and one stale line-pinned exemption).

None of this branch's contribution had been absorbed by intervening work — `try_forward_prefill`, `forward_prefill_with_hidden` and `forward_step_with_hidden` still do not exist on `main`. Of the failing checks here, two were real code faults, now fixed; the rest were infrastructure failures and downstream gates inheriting an abandoned state from starved dependency jobs.

Closing in favour of #1339.

Issue #322, opened 2026-06-25
bug(inference): grammar-decoding runtime (pda/engine/vocab_partition) — review+TBV findings (F1 unknown-state soundness fixed, F2-F6 tracked)


ohdearquant:
> _Issue filed by Claude (Anthropic agent) on behalf of @ohdearquant._

## Grammar-constrained decoding runtime — discovery + verification pass

Adversarial sweep over the grammar-decoding **runtime** (`pda.rs`, `engine.rs`, `vocab_partition.rs`) — distinct from the JSON-schema compiler defects tracked in #310. Each finding below was verified against source. Severities are post-verification (some downgraded from the raw initial rating).

### F1 — [Medium] `find_state_id` unknown-state fallback was unsound → **FIXED** (PR linked below)

`crates/inference/src/grammar/engine.rs` `find_state_id` returned `0` (initial state) on no-match instead of a fail path. When a grammar exceeds `MAX_GRAMMAR_STATES` (256), BFS enumeration truncates and a deep runtime state has no precomputed mask, so `mask_logits` applied the **initial-state** mask at a deep position. Single-byte tokens valid only at state 0 (e.g. an opening `"`) stayed allowed and are not in the context-dependent recheck set → grammar could emit a forbidden token. Production-reachable (`mask_logits` is wired into `generate.rs` + `metal_qwen35.rs`) for any schema/GBNF with >256 reachable states. **Fixed**: `find_state_id → Option`, `None` computes the exact mask by simulation. Red/green test included.

### F2 — [Medium, perf] Redundant per-byte stack clone in `advance_byte`

`pda.rs:175` clones `state.stack` into `snapshot`, but `try_advance_byte` only writes `state.stack` on success and never touches it on failure — so the `state.stack = snapshot` restore at line 184 is a provable no-op. The outer snapshot is dead; removing it halves per-byte heap cloning on the grammar decode path with zero behavior change. (Perf-class: warrants a bench-compare on the grammar advance path.)

### F3 — [Low/architectural] Epsilon-first alternative false-reject

[...]

PR #1235, opened 2026-07-29
refactor(inference): make PDA fallback iterative

ohdearquant:

## Summary

- Replace recursive parent-frame fallback in `try_next_alt` with an iterative walk over the existing PDA stack.
- Keep alternative priority, consumed-frame refusal, stack truncation, clone-on-write rollback, same-byte retry, and `MAX_PDA_DEPTH` unchanged.
- Add a direct parent-fallback behavior test and a production-valid 4,096-level regression on a 64 KiB native thread stack.

Addresses F4 of #322. #322 remains open for F3, F5, and F6; this PR intentionally does not change epsilon-first/shared-prefix behavior, malformed direct-construction handling, or state-local token rechecks.

## Validation

- `cargo test -p lattice-inference --lib` — 2415 passed, 20 ignored, 0 failed.
- PDA unit tests — 19 passed.
- `grammar_tbv_probe` + `grammar_json_schema_regression` — 26 passed, 3 intentional ignores.
- `cargo clippy --workspace --all-targets -- -D warnings` — passed.
- `cargo fmt --all -- --check`, `git diff --check`, and repository pre-commit hook — passed.
- Mutation proof: restoring recursive parent traversal made `deeply_nested_parent_fallback_accepts_on_bounded_stack` terminate the isolated test process with stack overflow/SIGABRT; restoring the iterative loop made the same test pass.
- Three independent read-only reviews: APPROVE, no blockers.
- Hosted `e2e-parity.yml` remains the required model-backed macOS validation.

## bench-compare disposition

A/B run, quick resolution, on an otherwise idle dedicated macOS arm64 host (Apple M4, 10 cores).

```
base: origin/main (0f86f5125a)   head: 09657e0830
resolution: --quick
inference features: <none>   filters: inference='<all>' embed='<all>'
harness: pinned to origin/main (the bench script and gate run from the invoking checkout)
locks: bench-window acquired; Metal GPU acquired (uncontended)
ambient idle: before base 97.2%  between phases 94.4%  after head 95.8%   (floor 70.0%)   spread 2.8pp
```

The three ambient samples are read as a series and not only against the floor, because a per-sample
threshold cannot see a load that sat on one phase and not the other.

### Gate verdict

The repository's target-qualified gate report for the gating target
`lattice-inference:elementwise_cpu_bench` is **5 WARN** (a regression in the 3.0-7.0% band by the
two-sided-95%-CI lower bound). No row reached the >7.0% fail threshold.

| Bench | Δ point | 95% CI | verdict |
| --- | ---: | --- | --- |
| `residual_add_layer_norm/unfused/hidden768_seq128` | +4.67% | [+3.94%, +5.42%] | WARN |
| `add_bias_gelu/896` | +4.26% | [+3.93%, +4.59%] | WARN |
| `layer_norm/896` | +4.04% | [+3.84%, +4.24%] | WARN |
| `add_bias_gelu/4096` | +3.70% | [+3.20%, +4.21%] | WARN |
| `rms_norm/4096` | +3.40% | [+3.36%, +3.43%] | WARN |

An earlier revision of this description reported this target as "n=34, 0 significant, clean". That
was read off Criterion's own per-group significance markers rather than the repository's
target-qualified gate report. The two apply different rules: Criterion found no significant change at
its default threshold, while the repository gate classifies by the 95%-CI lower bound against a 3%
band and returns the five rows above. The gate report is the authoritative reading for this section,
so the earlier figures are withdrawn.

This diff changes only `crates/inference/src/grammar/pda.rs`, which contains none of the five flagged
kernels. The controls below establish that all five rows are measurement-order artifacts: a run with
the arms swapped produces positive WARNs too, in the direction that would mean main is slower than
this branch. None of the five is treated as a regression, and nothing reached the fail threshold.

### The controls, and what the flagged rows actually mean

Two controls were run on the same host at the same resolution and filters.

**Base versus base**, on byte-identical source (`base_sha == head_sha == 0f86f5125a`, ambient idle
96.3 / 96.5 / 96.9, spread 0.6pp): all 17 gating groups inside the 3% band, maximum point delta
+2.38% on `layer_norm/896`. That bounds a single run's spread but cannot separate a code effect from a
measurement-order effect, because both of its arms are identical by construction and the head arm
still runs second.

**Arms swapped**, the same two trees compared in the opposite order (base `a101ceba10`, head
`0f86f5125a`, the pair verified equal to the forward run's recorded `base_sha` before launching;
ambient idle 96.4 / 96.8 / 96.6, spread 0.4pp): the gate reports **4 WARN, every one positive** —
`softmax_attention/512` +5.15%, `gelu/896` +4.67%, `layer_norm/4096` +3.67%,
`residual_add_layer_norm/unfused/hidden384_seq128` +3.34%.

That is the decisive result. In the forward order a WARN says the PR tree is slower than main. In the
swapped order the head arm *is* main, so a WARN says main is slower than the PR tree. Both orders
produced positive WARNs, and those two readings cannot both be statements about the code. Counted
mechanically across every run on this host in this session, **13 of 13 WARN rows are positive and none
is negative**. The arm that is measured second is the slower one, whichever tree occupies it.

So quick mode's 3.0-7.0% WARN band on this host sits inside the host's own phase-order bias, and a
WARN at this resolution is not evidence about the diff. The >7.0% fail threshold was never approached;
the largest excursion measured in either direction was 5.15%.

What this does not establish: the size of the bias per group. Which groups cross the band varies from
run to run, so the systematic part is the sign and not the magnitude, and a real code effect inside the
3-5% range cannot be resolved at this resolution at all. Separating one would need `--full`, or a
harness that alternates or interleaves the arm order rather than always measuring base first.

The two improvements measured in this session point the other way, against the bias, so they are not
artifacts of it.

### `lattice-embed:simd`

`lattice-embed:simd` is the sole entry in `scripts/lib/bench-quick-informational-targets.txt`, so
quick mode measures it and reports it but excludes it from the regression verdict, which is why its
gate report reads "All 0 gated benches". Today's runs show why the manifest lists it: across PRs that
change none of its source, its `simd_normalize` group swung between -14% and +33% at this resolution.

It also cannot serve as a source-identical control for a `crates/inference` change.
`crates/embed/Cargo.toml` compiles `lattice-inference` with `["std", "f16"]` under its default
`native` feature, and the dependency runs embed to inference and not the reverse, so an inference diff
changes embed's bench binary too.

### Base equivalence across the interim merge

Main advanced from `0f86f5125a` to `ab48dee49b` while these runs were in flight. `git diff --stat`
over that range is one file, `docs/benchmarks.md | 20 ++++`, with zero `.rs`, `.toml`, `.yml` or
`.yaml` files in it, so the bench binaries at the two shas are built from identical effective source
and the A/B above remains valid against current main. Verified as a diff rather than asserted.


ohdearquant:
## Correction to the bench-compare disposition in this merged PR's description

The description of this PR, in the section "The controls, and what the flagged rows actually mean",
attributed the flagged warn-band rows to measurement arm order. It stated that the arm measured
second is the slower one whichever tree occupies it, and supported that with a count of warn rows
by sign. **Both statements are withdrawn.** This comment records the correction rather than editing
the description, so that anyone who already read the original sees it.

**The sign count carried no information.** The gate defines WARN as a confidence-interval lower
bound above the warn threshold, so a WARN row is positive by construction. A tally of warn rows by
sign is true of any dataset and said nothing about this one.

**Arm order is refuted by measurement.** A base-versus-base control on byte-identical source
(`base_sha == head_sha == 0f86f5125a`) returns 134 positive and 127 negative across 264 groups at a
median of +0.01%. Arm order predicts a one-sided result there and does not produce one. A forward
run in the same session landed negative at its gated median (-0.94%, 6 positive / 11 negative) while
the branch under test occupied the head arm, which arm order predicts cannot happen.

**What the data supports instead.** Each quick-resolution run carries a coherent offset shared
across its groups, and the sign of that offset differs between runs: gated medians measured on this
host in this session range from -0.94% to +3.00%. Criterion computes its intervals within a single
run, so they are narrower than the run-to-run spread and can be confidently wrong about it. On the
byte-identical control the per-group spread is 1.90pp standard deviation over a -6.07% to +9.55%
range.

**What this does not change.** The operative conclusion of the original disposition stands: a
warn-band row at quick resolution is not evidence about the diff, no gated group approached the >7%
FAIL threshold, and the merged change is unaffected. What changes is the reason. The cause is
run-level variance rather than arm position, and the corrected reading is being applied to the
remaining open PRs in this family. #1269 relabels quick-resolution warn-band rows as UNATTRIBUTABLE
on this basis.

PR #1244, opened 2026-07-29
refactor(serve): share vision prompt primitives

### LINKED ISSUE #1027: refactor(serve): vision prompt builder duplicates ChatML turn literals instead of sharing format_chat_template primitives
Found in PR #1021 review. `build_vision_prompt_ids` re-implements the ChatML renderer with its own `<|im_start|>`/role/turn literals instead of sharing the primitives behind `format_chat_template`. Any future template change must be duplicated by hand to keep vision requests aligned, and no test pins the relationship.

Expected: extract shared turn/prefix rendering helpers and add a test asserting the text portions around the vision-token splice match the plain-template rendering of the same messages.

Sites: `crates/inference/src/serve/metal_worker.rs` (`build_vision_prompt_ids`), `crates/inference/src/forward/metal_qwen35.rs` (template renderer).

ohdearquant:

## Summary

- centralize the ChatML assistant-generation turn opener and use it from both plain and vision prompt rendering
- split vision prompt text into explicit before/after fragments around the image-token insertion point
- prove that removing the image splice reproduces the plain chat template byte-for-byte
- independently pin the original content-part splice offset so an equivalent-but-misplaced split fails tests

Fixes #1027.

## Stack

Depends on #1231 and intentionally targets `codex/serve-image-content-parts`, whose shared image content-part representation this refactor consumes. Retarget to `main` after #1231 merges.

## Validation

- independent review: one test-coverage finding fixed, then APPROVE with no remaining findings
- Metal-worker suite: 30 passed
- focused vision-prompt regressions: 3 passed
- full default inference library suite: 2,438 passed, 20 ignored
- strict `metal-gpu,f16,test-utils` library/test Clippy with `-D warnings`: passed
- pre-commit hook: passed (workspace check plus 179-file documentation lint)
- `cargo fmt --all -- --check` / `git diff --check`: passed
- structural search: no remaining raw `<|im_start|>assistant\n` duplicate literal

Mutation evidence:

- rendering the final generation turn as `user` fails the byte-equivalence regression
- moving the image splice one byte earlier fails the independent before/after boundary assertion

One broad Metal-feature test requires `LATTICE_KV_F16=1`; it fails identically on unchanged #1231 without that environment and passes under the repository GPU lock with the required path-proof variables. This PR does not claim an unqualified full Metal-feature aggregate.

## bench-compare disposition

Required because the shared prompt-rendering code is compiled into the default inference crate.

Attempted the exact-SHA command:

```text
CARGO_TARGET_DIR=$SHARED_TARGET_DIR make bench-compare BASE=228417020736ae5b3a46aff6954b4ba7fb064f04 HEAD=1e8ba317241d7b7dc674516a67d0ef10c5dcfe03
```

The harness refused before measuring base because machine idle was 16.9% versus the required 70% floor; concurrent repository/storage services made the host uncertifiable. I did not lower the floor. This draft makes no performance claim; a certified A/B remains pending on a quiet host.

- Pull request head evaluated: `1e8ba317241d7b7dc674516a67d0ef10c5dcfe03`

re-derived.


ohdearquant:
Superseded: the shared chat-generation opener and the vision prompt splice now live on main. `push_chat_generation_open` is exported from the forward module and used by `build_vision_prompt_text` (crates/inference/src/serve/metal_worker.rs:796), and the before/after splice equivalence test `vision_prompt_text_matches_plain_template_around_image_splice` is present in the same file. A diff of this branch against current main carries no remaining unique change.

Issue #1027 was closed by the merge that introduced those primitives.

Issue #922, opened 2026-07-13
bug(inference): Metal Qwen3.5 context admission checks prompt_len only, unlike CPU's prompt+decode-budget bound


ohdearquant:
> _Issue authored by Claude (Anthropic agent) on behalf of @ohdearquant._

## What

The Metal Qwen3.5 generation entry points admit a request when `prompt_len <= max_context()`, ignoring the decode budget. The CPU path enforces the stricter total bound `prompt_len + effective_new_tokens <= max_context`.

- `crates/inference/src/forward/metal_qwen35.rs:8818` — `generate`: `if prompt_len > self.max_context()`
- `crates/inference/src/forward/metal_qwen35.rs:13362` — same shape in the multimodal entry point
- `crates/inference/src/forward/metal_qwen35.rs:16212` — same shape in the prefix-cache streaming entry point
- `crates/inference/src/model/qwen35/generation.rs:96-99` — CPU preflight includes the decode budget

## Why it matters

A prompt that fits the window but whose prompt+decode budget exceeds it passes Metal admission and only fails later (cache exhaustion mid-decode or silent truncation of the requested budget), while the identical request on CPU is rejected up front with a typed error. HTTP callers of `lattice_serve` are shielded by that binary's own preflight, but direct library callers, `chat_metal`, and any future binary hit the divergence.

## Suggested fix

Port the CPU total-context bound into one shared admission helper and call it from all three Metal entry points (they already share the `prompt_len > 0` and window checks). A test asserting Metal and CPU reject the same boundary request keeps them aligned.

Issue #743, opened 2026-07-09
bug(inference): Metal Qwen3.5 generation admits over-budget requests CPU rejects


ohdearquant:
**Defect**: Qwen3.5 CPU generation rejects a request whose `prompt_len + effective_new_tokens` exceeds `max_context`, while every live Metal generation entry point (plain, streaming, and prefix-cache streaming) guards only `prompt_len > max_context`, ignoring the requested completion length. For a request at exactly the context capacity, CPU returns an error before prefill while Metal admits the full prefill and returns a sampled token — a different success/failure outcome for the same request depending on backend.

Note: the Metal check is defended in a nearby code comment as intentional graceful degradation for prompt-only admission, but that does not resolve the cross-backend contract inconsistency — the same request produces different accept/reject outcomes depending on which backend serves it.

**Evidence**:
- `crates/inference/src/model/qwen35/generation.rs:98-99` (mirrored at `:401-402`) — `let effective_new = decode_cap(...); if prompt_len.saturating_add(effective_new) > max_context { ... }` (prompt + completion check).
- `crates/inference/src/forward/metal_qwen35.rs:8461` (plain `generate`), `:12799` (`generate_streaming`), `:15274` (prefix-cache streaming) — all guard only `if prompt_len > self.max_context()`, no completion-length term.

**Blast radius**: Ordinary Qwen3.5 direct and streaming generation — CPU and server paths reject over-budget requests while the corresponding Metal paths admit and complete them.

**Suggested fix**: Extract the CPU total-token preflight (including `decode_cap`) into a shared helper and invoke it before every Metal generation state mutation, so both backends apply the same admission contract.

Provenance: Audited at 13c8de8a3; re-verified at 0699e60cc.

ohdearquant:
Done on main: Metal now mirrors CPU's prompt+decode-budget admission bound — 604f911964 (PR #940, from #922). Over-budget requests are rejected on the Metal path to match CPU.

Issue #827, opened 2026-07-10
chore(inference): close out generation entry-point preparation series — rg gate + epic #614 update


ohdearquant:
> _Issue authored by Claude (Anthropic agent) on behalf of @ohdearquant._

## What

Close out this series of generation-preparation consolidation issues:

1. Run a targeted `rg` sweep confirming the RNG-normalization, tokenization,
   zero-budget-output, and context-preflight patterns extracted by the earlier
   issues in this series remain only in the shared preparation module, its
   documented exclusions, and tests — and that the three Metal sampling-route
   blocks remain only in `configure_sampling_route`.
2. Update epic #614: mark stage 1 (shared generation preflight/request setup)
   complete for the seven entry points this series covers, and explicitly list what
   is not covered.

## Why

This series migrates seven named entry points
(`generate_f16`/`generate_q8`/`generate_q8_neon`, `MetalQwen35State::generate`,
`generate_streaming`/`generate_streaming_with_cancel`,
`generate_streaming_with_prefix_cache_and_cancel`) to shared preparation. It does
not migrate, and must not be reported as covering:

- Canonical `Qwen35Model::generate` / `generate_streaming_with_cancel`
  (`crates/inference/src/model/qwen35/generation.rs`) — the RNG-seed helper
  (`initial_rng_state`, ~line 1580) may be donated to the shared module, but these
  entry points' own request setup is untouched by this series.
- `generate_with_batch_prefill` (`forward/batch_prefill.rs`) — owned by #807/#810
  (deprecate-then-remove); this series must not have spent any migration effort on
  it.
- Multimodal generation entry points, if any exist outside the Qwen3.5 text path.
- Epic #614 stages 2 (canonical sampling/token-decision entry point) and 3
  (backend-agnostic decode-loop skeleton) — both remain separate architecture work,
  gated on their own ADR sign-off.

## Checklist

- [ ] `rg` for the pre-extraction RNG-seed match arm shape
      (`SystemTime::now().duration_since(SystemTime::UNIX_EPOCH)` paired with the
      `if t == 0 { 1 } else { t }` fallback) across `crates/inference/src/forward/`
      and `crates/inference/src/model/qwen35/generation.rs`; every remaining hit
      must be either the shared preparation module, a documented canonical-path
      exclusion, or a test.
- [ ] `rg` for the three Metal route-selection call shapes
      (`choose_gpu_block_topk_route` followed by `choose_gpu_topk_route` in the same
[...]

PR #1246, opened 2026-07-29
refactor(inference): share prefix-cache preparation

### LINKED ISSUE #826: refactor(inference): move prefix-cache generation preparation to the non-destructive outer boundary

## What

Move the seed-normalization, tokenization, and the three empty/zero/context guards
currently inside `generate_streaming_with_prefix_cache_and_cancel_inner`
(`crates/inference/src/forward/metal_qwen35.rs`, ~lines 15403-15461) to execute in
the public wrapper `generate_streaming_with_prefix_cache_and_cancel` (~line 15330),
before the `_inner` call, using the shared `prepare_generation` helper. Convert a
`PreparedEntry::Early` outcome to the wrapper's `CachedGenerateOutput` shape
(`FullRefill` mode, zero reuse/prefetch stats) once, in the wrapper; pass
`PreparedEntry::Ready` into `_inner`, which no longer seeds, tokenizes, or repeats
the three guards itself.

## Why

PR #787 already moved the `logprobs`/`enable_mtp` capability preflight
(`check_logprobs_not_set`, `check_mtp_not_requested`) to the public wrapper,
specifically so a rejected request never reaches `_inner`'s `Err` arm, which
unconditionally calls `self.reset_state()` and `self.cross_turn_prefix_cache.remove(slot_id)`
on any error (~lines 15343-15346) — evicting a valid warmed cache entry the request
never actually mutated. The wrapper's own comment (~lines 15340-15351) states this
rationale explicitly.

The remaining seed/tokenize/empty/zero/context guards are still inside `_inner`
(~lines 15403-15461), *after* that same destructive-recovery boundary conceptually
governs the call — an empty-prompt or zero-budget or over-context request is
currently a normal early `Ok` return from `_inner` and so never triggers the error
arm today, but it does perform seed generation and tokenization before its own
`Ok` return, work the shared preparation step should absorb along with the two
already-hoisted checks. This issue finishes what #787 started for this call site:
one preparation step, at one boundary, before any state mutation.

## What does NOT move

- `check_logprobs_not_set` / `check_mtp_not_requested` already live correctly in the
  public wrapper (#787) — do not duplicate them into the new preparation call; fold
  them into the prefix-cache entry's capability contract instead so there is exactly
  one call performing all of this path's fail-closed checks.
- Cache-restore (`restore_cross_turn_prefix`), the reuse plan (`plan_cross_turn_reuse`),
  the cancellation checks around prefill, and the suffix-prefill call all stay in
  `_inner`, unchanged in order and behavior.

## Checklist

- [ ] Add a prefix-cache-specific capability contract to the shared preparation
      helper (or extend an existing one) that includes `logprobs` and `enable_mtp`
      rejection in addition to the standard empty/zero/context guards, matching the
      signed-off behavior matrix for this path exactly.
- [ ] Move preparation into the public wrapper, before the `_inner` call.
- [ ] Convert `PreparedEntry::Early` to `CachedGenerateOutput` once, in the wrapper,
      with `CrossTurnCacheStats { mode: PrefixReuseMode::FullRefill, reused_tokens: 0,
      prefetched_tokens: 0, .. }`.
- [ ] `_inner` receives `PreparedGeneration` directly; remove its own seed/tokenize/
      three-guard copies.
- [ ] Preserve the exact current `KvFull` context-guard behavior for this path (see
      #743 — Metal's prefix-cache context bound is `prompt_len` alone, not
      `prompt_len + max_new_tokens`; do not converge it to the CPU bound here).
- [ ] Add or extend a test that warms a cache slot, then submits each rejected
      config (empty prompt, zero budget, over-context, logprobs, `enable_mtp|top
      LATTICE_MTP`), and asserts the warmed slot is byte-for-byte/logically
      unchanged after each rejection.
- [ ] Cover the cancellation-after-restore path (~lines 15505-15529) with its own
      test, separate from the preparation-rejection tests above, since it happens
      after cache restore has already mutated state and is not something
      preparation-boundary hoisting affects.

## Gates

- Warmed-slot-survives test passes for every rejected/early config.
- Exact empty/zero/`KvFull` output and `CrossTurnCacheStats` for both early and
  normal paths.
- Full prefix-cache test suite (`crates/inference/src/forward/metal_qwen35.rs`
  prefix-cache test module and any dedicated prefix-cache integration tests).
- e2e parity.
- Workspace `fmt`/`clippy`.
- `make bench-compare` output in the PR description per ADR-058.

## Sequencing

- Depends on: the behavior-matrix + maintainer sign-off issue, the shared CPU-path
  preparation issue (reuses its types/module), and the Metal direct/streaming
  migration issue (establishes the pattern this issue applies to the prefix-cache
  call site).
- Last implementation issue in this series before closeout.

ohdearquant:

## Summary

- add a `MetalPrefixCache` contract to the shared generation-preparation seam
- run capability checks, seed normalization, tokenization, empty/zero handling, and context admission at the non-destructive public wrapper
- convert zero-budget completion to the exact cached-output shape once, then pass a prepared `GenerationPlan` into the inner path
- preserve suffix validation, cache restore/reset, cancellation, and suffix-prefill ordering
- add warmed-slot state/snapshot preservation and exact-restore cancellation regressions

Fixes #826.

This is stacked on #1245 (`423d7b322d0b3f9f187cc40a5b096458189c3160`), which establishes the shared Metal preparation pattern.

## Preserved behavior

- capability order remains logprobs → MTP → tokenization/empty → zero → context
- zero-budget calls return `FullRefill` with zero reused/prefetched tokens and do not mutate a warm slot
- the Metal prefix path retains its existing lack of whole-prompt OOV admission; the existing suffix preflight remains in place
- current #922 reasoning-aware context admission is preserved exactly; the older #743 sentence in #826 was already superseded on the base branch by `check_context_budget`
- cancellation after an exact-prefix restore still consumes the entry and leaves the slot fail-closed

## Mutation proof

Each mutation was applied independently and restored:

- moving prefix capability checks behind empty handling made the warmed-slot ordering assertion fail
- changing the zero-budget `reused_tokens` statistic from 0 to 1 made the exact-output assertion fail
- bypassing exact-prefix restore made the cancellation regression fail because the warm entry survived

## Validation

All Cargo runs used the repository-wide shared target. Because concurrent worktrees can alias same-package artifacts, the final suites were forced to rebuild this exact committed source and the new test names were observed directly.

- shared preparation suite: 14 passed
- exact real-Metal warmed-slot regression: passed
- exact real-Metal restore-time cancellation regression: passed
- full default `lattice-inference` library suite: 2,427 passed, 20 ignored
- strict `f16,metal-gpu` library/test Clippy with warnings denied
- repository pre-commit hook: passed (format, workspace/all-target Clippy, docs lint)
- independent review: approved with no correctness findings

Model-backed e2e parity was not run locally; the hosted parity gate remains the fail-closed proof for that artifact-dependent leg.

## bench-compare disposition

**Not owed. The default A/B cannot observe this change.** Under ADR-087 D3-A step 1, which is
file-level and mechanical: none of this pull request's seven changed paths falls inside the
reachable surface of the DEFAULT `bench-compare` A/B — the corrected surface recorded in
Amendment 1 of `docs/adr/ADR-087-bench-compare-gate-calibration-and-coverage.md`. Paths are
classified on both `filename` and `previous_filename`. Because the file-level answer is no,
step 2 is not owed.

Worth naming rather than leaving as a bare no, because this pull request sits exactly on the
distinction the corrected surface fixed: it changes `crates/inference/src/forward/cpu_f16.rs`
and `crates/inference/src/forward/cpu_q8.rs`. Those are sibling modules of the reached
directory `crates/inference/src/forward/cpu/`, not files inside it. An earlier form of the
predicate used the prefix without the trailing separator and matched them, which would have
made this pull request read as reached and owed a step-2 discharge it does not owe.

- Pull request head evaluated: `e6141e58d3c54b6b8625f4346ec967dbb5c79a42`

Void if either ref moves. A moved head means this has discharged nothing and must be re-derived.

**Historical, and not a disposition.** An earlier attempt ran the A/B against a base that was
this pull request's parent at a superseded head, and the gate refused before measuring base
because the machine was 21.4% idle against the repository's unchanged 70% floor. The floor was
not lowered and no measurements were produced. That is a record of a machine state at refs this
pull request no longer carries, and nothing in the statement above rests on it.

Residual, stated rather than implied: an unreachable change is not a safe change, it is an
unmeasured one. Nothing here claims the change is neutral, only that the default A/B could not
have told anyone either way.

### DIFF
--- a/crates/inference/src/forward/cpu_f16.rs
+++ b/crates/inference/src/forward/cpu_f16.rs
@@ -14,8 +14,8 @@ use crate::attention::gdn_fused::{
 };
 use crate::forward::cpu::{elementwise_mul, silu_inplace};
 use crate::model::qwen35::{
-    ForwardScratch, KvCache, decode_tokens, qwen35_rms_norm, resize, sample_token,
-    should_stop_token,
+    ForwardScratch, GenerationEntryContract, GenerationPlan, GenerationPreparation, KvCache,
+    decode_tokens, prepare_generation, qwen35_rms_norm, resize, sample_token, should_stop_token,
 };
 use crate::model::qwen35_config::{GenerateConfig, GenerateOutput, Qwen35Config};
 use crate::rope::RopeTable;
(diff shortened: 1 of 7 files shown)


ohdearquant:
Closing this in favour of #1353, and recording what it carried that #1353 does not, so the idea is not lost with the branch.

State as of now: this PR is a draft last updated 2026-08-04, `CONFLICTING` / `DIRTY` against `main`.
Its parent in the stack, #1245, merged. The issue it says it fixes, #826, is closed.

#1353 does the narrow remaining piece under the still-open #827: `generate_streaming_with_prefix_cache_and_cancel`'s private delegate was still computing its RNG seed with an inline copy of the normalize-seed arithmetic rather than routing through `prepare_generation`. So the work was genuinely still outstanding — closing #826 did not migrate that path — but it is one file rather than seven, and it reuses the existing `MetalStreaming` contract instead of adding a variant.

**The design difference worth preserving.** This PR proposed a `MetalPrefixCache` contract variant that folds the capability checks into the seam while keeping the order `logprobs → MTP → tokenization/empty → zero → context`. #1353 deliberately did not do that, and the reason is the same fact from the other side: `prepare_generation`'s internal order is fixed at tokenize → `check_prompt_not_empty` → zero-budget → `validate_capabilities` → `validate_context`, while the wrapper ran the logprobs and MTP checks *before* tokenizing. Folding them into `validate_capabilities` as it stands would move them behind the empty-prompt check, changing which error a caller gets when it violates more than one guard at once. #1353 works around that by leaving the two checks as explicit preflights in their original position; this PR proposed solving it properly by making the seam's order accommodate them.

That is the better long-term shape, and it is a change to `generation_setup.rs` rather than to any one entry point. Filing it as follow-up work is the right home for it — a seam-ordering change deserves its own review and its own regression coverage, not a ride on a one-file migration.

Closing rather than rebasing: the branch is three days stale across a merged parent, conflicting, and its stated issue is closed, so a rebase would be reconstructing the design against a moved base with no reviewer waiting on it. Reconstructing from this description, when the seam-ordering work is scheduled, is cheaper and starts from a current base.

PR #1253, opened 2026-07-30
feat(serve): align reasoning and stop policies

### LINKED ISSUE #831: feat(serve): implement reasoning_budget on lattice serve; align stop and full-window accounting

## What

Today the two serve surfaces silently diverge on three related fields.
Verified at current `main`:

- `reasoning_budget`: `lattice serve`'s `ChatCompletionRequest` DTO
  (`crates/inference/src/bin/lattice.rs`) has no such field, and serde
  accepts unknown fields by default, so a client-sent `reasoning_budget` is
  silently accepted at the JSON layer and then silently ignored — worse than
  an explicit rejection. `lattice_serve`'s `ChatReq` DTO does have a
  `reasoning_budget: Option<usize>` field and maps it into `GenerateConfig`.
- `stop`: both DTOs already accept a `stop` field and parse it into stop
  strings on the surfaces where it is implemented; confirm which side (if
  either) is currently missing support before propagating in the shared
  contract, and align both to accept one string or 1-4 non-empty strings.
- Full-window accounting: the two servers must use one saturating formula —
  `prompt + max_new_tokens + reasoning_budget + 1 <= max_context` — rather
  than each computing (or omitting) the `reasoning_budget` and `+1` reserve
  terms independently.

**Maintainer decision (binding):** implement `reasoning_budget` on `lattice
serve` per `lattice_serve`'s existing behavior — accept, normalize positive
values, and apply to `GenerateConfig` using the shared full-window formula
above. Silent ignore is not an acceptable end state. Release notes for the
PR that lands this must call out that `lattice serve` changes from silently
ignoring `reasoning_budget` to actually applying it.

This work must go through the ADR-080 C3 backend-neutral decode-policy
guards (`check_reasoning_budget_not_set` / stop-string guards in
`crates/inference/src/model/qwen35/generation.rs`, shipped in #787, merged
2026-07-10) rather than bypassing them — every CPU decode variant and both
Metal entry points now apply-or-fail-closed on `reasoning_budget` and
`stop_strings` at the engine layer, and this issue's job is to make the HTTP
DTOs on both binaries actually populate those fields instead of dropping
them before they reach the engine.

## Why

The capability matrix documents this as a named cross-server difference,
and a 200 status alone cannot establish success parity when one server
silently no-ops a field the client explicitly set — that is a worse failure
mode than an honest 400. The full-window `+1` reserve term is also currently
at risk of being computed two different ways (or omitted) if each binary
keeps its own accounting; one shared formula removes that drift source
permanently.

## Checklist

- [ ] Add `reasoning_budget` to `lattice serve`'s request DTO; normalize and map positive values into `GenerateConfig`.
- [ ] Add or align `stop` string/array support so both binaries accept the same shape and both map into `GenerateConfig`'s `stop_strings`.
- [ ] Implement the shared saturating full-window formula (`prompt + max_new_tokens + reasoning_budget + 1 <= max_context`) as one function used by both binaries.
- [ ] Update `docs/capability-matrix.md` rows for `reasoning_budget`, `stop`, and window accounting.
- [ ] Add a release-note entry describing the `lattice serve` behavior change (silent-ignore → applied-or-rejected).
- [ ] Do not modify the ADR-080 C3 guard functions themselves (`check_reasoning_budget_not_set`, stop-string guards) — only the HTTP-layer DTOs and config construction that feed them.

## Gates

- Mutation sensitivity: delete the `reasoning_budget` field assignment, zero the budget value, drop a stop string, remove the `+1` window term, or revert to either binary's old accounting formula — the corresponding config-capture and exact-boundary parity-harness cases (from issue #1 in this cluster) must go red.
- New parity-harness rows: `reasoning_budget` propagation into `GenerateConfig` plus exact context-boundary accept/reject cases; `stop` string and array propagation.
- `cargo test -p lattice-inference --lib`, capability-matrix check, and both binaries' targeted + Metal-feature tests (where gated) pass.

## Sequencing

Depends on the shared-DTO issue (#3 in this cluster) landing first. Depends on ADR-080 C3 (#787) — **this dependency is now satisfied**, #787 merged 2026-07-10; branch for this work from current `main`, not from the pre-#787 snapshot the original design was scoped against.

ohdearquant:

## Summary

- Add `reasoning_budget` to the unified `lattice serve` request path, normalize positive values into `GenerateConfig`, and treat zero as absent.
- Give both HTTP binaries the same `stop` contract: one non-empty string or an array of 1–4 non-empty strings, propagated without truncation.
- Centralize saturating full-window admission as `prompt + max_tokens + reasoning_budget + 1 <= max_context` and use it in both production adapters.
- Make the unified server's generated-token invariant account for reasoning tokens plus the forced close delimiter, while retaining the prior bound when no reasoning budget is active.
- Update ADR evidence, the capability matrix, HTTP documentation, and release notes for the silent-ignore → applied-or-rejected behavior change.

The ADR-080 C3 engine guards are unchanged.

## Production adapter parity

A shared three-row table is consumed by the real unified and daemon router/adapter paths with the same measured prompt length and 1024-token context:

1. scalar stop + reasoning at the exact boundary: accepted and full config captured;
2. three-element stop array + reasoning at the exact boundary: accepted and full config captured;
3. two-element stop array one token past the boundary: rejected before generation.

The deprecated two-argument context callback now fails closed when a reasoning budget is present instead of silently undercounting it.

## Mutation sensitivity

Verified red before restoring the source:

- deleting the reasoning assignment fails both production-adapter consumers;
- truncating stop arrays to one element fails both consumers;
- removing the `+1` reserve makes both one-past rows return 200;
- restoring the old `max_tokens`-only output guard rejects the valid exact-budget case;
- accepting every output count admits the one-past case;
- removing the legacy-helper fail-closed branch lets a reasoning request bypass accounting.

## Validation

- `cargo test -p lattice-inference --lib`: 2442 passed, 20 ignored.
- `cargo test -p lattice-inference --bin lattice --features f16,metal-gpu,test-utils`: 154 passed.
- `cargo test -p lattice-inference --bin lattice_serve --features f16,metal-gpu,test-utils`: 100 passed.
- Shared production-adapter observations: 2/2 in each binary; shared contract suite: 53/53; full-window arithmetic: 2/2; unified output-cap regression suite: 4/4.
- Strict Clippy for `--lib --bins --features f16,metal-gpu,test-utils` with `-D warnings` passed.
- Metal-feature no-run compilation, `cargo fmt --check`, `git diff --check`, `make lint-docs`, and the repository pre-commit hook passed.
- Independent review: APPROVE; no correctness or contract findings remain.

The configured remote Mac mini was unreachable (`No route to host`). The Metal-feature validation above ran on the local macOS host; no engine math or live-model behavior changed.

## bench-compare disposition

Attempted the exact comparison:

```text
make bench-compare BASE=1e8ba317241d7b7dc674516a67d0ef10c5dcfe03 HEAD=42bcb2299e2e4a7c818c15ff9b5bfc4fbe9bd858
```

The merged machine-state gate refused before the base arm because idle CPU was 0.4%, below the unchanged 70% floor. No A/B result was certified or quoted. This change is HTTP normalization, validation, adapter plumbing, tests, and documentation; it does not modify timed kernels or hot-path math.

Fixes #831

- Pull request head evaluated: `42bcb2299e2e4a7c818c15ff9b5bfc4fbe9bd858`

re-derived.


ohdearquant:
Superseded by the reasoning and stop parity work merged in #1324, which closed #831. This branch's shared-contract and normalization changes are represented on main.

One behavioral slice of this branch is not on main: the post-generation completion-length invariant still checks only `generated_tokens > max_tokens`, so a completion that uses the reasoning allowance can pass admission and then be reported as a failure. That slice is filed as #1334 rather than lost here.

PR #1262, opened 2026-07-30
perf(inference): route compact top-k 50 through stage 1

### LINKED ISSUE #546: lm_head compact path: add local_k=50 Stage-1 pipeline variant

Follow-up to #171. The env-gated (`LATTICE_COMPACT_TOPK`) two-stage block-top-k lm_head path precompiles Stage-1 pipeline variants for `LM_HEAD_LOCAL_KS = {1, 8, 16, 40, 64}`. `top_k = 50` — a common sampler default — intentionally falls through to the older `HierarchicalK50` route, which still performs a full-logit GEMV (62,080 TGs Q8 / 124,160 TGs Q4) before its compact dispatch, so the threadgroup-reduction claim is explicitly scoped to the precompiled set.

Work:
- Add a `local_k=50` MSL function-constant variant to the Stage-1 precompile loop and route `top_k=50` to `BlockTopK { local_k: 50 }` under the same env gates.
- Top-k set agreement coverage for K=50 (real checkpoint, incl. adversarial clustered-tile case), plus the idle bench row.
- Re-check whether `HierarchicalK50` becomes dead code once routed away, and clean up if so.

This matters before any promotion of the compact path to default routing (which is itself gated on a measured end-to-end contention win, see #77).

ohdearquant:

## Summary

- add and precompile the Stage-1 `LOCAL_K=50` variant for F16 and Q4
- route `top_k=50` through Stage 1 only when both existing compact-selection gates are enabled
- remove the now-dead Rust `HierarchicalK50` route, pipeline, precompile, and dispatch plumbing
- add a fail-closed route resolver, execution-path marker, K50 benchmark case, route tests, and clustered-tile adversarial coverage

This does not change Metal math, intrinsics, SIMD behavior, buffer layouts, or threadgroup geometry for any kernel that survives.

**Correction to an earlier version of this description.** It said "the legacy MSL
kernel bodies are unchanged; the shader diff is comment-only." That was true when
written and is no longer true. The shader diff against `main` is now a 171-line
deletion, described in the next section.

## Why a shader deletion belongs in this PR

This PR removes `logits_topk_select50_first` and `logits_topk_select50_merge`
from `shaders/qwen35.metal`, together with the constants used only by them, and
adds a guard test asserting they do not come back.

That deletion is a consequence of this PR rather than an unrelated tidy-up, and
the distinction is checkable in one command. On `main`, both kernels are still
constructed unconditionally:

```
$ git grep -n 'make_pipeline("logits_topk_select50' origin/main -- crates/inference/src/forward/metal_qwen35.rs
2719:            topk_select50_first: make_pipeline("logits_topk_select50_first")?,
2720:            topk_select50_merge: make_pipeline("logits_topk_select50_merge")?,
```

There is no `cfg` gate on either; they sit in the same struct-initializer list as
the argmax and decode-attention pipelines. Deleting the kernels from the shader
without also removing those call sites would make both `?` operators fail during
pipeline construction, which breaks every Metal state construction.

The same grep against this branch's head returns nothing. Routing `top_k=50`
through Stage 1 is what removes the callers, so this branch is where the kernels
become dead, and it is the only ref where deleting them is correct.

The retention comment in the shader file claimed the kernels were "kept for the
`topk_readback` bench". That was inaccurate: `benches/topk_readback.rs` declares
its own `kernel void logits_topk_select50_first` and `..._merge` in an embedded
MSL string and compiles them through its own pipeline helper, so it never reads
the production shader file. The bench is unaffected by this deletion, and the
comment is removed with the kernels.

Fixes #546.

## Correctness evidence

- real cached Q8 checkpoint: exact top-50 set agreement on 8/8 cases, with zero accepted ties
- fail-closed K50 marker observed: `stage1_dispatches=1 candidates=50`
- clustered single-tile adversarial top-50 agreement passed
- route-gate unit tests passed
- production call-site audit confirms all three callers pass the compact-selection gate consistently

## bench-compare disposition

Measurement owed, not waived. The population and the reachability result are
stated rather than a bare N/A.

**The default A/B population.** `scripts/lib/bench-compare-impl.sh:441-443` sets
the default targets to `lattice-inference:elementwise_cpu_bench` and
`lattice-embed:simd`, and leaves `CARGO_FEATURES_INFERENCE` empty.
`crates/inference/Cargo.toml:16` gives `default = ["std", "download", "serve"]`,
so neither `metal-gpu` nor `f16` is enabled in a default bench build.

**Reachability against that population.** Every changed line in
`crates/inference/src/forward/metal_qwen35.rs` falls inside `mod inner`, which is
gated `#[cfg(all(target_os = "macos", feature = "metal-gpu"))]` at lines
1052-1053. Checked by range rather than by eye: the changed head-side lines span
1772 to 27358, the nearest preceding column-zero item declaration for that whole
span is `mod inner` at 1053, and the next column-zero items in the file are the
`#[cfg(test)]` blocks starting at 36299. The first column-zero `}` after 1053 sits
inside an embedded Metal shader raw string and is not the module's closer, which
is why a naive brace scan is not the instrument here. `shaders/qwen35.metal`
enters the crate through exactly one `include_str!`, at line 1092, inside that
same gated module. Base and head default bench binaries are therefore built from
identical effective source, and no A/B over the default targets could attribute
any delta to this diff.

**A declared target does reach the change, so this is not a waiver.**
`crates/inference/Cargo.toml` declares `lm_head_bench`, which measures the final
head route this PR changes. It is outside the default set, its real body compiles
only under `--features metal-gpu,f16,bench-internals` on macOS, and it requires a
checkpoint on disk. The rule grants a waiver only when no declared target reaches
the change; one does, so the measurement is owed on `lm_head_bench` under that
feature trio, on a quiet machine holding the bench-window and Metal locks.

**Residual risk.** Until that run exists this PR carries no performance evidence
about the K=50 route it changes. An unmeasured change is not a safe change, and
nothing above claims a speedup, a regression, or a neutral result.

**Reach re-derived under ADR-087 D3-A step 1, and it does NOT discharge what is owed above.**
Step 1 is file-level and mechanical: none of this pull request's four changed paths falls
inside the reachable surface of the DEFAULT `bench-compare` A/B — the corrected surface
recorded in Amendment 1 of
`docs/adr/ADR-087-bench-compare-gate-calibration-and-coverage.md`. That is a statement about
the default surface only. The section above establishes the opposite conclusion on a wider
population: `lm_head_bench` is a declared target that does reach this change, so the
measurement stands as owed. The narrow step-1 result and the wider reach result point in
different directions here, and the narrow one does not settle the wider one.

- Pull request head evaluated: `a1a6e8eeb0f4676b6abd82cc995757aed91fea4f`

re-derived.

**Historical, and not a disposition.** The exact locked A/B benchmark was attempted at the repository's unchanged 70% idle floor. The harness refused before measurement because ambient CPU idle was 3.0%, so no benchmark numbers are claimed.

The real Q4 runtime proof could not run because both locally cached Q4 checkpoints contain a pre-existing invalid block scale of zero. Q4 compilation and route construction are covered; runtime Q4 evidence is left to draft CI/review.

## Validation

- `cargo test -p lattice-inference --lib`: 2,427 passed, 0 failed, 21 ignored, as reported by the CI run for this head
- feature-enabled benchmark target compiled
- feature check, strict clippy, formatting, `git diff --check`, and the full pre-commit hook passed
- route and precompiled-variant mutations made the focused tests fail, then passed after restoration


ohdearquant:
Superseded: the compact top-k 50 Stage-1 routing this PR adds is already on main (local_k slot machinery in crates/inference/src/forward/metal_qwen35/inner/dispatch.rs, validated against LM_HEAD_LOCAL_KS via resolve_block_local_k; see #546, closed as completed). The one piece main lacks — a block_topk_k50 RouteCase in lm_head_bench — will ride a future bench PR.

PR #1286, opened 2026-08-01
ci: pause the post-merge perf lane's automatic trigger

ohdearquant:
## Summary

Removes the automatic `push` trigger from the post-merge perf lane. The lane is **paused, not fixed**: `workflow_dispatch` is unchanged, and the paths list is retained verbatim as a comment so that restoring the trigger is a revert of one block rather than a re-derivation of coverage.

## Why

The lane has produced 15 verdict-bearing runs. All 15 reported a confirmed regression, on 15 unrelated merges.

The cause is the decision rule, not the plumbing. The rule fails a run when a group's **lower** confidence bound exceeds 7%, applied one-sidedly across **263 group comparisons per run**. Reading the gate's own published tables across all 15 runs (3,948 group comparisons):

- Criterion's intervals are tight **within** a run: median width across the 82 FAIL rows is 0.280 percentage points.
- The same groups swing widely **between** runs. `simd_throughput_384/normalize` flagged in 7 of the 15 runs; `simd_dot_product/simd/768` in 5; `simd_query_batch_dot_product/simd_batch/768d_16c` and `layer_norm/4096` in 4 each.
- A within-run interval cannot bound between-run variability, and the gate asks it to. With 263 groups and that spread, at least one positive excursion past 7% per run is close to certain, so a one-sided rule converts roughly half of a symmetric noise distribution into confirmed regressions.

Two alternative explanations were tested and rejected rather than assumed:

- **Real compounding regressions.** If those recurrences were genuine per-merge regressions, `normalize` alone would have compounded roughly 8–11% seven times over and be about twice as slow today. It is not.
- **Arm-ordering bias.** The harness on `main` runs base-then-head, so drift over a run would push head systematically slower. Measured, it does not: among the 3,948 comparisons, excursions past ±7% split 85 head-slower against 119 head-**faster**, and past ±10%, 49 against 86. Large deltas are symmetric and slightly favour improvement, which refutes a sign-aligned whole-run artifact by its own signature.

So each individual verdict is a correct application of a stated rule to a tightly-measured number. The rule is what is wrong.

## Cost of leaving it on

Roughly 88 minutes of wall clock per run on each of two runner jobs, for every qualifying merge, with observed excursions to 116, 145 and 169 minutes. Four recent runs completed the full A/B and then discarded it at the upload step, burning the entire duration for nothing.

This repository is public, so hosted runner minutes are free and the billing field reads zero. That is a billing fact, not a time fact; the wall-clock occupancy and the serial concurrency group are real.

## What is deliberately not done here

Making the failing upload step non-fatal would be the tempting smaller change. It is the wrong one: with that defect masked the lane would resume reaching its gate, and the gate's demonstrated behaviour is to call a regression on essentially every merge. That does not produce a false all-clear, it restores a false alarm.

## Restoring the trigger

Named in the workflow comment so the conditions travel with the file:

1. a between-run variance estimate rather than a within-run confidence interval;
2. multiple-comparison handling across the 263 groups;
3. identifying the chronically unstable groups and either stabilising or excluding them — and any change to which groups are gated requires a repeated A/A on the target machine class plus a re-derivation of the threshold first.

## Safety

`perf-postmerge-gate` is not a required status check. The required contexts on `main` are `ci-gate`, `cargo-deny`, `parity-gate`, `app-binaries-gate`, `typos`, `actionlint`, `cargo-audit-gate`; this lane is in none of them, and the workflow's own header already instructs that it must never be added, since a post-merge job cannot be satisfied pre-merge. Pausing it therefore cannot affect any merge.

`actionlint` passes on the edited file. The check was validated against a deliberately broken copy of the same file, which it rejects with a parse error, so a clean result is a working linter rather than a silent one.

## bench-compare disposition

Not applicable: this PR changes one file under `.github/workflows/` and touches no code in `crates/inference/`, `crates/embed/` or `crates/fann/`.

### DIFF
--- a/.github/workflows/perf-postmerge-gate.yml
+++ b/.github/workflows/perf-postmerge-gate.yml
@@ -1,4 +1,9 @@
-name: perf post-merge detection (informational)
+name: perf post-merge detection (informational, PAUSED)
+#
+# PAUSED 2026-08-01: this lane no longer runs automatically. It is on-demand only
+# (workflow_dispatch). The rest of this header describes the lane's design and is
+# unchanged; the reason for the pause and the conditions for lifting it are in the
+# block immediately above `on:`. Read that before restoring the trigger.
 #
 # WHAT THIS IS, AND WHAT IT IS NOT (lattice#1105).
 #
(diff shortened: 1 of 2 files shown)


ohdearquant:
Second commit added: the contract test in `tests/test_perf_postmerge_workflow.py` parsed the live `push` block and asserted its paths equal the bench-binary input set, so removing the trigger made it fail with `push paths block is missing`. The test fired correctly — its purpose is that this lane's coverage cannot change silently.

It is adapted rather than weakened. It now recognises both states and pins the same set in each: while the lane is paused it parses the retained commented block, so the path list cannot drift while the trigger is off and restoring the trigger stays a revert. Two contracts are added — the file must be in exactly one trigger state (never both, never neither, which rejects a half-restored workflow), and while paused the `on:` block must contain `workflow_dispatch` alone with `PAUSED` in the workflow name.

Both arms were mutation-checked rather than assumed: dropping one path from the retained comment fails the path contract, and adding a live `push` trigger beside the commented one fails the state contract. Each mutation was reverted by its inverse (not by `git checkout`, which would have discarded the change under test) and the suite returns to green, 5 tests.

PR #1289, opened 2026-08-02
feat(embed): one-command embedding-drift gate across the production models

ohdearquant:
## Why

Answering "does this ref change embedding output versus the last published release?" took two
worktrees, two release builds, two example runs and a third example to compare them. It is now one
command.

That question is not the same one `embed_parity_vs_hf` answers. HF parity can hold while the
lattice-versus-lattice geometry shifts inside the HF tolerance, and a consumer that has persisted
an ANN index over these vectors cares about the second property: a numeric change
invalidates the geometry that index was built against, which is a retrieval and
index-version risk rather than a compile question. Nothing in this tree measures
recall over the served models, so this PR makes no claim about the size of that
risk; the only Recall@K computation here evaluates SIMD-tier agreement against
synthetic corpora and is unrelated to model or version drift.

The frozen drift baseline covered `bge_small_en_v15` alone. The two models actually served,
`all-minilm-l6-v2` and `paraphrase-multilingual-minilm-l12-v2`, had no lattice-versus-lattice
drift coverage at all.

## What landed

**`lattice_embed::drift`** — the comparison moves out of the test and into the library, so the test
and the new binary share one implementation and one definition of `MAX_COSINE_DRIFT`. Two
implementations of one comparison is the sibling-path failure this repo already has a convention
about.

The per-model outcome is three-state:

```rust
pub enum ModelDriftOutcome {
    Checked { max_one_minus_cos: f32, worst_index: usize },
    WeightsAbsent { model: String },
    NoBaseline { model: String },
}
```

A skip is not representable as a pass, and the `WeightsAbsent` value is written by the failure path
itself rather than assembled later when the overall result is computed. `embed_drift_baseline` retains its threshold and its
skip/enforce semantics, broadens from BGE alone to every committed baseline fixture
(four at this head, matching the binary's required-model set), and now calls this
module.

**`embed-drift` binary** — `required-features = ["native"]`, following the existing `embed` bin's
parser style and its `@@lattice {"ev":...}` json-line convention.

```
embed-drift [--model NAME]... [--json] [--enforce] [--update-baseline] [-h|--help]
```

| exit | meaning |
|---|---|
| 3 | a requested model has no baseline fixture (`NoBaseline`); also usage or IO errors |
| 1 | drift at or above threshold on at least one checked model |
| 2 | a model's weights were absent while `--enforce` was set |
| 0 | none of the above |

Precedence is fixed in code and evaluated in this order (3 → 1 → 2 → 0) over all
rows in one run; the first matching condition wins, so drift outranks enforced
missing weights when both occur.

The `--json` line carries explicit `checked` / `skipped` / `no_baseline` counts, so a consumer can
tell "all clear" from "nothing ran" from the line alone. Without `--enforce`, absent weights are a
visible skip: such a run can exit 0 with `checked: 0`, and the JSON counts are what distinguish it
from an all-clear — enforcement is what makes absent weights fatal. `--update-baseline` refuses to
run alongside `--enforce`.

**Baselines for all four fixture models**, generated at the `v0.7.0` release commit — a tree
measured bit-identical to published `v0.6.1` on all four models, so the baselines are anchored to a
published release rather than to an arbitrary commit. The current release is `v0.7.1`; the fixture
schema records model, texts and embeddings but no release provenance, so the gate establishes
drift against the committed `v0.7.0` baseline specifically, not automatically against whatever
release is latest.

**Gate hardening** — the checked model set is bound in the binary (a `REQUIRED_MODELS` const
naming the four fixture models): with no `--model` given, deleting a fixture surfaces as
`NoBaseline`/exit 3 instead of silently shrinking the checked set. The drift comparison's private
cosine now classifies a zero- or non-finite-norm current embedding as maximum drift rather than
identical. Both changes carry regression tests that fail with the fix reverted.

**Two defects in the dump and compare examples**, both found by using them:

- `drift_compare` reported biased entropic optimal transport. Entropic regularisation has a
  self-distance floor on small point sets, so running it on two byte-identical inputs printed
  nonzero wasserstein and displacement — up to `2.9e-2` on one model — with nothing in the output
  saying so. Anyone comparing two genuinely different dumps would read that floor as drift. It now
  reports the debiased `sinkhorn_divergence`, which is zero at identity. The header also states that
  `max 1-cos` is f32-saturated near 1.0 and cannot resolve below roughly `1e-7`, so it is not on its
  own proof of identity.
- `dump_parity_embeddings` serialised a `HashMap`, so JSON key order and therefore the file digest
  changed between runs over identical vectors. A digest of the dump was not a valid identity check.
  It now serialises deterministically.

**CI** — the `embed-drift` job runs the binary and gains weight caches for the added models
following the existing cache-key pattern; `embed-drift-x86` stays an observation job. The release
workflow gains a drift step with `--enforce` beside the existing HF parity step. The two answer
different questions and both are kept.

The `feature-matrix` job also gains a step running `cargo test -p lattice-transport --example
drift_compare` on both its Linux and macOS legs. `cargo test --workspace` builds libs, bins, test
targets, benches and doctests, and does not build `[[example]]` targets, so the four degenerate-case
regressions guarding `drift_compare`'s `cosine()` — including the one-sided-collapse cases — lived in
a `#[cfg(test)]` module that no required job ever compiled. Written but never run is the failure mode
those tests exist to prevent, so they are now invoked by target explicitly. The step runs 4 tests.

## Verification

Every arm below is command output from this branch, not a description of one.

**Debiasing is real, and it did not blind the tool.** On two byte-identical dumps every row now
reads `0.000000e0`. Perturbing one coordinate of one `all_minilm` vector by `+0.05` moves that row
to `4.999936e-4` divergence and `1.239955e-3` max `1-cos` while the other four rows stay pinned at
zero, so the comparator still isolates per model.

| arm | result |
|---|---|
| must-MOVE: perturbed fixture | exit **1**, names the model |
| must-PASS: same series, unperturbed | exit **0** |
| skip control, enforcing, empty model root | exit **2** |
| skip control, not enforcing | exit **0**, every model shown skipped, json `checked: 0` |
| single-definition of the threshold constant | one definition, positive control in the same invocation |
| rewritten test still bites under reverse-applied change | fails, then passes on restore |
| four models, one command | `checked:4 skipped:0 no_baseline:0`, all four PASS under the `1.0e-3` threshold: `bge-small-en-v1.5` and `multilingual-e5-small` at `max_one_minus_cos = 0.0`, `all-minilm-l6-v2` and `paraphrase-multilingual-minilm-l12-v2` at `5.960464477539063e-8` |
| `clippy --workspace --all-targets -- -D warnings`, `fmt --check` | clean |
| `actionlint` on both edited workflows | clean |

## Bench-compare disposition — A/B owed, not waived

This touches `crates/embed/`, so the bench policy (ADR-087, superseding ADR-058) applies. The structural waiver does **not**: `drift.rs` is
a new module in the lib and the `simd` bench binary links the lib, so the changed lines are compiled
into the bench build. Not being *called* is a different property from not being *compiled*, and the
rule is written against the compiled test deliberately.

The expectation is a null — `crates/embed/benches/simd.rs` has zero references to the drift path,
and nothing on the existing embedding hot path changed. Stating that here so a null reads as
confirmation rather than as an absence of signal. The A/B runs before this merges; it needs a quiet
machine, which is why this opens as a draft.


oceanwaves630 (approved):
Approving on the recorded review result for this head together with a check-run verification (29 success, 1 skipped, 0 failing). This is a stamp of that result, not an independent line-by-line read of the diff.

Issue #1294, opened 2026-08-02
bench: no group reaches simd::manhattan_distance; calibrate before it can gate


ohdearquant:
## What

`crates/embed/src/simd/manhattan.rs` (added in #1293) has no bench coverage. No
bench group's measured path reaches `manhattan_distance`: across the 30 bench
files in the repo the string `manhattan` appears 0 times, and in
`crates/embed/benches/simd.rs` — the gated `lattice-embed:simd` target — it
appears 0 times against a control of 19 occurrences of `euclidean` in the same
file.

## Why this is a tracked row and not a promise

#1293's bench-compare obligation was discharged as **unmeasurable-now**, not
waived. Three facts made it unmeasurable rather than unnecessary:

1. `crates/embed/src/simd/` is *reachable surface* under ADR-087's own coverage
   predicate. The precise position is **reachable module, unreached function** —
   not "the instrument cannot see this directory".
2. The measurable question that leaves is **neighbor perturbation**: new code
   compiled into the bench binary can shift the layout of the kernels the bench
   does measure, whether or not anything calls it.
3. That question is exactly what an uncalibrated instrument cannot separate from
   its own noise. Under ADR-087 D1 the gate is advisory pending threshold
   re-derivation, and the diff is additive-only, so the perturbation risk is
   layout noise.

The obligation transfers here whole.

## Work

Per ADR-087 D3, widening the bench target set is itself an instrument change, so
this is not "add a group":

- [ ] Add a Criterion group exercising `simd::manhattan_distance` across the
      dimensions that straddle the 4/8/16-lane widths and the 4x-unrolled chunk
      sizes, matching how the existing distance groups are shaped.
- [ ] Run the A/A null on the new group on a machine that clears the idle floor
      for the whole run, and record the `Run conditions` block.
- [ ] Re-derive the threshold against that null before the group can contribute
      to any FAIL verdict.
- [ ] Only then may bench-compare output be cited as evidence about L1 kernel
      changes.

## Related

- #1293 adds the kernels
- ADR-087 D1 (advisory pending re-derivation), D3 (coverage, and target-set
  widening as an instrument change)

ohdearquant:
## The A/A null has been measured, and it exceeds the gate's own reject region

This issue asked for an A/A null before re-deriving thresholds. That null now
exists as data rather than as an assumption, taken on a dedicated idle Apple
Silicon host while landing #1293.

[...]

Issue #702, opened 2026-07-07
perf(embed): head-batch the per-head score/context GEMMs in BERT attention (dispatch-tax reduction)


ohdearquant:
## Motivation

Attribution profiling (from #673) breaks BERT attention time at small M into three cost centers, measured per-layer at bge-small-en-v1.5 geometry (hidden 384, 12 heads, head_dim 32), consistent across single-text (seq 16/32) and the packed batch path (total 1024 rows):

| cost center | share of attention time |
|---|---|
| per-head Q/K reshape copies | ~2-3% |
| tiny per-head score/context GEMMs (24/layer) | ~10-13% |
| large Q/K/V + output projection GEMMs | ~85-89% |

The per-head score (`Q@K^T`) and context (`scores@V`) GEMMs run as `num_heads` separate tiny `matmul_bt` calls per layer (M≈seq_len, K=head_dim=32, N≈seq_len). At ~0.30µs/call these are dominated by per-call dispatch overhead, not compute — the small-M Accelerate/AMX dispatch tax also documented in #688. This issue proposes batching the per-head GEMMs into fewer, larger dispatches (a batched/blocked GEMM across heads, or a strided-batch formulation) to reclaim that overhead.

## Scope and measured ceiling

The recoverable ceiling is **~11-13% of attention time** (the tiny-GEMM share). This is deliberately stated up front so the work is not inherited under an inflated premise: this is **not** the single-text latency lever. That remains the large projection GEMMs (~85% of attention time), which are Accelerate/AMX-dispatch-bound at these shapes (#688) and are untouched by this change.

## Mandate: measure first

- **Profile the batch case first.** Head-batching compounds with the packed batched-encode path (#677): the per-head GEMMs are the same per-sequence cost, but the large projection GEMMs amortize over `total` rows, so the recoverable fraction is at its most favorable in batch. Confirm the split there before writing the kernel.
- **Parity is accepted-reassociation, not bit-exact.** Batching heads changes the reduction grouping of the score/context sums, which reorders f32 additions. Gate parity with a tight tolerance justified as N× the measured max (same convention as #677), backed by a mutation-sensitive test; do not assert `to_bits()` equality.
- Include before/after `bench-compare` numbers per the repo perf-PR rule.

## Methodology note

[...]

PR #1332, opened 2026-08-06
perf(inference): head-batch per-head attention GEMMs; bench(embed): manhattan group

ohdearquant:
Part of #702. (An earlier revision also carried the #1294 manhattan bench group; that landed on main independently via #1344, so this branch was rebuilt on top of it and no longer touches crates/embed.)

## perf(inference): head-batch per-head Q@K^T / scores@V GEMMs (#702)

`multi_head_attention_batched` issued `num_heads` separate tiny `matmul_bt` calls per sequence for the score (`Q@K^T`) and context (`scores@V`) products — at bge-small geometry (head_dim 32) these are dispatch-dominated, not compute-dominated. This change stacks each product into ONE `matmul_bt` call per sequence and extracts the diagonal `(h,h)` blocks.

Attribution reproduction (measure-first, packed-batch path, bge-small geometry, total=1024 rows, script + raw output retained in the workspace): tiny-GEMM share measured **30.5%** of attention time at batch=64/seq=16 and **27.5%** at batch=8/seq=128 on this host — materially above the ~10-13% the issue cites; reported as measured, not reconciled. Either figure clears the "real recoverable dispatch tax" bar.

Trade-off, stated explicitly: the stacked formulation computes the off-diagonal cross-head blocks and discards them — O(num_heads) more FLOPs and scratch per sequence in exchange for `2×num_heads → 2` dispatches. Expected to be favorable at short-seq/many-sequence shapes (the embedding serving shape) and **unverified at every sequence length, short or long** — see the bench-compare section below, which no longer claims otherwise. Apple Accelerate on this class of host exports no `cblas_sgemm_batch` (probed at link time), so a true strided-batch GEMM is not available without new FFI surface.

Also bounds the transient head-batched scratch. Six buffers are live at once per packed sequence and the dominant term is `(num_heads*seq_len)^2`, quadratic in a value the caller controls through `cu_seqlens`, so `head_batch_scratch_plan` sizes all six with checked arithmetic and rejects the shape with a typed error when their **total** would exceed 384 MiB — before allocating anything. 384 MiB admits BERT-large at its standard 512-token ceiling (310.00 MiB live), which is the largest shape in routine use here; bge-small at 512 tokens is 167.25 MiB. This bounds one call; bounding N concurrent calls is #1480.

Parity: new `head_batched_attention_matches_per_head_reference` test compares the live path against a preserved verbatim per-head reference implementation (same pattern as the file's existing padded reference). Measured max-abs-diff at the test shape: 0.0; tolerance pinned at 1e-5 matching the crate's other reassociation checks, with the pin's rationale in the test. Mutation check run: corrupting the diagonal-block offset fails both this test (max diff 2.58) and the pre-existing packed-vs-single cross-check; revert returns green. Full `attention::standard::` module 8/8; `cargo test -p lattice-embed` and `-p lattice-inference --lib` both clean.

## bench-compare disposition (ADR-087)

Three A/B runs, base `3671c5f0e109be5e0b4859234ec6974fa3ed1801` (origin/main) vs head `e74625aafa1558555b62745dd67f093cd336d37a`, on a dedicated quiet bench host. All three refused a source-performance classification under the harness's AB/BA order-bias gate (fail margin 7.0%), each time on a **different** ns-scale group, with every machine-state checkpoint passing (AC power, thermal nominal, HID idle, ambient idle floor, both machine-wide locks held uncontended):

1. `--quick`, default targets: worst order bias `matmul_bt/attn_proj_seq16_hidden384` at 14.28%.
2. `--full`, default targets: `matmul_bt` at 16.98%; an unrelated embed `normalize` group also at 15.66%.
3. `--full`, filters `inference='attention'` `embed='float32_simd'`: worst `layer_level_ops/softmax_attention/heads12_seq64` at 15.41%; the other 7 of 8 inference groups had order-bias bounds ≤ 3.97%.

A different offender on each run of a certified-quiet machine is a property of measurement resolution on ns-scale groups, not of this diff. The harness suppressed WARN/FAIL classification and this section reports that refusal rather than overriding it; no run below is quoted as a performance claim.

What run 3 (full resolution) does render and show:

- **`lattice-embed:simd` (`float32_simd`): ✅ all 15 gated benches within the ±3.0% noise band** — a rendered pass for the target the embed-side change sits nearest to.
- Inference `attention` filter, diagnosis-only: 7 of 8 groups clean, point deltas in [-2.63%, +0.43%]; the 8th is the order-bias-invalidated group above. These groups are a regression control, not evidence about this diff: `attention_kernel` calls `multi_head_attention` and `softmax_attention` times the softmax kernel directly, and neither function is modified here.
- **Correction (supersedes an earlier revision of this section).** An earlier version of this body offered `attention_kernel/seq_len/128` at -0.28% and `softmax_attention/heads12_seq128` at +0.13% as the long-seq arm promised in the trade-off paragraph. That was wrong and is withdrawn: the only in-tree caller of the changed function is `BertModel::forward_batch`, and neither of those two groups reaches it. They exercise code this diff does not touch, so they cannot express the regression they were quoted as ruling out. The long-seq arm has not been run.

Coverage notes:

- Run 3's filter drops `matmul_bt` deliberately: this diff does not modify `matmul_bt` itself, only its caller's batching in `multi_head_attention_batched`, so `matmul_bt` groups time the unchanged primitive (they were also the order-bias offenders in runs 1-2).
- The group this PR adds (`attention_batched_kernel`) is the only declared bench target that calls `multi_head_attention_batched`, and it has no base-side counterpart, so no A/B exists for it. Combined with the correction above, the honest coverage statement is that **no A/B in this section measures the changed function at any shape**; the bench group is a coverage addition, not a measured claim. Landing the bench group on the base first is what would make that A/B possible. (The manhattan group measured in run 3's embed target is main's own, via #1344.)

<details><summary>Run conditions (run 3)</summary>

    base: origin/main (3671c5f0e1)   head: e74625aafa1558555b62745dd67f093cd336d37a
    head arm: detached worktree
    resolution: --full
    targets: lattice-inference:inference_bench, lattice-embed:simd
    inference features: <none>
    filters: inference='attention' embed='float32_simd'
    enforcement: report-only (regressions reported; measurement failures still refuse)
    arm order: ABBA (base1 -> head1 -> head2 -> base2)
    locks: bench-window acquired immediately (uncontended); Metal GPU acquired immediately (uncontended)
    ambient: before first arm idle 78.2% | between order strata idle 96.7% | after final arm idle 96.2% (floor 70.0%)
    machine state: all three checkpoints passed (AC via pmset, thermal nominal, HID idle >398000s, kill-switch clear)
    rustc 1.94.1 / cargo 1.94.1 / criterion 0.5.1 on both arms
    started 2026-08-11T02:25:22Z, finished 2026-08-11T02:40:22Z

</details>


ohdearquant:
## bench-compare disposition: measured regression at 100-token sequences

The A/B has landed and it does not support merging this as it stands. The trade-off this PR
flagged as unverified at longer sequences is real and large.

### The default comparison could not have found it

`make bench-compare` runs `lattice-inference:elementwise_cpu_bench` and `lattice-embed:simd`
by default. Neither reaches `multi_head_attention_batched`. I searched all 24 bench targets
declared in `crates/inference/Cargo.toml`: `inference_bench`'s `attention_kernel` group calls
`multi_head_attention` (the single-sequence path), and the bench file defines its own local
`forward_pass` that does the same, so the whole of `inference_bench` is blind to this change.
The one declared target that reaches it is `e2e_bench`, whose `encode_batch` groups go
through `BertModel::encode_batch` → `forward_batch` → `multi_head_attention_batched`.

For the record, a full A/B on the two default targets was run first and reported all 31
groups within the noise band. That result is not evidence about this change; it is a
statement about coverage.

### e2e_bench, base b76c4278c0 vs head 2cdc620d5e

| Bench | Δ | ABBA bound | order bias ≤ | new ns | base ns | verdict |
|---|---:|---|---:|---:|---:|---|
| `e2e_encode_batch_padding_variance/uniform_100` | +19.75% | [+19.54%, +19.97%] | +0.12% | 16198035.9 | 13526758.2 | FAIL |
| `e2e_encode_batch_padding_variance/mixed_5_30_100` | +13.65% | [+13.10%, +14.24%] | +0.28% | 8594014.7 | 7561908.7 | FAIL |
| `e2e_encode_batch/bge_small/batch_1` | -0.74% | [-3.62%, +2.28%] | +2.62% | | | pass |
| `e2e_encode_batch/bge_small/batch_4` | +1.91% | [-0.92%, +4.60%] | +2.21% | | | pass |
| `e2e_encode_batch/bge_small/batch_8` | +1.02% | [+0.64%, +1.44%] | +0.25% | | | pass |

Higher ns is slower. The two failures are not marginal: the confidence intervals are disjoint
from zero, they are an order of magnitude wider than the measured order-bias envelope, and
they exceed the 7.0% fail threshold by a factor of two to three.

The split matches the mechanism exactly. The `e2e_encode_batch` groups encode sentences of
roughly ten tokens, where dispatch cost dominates and the stacked formulation is neutral to
slightly negative. The `padding_variance` groups carry 100-token texts, where the discarded
off-diagonal cross-head arithmetic dominates, and there the cost is 13.6% to 19.8% of
end-to-end encode time.

### Run conditions

```
resolution: --full
targets: lattice-inference:e2e_bench, lattice-embed:simd
filters: inference='e2e_encode_batch' embed='^simd_dot_product/simd/384$'
arm order: ABBA (base₁ → head₁ → head₂ → base₂)
locks: bench-window acquired immediately (uncontended); Metal GPU acquired immediately (uncontended)
ambient: before first arm idle 95.8% | between order strata 96.7% | after final arm 97.1% (floor 70.0%)
machine state: AC power, thermal nominal, HID idle, gate passed at all three boundaries
samples: 200 per arm on every group
```

Host is a dedicated idle machine, not a developer workstation. `lattice-embed:simd` was
filtered to a single group as a sanity control and passed within the noise band.

Base `b76c4278c0` is behind current `main` (`35f1736833`), but `main`'s delta does not touch
the measured call path: `crates/inference/src/model/bert.rs` and
`crates/inference/src/attention/standard.rs` are unchanged between the two, and the only bench
file that moved is `metal_decode_bench.rs`, which is not among the measured targets.

### What this suggests

The stacked formulation wins where it was designed to win and loses where the PR predicted it
might. A sequence-length threshold that selects the stacked path below a measured crossover
and keeps the per-head path above it would capture the dispatch saving without the long-
sequence cost. Finding that crossover needs a bench group that reaches
`multi_head_attention_batched` directly across a sequence-length sweep, which does not exist
today; that gap is worth closing regardless of which way this PR goes, since it is the reason
the regression was invisible to the standard comparison.

100-token inputs are an ordinary embedding-serving shape, so the failing groups are not a
corner case.

ohdearquant:
## Sequence-length sweep: there is no useful crossover

Follow-up to the end-to-end regression reported above. The suggestion there was a sequence-length threshold selecting the stacked path below a crossover and the per-head path above it. The bench that makes that measurable now exists on this branch (`attention_batched_kernel`, behind `bench-internals`), so the crossover is measured rather than assumed. It sits at roughly 32 tokens, and the win below it is small enough that a threshold is not worth building.

Both arms carry the same bench commit, so the group exists identically on each side; the arms differ by exactly the two files this PR changes (`crates/inference/src/attention/standard.rs`, `crates/embed/benches/simd.rs`), verified by `git diff --name-only` between the two SHAs.

### `attention_batched_kernel`, base `e09149c07f` vs head `e74625aafa`

| Bench | Δ | ABBA bound | order bias ≤ | new ns | base ns | verdict |
|---|---:|---|---:|---:|---:|---|
| `seq_len/16` | −8.74% | [−14.73%, −2.91%] | +4.03% | 62129.9 | 68081.1 | improvement |
| `seq_len/32` | −0.65% | [−2.08%, +0.80%] | +0.96% | | | within noise |
| `seq_len/64` | +25.87% | [+22.66%, +28.82%] | +1.71% | 178488.0 | 141801.4 | FAIL |
| `seq_len/128` | +68.81% | [+59.28%, +77.56%] | +3.66% | 509525.3 | 301835.0 | FAIL |
| `seq_len/256` | +234.33% | [+229.47%, +239.37%] | +1.29% | 2601597.2 | 778156.2 | FAIL |
| `seq_len/384` | +308.12% | [+303.91%, +312.39%] | +0.77% | 5738725.4 | 1406144.6 | FAIL |
| `seq_len/512` | +363.83% | [+356.87%, +370.78%] | +1.11% | 11528017.9 | 2485409.7 | FAIL |

The `batch_size` sweep (1, 4, 8, 16) is flat: −0.93% to +1.58%, every interval inside the ±3.0% noise band. Higher ns is slower.

### Reading

The cost grows faster than the sequence length: 1.26× at 64 tokens, 1.69× at 128, 3.34× at 256, 4.64× at 512. That is the shape of discarded off-diagonal work scaling with the square of the sequence, and it matches the end-to-end split reported earlier, where the ~10-token `encode_batch` groups were neutral and the 100-token `padding_variance` groups lost 13.6% to 19.8%.

A threshold at 32 would therefore buy at most the 8.74% win measured at 16 tokens, on sequences of 16 tokens or fewer, at the cost of carrying two attention implementations and a length-dependent branch. The `batch_size` sweep shows the dispatch saving the stacked formulation was designed to capture does not appear even where batching is the varied axis. On this evidence the trade is not worth making, and I do not think a threshold rescues this PR.

What is worth keeping regardless of what happens to the rest: the `attention_batched_kernel` bench group itself. Before it existed, no declared bench target reached `multi_head_attention_batched`, which is the reason a full A/B on the default targets reported 31 of 31 groups clean against a change that costs 3.6× at 512 tokens. If the rest is dropped, that group should be split out and landed on its own.

### Run conditions

```
resolution: --full
targets: lattice-inference:inference_bench, lattice-embed:simd
inference features: bench-internals
filters: inference='attention_batched_kernel' embed='^simd_dot_product/simd/384$'
arm order: ABBA (base₁ → head₁ → head₂ → base₂)
locks: bench-window acquired immediately (uncontended); Metal GPU acquired immediately (uncontended)
ambient: before first arm idle 96.3% | between order strata 97.6% | after final arm 96.8% (floor 70.0%)
machine state: AC power, thermal nominal, HID idle, gate passed at all three boundaries
samples: 20 Linear per arm on every group
rustc 1.94.1 both arms; criterion 0.5.1 both arms
```

Host is a dedicated idle machine. `lattice-embed:simd` was filtered to a single group as a sanity control and passed within the noise band.

ohdearquant:
## Measurement-conditions disclosure on the end-to-end table above

The operator of the other workload on the bench host has since produced a per-event ledger for the evening, and it changes what the first table's run conditions can be taken to mean. Recording it here rather than leaving the number to stand on a quiet-machine claim it does not fully support.

The end-to-end run (`e2e_bench`, the +19.75% / +13.65% table) ran `00:00:31Z` to `00:12:13Z`. The other workload's ledger records 9 browser-driving events in the `00:00Z` ten-minute bucket and 9 in the `00:10Z` bucket, so up to 18 fall inside that window. The three machine-state checkpoints in that run all certified — 95.8%, 96.7%, 97.1% idle against a 70% floor — but those are point samples taken at phase boundaries, and a burst landing mid-phase is exactly what they cannot see. So the run's own certification is not the reassurance it looks like, and I am not going to present it as one.

Two things keep the conclusion standing while the specific numbers get a caveat.

The sequence-length sweep in the comment above is an independent measurement of the same mechanism, run `00:34:24Z` to `00:42:19Z`, a window carrying 4 events rather than 18. It agrees in direction and its effects are far outside anything ambient load produces: +364% at 512 tokens, +234% at 256, with confidence intervals disjoint from zero and an order-bias envelope of about 1%.

The two results also explain each other mechanically. A cost that grows super-linearly in sequence length predicts exactly the end-to-end split that was observed — neutral on ~10-token `encode_batch` inputs, badly negative on 100-token `padding_variance` inputs. Ambient browser load does not produce a pattern that tracks sequence length.

What I would not do is quote `+19.75%` as a precise figure. Treat it as directionally right and numerically soft; the sweep is the measurement to cite. I will re-run the end-to-end arm under a coordinated hold on the host and replace the table when I have a clean one.

For the record on the default-target run mentioned earlier: it ran `23:30:49Z` to `23:49:26Z`, against 6 events in the `23:40Z` bucket. It was already labelled as a statement about coverage rather than evidence about this change, so its conditions do not affect any conclusion here.

ohdearquant:
## Clean end-to-end re-run: the caveat retires, the numbers hold

Discharging the commitment from the disclosure comment above — "I will re-run the end-to-end arm under a coordinated hold on the host and replace the table when I have a clean one."

One honesty point before the numbers, because it changes what this table is. **This ran on a different host** than the contaminated one. That host had the concurrent workload; keeping the measurement there would have meant negotiating a hold over someone else's evening. Running it on a machine that had no other workload at all was both cleaner and cheaper. So this corroborates the original at a different point in configuration space rather than literally replacing it. For a claim about a mechanism that direction is a strength, not a weakness — but it is not the same run repeated, and I am not going to present it as one.

Same refs, `e09149c07f` → `e74625aafa`, `--full`, ABBA, `e2e_bench` unfiltered.

### The two padding-variance groups

| Bench | Δ point | ABBA bound | order bias ≤ | base ns | head ns |
|---|---:|---|---:|---:|---:|
| `e2e_encode_batch_padding_variance/uniform_100` | **+19.95%** | [+19.53%, +20.36%] | +0.25% | 13,520,725.9 | 16,217,663.6 |
| `e2e_encode_batch_padding_variance/mixed_5_30_100` | **+14.01%** | [+12.63%, +15.37%] | +0.71% | 7,601,563.5 | 8,666,889.3 |

The contaminated run put these at +19.75% and +13.65%. Both land inside the new intervals. The order-bias envelope — the thing that gets large when the machine is drifting under the measurement — is a quarter of a percent on the first group and under a percent on the second, against ABBA intervals that are nowhere near zero.

### The other six groups, which are the part that matters

| Bench | Δ point | bound-lower | bound-upper | order bias ≤ |
|---|---:|---:|---:|---:|
| `e2e_encode/bge_small/long_100tok` | +0.55% | -4.95% | +7.15% | +4.50% |
| `e2e_encode/bge_small/medium_30tok` | -0.78% | -4.77% | +2.96% | +2.79% |
| `e2e_encode/bge_small/short_5tok` | -1.79% | -3.72% | +0.22% | +1.64% |
| `e2e_encode_batch/bge_small/batch_1` | +0.36% | -3.92% | +5.13% | +3.43% |
| `e2e_encode_batch/bge_small/batch_4` | +1.41% | -0.40% | +3.30% | +1.22% |
| `e2e_encode_batch/bge_small/batch_8` | +1.18% | +0.27% | +2.16% | +0.52% |

Every one within the noise band. That split — flat on the short and uniform-batch inputs, badly negative on the two padding-variance inputs — is the same shape reported before, and it is what a cost growing super-linearly in sequence length predicts. Ambient load does not produce a pattern that selects for padding variance.

`simd_dot_product/simd/384` also ran as a control on the embed side: +2.33%, bound `[-0.61%, +5.75%]`, inside the band.

### Run conditions

This is the part the previous table could not support, so it is stated in full rather than summarised.

```
started_utc=2026-08-07T02:42:09Z    finished_utc=2026-08-07T02:57:25Z
os=Darwin 25.4.0 arm64              criterion_mode=full
base_sha=e09149c07f866558e1db4f2429b6bf07038d02d8
head_sha=e74625aafa1558555b62745dd67f093cd336d37a
base_rustc=head_rustc=1.94.1 (e408947bf 2026-03-25)   base_criterion=head_criterion=0.5.1
targets=lattice-inference:e2e_bench, lattice-embed:simd
inference_features=<none>   filters=inference='<all>'   enforcement=report-only
arm order=ABBA (base₁ → head₁ → head₂ → base₂)
lock  bench-window   (/tmp/lion-bench-window.lock):     acquired immediately (uncontended)
lock  Metal GPU      (/tmp/lion-metal-gpu-test.lock):   acquired immediately (uncontended)
ambient [quiet] before first arm:      idle 97.1% (floor 70.0%) ok
ambient [quiet] between order strata:  idle 97.1% (floor 70.0%) ok
ambient [quiet] after final arm:       idle 97.4% (floor 70.0%) ok
machine_state[before first arm]:      power ac, thermal nominal, HID idle 53499.9s, gate passed
machine_state[between order strata]:  power ac, thermal nominal, HID idle 53994.4s, gate passed
machine_state[after final arm]:       power ac, thermal nominal, HID idle 54383.9s, gate passed
criterion_base_samples=200 Flat (4 benchmarks), 200 Linear (4 benchmarks)
criterion_head_samples=200 Flat (4 benchmarks), 200 Linear (4 benchmarks)
```

Both machine-wide advisory locks were uncontended on acquisition, so nothing else on the box was holding a bench or GPU window. The largest ambient consumer at every checkpoint was a wallpaper extension at 6–8%. HID idle above fifteen hours at all three boundaries means nobody touched the machine during or before the run.

The checkpoint limitation from the disclosure comment still applies in principle: three point samples cannot see a burst that lands entirely between them. What is different here is that there is no candidate burst — the box was running nothing else, both locks were free, and the order-bias envelope, which is the in-band instrument for exactly this drift, came in at a quarter of a percent on the worst group. A load asymmetry large enough to manufacture a 20% effect would have shown up there.

### Disposition

The `+19.75%` / `+13.65%` figures stand, and the "directionally right, numerically soft" caveat on them is withdrawn. Both are now supported by an independent run whose conditions are recorded rather than asserted. The sequence-length sweep remains the sharper measurement of the mechanism; this table is what says the mechanism reaches the end-to-end path at a magnitude worth acting on.

ohdearquant:
Status: the resource bound this needed has landed on the branch, but the
performance claim in the title is still unmeasured, so this is not ready to
merge yet.

**What landed** (`b5d07bc`): `checked_head_batch_scratch_len` sizes every
head-batched scratch buffer through two checked multiplications against a
256 MiB ceiling and returns `InferenceError::InvalidInput` instead of
allocating, so an oversized `num_heads * seq_len` is rejected rather than
aborting the process. `multi_head_attention_batched` and
`BertModel::forward_batch` propagate the `Result`. The ceiling is sized to admit
BERT-large at 16 heads and 512 tokens exactly (8192² × 4 B). Verified locally:
fmt and clippy clean, 239 attention tests pass, and the guard is
mutation-sensitive — raising the constant to `usize::MAX` makes the bound test
fail, restoring it makes it pass.

**What still blocks the merge.** The stacked Q·Kᵀ computes the full
`(num_heads * seq_len)²` score matrix and then copies out only the
`num_heads` diagonal blocks. Every off-diagonal block is computed and
discarded, so the change trades a fixed per-call dispatch cost for
`num_heads` times the QK arithmetic and `num_heads` times the scratch memory.
The source comment on the branch says as much and asks for long-`seq_len`
shapes to be re-measured before assuming the trade wins uniformly.

That measurement has not been made. The bench group added here exercises only
the new implementation, so there is no before/after pair at any shape, and the
repository requires a paired comparison for a change to `crates/inference/`
whose premise is speed. The reachable target is the attention group inside
`inference_bench`, not either of the two targets the default comparison runs, so
it needs `--bench inference_bench` with the group filter at both refs.

**Why it is not running now.** The machine reserved for timed runs is offline
(last seen three days ago), and running the comparison here would produce
numbers taken on a machine that is also doing other work — the exact condition
the harness refuses to certify. Merging on the unmeasured side of that is
what the rule exists to prevent, so this waits for the bench host rather than
substituting a weaker measurement.

Next step: run the attention group at the merge base and at this head on the
bench host, at a short shape and at a long one (`seq_len` 512 and above, where
the extra arithmetic is largest), and post the pair here. If the long shape
regresses, the head-batched path needs a length threshold that falls back to
the per-head loop rather than applying unconditionally.

ohdearquant:
Status check before this moves again, so the next pass starts from measurement rather than rediscovery.

**The branch is 27 commits behind `main`, and all three of its files have changed there since the merge base** (`c9b6ade4`, 2026-09-02):

- `crates/inference/benches/inference_bench.rs`
- `crates/inference/src/attention/standard.rs` — changed by the packed-batch attention kernel group (#1482)
- `crates/inference/src/model/bert.rs`

A trial merge (`git merge-tree --write-tree`) names two real conflicts, in the bench file and in `attention/standard.rs`; `bert.rs` auto-merges.

**Consequence for the numbers: the A/B in this PR is void.** Every path it measured moved underneath it, so the comparison no longer has a fixed base — it has to be re-run after the rebase, not carried over. The memory-budget constant and the shape table in the doc comment are arithmetic over shapes and survive; the timing claims do not.

Next pass is therefore: rebase onto current `main`, resolve the two conflicts (the attention one needs reading against #1482 rather than taking either side), then re-measure.

ohdearquant:
Merged `main` in and resolved the two conflicts. Head is now `40a5c0d1`, `MERGEABLE`.

A merge rather than a rebase, since rebasing a shared PR branch needs a force-push.

## Both conflicts were decided by the code, not by preference

**`attention/standard.rs`** — the `bench_support` re-export's return type. The merge left the inner `multi_head_attention_batched` fallible, so the wrapper mirrors it; taking `main`'s `()` would not have compiled.

**`inference_bench.rs`**, three hunks. Two are the same call site, where `main`'s bare `);` cannot stand against a fallible callee. The third is two independently written comments describing the same bench group, and neither is redundant, so they are **merged rather than sided**: `main`'s text is the base because it carries the fact the other lacks — the group is in no default bench workflow, so a comparison that does not name both the target and `bench-internals` is silently empty. Folded back in from this branch: that the group is now the direct caller a coverage audit should count, and the crossover algebra specific to head-batching, which is what makes the sweep's endpoints the right ones.

## The clean auto-merge was checked, not trusted

A clean auto-merge means git found no textual conflict, not that the two edits agree. The scratch budget constant, its guard, and all **five** tests that pin the boundary — including the at-the-cap acceptance and the sum-versus-largest rejection — survive.

## The compile check was made to prove it reaches the resolved code

`cargo check -p lattice-inference --bench inference_bench --features bench-internals` returns 0 — but a cached `Finished` line cannot distinguish a compile from a skip. So: breaking the wrapper's return type gives **rc 101** and `error[E0308]` naming `standard.rs`; restoring the file to its exact pre-mutation digest gives **rc 0** with `Checking lattice-inference` in the output.

`cargo test -p lattice-inference --lib attention::` — **209 passed, 0 failed**, with all five budget tests present by name rather than in the filtered-out count.

## The bench comparison is still outstanding, and it is blocked

Every path this PR measured changed on `main` since the merge base, so the numbers in the description are void and have to be re-run rather than carried over.

That re-run cannot go through the default comparison route as things stand. The harness reaches a non-default bench target only if `BENCHES_INFERENCE` and `CARGO_FEATURES_INFERENCE` are set — both are overridable, and the defaults are `elementwise_cpu_bench` with no features. Filtering for this group against that target is exactly the silently-empty case the merged comment describes. Extending the comparison route to pass those two through is the prerequisite, and it wants an arm proving the empty case fails loudly rather than reading as clean.

ohdearquant (commented):

Verdict on head 40a5c0d10d42b93344ecdb6af8735ca2a192b1a8: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

ohdearquant:
> **Corrected after posting.** Two claims in the first version of this comment were wrong: it said
> this PR's description never disclosed the peak-memory cost, and it presented the cross-call bound
> as an unfiled follow-up. The description already states 310.00 MiB and 167.25 MiB and already
> links #1480. Both are fixed below; the measurements are unchanged.

## Peak attention scratch, and where the raised concern actually lands

A concurrent memory-exhaustion concern was raised against the per-call scratch budget here.
Working it at source produced one confirmation, one refutation, and one item that belongs to
`main` rather than to this branch.

### Confirmed: this branch raises peak per-call scratch by exactly `num_heads`

`main` has no stacked buffers at all — `git show origin/main:crates/inference/src/attention/standard.rs | grep -c '_stacked'`
returns **0**. The per-head loop's dominant buffer is `scores` at `num_heads * seq_len^2`; the
stacked path's is `scores_stacked` at `(num_heads * seq_len)^2`. That is a factor of `num_heads`,
and summing the six live buffers measures it out:

| model | seq_len | `main` peak | this branch | ratio |
|---|---|---|---|---|
| bert-large (16 heads) | 128 | 1.7 MiB | 26.5 MiB | 16.0x |
| bert-large | 256 | 5.4 MiB | 87.0 MiB | 16.0x |
| bert-large | 512 | 19.4 MiB | 310.0 MiB | 16.0x |
| bge-small (12 heads) | 512 | 13.9 MiB | 167.2 MiB | 12.0x |

`seq_len` there is per sequence, not per request: the plan is computed inside the `cu_seqlens`
loop (`let seq_len = end - start`) and the buffers are allocated per iteration, so a packed
request of many sequences peaks at one sequence's plan rather than at their sum.

This is disclosed rather than new. The description states the same 310.00 MiB and 167.25 MiB
figures, calls the trade-off `O(num_heads)` more scratch per sequence, and names #1480 —
"head-batched attention scratch is bounded per call, not across concurrent calls" — as the
tracked work for bounding N concurrent calls. The concern's arithmetic agrees with that table;
what it reports as unresolved is a filed and linked follow-up, deliberately not widened into a
branch whose subject is the kernel.

### Refuted: the serving path's concurrency cap is not unrelated to this budget

The claim was that concurrent callers can exhaust process memory because the limit is per-call,
and that the server's fixed four-job cap is unrelated to it. Two facts in the tree say otherwise:

- `EMBEDDING_MAX_CONCURRENT_JOBS = 4`, and the doc on the semaphore it feeds states the purpose
  in its own words: each outstanding job can already allocate up to `MAX_EMBEDDINGS_TOTAL_TOKENS`
  worth of activation buffers, "so total memory pressure across concurrent requests is bounded by
  (this cap) x (that per-request budget)". Bounding aggregate memory is what it is for.
- The permit is held across the encode rather than released at admission: `run_embedding_encode`
  moves the `OwnedSemaphorePermit` into the `spawn_blocking` closure (`let _permit = permit;`), so
  it drops only when `encode_batch` returns.

The sole production caller of the changed function is `BertModel::forward_batch`, reached from
`BertModel::encode_batch`; everything else that names it is a bench, the bench-only re-export, or
a test. `encode_batch` has exactly two consumers: that bounded route, and the library type
`NativeEmbeddingService`, which has no admission control and is the case #1480 covers.

The other embeddings route, the one #1397 correctly reports as having no admission cap, does not
reach this code: it serves a Qwen VL embedder, and the head-batched path is BERT-only. So the
open unbounded route and this change do not intersect.

What the four-job cap does deserve on the back of this change: its arithmetic is
`(cap) x (per-request budget)`, and this branch moves the second factor by `num_heads`, taking
four concurrent jobs from roughly 78 MiB of attention scratch to roughly 1.21 GiB. Bounded and
survivable, but the constant was chosen against the smaller number. Noted on #1480.

### Out of subject

The note about 14 `prepared_*` module declarations and roughly 20k dormant lines is accurate
about the file, but all 14 are already on `main`: this branch's `bert.rs` delta is 4 insertions
and 4 deletions and adds none of them. `git diff --name-only origin/main HEAD` is three files.

### State

Unchanged. Still blocked on the A/B, for the reason given earlier: the comparison route passes
group filters only, so it cannot name this bench target or its feature, and filtering for this
group against the default target is exactly the silently-empty case.

ohdearquant:
Progress on the A/B blocker, and a second thing it exposed.

**The driver gap is fixed and in review.** The lane that runs PR-time A/Bs exported the two group
filters and nothing else, so every run reached `bench-compare`'s default targets with no features.
Filtering for this PR's group against `elementwise_cpu_bench` was therefore a no-op. The harness
itself was never the problem — `BENCHES_INFERENCE`, `CARGO_FEATURES_INFERENCE` and their embed
counterparts are honored at `origin/main` (`scripts/lib/bench-compare-impl.sh:461-464`); only the
caller failed to pass them.

**One claim of mine there was wrong and is worth retracting.** I wrote earlier that this "wants an
arm proving the empty case fails loudly rather than reading as clean". That arm already exists:
`require_measured` refuses on `rc != 0` **and** on "exited 0 but produced no measurements", with a
comment naming exactly this case ("a filter that matches nothing is the ordinary way to get here"),
`run_bench` always sets the line count so the guard cannot be skipped by an unset value, and
`tests/test_bench_compare_measurement.py` carries
`test_enforcing_mode_refuses_a_run_that_measured_nothing` and its default-mode twin. I asserted a
gap without reading the guard.

**#1513, found while checking whether the A/B would even be gate-eligible.** `run_target_gate`
makes an uncalibrated target informational only when the target is an embed target; the other
branch runs only under `--quick`. So at full resolution a non-default `lattice-inference:` target
votes on thresholds nobody calibrated for it. That was unreachable while the driver could not
select an inference target, and becomes reachable the moment it can.

**What that means for this PR's numbers.** The honest sequence is a same-configuration A/A on
`inference_bench` with `bench-internals` to establish this target's own noise floor, then the A/B
read against it. Until #1513 is settled I will treat whatever the gate says about this target as
informational regardless of how it votes, which is what ADR-087 D3/D5 asks for anyway.

ohdearquant:
> **Correction, posted about ten minutes after this comment and leading it because it changes what option 2 means.** The last section originally said the scratch-budget work "should not be held hostage to the stacking result," and listed the budget plan, its five boundary tests and the parity test as things that stand on their own if the stacking is withdrawn. That is wrong, and checking the code rather than my memory of it is what showed it.
>
> `head_batch_scratch_plan` sums six buffers and **four of them — `q_stacked`, `k_stacked`, `scores_stacked`, `context_stacked` — exist only in the stacked path.** The `(num_heads*seq_len)^2` term that produces the 310.00 MiB figure and motivates the 384 MiB ceiling is `scores_stacked`. Remove the stacking and the budget is guarding buffers that are not allocated. The parity test likewise compares the stacked path against the per-head reference, so it has no subject either.
>
> What `main`'s per-head path actually allocates per sequence is `scores` at `num_heads*seq_len^2` plus `v_all_t` at `hidden_size*seq_len` — about **18 MiB** at BERT-large's 512-token shape, against 310 MiB for the stacked form. Still quadratic in a caller-controlled value, so a bound there is arguable, but it is a different claim by a factor of seventeen and it should be argued on its own rather than carried over on this PR's evidence.
>
> The measurement below is unaffected. The correction is entirely about what would remain if the stacking is withdrawn.

## The A/B this PR has been waiting for now exists, and it is a blocking result

The body's coverage statement said `attention_batched_kernel` "has no base-side counterpart, so no A/B exists for it," and named the precondition: *landing the bench group on the base first is what would make that A/B possible.* That precondition is met — the group is on `main` now (`inference_bench.rs:1040`, calling `multi_head_attention_batched` at `:1062` and `:1106`), which is why merging `main` in produced a comment conflict on it. So the changed function can finally be measured directly, and it has been.

**Head-batching more than doubles this kernel's time at long sequences.**

| bench | Δ (head vs base) | swapped arms | A/A floor |
|---|---:|---:|---:|
| `seq_len/16` | -4.62% | +3.34% | ±1.1% |
| `seq_len/32` | -2.67% | +0.56% | ±1.1% |
| `seq_len/64` | **+11.89%** | -10.81% | ±1.1% |
| `seq_len/128` | **+31.04%** | -23.43% | ±1.1% |
| `seq_len/256` | **+82.15%** | -45.07% | ±1.1% |
| `seq_len/384` | **+102.81%** | -50.24% | ±1.1% |
| `seq_len/512` | **+116.25%** | -53.57% | ±1.1% |
| `batch_size/1,4,8,16` (at seq 32) | -1.45%, -1.48%, -0.28%, -2.01% | +1.41%, -0.38%, -0.12%, +1.26% | ±1.1% |

Positive = the stacked path is slower.

### Why this is a code effect and not measurement noise

Three runs on a dedicated quiet bench host, same target, same feature set, same filter, quick resolution, ABBA order in every one.

1. **A/A** (`main` vs `main`) establishes the floor for this instrument, which had never been characterized: every row within **1.1%**, order bias at most **1.60%**. That is the resolution available here, and the long-sequence deltas above are 70-100× it.
2. **Forward A/B**, run twice — the first arm of the first run overlapped an unrelated multi-core compile on the host for 23 s, so it was discarded and re-run clean rather than kept. The two runs agree: +115.34% and +116.25% at `seq_len/512`.
3. **Swapped arms** — the PR head in the base position, `main` in the head position. The sign flips on every long-sequence row, and the magnitudes are the exact reciprocals: `1/(1 + 1.1625) = -53.76%` predicted against **-53.57%** measured at 512; -50.69% vs -50.24% at 384; -45.10% vs -45.07% at 256. A phase-order artifact keeps its sign when the arms are reversed. This inverts, so the delta tracks the code.

The wide `order bias ≤` column on the long-sequence rows in the forward run is a consequence of the effect being large, not evidence of an unquiet machine: the same machine, filter and phase structure produced 1.60% bias in the A/A five minutes earlier. Ambient idle was 98%+ in every phase of every run, with all machine-state checkpoints passing.

### What it means for the change

The trade-off paragraph predicted the shape of this exactly — *O(num_heads) more FLOPs and scratch per sequence in exchange for `2×num_heads → 2` dispatches*, favorable at short sequences — and the measurement locates the crossover between **seq 32 and seq 64**. Below it the change is worth about 1-2%, which is barely outside the A/A floor. Above it the extra arithmetic dominates immediately and does not stop growing.

The batch_size sweep runs at a fixed `seq_len` of 32, i.e. right at the crossover, which is why those four rows are flat. They are not evidence that the many-sequence serving shape is safe; they are a measurement at the one sequence length where the trade is neutral.

So at bge-small's 512-token ceiling this costs 2.16× the kernel time, on top of the `num_heads`× scratch the budget section already documents (310.00 MiB for BERT-large at 512, 167.25 MiB for bge-small). Both costs point the same way at the same shapes.

### Disposition

Blocking, and it is my own PR's premise that the measurement refutes. I am not merging this. The options as I see them:

1. **Gate the stacked path on sequence length**, using the crossover this run locates, so short packed sequences take the batched form and anything above ~32-64 keeps the per-head loop. That preserves the measured win where it exists and removes the regression everywhere else, at the cost of two live paths.
2. **Withdraw the stacking.** Almost nothing else in this PR survives that, and the correction above says why: the budget, its five boundary tests and the parity test all take the stacked path as their subject. What is left is the bench group, which is on `main` already and is what made this measurable.

I lean to (2) for this PR and (1) as a separate, measured change, because the win being protected is 1-2% at one sequence length and the machinery to protect it is a permanent fork in the kernel.

Raw artifacts and the A/A characterization are retained; the numbers above are quoted from them rather than retyped from a prior revision of this description.

ohdearquant:
Closing on the measurement above rather than parking it as a draft.

With the stacking withdrawn there is no diff left. `attention_batched_kernel` is already on `main`; `standard.rs` and the `bert.rs` fallible-signature plumbing exist only to carry the stacked path, and as the correction on the previous comment records, four of the scratch budget's six buffers and the parity test's entire subject go with it. A branch reverted to nothing is not a PR.

What this leaves behind is not nothing:

- **The finding**, on #702, with the crossover located between `seq_len` 32 and 64 and the reciprocal-under-swap check that makes it a code effect rather than an ordering artifact. The dispatch tax the issue identified is real; stacking is not the way to recover it, and a strided-batch GEMM is the option whose payoff does not shrink as sequences grow.
- **The bench group**, on `main` — the only declared bench target that calls `multi_head_attention_batched` directly. Its absence on the base side is exactly why three earlier A/B attempts here could not say anything about the changed function, and why this one could.
- **#1480** stays open on its own terms; it was always about bounding concurrent calls rather than one call, and it does not depend on this implementation.
- **#1515**, filed out of this run: the bench driver certifies machine quiet from three phase-boundary samples, so a load living entirely between two of them reads as perfectly idle. That is how a 23-second compile inside one of these runs went unrecorded.

Reopen-able if the sequence-length-gated variant is worth pursuing, but that wants its own PR and its own justification — 1-2% at one sequence length, against a permanent second path through the kernel, is a trade to argue on its own evidence rather than to inherit from this branch.

PR #1350, opened 2026-08-07
fix(serve): restore three vision error-taxonomy distinctions and make a vision-weight load failure retryable

### LINKED ISSUE #1335: serve: preserve structured image errors and vision-load retry semantics
The inline-image capability is present on main, but the surviving serving path does not carry three API distinctions that an earlier implementation had.

1. Vision-weight load failures are stored as a permanent failed state and surface as a generic `500 internal_error`. Return a client-safe `400 vision_load_failed` that does not expose checkpoint paths, and keep a failed lazy load retryable unless the runtime has an explicit permanent-failure policy.
2. Image preprocessing maps dimension-budget failures and malformed images to the same `invalid_image` code. Give the serving patch and dimension budget a dedicated `image_dimensions_exceeded` code.
3. An image combined with `response_format.json_schema` or a nonzero `reasoning_budget` is rejected through the generic unsupported-feature path. Return a stable `image_unsupported_combination` code, and keep the rejection ahead of worker dispatch.

Implement this in the shared contract and worker path so the two serving binaries cannot drift. Add tests for the three codes, assert that client-visible messages contain no model or checkpoint paths, and assert that a transient vision-load failure can be retried.

ohdearquant:
Restores three distinctions on the image-serving path that the current error taxonomy collapses,
and makes a vision-weight load failure retryable instead of permanently disabling image support
for the process.

Closes #1335.

All three changes land in the shared contract and worker path, so both server binaries pick them
up: `serve/contract.rs`'s `normalize_request_inner`, and the `MetalWorker` that both binaries
spawn and submit to. Neither binary carries its own copy of the vision error mapping.

## 1. A vision-weight load failure no longer poisons the process

`VisionRuntime::get_or_load` cached a load failure in `VisionState::Failed(message)` for the
lifetime of the process and permanently cleared the `vision_supported` flag both binaries read
when admitting image requests. One transient failure therefore rejected every subsequent image
request, and the failing request itself surfaced as a generic 500 `internal_error`.

`VisionState::Failed` is gone. On a load error `get_or_load` returns `Err(detail)` for that one
call and leaves the state `Pending` with `vision_supported` untouched, so the next admitted image
request retries the load. The call site maps it to a 400 `vision_load_failed` with a fixed,
path-free client message; the underlying detail, which routinely names the checkpoint directory,
is logged server-side only.

On whether a load failure should ever be permanent: a search for an existing permanent-failure
policy for this runtime found none, so none was invented. The lazy load is unconditionally
retryable.

## 2. `image_dimensions_exceeded` split out of `invalid_image`

Every failure out of `preprocess_qwen35_image_for_serve` was flattened into `invalid_image` — a
genuinely malformed payload and a request that merely exceeds a serving budget were
indistinguishable to a client deciding whether to resize and retry.

A new `VisionError::DimensionsExceeded` now covers the three serving budgets: the 2048px hard cap
per side, the pre-merge patch-count budget, and the preprocessed-byte budget. The hard-cap case is
distinguished from an unparsable header by matching `image::ImageError::Limits(_)` specifically,
which `into_dimensions()` returns only when the declared size exceeds the configured limits.

The "dimensions not a multiple of `patch_size * spatial_merge_size`" check deliberately stays
`InvalidConfig`. It is a structural resize-scope limitation shared with the non-serving embedding
path, not a serving budget.

## 3. `image_unsupported_combination` split out of `unsupported_feature`

An image combined with `response_format.json_schema` or a nonzero `reasoning_budget` reported the
generic `unsupported_feature`, the same code an unrelated unimplemented feature would produce.
Both checks already ran inside `normalize_request_inner`, ahead of worker submission in both
binaries, so only the code changed.

The pre-existing `image` + `stream` and `image` + `logprobs` checks keep `unsupported_feature`;
the issue names only the two above.

## Consumers

`ApiError::BadRequest`'s `IntoResponse` maps every `code` string to the same envelope and HTTP 400
with no per-code match, so a new code cannot land in an unhandled arm at the HTTP layer.
`serve/metrics.rs`'s `record_error` takes an arbitrary string. No client binding or JSON schema in
the repository enumerates these codes. `docs/serve-http-api.md` documents them in prose and is
updated to name all three new codes and which old code each one narrows. Existing test fixtures
that hard-coded the old code for a now-split case were updated rather than left beside the new
expectation.

`docs/releases/v0.8.0.md` records the original v0.8.0 code set as a historical release note and is
deliberately unchanged.

## Tests

Five, four new and one rewritten:

- an oversized image at the hard cap produces `DimensionsExceeded` naming 2048px, deterministically
  regardless of the patch-count environment override;
- the same image through the real `build_vision_request` produces `image_dimensions_exceeded` and
  status 400;
- a control arm: `b"not an image"` still produces `invalid_image` and 400;
- a checkpoint that passes the capability preflight but carries an undecodable tensor produces
  `vision_load_failed` and 400, with a positive assertion that the message does not contain the
  test's real temporary path, and `is_supported()` still true afterwards;
- the existing capability test, which previously asserted the permanent-failure behaviour, now
  asserts the failure stays retryable, repairs the checkpoint in place with genuinely valid q4
  tensors for all 21 expected vision weights, retries, and asserts the load reaches `Ready`.
  Reaching `Ready` rather than a second error is what proves the retry re-attempted the load
  instead of replaying a cached failure.

The path-leak assertion exists only where a leak was possible. The other two codes build their
messages from image and request field values and never touch a filesystem path, so an equivalent
check there would assert that a static literal contains no path, which reading the code already
settles.

## Verification

Scoped `lattice-inference` tests covering every module this diff touches: green.

The unscoped `--features metal-gpu` run was **not** completed and no result is claimed for it. The
host was running several concurrent `cargo test` and `cargo build` invocations across other
worktrees for the duration; one attempt exceeded 45 minutes and was killed. Its partial output
showed three failures, all in `forward::metal_qwen35::inner::tests` (MRoPE logit injection, f16 KV
Metal dispatch), a module this diff does not touch, consistent with real-GPU contention from the
concurrent processes. That was not chased down; it is outside an error-taxonomy change.

## bench-compare disposition

Structural waiver, no A/B table.

Population searched: all 24 bench targets declared in `crates/inference/Cargo.toml`, including
those gated behind `required-features` (`f16`, `bench-internals`, `metal-gpu`). None reaches the
changed code. The diff touches the HTTP request-normalization and vision-preprocessing error
paths — `serve/contract.rs`, `serve/metal_worker.rs`, `vision/mod.rs`, `vision/qwen35_vit.rs` —
and no declared bench group constructs a serve request or drives image preprocessing.

Build inputs are identical between base and head: no manifest, dependency, lockfile, feature,
profile, build-script, generated-source, bench-target-definition or harness change.

Residual risk: this change is unmeasured rather than measured-safe. The error paths it edits are
by construction the paths a successful request does not take, so the throughput risk is low, but
no bench in this repository would show it either way.

### DIFF
--- a/crates/inference/src/serve/contract.rs
+++ b/crates/inference/src/serve/contract.rs
@@ -529,7 +529,9 @@ fn normalize_request_inner<C>(
             .as_ref()
             .is_some_and(|format| format.r#type == "json_schema")
     {
-        unsupported("json_schema response format is not supported for image requests")?;
+        image_unsupported_combination(
+            "json_schema response format is not supported for image requests",
+        )?;
     }
     // Resolved ahead of `check_context` (rather than in its pre-refactor spot
     // after `check_context`/`stop`) so the shared full-window formula
@@ -550,7 +552,7 @@ fn normalize_request_inner<C>(
             .filter(|&value| value > 0);
     }
     if has_image && reasoning_budget.is_some() {
-        unsupported("reasoning_budget is not supported for image requests")?;
+        image_unsupported_combination("reasoning_budget is not supported for image requests")?;
     }
 
     let context = check_context(&messages, max_tokens, reasoning_budget)?;
@@ -822,6 +824,20 @@ fn unsupported(message: impl Into<String>) -> Result<(), ApiError> {
     Err(ApiError::BadRequest {
         message: message.into(),
         code: "unsupported_feature",
+    })
+}
+
+/// Rejects an image combined with a specific incompatible request feature
+/// (`response_format.json_schema`, a nonzero `reasoning_budget`) with a
+/// stable, distinct code from the generic [`unsupported`] path, so a client
+/// can branch on "this combination is unsupported" without string-matching
+/// the message. Runs from [`normalize_request_inner`], ahead of worker
+/// dispatch on both server binaries -- a request that cannot be served never
+/// reaches the shared Metal worker.
+fn image_unsupported_combination(message: impl Into<String>) -> Result<(), ApiError> {
+    Err(ApiError::BadRequest {
+        message: message.into(),
+        code: "image_unsupported_combination",
     })
 }
 
@@ -1376,35 +1392,44 @@ mod tests {
             "unsupported_feature"
         );
 
+        // Pinned to the exact `BadRequest` variant (not just `.code()`), since
+        // that variant is what `ApiError`'s `IntoResponse` impl maps to HTTP
+        // 400 (`serve/mod.rs`, exercised there and in each binary's own
+        // envelope tests) -- this is the proof that `image_unsupported_combination`
+        // is a 400, not merely that some error carries that code string.
         let mut req = inline_image_request("user", &uri);
         req.reasoning_budget =
             Some(serde_json::value::RawValue::from_string("4".to_string()).unwrap());
-        assert_eq!(
+        assert!(matches!(
             normalize_request(
                 &req,
                 defaults(),
                 ServeProfile::lattice_serve("model", 32).with_vision_support(true),
             )
-            .unwrap_err()
-            .code(),
-            "unsupported_feature"
-        );
+            .unwrap_err(),
+            ApiError::BadRequest {
+                code: "image_unsupported_combination",
+                ..
+            }
+        ));
 
         let mut req = inline_image_request("user", &uri);
         req.response_format = Some(ResponseFormat {
             r#type: "json_schema".to_string(),
             json_schema: None,
         });
-        assert_eq!(
+        assert!(matches!(
             normalize_request(
                 &req,
                 defaults(),
                 ServeProfile::lattice_serve("model", 32).with_vision_support(true),
             )
-            .unwrap_err()
-            .code(),
-            "unsupported_feature"
-        );
+            .unwrap_err(),
+            ApiError::BadRequest {
+                code: "image_unsupported_combination",
+                ..
+            }
+        ));
     }
 
     #[test]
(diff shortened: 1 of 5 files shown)


ohdearquant:
## This PR requires a 0.9.0 release, and the semver gate is right to say so

`semver-checks` failed at `96879753ef`, and the failure is real rather than a gate artefact:

```
Checked  196 checks: 195 pass, 1 fail, 0 warn, 58 skip
Summary  semver requires new major version: 1 major and 0 minor checks failed
--- failure enum_variant_added: enum variant added on exhaustive enum ---
Failed in:
  variant VisionError:DimensionsExceeded in crates/inference/src/vision/mod.rs:117
```

Note the executed-check count: 196 checks actually ran, so this is a gate that measured something, not one that skipped everything and reported success.

`VisionError` is public through `pub mod vision` and was exhaustive, so distinguishing a serving budget from a malformed payload — the entire point of #1335 — is a breaking change under semver. Under 0.x that means `0.8.0` → `0.9.0`.

There is no version of this change that avoids the break. Marking the enum `#[non_exhaustive]` instead is itself a major change (`enum_marked_non_exhaustive`), and reusing an existing variant to carry the distinction would defeat the issue.

## What I changed in response

Marked `VisionError` `#[non_exhaustive]` at `9aa4ed2791`. That does not make this PR pass — it is a second major change in the same release — but it means the cost is paid once. Every future failure-mode distinction on this type becomes a minor change instead of another release-blocking break. No crate in the workspace matches on `VisionError`, so nothing needed a wildcard arm; `cargo build -p lattice-inference` is clean.

## What this needs from a human

A release decision, which is not this PR's to make. Either:

- cut `0.9.0` and land this and any other pending breaks together, or
- hold this PR until the next planned minor bump and land it in that batch.

Blocking on that. The code side is done and the rest of the checks were green.

## The generalisable part

This is worth recording independently of #1335. Every public error enum in this crate that is exhaustive will block its own future refinement the same way, and the refinement it blocks is exactly the useful kind: splitting one over-broad error into two that a client can act on differently. The exhaustive default converts "make this error more informative" into "cut a release". Auditing the crate's public enums for this and marking the ones that should stay open would be cheaper than discovering each one at a gate failure.

PR #1366, opened 2026-08-07
refactor(lora): extract shared adapter descriptor and blend validation into lattice-fann

### LINKED ISSUE #615: arch(lora): extract shared adapter descriptor + blend validation — tune and Metal inference maintain drifting copies

ohdearquant:
Closes #615.

## What changed

`lattice-fann` gains a `lora` module holding the LoRA adapter descriptor and the blend-validation logic that `lattice-tune` and the Metal inference path had each been carrying their own copy of. Both now call the shared module.

`lattice-fann` is the right home rather than `lattice-inference`: it is the only leaf crate `lattice-tune` depends on unconditionally, whereas the `tune -> inference` edge is behind the `inference-hook` / `train-backward` features and is off in a default build. A module owned by `inference` would have been unreachable from tune's default build, so the dependency direction picks the location here.

## The duplication that existed, and the one disagreement

The caps were genuinely duplicated but did not disagree: `MAX_BLEND_RANK_TOTAL = 4096` and `MAX_BLEND_TOTAL_ELEMENTS = 1 << 30` were identical in `tune`'s `blend_lora_adapters` / `blend_layer_entries` and in inference's Metal-gated `blend_lora_layer_data`. Every existing error message is preserved byte-for-byte.

One thing did disagree, and it is a behavior change worth reading closely. `chat_metal.rs` computed the scale locally as:

```rust
let scale = if rank > 0 { alpha / rank as f32 } else { 1.0 };
```

so a zero rank produced a scale of `1.0`. `tune`'s documented and tested `LoraConfig::scale()` produces `0.0` for the same input. The shared helper adopts tune's value:

```rust
pub fn effective_scale(rank: usize, alpha: f32) -> f32 {
    let scale = if rank == 0 { 0.0 } else { alpha / rank as f32 };
    ...
}
```

Zero rank means an empty factorization, which contributes nothing, so `0.0` is the defensible value and `1.0` was the outlier. The helper also collapses a non-finite `alpha` or a non-finite resulting scale to `0.0` rather than letting NaN or Inf propagate into the forward pass. This is called out rather than folded in silently because it is a real change to what `chat_metal` computes on that input, not a pure refactor.

## Verification

Mutation arm: weakening the rank cap in the shared module reddens exactly the predicted tests in `tune` and in the shared module, and leaves the neighbouring assertions green — including `blend_aggregate_budget_exceeded_returns_err`, which stays passing while `blend_rank_exceeds_cap_returns_error` fails. The mutation was restored and the tree verified clean afterwards.

The inference-side Metal-gated tests call the identical mutated function but were **not executed** this round: the machine-wide GPU test lock was held by another job for the duration. That is a gap in the evidence, not a claim that those tests pass. The shared function they call is the one the mutation arm did cover.

Workspace builds clean on default features; `cargo fmt` and clippy clean.

## bench-compare disposition

Structural waiver, no A/B table, and the population is stated because the two targets `bench-compare` runs by default are not the population.

**Corrected.** An earlier revision of this section searched all 24 declared `[[bench]]` targets in `crates/inference/Cargo.toml` plus the 1 in `crates/fann`, found no reference to `blend_lora`, `effective_scale`, `LoraDescriptor`, or `lora_mixture`, and concluded no measurement of the changed path exists. That search was run with a positive control and did reach every file it enumerated — but it enumerated the wrong population.

`crates/inference/Cargo.toml:159` declares a `[[bin]]`, not a `[[bench]]`:

```toml
[[bin]]
name = "bench_lora_mixture"
path = "src/bin/bench_lora_mixture.rs"
required-features = ["f16", "metal-gpu", "bench-internals"]
```

It calls `blend_lora_layer_data(&refs)` at `src/bin/bench_lora_mixture.rs:141` and `:148` — the function this PR changes. It is a bin rather than a bench because it needs a real Metal device and a model directory, which a Criterion harness cannot supply. So a measurement of the changed path does exist, and enumerating by declaration kind stepped past it.

The manifest change gets its own paragraph, because a build-input change is exactly the case where a structural waiver normally does not apply. `metal-gpu` now carries `dep:lattice-fann`, since `blend_lora_layer_data` and `chat_metal` call the shared module whenever that feature is on rather than only under `mixture`. `metal-gpu` is not a default feature (`default = ["std", "download", "serve"]`), so this adds no dependency to the default bench build and cannot change what those binaries compile. Under a `metal-gpu` bench build it does add a dependency, and no declared bench target reaches the changed code there either.

Residual risk, restated correctly: the LoRA blend path is **not measured by this round**, which is a different and weaker claim than the earlier "no target in the repository can notice". `bench_lora_mixture` reaches the changed function and has not been run against this branch, so the performance of the blend path is unknown here rather than unmeasurable. Treat that as an owed measurement, not as an absence of one.

## Why a test file is in this diff

`metal_measurement_lock_contract.rs` records a `recorded_position` for each construction exemption, and the shared-helper change adds three lines above the two `MetalQwen35State` construction sites in `chat_metal.rs::run`. CI failed with the guard's own wording: *"recorded a position that no longer matches (not a locking violation)"*, `770:39 -> 773:39` and `795:39 -> 798:39`.

Both call sites are byte-identical to their previous form (compared against `origin/main` before touching the numbers, because "move the numbers until it goes green" is exactly how a real violation would hide here). Only the recorded positions move. Locally: `cargo test -p lattice-inference --test metal_measurement_lock_contract` → **53 passed, 0 failed**; the pre-fix state failing in CI is the paired negative arm.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -248,6 +248,31 @@ jobs:
           cache-on-failure: true
       - name: tune — safetensors + inference-hook + serde (compile tests)
         run: cargo test -p lattice-tune --features safetensors,inference-hook,serde --no-run
+      # `LoraConfig::validate` (the check `LoraAdapter::new` runs on every
+      # construction path) deliberately does not enforce a known-module-name
+      # allowlist — the adapter representation is model-neutral. The only
+      # remaining enforcement of unknown/wrong-architecture target-module
+      # names is `LoraAdapter::validate_against`/`validate_against_bert`,
+      # gated behind `inference-hook` like the rest of this step's compile
+      # target above, so the `--no-run` step only ever type-checks them.
+      # Build AND run this specific surface, same as the `mixture` and
+      # `simulated-teacher` steps below, so the fail-closed rejection tests
+      # cannot rot silently.
+      - name: tune — LoRA validate_against rejection tests (build + run)
+        run: |
+          log="${RUNNER_TEMP:-/tmp}/lora-validate-against-tests.log"
+          if cargo test -p lattice-tune --lib --features inference-hook \
+            validate_against >"$log" 2>&1
+          then
+            cat "$log"
+          else
+            status=$?
+            cat "$log"
+            exit "$status"
+          fi
+          grep -q 'test lora::tests::validate_against_tests::test_validate_against_unknown_module_errors ... ok' "$log"
+          grep -q 'test lora::tests::validate_against_bert_tests::test_validate_against_bert_unknown_module_rejected ... ok' "$log"
+          grep -Eq '^test result: ok\. [1-9][0-9]* passed; 0 failed; 0 ignored; 0 measured; [0-9]+ filtered out;' "$log"
       - name: tune — train-backward (compile tests)
         run: cargo test -p lattice-tune --features train-backward --no-run
       # The `mixture` feature gates the experimental online router-refit module
--- a/crates/fann/src/lib.rs
+++ b/crates/fann/src/lib.rs
@@ -14,6 +14,7 @@
 mod activation;
 mod error;
 mod layer;
+pub mod lora;
 mod network;
 pub mod training;
 
@@ -23,6 +24,7 @@ pub mod gpu;
 pub use activation::Activation;
 pub use error::{FannError, FannResult};
 pub use layer::Layer;
+pub use lora::LoraDescriptor;
 pub use network::{Network, NetworkBuilder};
 pub use training::{
     BackpropTrainer, GradientGuardStrategy, Trainer, TrainingConfig, TrainingResult,
--- a/crates/fann/src/lora.rs
+++ b/crates/fann/src/lora.rs
@@ -54,6 +54,58 @@ impl LoraDescriptor {
     }
 }
 
+/// Recognized LoRA target-module names across every architecture this
+/// project trains or serves adapters for: full-attention (GQA) `q_proj`,
+/// `k_proj`, `v_proj`, `o_proj`; linear-attention (GDN) `in_proj_qkv`,
+/// `in_proj_z`, `in_proj_b`, `in_proj_a`, `out_proj`; MLP `gate_proj`,
+/// `up_proj`, `down_proj`; BERT `query`, `key`, `value`, `attn_output`,
+/// `ffn_intermediate`, `ffn_output`.
+///
+/// This is a flat name allowlist, not an architecture-aware shape check —
+/// whether a given module is valid for a *specific* layer's type (e.g. a GDN
+/// module on a full-attention layer) is `qwen35_projection_shape`'s job in
+/// `lattice-inference`, which both `lattice-tune` and the Metal load path
+/// already call. This list exists so a descriptor's declared
+/// `target_modules` can be checked for typos or unrecognized names before
+/// any model architecture is known, in the one leaf crate both consumers share.
+pub const KNOWN_LORA_TARGET_MODULES: &[&str] = &[
+    "q_proj",
+    "k_proj",
+    "v_proj",
+    "o_proj",
+    "in_proj_qkv",
+    "in_proj_z",
+    "in_proj_b",
+    "in_proj_a",
+    "out_proj",
+    "gate_proj",
+    "up_proj",
+    "down_proj",
+    "query",
+    "key",
+    "value",
+    "attn_output",
+    "ffn_intermediate",
+    "ffn_output",
+];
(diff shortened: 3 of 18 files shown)


ohdearquant:
Four items to settle at `d4d88e760f3baee0c6b7a19efd3121eaa2fb21fa`. The first is the one the branch exists for.

**1. The shared descriptor has no production consumer.**

`crates/fann/src/lora.rs:29-37` defines `LoraDescriptor`, and a whole-tree search across `crates/fann/src`, `crates/tune/src` and `crates/inference/src` finds it only in its own definition, its export, and its tests. Both consumers still carry their own shapes: tune constructs `LoraConfig` without `dtype` (`crates/tune/src/lora/mod.rs:48-57`), and the Metal path consumes `LoraLayerData` without alpha, target modules, or dtype (`crates/inference/src/forward/metal_qwen35.rs:1691-1707`). Target-module validation is duplicated and divergent between tune's opt-in `validate_modules` (`crates/tune/src/lora/mod.rs:266-275`) and Metal's `expected_lora_shape` (`crates/inference/src/forward/metal_qwen35.rs:3693-3714`), while the shared `validate` (`crates/fann/src/lora.rs:51-54`) checks only alpha/scale finiteness.

So a public type introduced to unify two consumers is currently used by neither. Whatever shape the fix takes, it wants one cross-crate test that fails if either consumer stops carrying the shared descriptor — otherwise the next change re-forks the types silently.

**2. An empty inner adapter passes blend admission and can erase a loaded adapter.**

With `adapter_weights = &[(&[] as &[LoraLayerData], 1.0)]`: the outer-slice guard at `crates/inference/src/forward/metal_qwen35.rs:1755-1759` passes, the loop at `:1768-1774` leaves `grouped` empty, and the function returns `Ok(Vec::new())` at `:1813-1814` and `:1905-1907`. `generate_with_lora_mixture` unloads the currently loaded adapter before the blend (`:3991-4005`), and `load_lora_adapter` then rejects empty layers (`:3757-3759`). The request fails after the previous adapter is already gone, which contradicts the admission-error promise stated at `:3971-3976`.

The broad path predates this branch, but it sits inside the blend-validation surface this change claims, so it is worth closing here: reject an empty inner slice before any state mutation, and test adapter preservation across a rejected request separately from the blend itself.

**3. The rank-zero test observes the helper, not the changed call site.**

The production change is at `crates/inference/src/bin/chat_metal.rs:386-394`, but `cm_chat_metal_zero_rank_scale_matches_tune_not_the_old_local_fallback` (`:1348-1351`) calls `lattice_fann::lora::effective_scale` directly and never reaches `load_lora_safetensors`. Restoring the old `if rank > 0 { alpha / rank as f32 } else { 1.0 }` at `:394` leaves the test body unchanged and passing, so it does not guard the fix. Either extract the loader's scale computation into a function the test can drive through the loader path, or assert an observed returned scale from a rank-zero metadata fixture.

**4. The shared error formatter overflows on 32-bit targets.**

`crates/fann/src/lora.rs:210-218` evaluates `MAX_BLEND_TOTAL_ELEMENTS * 4` with the constant at `1 << 30`; under `rustc --target wasm32-unknown-unknown` the equivalent expression fails with `arithmetic operation will overflow`. Any aggregate above the cap reaches it. Doing the GiB arithmetic in `u64`, or dividing before multiplying, fixes it; since `fann` is a public leaf crate it is worth a check that would catch the regression rather than only correcting the expression.

CI is green at this head (37 newest-per-name checks, 36 success, 1 skipped — `build macOS Studio Swift release`, which never ran and is not evidence for a Rust change). The changed Metal blend unit module is under `#[cfg(test)]` at `crates/inference/src/forward/metal_qwen35.rs:34117` and no execution marker for it appears in the exact-head feature-matrix log, so that surface is unmeasured rather than passing.

Issue #1336, opened 2026-08-06
serve: add vision preload control and skip redundant multimodal head work


ohdearquant:
The image-serving path always loads vision weights lazily, and runs the terminal RMSNorm, lm-head, and readback tail for every multimodal prefill position. Two capabilities are worth recovering.

1. An operator-controlled `--preload-vision` option. Lazy stays the default so text-only startup cost and memory are unchanged; the option loads vision weights at startup for predictable first-image latency, warns and falls back to lazy loading if preload fails, and documents the startup and resident-memory tradeoff.
2. An `emit_head`-style flag threaded through the Metal multimodal prefill path. Only the final prefill position's logits are sampled, so non-final positions can advance KV and GDN state without terminal head work or vocabulary readback. Text-only behavior must be unchanged, and final-logit and state equivalence must be proven against the always-emitting reference path.

Add CLI, fallback, and equivalence tests. Capture a before/after measurement for the multimodal prefill path; no performance claim without it.

ohdearquant:
> _Comment authored by Claude (Anthropic agent) on behalf of @ohdearquant._

Both parts of this are on main (#1367):

- `--preload-vision` on `lattice serve` and `lattice_serve`. Lazy loading stays the default, and a failed eager load warns and falls back to lazy. The CLI and fallback contract have tests.
- The `emit_head` skip on non-final multimodal prefill positions. `generate_multimodal_vision_emit_head_skip_matches_always_emit_reference` checks it against the always-emitting path: final logits, GDN state, KV, and the following decode step. That test runs in the checkpoint-backed parity workflow.

The measurement comes from the `bench_vision_prefill_ab` harness, recorded on #1561. Setup: bench host, ABBA order, Qwen3.5-0.8B, 1024 visual rows, medians of 7. Decoder multimodal prefill takes about 15.04 s when every position emits the head and about 10.22 s when only the final position does, roughly 32% less. The harness excludes ViT and merger cost by design.

PR #1367, opened 2026-08-07
feat(inference): vision weight preload control and skip redundant multimodal head work

ohdearquant:
Vision prefill runs many positions before the first token is emitted, and only the last of them needs the terminal RMSNorm and `lm_head` dispatch. This lets the earlier positions skip that work.

## What changed

`forward_step_mrope` and `forward_step_injected_mrope` take an `emit_head` flag; the vision prefill loop passes `false` for every position except the last. When `emit_head` is false the terminal head is not encoded at all, rather than encoded and discarded.

## The guard that decides whether the head runs

The head runs when the caller asks for it, when hidden capture is requested, or when this request has activated MTP:

```rust
let run_head = emit_head || capture_hidden || self.session.mtp_active;
```

The third operand is a per-request flag, not a property of the checkpoint. That distinction is the substance of this change. `session.mtp` reports whether MTP weights are resident, and weights load whenever `mtp_num_hidden_layers != 0`, independently of whether a request enables MTP. Keying the guard on residency meant that any vision request against an MTP-capable checkpoint ran the terminal head at every prefill position, which is exactly the work this change exists to skip.

`mtp_active` defaults to false, is set true only in the branch that commits to MTP for a request (immediately before `mtp_prefill`, the sole production call site), and is cleared by `reset_state()`. Every generate entry point calls `reset_state()` at its top, so no request inherits activation from a prior call on the same session — `generate_multimodal_vision_impl` clears it at :9491, ahead of the prefill loop's `forward_step_mrope` calls at :9550 and :9558. The protection the original operand provided is preserved: while MTP is genuinely active, the head still runs for any caller of the mrope entry points.

## Also in this change

- The bench binary's `BENCH_REPEATS` rejects `0` before model initialization instead of panicking on an out-of-bounds median index after the slow model load. The median helper returns a `Result` on empty input rather than indexing.
- The `#[cfg(test)] mod tests` block inside `bench_support` moved below the helper functions it precedes, so the file has no items after a test module.

## CI coverage for the bench surface

`--all-targets` does not reach the bench binaries under `src/bin/` unless `bench-internals` is in the feature list, so that surface was going unlinted. It is now linted by its own macOS-only step that passes `--release`, rather than by adding the feature to the existing debug lint. The reason is specific: `bench_vision_prefill_ab.rs` carries a `compile_error!` gated on `debug_assertions`, so that surface cannot compile in a debug clippy pass at all, and forcing `--release` onto the existing step would slow it for everything else it covers. The existing metal-gpu clippy step is unchanged.

The MP2 Metal injection gate asserts an exact test count, and this change adds tests that its filter selects, so the expected count moves from 12 to 13. The count in the committed workflow was measured by running the gate's own command, not computed; a stale explanatory comment beside it, which undercounted the `emit_head` matches, was corrected at the same time.

## Verification

The regression fixture builds a session with MTP weights resident and `mtp_active` false, calls `forward_step_mrope` directly with `emit_head=false` — the same call the vision prefill loop makes — and asserts no logits are returned and a sentinel written into the shared logits buffer survives untouched. Poisoning the buffer is necessary because `emit_head=false` returns empty logits by contract either way, so the return value alone cannot distinguish "skipped" from "ran and was discarded".

Executed on an Apple M2 Max, green. Reverting the guard's third operand to the residency check reddens it (exit 101) at the `must return no logits` assertion, because the mutated guard dispatches the terminal head; restoring the operand returns it to green. Three runs, green then red then green.

Both assertions were then shown to be load-bearing individually, which the run above does not establish on its own: the first assertion aborts the test before the sentinel comparison executes, so that comparison passing proves nothing about whether it can detect anything. A second mutation disabled the first assertion in this one test so control could reach the second, and reverted the guard as before. The test then failed at the sentinel `assert_eq!` with its own message, the observed buffer holding zeros where the sentinel had been written. The sentinel was overwritten, so the terminal head demonstrably dispatched, and the sentinel comparison is what caught it. Both edits were reverted and the test returns to green.

Clippy is clean under `f16,metal-gpu,bench-internals` and under `f16,metal-gpu,serve`; five unit tests cover the repeats validation and the median helper.

## Semver gate disclosure

The `semver-checks` job on this PR is green and that green is not coverage. The job's own annotation:

> **SEMVER: 0 checks executed** (0.8.0 -> 0.9.0; a 0.x minor bump reads as an assumed-major change, so cargo-semver-checks skipped all 254 checks). A green result here is NOT coverage. It becomes coverage again once the bumped version is published and becomes the crates.io baseline. Observed `lattice-transport` directly; the other workspace crates are inferred to be in the same state because they share one `[workspace.package] version`.

The workspace sits at an unpublished 0.9.0 against a published 0.8.0, so under 0.x rules the bump already declares breaking-allowed and the comparison is moot rather than merely unmeasured. Tracked in #1378. No semver claim of any sign is being made about this change.

## bench-compare disposition

A reachable target now exists and the measurement is unrun. This section replaces an earlier
structural waiver that is no longer accurate at this head.

### Measurement (issue #1336's acceptance criterion)

#1336 requires a before/after multimodal-prefill measurement and states there is no performance
claim without it. That criterion is **not discharged by this PR**, and this section replaces an
earlier structural waiver that is no longer accurate: at this head the branch itself adds the
missing bench target, so "no declared bench target reaches this path" is false.

What now exists: `crates/inference/src/bin/bench_vision_prefill_ab.rs`, an A/B harness whose two
arms are the same per-position forward-step calls differing only in a runtime `emit_head` flag,
so the two arms are one binary rather than two builds. It is registered with
`required-features = ["f16", "metal-gpu", "bench-internals"]`, matching the existing
`bench_decode_ab` and `bench_gdn_prefill_ab` targets.

Operator command:

```
LATTICE_MODEL_DIR=<checkpoint dir> \
BENCH_VISUAL_ROWS=1024 BENCH_WARMUP=1 BENCH_REPEATS=5 \
cargo run --release -p lattice-inference --features f16,metal-gpu,bench-internals \
  --bin bench_vision_prefill_ab
```

What the harness does not control for, stated by the harness's own module doc: it excludes
ViT/merger cost (synthetic checkpoint-shaped `post_merger_rows`, so the number is decoder
multimodal-prefill time and not full request latency), it is sensitive to machine contention and
thermal state, one untimed warmup per arm reduces but does not remove first-dispatch overhead,
and it models a single request with no batching or KV reuse.

**No number is claimed anywhere in this PR.** The harness compiles under
`--features f16,metal-gpu,bench-internals` and its debug-build guard was confirmed live, but it
has not been executed: timed runs happen on the dedicated bench machine under an exclusive lock,
not here. #1336 stays open until that run produces the before/after figures.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -327,15 +327,24 @@ jobs:
       # gate to examples, benches, and #[cfg(test)] modules — the same surface
       # where a second wave of lint debt (manual_let_else, approx_constant,
       # dead bench fields) had silently accumulated past the lib+bins gate.
+      - name: inference — metal-gpu clippy (macOS only)
+        if: runner.os == 'macOS'
+        run: cargo clippy -p lattice-inference --all-targets --features f16,metal-gpu -- -D warnings
       # `bench-internals` gates the bench binaries under `src/bin/` (e.g.
       # `bench_vision_prefill_ab`) and their `bench_support` module — without
       # it in this feature list, `--all-targets` never compiles that surface,
       # so lint debt there (an "items after a test module" error, #1336
       # round 2) went undetected by this exact command until a manually
-      # widened invocation caught it.
-      - name: inference — metal-gpu clippy (macOS only)
+      # widened invocation caught it. This is a separate --release step
+      # (rather than adding bench-internals to the step above) because
+      # `bench_vision_prefill_ab.rs:95` has a `compile_error!` gated on
+      # `debug_assertions` — "bench_vision_prefill_ab: MUST build --release
+      # (debug Metal timing is meaningless)" — so this surface cannot compile
+      # in a plain debug clippy pass, and forcing --release onto the fast
+      # debug lint above would slow it for everything else it covers.
+      - name: inference — metal-gpu bench clippy, release (macOS only)
         if: runner.os == 'macOS'
-        run: cargo clippy -p lattice-inference --all-targets --features f16,metal-gpu,bench-internals -- -D warnings
+        run: cargo clippy --release -p lattice-inference --all-targets --features f16,metal-gpu,bench-internals -- -D warnings
       # The worker module is macOS/metal-gpu-gated, but this lifecycle test
       # uses only a blocking thread-local destructor and never acquires a
       # Metal device. Run it explicitly so compiling the test is not mistaken
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -327,9 +327,15 @@ jobs:
       # gate to examples, benches, and #[cfg(test)] modules — the same surface
       # where a second wave of lint debt (manual_let_else, approx_constant,
       # dead bench fields) had silently accumulated past the lib+bins gate.
+      # `bench-internals` gates the bench binaries under `src/bin/` (e.g.
+      # `bench_vision_prefill_ab`) and their `bench_support` module — without
+      # it in this feature list, `--all-targets` never compiles that surface,
+      # so lint debt there (an "items after a test module" error, #1336
+      # round 2) went undetected by this exact command until a manually
+      # widened invocation caught it.
       - name: inference — metal-gpu clippy (macOS only)
         if: runner.os == 'macOS'
-        run: cargo clippy -p lattice-inference --all-targets --features f16,metal-gpu -- -D warnings
+        run: cargo clippy -p lattice-inference --all-targets --features f16,metal-gpu,bench-internals -- -D warnings
       # The worker module is macOS/metal-gpu-gated, but this lifecycle test
       # uses only a blocking thread-local destructor and never acquires a
       # Metal device. Run it explicitly so compiling the test is not mistaken
--- a/.github/workflows/e2e-parity.yml
+++ b/.github/workflows/e2e-parity.yml
@@ -1073,11 +1073,12 @@ jobs:
       # documents. LATTICE_METAL_TEST_ENFORCE=1 makes a missing device or
       # checkpoint a hard failure instead of the default dev-run skip.
       #
-      # The filter also selects two later additions that reuse this same
+      # The filter also selects later additions that reuse this same
       # checkpoint+device: `forward_step_mrope_emit_head_false_skips_terminal_head_dispatch`
       # (mechanism test: `emit_head=false` must skip the terminal RMSNorm +
-      # lm_head dispatch entirely, not merely its readback) and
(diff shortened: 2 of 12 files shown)


ohdearquant:
Holding this one short of merge on the measurement, not on the code.

#1336 states the acceptance criterion plainly: "Capture a before/after measurement for the multimodal prefill path; no performance claim without it." The behaviour half is discharged here — the flag is wired through both serving entry points, the failure path leaves the pending state intact and both callers warn and continue, the non-final prefill still advances the model layers before the optional head, and a sibling search over `forward_step_(injected_)?mrope|emit_head|preload_vision` finds no unguarded M-RoPE path. The performance half is not, and this PR's whole reason for existing is that skipping the terminal head on non-final multimodal prefill positions is faster.

The PR body currently carries a structural waiver instead of an A/B table. That waiver does not apply. The waiver exists for changes no declared bench target can execute, and it requires naming the targets searched and the gate that excludes them. Here the honest reading is different and worse: `/usr/bin/grep -R -n -E "generate_multimodal_vision|bench_" crates/inference/benches crates/inference/src/bin crates/inference/examples` finds no multimodal-prefill benchmark target at all, and no bench target in `crates/inference/Cargo.toml` names vision, mrope, or multimodal. So the change is unmeasurable with the targets that exist, which is a reason to add one, not a reason to waive. An unmeasured change is not a safe change.

Second point, smaller but worth fixing in the same round. The new `emit_head` mechanism and equivalence tests and both preload tests are not reached by any required check at this head. The macOS vision job runs `cargo test --release -p lattice-inference --lib inject`, whose 8 tests are injection tests; the other green vision gates are `vision_s5b_e2e_gate_test` and `vision_serve_e2e_test`. So a future edit can break the new guard while CI stays green. The two CLI parser tests are genuinely mutation-sensitive — mutating the default to `true` reddens `preload_vision_omitted_defaults_to_false` with the expected assertion — and the equivalence test is intentionally green under an always-emit implementation, since the paired mechanism test is what detects that no-op. That pairing only works if both actually run somewhere required.

What unblocks this: a real-checkpoint multimodal-prefill measurement with an always-emit baseline against the head-skipping path, run on the dedicated bench host under the quiet-machine discipline, plus the four new tests added to a required Metal command. The measurement is booked; this is not a rewrite request.

CI at this head reads 37 checks, 33 success and 4 skipped, no failures. The skips are `build macOS Studio Swift release`, both `cargo audit` jobs, and `layer-23 golden bridge` — those surfaces are unmeasured here, which is a coverage statement rather than a safety one.

PR #1369, opened 2026-08-07
fix(inference): route all checkpoint mmap sites through the trust boundary

### LINKED ISSUE #1368: mmap trust boundary is wired into only the Q4 loader; three checkpoint mmap sites bypass it
`crates/inference/src/weights/mmap_trust.rs` implements a trust boundary for memory-mapped checkpoint files: it rejects group- or other-writable model files and walks the parent-directory chain for the same property, with a documented carve-out for root-owned files and for sticky-bit directories. The module is 1724 lines and thoroughly unit-tested.

It has exactly one caller.

```
$ git grep -n 'open_trusted_mmap_file' -- crates/ | grep -v 'mmap_trust.rs'
crates/inference/src/weights/q4_weights.rs:969:    let (mut file, metadata) = crate::weights::mmap_trust::open_trusted_mmap_file(path)?;
```

Every other zero-copy checkpoint loader opens and maps its file directly, bypassing the boundary:

| Site | What it maps |
|---|---|
| `crates/inference/src/weights/f32_weights.rs:233` | safetensors shard, reached from `open_manifest_entry_once` |
| `crates/inference/src/quant/quarot/io.rs:140` | safetensors shard for a QuaRot artifact |
| `crates/inference/src/forward/metal_qwen35.rs:789` | Q3 weight file, mapped into a Metal no-copy buffer |

The Metal Q3 site is the clearest case, because its own safety comment names the exact invariant the guard exists to enforce:

> the caller must not mutate the on-disk file while this process runs — the same invariant `mmap_q4_weight` relies on for its no-copy Metal buffer

`mmap_q4_weight` gets that invariant checked by `open_trusted_mmap_file`. The Q3 path asserts the same invariant and checks nothing. A world-writable Q3 checkpoint is accepted where the equivalent Q4 checkpoint is rejected, and because the buffer is no-copy, a concurrent writer's edits are visible to the running model.

This is a reachability gap rather than a missing mechanism. The guard is correct and tested; it is simply not on three of the four paths that need it, so its test suite passes while most of the surface it names is unprotected.

### Suggested shape

Route the three sites above through `open_trusted_mmap_file`, keeping each loader's existing error type by mapping at the call site. `f32_weights::from_open_file` takes an already-open `File` on purpose — the comment there explains that reopening by path would reintroduce a TOCTOU window — so that one needs the check applied at the `open_manifest_entry_once` open rather than inside `from_open_file`.

Worth adding alongside the fix: a test that enumerates mmap construction sites and requires each to be either guarded or explicitly exempted with a stated reason, in the same spirit as the existing measurement-lock contract test. Without it the next loader added has nothing stopping it from repeating this, and a per-site fix leaves no evidence for the next reader that the set is complete.

### Note on prior work

PR #1037 set out to do exactly this ("every zero-copy weight loader"). Its trust-boundary module has since landed on main independently, but only wired into the Q4 loader, so the wiring half of that PR was never delivered.

ohdearquant:
Closes #1368.

The mmap trust boundary was wired into the Q4 loader only. `grep -rn "MmapOptions::new()\|Mmap::map("` over `crates/inference/src` finds six constructions: four production sites plus two fixtures inside `mmap_trust`'s own tests. One of the four routed through `open_trusted_mmap_file`; three opened the file directly and mapped it with no ownership, mode or ACL check.

## Why this is two chokepoints rather than one

`open_trusted_mmap_file` opens the path itself with `O_NOFOLLOW`, which rejects a symlink at the final path component. Two of the three unguarded sites must follow one. A HuggingFace hub-cache checkpoint stores the tensor file as a symlink into the blob store (`snapshots/<rev>/model.safetensors -> ../../blobs/<sha>`), and the symlink is the final component, so routing `SafetensorsFile::open` and `open_manifest_entry_once` through `open_trusted_mmap_file` would have made every hub-cache checkpoint unloadable. That path is covered by an existing test, `sharded_loader_follows_hub_cache_snapshot_layout`, which stays green here.

So this adds `reject_if_open_mmap_file_untrusted(file, path)`, which runs the same mode, uid and ACL checks plus the parent-directory check against a `File` the caller has already opened — by which point the symlink is resolved — without the open-time `O_NOFOLLOW`/`O_NONBLOCK` guards. `open_trusted_mmap_file`'s open-time behaviour is unchanged: it still opens with `O_NOFOLLOW` and `O_NONBLOCK` and still applies the same regular-file, mode, uid and ACL checks, in the same order.

**Corrected.** An earlier revision of this section said its "signature and behaviour are unchanged". The behaviour half is accurate; the signature half is not. Its return type changes from `Result<(File, std::fs::Metadata), String>` to `Result<TrustedMmapFile, String>`, in both the unix and the non-unix definition, so that the opened descriptor and its metadata travel together to the mapping call instead of being passed as a loose pair. It is a crate-private function, so this is not a public API change, but the sentence as written was wrong about the diff it was describing.

Placement:

| Site | Guard |
|---|---|
| `weights/q4_weights.rs` `open_and_mmap_q4_file` | unchanged, already guarded: `open_trusted_mmap_file` then `map_and_verify_trusted` |
| `weights/f32_weights.rs` `SafetensorsFile::open` | opens, then delegates to `from_open_file` |
| `weights/f32_weights.rs` `SafetensorsFile::from_open_file` | `map_after_untrusted_open` (this is the site that performs the `Mmap::map`) |
| `weights/f32_weights.rs` `open_manifest_entry_once` | `reject_if_open_mmap_file_untrusted` — it returns a `File` and never maps; its consumer maps through `from_open_file` |
| `quant/quarot/io.rs` `Shard::open` | `map_after_untrusted_open` |
| `forward/metal_qwen35.rs` `mmap_q3_weight` | `open_trusted_mmap_file` (Q3 is this project's own format, never a hub-cache symlink target, so it matches its Q4 sibling exactly) |

An earlier revision of this section placed `reject_if_open_mmap_file_untrusted` at
`SafetensorsFile::open` and `Shard::open`, and said of `from_open_file` that it "deliberately does
not carry the call". Both statements described a design that no longer matches this head, and the
second is now the opposite of the truth: `from_open_file` is precisely where the guarded mapping
happens. The table above is the placement at this head.

## The mapping call is no longer reachable unguarded

Two changes make that a structural property rather than a convention every future caller has to
remember.

First, the trusted handle no longer surrenders the file it guarded. `TrustedMmapFile`'s entire
surface is `len()`, `impl Read` and `impl Seek`; there is no accessor returning the underlying
`File`. A caller therefore cannot take a guarded handle, extract the descriptor, and map it itself,
which was the one path by which the guard could be satisfied and then bypassed.

Second, the two mapping producers are now symmetric about the post-map recheck. Previously only the
`open_trusted_mmap_file` flavour rechecked identity after mapping. `map_after_untrusted_open` now
snapshots metadata *before* running the guard, maps, and then calls `verify_mmap_target_unchanged`
against that snapshot, so a truncate-or-replace landed in the guard-to-map window is caught on this
path too. The window was real and is now covered on both flavours rather than one.

## Contract test

`crates/inference/tests/mmap_trust_boundary_contract.rs` parses every file under `crates/inference/src` with `syn`, finds each mmap construction, and keys it by enclosing module/function path plus an ordinal rather than by line or column, following the convention already used by `metal_measurement_lock_contract.rs`. A site fails unless its enclosing function also reaches a trust-boundary check or carries a reviewed exemption. The discovered inventory is additionally asserted against an expected set, so adding, removing or renaming any mmap site anywhere in the crate fails the test until it is consciously updated. A second test rejects an exemption that no longer matches a live site.

## Verification

```
cargo test -p lattice-inference --features metal-gpu --test mmap_trust_boundary_contract
  test result: ok. 2 passed; 0 failed
cargo test -p lattice-inference --lib --features metal-gpu mmap
  test result: ok. 31 passed; 0 failed
cargo clippy --workspace --all-targets -- -D warnings   # clean
```

Mutation arm, run at this head with the prediction stated before the run: reverting `mmap_q3_weight` to a plain `File::open` must redden `every_mmap_construction_site_is_guarded_or_explicitly_exempted` and nothing else. Observed exactly that — the failure names `src/forward/metal_qwen35.rs::inner::mmap_q3_weight::MmapOptions::new().map()#1`, and `every_exemption_matches_a_currently_expected_site` stayed green. Restored, re-ran, both green.

Each newly guarded site also gained an integration-level regression test that constructs a group- or other-writable checkpoint and asserts the loader rejects it, rather than relying on the underlying predicate's own unit tests.

## bench-compare disposition

No A/B was run; the argument is structural, and it does not claim safety.

Population searched: all 24 `[[bench]]` targets declared in `crates/inference/Cargo.toml`, partitioned by `required-features` (18 with none, one under `f16`, two under `bench-internals`, three under `metal-gpu`+`f16`) — not just the two targets `bench-compare` runs by default. Sweeping all 24 sources for identifiers that reach the changed loaders finds five that do, including `lm_head_bench`, which carries no required feature and so builds in a plain `cargo bench`. The sweep's positive control (`grep -l "Device::system_default" crates/inference/benches/*.rs`, 2 hits) confirms it finds real matches.

So a reachable target exists and this change is measurable in principle. It was not measured here. The added work is one `stat()` and, on macOS, one `acl_get_fd_np()` per model load — before the first token in every one of the five reaching benches, not per tensor and not per step — so an effect on decode or prefill timing is not expected. That is reasoning from the shape of the change, not evidence. If it is wrong, the error runs toward an unmeasured one-time model-load cost rather than a decode-loop regression. Stating the residual risk is the point: an unmeasured change is unmeasured, not safe.

### DIFF
--- a/crates/inference/src/forward/metal_qwen35.rs
+++ b/crates/inference/src/forward/metal_qwen35.rs
@@ -776,16 +776,17 @@ mod inner {
 
         validate_q3_mlp_role(tensor_name).map_err(|e| e.to_string())?;
 
-        let (mut file, metadata) = crate::weights::mmap_trust::open_trusted_mmap_file(path)?;
-        let header = crate::weights::q3_weights::read_q3_header(&mut file)
+        let mut handle = crate::weights::mmap_trust::open_trusted_mmap_file(path)?;
+        let header = crate::weights::q3_weights::read_q3_header(handle.file_mut())
             .map_err(|e| format!("failed to parse Q3 header {}: {e}", path.display()))?;
-        let file_len = file
+        let file_len = handle
+            .file_mut()
             .metadata()
             .map_err(|e| format!("failed to stat {}: {e}", path.display()))?
             .len();
         crate::weights::q3_weights::validate_q3_header_payload_bounds(&header, file_len, path)
             .map_err(|e| format!("failed to validate Q3 payload {}: {e}", path.display()))?;
-        let mmap = crate::weights::mmap_trust::map_and_verify_trusted(&file, &metadata, path)?;
+        let mmap = crate::weights::mmap_trust::map_and_verify_trusted(&handle)?;
 
         let buf = device.new_buffer_with_bytes_no_copy(
             mmap.as_ptr().cast(),
--- a/crates/inference/src/forward/metal_qwen35.rs
+++ b/crates/inference/src/forward/metal_qwen35.rs
@@ -755,7 +755,12 @@ mod inner {
     ///
     /// # Safety invariant
     /// The mmap is read-only (`MAP_PRIVATE`) and the model files must not be
-    /// modified while the process is running.
+    /// modified while the process is running. [`crate::weights::mmap_trust::open_trusted_mmap_file`]
+    /// enforces the write-boundary half of that invariant (group/other-writable,
+    /// foreign ownership, a writable ancestor directory, or a permission-granting
+    /// extended ACL) the same way [`mmap_q4_weight`] does; same-uid mutation
+    /// through a second fd remains the documented residual, as for every mmap
+    /// site in this crate.
     ///
     /// Not yet called from live checkpoint loading (proven end-to-end by this
     /// module's `mmap_q3_weight_*` tests instead).
@@ -771,8 +776,7 @@ mod inner {
 
         validate_q3_mlp_role(tensor_name).map_err(|e| e.to_string())?;
 
-        let mut file = std::fs::File::open(path)
-            .map_err(|e| format!("failed to open {}: {e}", path.display()))?;
+        let (mut file, _metadata) = crate::weights::mmap_trust::open_trusted_mmap_file(path)?;
         let header = crate::weights::q3_weights::read_q3_header(&mut file)
             .map_err(|e| format!("failed to parse Q3 header {}: {e}", path.display()))?;
         let file_len = file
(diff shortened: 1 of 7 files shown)


ohdearquant:
The production wiring here holds up: each of the three loaders now validates the descriptor before mapping it, and removing either new guard individually reddens its own test (`weights::f32_weights::tests::open_rejects_a_group_or_other_writable_checkpoint_file` and the QuaRot equivalent at `crates/inference/src/quant/quarot/io.rs:1218-1235`).

One thing to strengthen before this lands, in the regression net rather than the fix.

`crates/inference/tests/mmap_trust_boundary_contract.rs` states its contract in the module doc as "must **first** call one of this crate's two mmap trust-boundary chokepoints." The implementation does not check "first". `GuardCallFinder` sets `found = true` when a guard call appears anywhere in the enclosing function body, including nested blocks and closures, and the site loop consumes only that one boolean. Its own doc comment says as much: "not a stricter before/after ordering."

Two shapes pass that check while violating the stated contract:

- a function that maps the file and then validates it, since the guard call is still present in the body;
- a function where one branch validates and a second branch maps unguarded, since a guard anywhere satisfies every site.

Both are the bypass the contract exists to catch, so the test emits a green that reads as evidence ordering was checked when nothing about ordering was checked. The disclosure in the comment does not close it, because the reader of the green is CI.

These are straight-line open-then-map functions, so a positional check that requires a guard earlier in the body than each construction site should be enough for the shapes that exist today. If a full control-flow dominance check costs more than this test warrants, the acceptable alternative is to implement the strongest check that is practical and make the module doc state exactly what is enforced and what is not. A weaker check with an accurate doc is fine; a strong-sounding doc over a weak check is the thing to remove.

Worth a mutation arm either way: insert an extra `Mmap::map(&file)` before the existing guard in one listed loader and confirm the contract test reddens.

PR #1370, opened 2026-08-07
fix(npm): fail closed on empty platform matrix and misnamed binaries

### LINKED ISSUE #696: npm: broaden @khive-ai/lattice-embed native prebuilt matrix beyond darwin-arm64

ohdearquant:
Closes #696

Closes the second of the two blockers found on #1039's publish path: `publish-npm.sh` now fails closed on an empty platform matrix and on misnamed or missing platform binaries, instead of only on a partial matrix.

## What was wrong

`EXPECTED_PLATFORMS` was derived from the main package's `optionalDependencies` and consumed directly by a `for` loop, with no assertion that the derived set was non-empty. Measured under `/bin/sh` with `set -e`, the script's own interpreter:

| `optionalDependencies` | derived set | loop iterations | rc | reached publish? |
|---|---|---|---|---|
| two platforms (control) | two entries | 2 | 0 | yes, guarded |
| `{}` | empty | 0 | 0 | **yes, unguarded** |
| key absent | — (throws) | — | 1 | no, fails closed |

An empty object is valid JSON that yields an empty string, and a loop over nothing succeeds, so the script would publish the WASM and native main packages with no platform binaries at all. The absent-key arm was already safe only incidentally, because `Object.keys(undefined)` throws and `errexit` catches it.

Separately, the per-platform check was an existence glob (`ls "$pkgdir"*.node`), so a directory holding any stale or misnamed `.node` passed even when the filename that subpackage's own `main` field names was absent.

## What changed

- Refuse an empty expected matrix before the loop, naming the package whose `optionalDependencies` was empty.
- Require the exact `.node` path each subpackage's `main` field declares, rather than any glob match.
- Verify each platform package's `name` and `version` agree with the main package.
- Run the platform packlist content guard for every platform inside the script, so `make publish-npm` gets the same protection the workflow job already had. The workflow assertions stay where they are, as defence in depth.

## Verification

Four guards were added for the defects above, each proven load-bearing by mutation: remove the
guard, watch its specific arm redden, restore, watch it pass. Predictions were written before each
run and reconciled arm by arm.

One test-design flaw surfaced and was fixed rather than smoothed over. The first exact-path arm
named a correctly-named file that did not exist on disk, so removing the exact-path guard left the
arm green — not because the guard was inert, but because the packed tarball then contained zero
`.node` files, which a different assertion also rejects. The arm was overdetermined and could not
isolate the guard it claimed to test. It was rewritten to name an actually-wrong filename so exactly
one `.node` is packed either way.

## A self-test that could not tell a dry run from a real publish

The self-test as first written was weaker than it read, in a way worth stating because it is the
kind of gap a passing count hides.

Its `npm` stubs branched on the command word alone, so `npm publish` and `npm publish --dry-run`
were indistinguishable to every arm that existed to exercise the dry-run guard. Those arms could not
have failed if the script had published for real. Every stub now records its full argv and rejects
an unexpected invocation, naming what it saw.

Removing a guard *call* from `main()` was also invisible, because the arms drove the guard functions
directly. An added arm runs the script's entry point against a recording stub and asserts the
expected guard invocations actually occurred, so deleting a call site reddens.

The main packlist guard was not status-preserving: it read `npm pack --dry-run --json` through a
pipeline whose exit status is only its last command's, so a failing pack that still emitted a
well-formed manifest on stdout reported success. It now captures that command's output, checks its
status, and only then hands the manifest to the assertion. Fixing the `package.json` script entry
instead was rejected because npm runs script entries through the platform's default shell, which is
not guaranteed to support `set -o pipefail`.

At this head the self-test reports **94 passed, 0 failed**, run end to end. End-to-end under
`/bin/sh`, an empty matrix exits 1 with a diagnostic and a one-platform matrix proceeds.
## Not in scope

The version sequencing problem is deliberately untouched here. The main package and all seven optional dependencies are declared at `0.1.0`, and `@khive-ai/lattice-embed` and `@khive-ai/lattice-embed-darwin-arm64` each have exactly that version published already; npm tuples are immutable, so the preflight correctly refuses before shipping anything. That needs a coordinated version bump across the manifests and lockfile and is being handled separately.

## The lockfile pinned a vulnerable dev dependency that the unlocked range does not select

`package.json`
declares `@napi-rs/cli: ^3.7.2` and did so before this branch, so it is tempting to conclude the
exposure predates the PR. It does not. The lockfile this PR adds is what decides the resolution,
and it pinned `@napi-rs/cli 3.7.3` with `js-yaml 4.3.0`, which carries
[GHSA-5p4m-2wfm-xmqj](https://github.com/advisories/GHSA-5p4m-2wfm-xmqj). Meanwhile `^3.7.2`
resolves to `3.8.2` today, and every `@napi-rs/cli` in range declares `js-yaml ^4.2.0`, which
admits the patched `4.3.1`. So the lockfile froze an exposure the floating range moves past.

Whether an added lockfile is a regression is not answered by whether the dependency was already
declared. It is answered by comparing what the lockfile pins against what the range resolves to
now. The lockfile is regenerated: it now pins `@napi-rs/cli 3.8.2` and `js-yaml 4.3.1`. Both audit
arms are clean — `npm audit --package-lock-only --omit=optional` and the same with `--omit=dev`
each exit 0, the second reporting zero vulnerabilities.

## The self-test supplies its own npm cache

The real `npm pack` and `npm publish --dry-run`
child processes previously inherited the ambient `~/.npm` cache, so the suite's result depended on
the ownership of a directory outside the repository. Each run now creates a cache it owns under its
own temporary directory. The discriminating check is that the suite passes with `NPM_CONFIG_CACHE`
unset and `HOME` pointed at a directory containing no `.npm` at all: 10 tests, 10 passed, 0 failed.

## The dry-run stubs require exact argv

The stub previously accepted any argument list whose
first two arguments were `publish --dry-run`, so an unexpected trailing flag passed silently. It
now requires the exact argument count and rejects anything else by name. The failure arms were also
one-sided — failure was injected only at the WASM directory — so there are now independent failure
arms for a platform directory and for the native directory, five cases in total.

## Semver gate disclosure

The `semver-checks` job on this PR is green and that green is not coverage. The job's own annotation:

> **SEMVER: 0 checks executed** (0.8.0 -> 0.9.0; a 0.x minor bump reads as an assumed-major change, so cargo-semver-checks skipped all 254 checks). A green result here is NOT coverage. It becomes coverage again once the bumped version is published and becomes the crates.io baseline. Observed `lattice-transport` directly; the other workspace crates are inferred to be in the same state because they share one `[workspace.package] version`.

The workspace sits at an unpublished 0.9.0 against a published 0.8.0, so under 0.x rules the bump already declares breaking-allowed and the comparison is moot rather than merely unmeasured. Tracked in #1378. No semver claim of any sign is being made about this change.


ohdearquant:
Findings at `ac585cc6a8d8b4471e7af83b419b9964632e6d82`. Three real, one refuted, and the refuted one was the most confidently stated of the set.

**Refuted: the artifact upload/download paths are correct.**

The concern raised was that the assembled platform packages are downloaded one directory level too high, quoting `path: npm/lattice-embed-native` at line 268. Line 268 reads `path: npm/lattice-embed-native/npm`. The quoted value belongs to **line 179**, which is a different download of a different artifact.

There are two artifact flows in this workflow, and they are easy to conflate because they have the same shape:

- `native-<target>` — raw `.node` files, uploaded from `npm/lattice-embed-native/*.node` (line 145), downloaded with `pattern: native-*` to `npm/lattice-embed-native` (line 179). Correct: the binaries belong at the package root.
- `npm-native-prebuilds` — assembled platform directories, uploaded from `npm/lattice-embed-native/npm/` (line 215), downloaded to `npm/lattice-embed-native/npm` (line 268). Correct: the platform directories land exactly where `scripts/publish-npm.sh:90` reads them.

Recording it because a value read from the wrong artifact of the same shape is specific, plausible, and survives a careful read.

**Real: two of the new guards are not load-bearing.**

Defeating the platform package-name check at `scripts/publish-npm.sh:115-119` — mutating the comparison so it can never fire — leaves the selftest at `26 passed, 0 failed`. Reproduced directly, with the source restored afterwards and `git diff` confirming no residue. The same holds for the main-package packlist call at `:222`: replacing it with `:` leaves the suite green.

So the suite's arm count does not mean every new guard is covered. Two of them are asserted rather than tested, and the count reads as if they were. Each needs a fixture: a platform package with a wrong `name`, and a main package whose packlist fails.

**Real: the lockstep check validates keys but not values.**

`scripts/publish-npm.sh:75` derives only the platform *names* from `optionalDependencies`, and `:121-127` compares each platform package's version against `NATIVE_VERSION`. Nothing ever reads the dependency *values* at `npm/lattice-embed-native/package.json:53-60`. Changing the fixture's optional-dependency values to `9.9.9` while leaving the platform packages at `1.2.3` still passes `(m) full valid matrix passes`.

That publishes a main package advertising `@khive-ai/lattice-embed-<platform>@9.9.9` alongside platform packages at `1.2.3` — an install-time mismatch that the local directory checks cannot see. Each platform's `optionalDependencies` value should be required to equal both the platform package's version and `NATIVE_VERSION`, with a missing key or a non-exact range rejected.

**Real: the checked-in versions cannot publish.**

A single `/bin/sh` `curl` invocation with an explicit User-Agent and `lodash@4.17.21` as a known-published control: `@khive-ai/lattice-embed-wasm@0.6.1` → 200, `@khive-ai/lattice-embed@0.1.0` → 200, `@khive-ai/lattice-embed-darwin-arm64@0.1.0` → 200, and the other six platform tuples → 404. Registry tuples are immutable, so the first release run stops on the already-published wasm tuple at `scripts/publish-npm.sh:198-202` before reaching the matrix this change exists to guard.

That is release sequencing rather than a defect in this diff, but it means the guards here have never been exercised end to end and cannot be until the versions move.

**Also worth fixing:** the comment at `scripts/publish-npm.sh:175-176` cites `fix_r4_report.md` as the evidence for the E404 behaviour, and no such file exists in the tree. Either drop the reference or land the artifact. And `npm audit` on the newly committed native lockfile reports one high advisory for `js-yaml` 4.3.0 reached through `@napi-rs/cli` 3.7.3; `--ignore-scripts` at `.github/workflows/npm-prebuild.yml:256-258` controls lifecycle execution, not the resolved version.

Issue #1299, opened 2026-08-03
embed: mandatory blake3 dependency breaks the no-C-toolchain claim for consumers on aarch64 cross-targets


ohdearquant:
## Problem

`lattice-embed` 0.7.1 (published) carries a **mandatory** `blake3` dependency. `blake3` 1.8.5's default build script compiles `c/blake3_neon.c` for little-endian AArch64 (`blake3-1.8.5/build.rs:367-372`). So any downstream consumer that enables a lattice-embed-backed feature and cross-compiles to e.g. `aarch64-unknown-linux-gnu` needs a target C toolchain — a "pure Rust, no C toolchain" claim about lattice-embed consumers is false on that target.

## Reproduction (external consumer, target CC deliberately absent)

```
CC_aarch64_unknown_linux_gnu=/usr/bin/false \
  cargo check -p <consumer-crate> --features <lattice-backed-feature> \
  --target aarch64-unknown-linux-gnu --locked
```

fails inside `blake3` while invoking `c/blake3_neon.c`. Found during downstream integration review on 2026-08-03; independently reproducible from any consumer of the published 0.7.1.

## Why it matters

Pure-Rust portability is a headline property of this workspace (Pi 5 / edge cross-compile targets included). A mandatory C-compiling transitive dep quietly voids it for the embed crate's consumers on our primary edge architecture.

## Options

1. Enable `blake3` with `default-features = false` (its `pure` / no-asm configuration) if the hashing use is not hot-path critical — measure first.
2. Make the `blake3`-using functionality an optional feature so kernel-only consumers (`simd` distance paths) never pull it.
3. If neither is viable, document the C-toolchain requirement per target explicitly in the crate README and release notes.

Option 2 is likely the right shape: distance-kernel consumers (the common downstream integration) have no need for content hashing.

## Acceptance

A downstream crate using only the SIMD distance surface cross-checks to `aarch64-unknown-linux-gnu` with `CC_aarch64_unknown_linux_gnu=/usr/bin/false` and passes; the arm asserting this runs in CI or is recorded as a release-checklist step.

PR #1403, opened 2026-08-11
refactor(embed): drop blake3, hash with sha2

### LINKED ISSUE #1402: refactor(embed): drop the blake3 dependency — replace with the workspace's existing sha2
Supersedes the original pure-hash-feature proposal on this issue (kept in the edit history).

lattice uses blake3 in exactly two places, both in lattice-embed, both cold and small-input: the embedding-cache key hasher (src/cache.rs:154) and the model provenance hash (src/model.rs:40, hex asserted 64 chars at src/model.rs:473). Nothing needs blake3 specifically, and its build story is a standing tax: the default backend compiles C (c/blake3_neon.c on aarch64 cross-targets, #1299), and the 'pure' escape hatch force-propagates a portable-only implementation into every downstream workspace that shares the crate (feature unification cannot be overridden downstream).

Fix: remove blake3 from the workspace entirely and use sha2 (already a workspace dependency, used by lattice-inference and lattice-tune; RustCrypto pure Rust, no build script, no C, no feature coupling). SHA-256 hex is also 64 chars, so the provenance length contract is unchanged.

Behavior notes:
- embedding-cache keys change value once → existing caches miss and repopulate; no correctness impact
- provenance hash values change → any stored provenance strings from earlier versions will not match newly computed ones; release notes must say so

Acceptance:
- no blake3 anywhere in the lockfile (cargo tree -i blake3 errors out)
- #1299's repro stays green trivially (no C-compiling dep left on that path)
- cache round-trip and provenance tests updated and passing

Target: 0.9.1.

ohdearquant:
## Problem

blake3's default backend compiles C (`c/blake3_neon.c` on aarch64 cross targets, #1299), and its `pure` escape hatch force-propagates a portable-only implementation into every downstream workspace that shares the crate via Cargo feature unification — downstream cannot override it.

## Fix

lattice-embed was the workspace's only blake3 consumer, in two cold-path sites: the embedding-cache key hasher (`crates/embed/src/cache.rs`) and the model provenance hash (`crates/embed/src/model.rs`). Both now use SHA-256 via `sha2` (already a workspace dependency; pure Rust, no build script), and blake3 is removed from the workspace entirely. SHA-256's 64-char hex keeps the provenance length contract unchanged. Crate docs that named BLAKE3 are updated.

## Behavior notes

- Embedding-cache keys change value once: existing caches miss and repopulate; no correctness impact.
- Provenance hash values change algorithm: provenance strings computed by earlier versions will not match newly computed ones.

## Verification

- `grep -rn blake3 crates/ Cargo.toml` → no hits outside the historical CHANGELOG entry; `cargo tree -i blake3` → no such package (both re-run at the commit).
- `cargo test -p lattice-embed` green; `test_model_provenance_unique_hash` discriminates a wrong-constant mutation (mutated → fail, restored → pass).
- `cargo clippy -p lattice-embed -- -D warnings` and `cargo fmt --check` clean; full workspace clippy ran in the pre-commit gate.

## Bench-compare disposition

Dependency swap on cold-path hashing (cache key construction, provenance). Manifest/lockfile changed, so a measured A/B on the default embed bench group will be attached as a comment before merge rather than claiming a structural waiver.

Closes #1402.


ohdearquant:
## bench-compare disposition

Measured A/B, `--quick`, ABBA arm order (base₁ → head₁ → head₂ → base₂), on an otherwise idle
Apple silicon machine. Base `44b733ac67`, head `d8e0ea5419`. The current head `3ab97e6296`
differs from the benched head by a changelog entry alone
(`git diff --name-only d8e0ea5419..3ab97e6296` returns only `crates/embed/CHANGELOG.md`), so no
bench-relevant build input moved after the run.

### Run conditions

- Locks: bench window and Metal GPU, both acquired uncontended.
- Machine state at all three arm boundaries: AC power, thermal nominal, HID idle ≈ 5.2 days,
  gate passed.
- Ambient CPU idle 96.8% / 97.3% / 97.2% against a 70.0% floor.
- Toolchain identical on both arms: rustc 1.94.1, cargo 1.94.1, criterion 0.5.1.
- Targets: `lattice-inference:elementwise_cpu_bench` and `lattice-embed:simd`, no filters.

### Result

`lattice-inference:elementwise_cpu_bench` — all 17 gated benches within the ±3.0% noise band.
Largest absolute point delta 0.83% (`residual_add_layer_norm/fused/hidden768_seq128`).

`lattice-embed:simd` — 260 measurements, classified informational-only at quick resolution
(#714), so they contribute nothing to the gated verdict. Reported for completeness: the point
deltas span −6.32% to +5.72%, are symmetric about zero, and show no directional shift. The one
cache-touching group, `cached_embedding_delegate/native_wrong_model_disabled_64x4096`, is
+1.30% with a confidence interval of [−3.26%, +6.07%] and a measured arm-order bias bound of
4.35% — wider than the point estimate, so the row carries no signal.

### What this run can and cannot say

It says nothing about the changed hashing code, because no declared bench target executes it.

`crates/embed` declares five bench targets: `simd_bench`, `simd`, `embeddings`,
`simsimd_comparison`, `simd_opt_bench` (`crates/embed/Cargo.toml`), matching the five files in
`crates/embed/benches/`. Only `simd` touches the cache subsystem, and its
`cached_embedding_delegate` group builds the service with capacity zero —
`CachedEmbeddingService::new(Arc::new(NativeEmbeddingService::default()), 0)`,
`crates/embed/benches/simd.rs:50` — so the cache is disabled, and `cache_and_embed` returns on
the explicit "bypass cache entirely when disabled" fast path,
`crates/embed/src/service/cached.rs:116-121`, before `compute_key` is ever called. The
provenance hash in `crates/embed/src/model.rs` is referenced by no bench target.

The structural waiver is nonetheless unavailable here, which is why the run was made: the change
edits `Cargo.toml`, `crates/embed/Cargo.toml`, and `Cargo.lock`, and the waiver covers only
runtime source changes whose build inputs are identical between base and head. So this A/B
measures the dependency-graph change — `blake3` dropped from the workspace, and `sha2`, already
a workspace dependency, now linked into `lattice-embed` — and found no movement in the measured
surface.

Residual risk, stated rather than waived: the hashing change itself is unmeasured. Both call
sites are cold — cache-key construction, and one provenance hash per model load — and each
hashes a single text plus a short formatted string, so neither sits on a hot loop. That is an
argument from the call graph, not a measurement, and an unreachable change is an unmeasured
change rather than a safe one.

ohdearquant:
## bench-compare disposition

Two things are true here and both belong in the record: an A/B was run and is clean, and the A/B cannot speak to the change itself. Stating only the first would misrepresent the coverage.

### Measurement

`scripts/bench-compare.sh` at base `afcc77786828cc6ba28a094ff0c309b105800aac`, head `fcbb662b3d8efdc9522dbd68810d9543856374a8`.

`local-compare/lattice-inference:elementwise_cpu_bench` — all 17 gated benches within the ±3.0% noise band. `lattice-embed:simd` is classified informational-only in quick mode (#714); its 255 measurements are excluded from the gate result.

<details><summary>All 17 measurements</summary>

| Bench | Δ point | bound-lower | bound-upper | order bias ≤ | base n/mode | head n/mode |
|---|---:|---:|---:|---:|---|---|
| `add_bias_gelu/4096` | -0.82% | -1.78% | +0.14% | +0.79% | 4 Linear | 4 Linear |
| `add_bias_gelu/896` | -0.51% | -1.49% | +0.47% | +0.71% | 4 Linear | 4 Linear |
| `elementwise_mul/4096` | -0.74% | -1.50% | +0.04% | +0.75% | 4 Linear | 4 Linear |
| `gelu/4096` | -0.43% | -2.11% | +1.29% | +1.26% | 4 Linear | 4 Linear |
| `gelu/896` | -1.50% | -1.69% | -1.31% | +0.12% | 4 Linear | 4 Linear |
| `layer_norm/4096` | -1.32% | -2.57% | -0.06% | +0.80% | 4 Linear | 4 Linear |
| `layer_norm/896` | -1.39% | -1.76% | -1.02% | +0.34% | 4 Linear | 4 Linear |
| `residual_add_layer_norm/fused/hidden384_seq128` | -0.66% | -1.94% | +0.64% | +1.19% | 4 Linear | 4 Linear |
| `residual_add_layer_norm/fused/hidden768_seq128` | -0.30% | -1.12% | +0.53% | +0.57% | 4 Linear | 4 Linear |
| `residual_add_layer_norm/unfused/hidden384_seq128` | -0.60% | -1.07% | -0.13% | +0.46% | 4 Linear | 4 Linear |
| `residual_add_layer_norm/unfused/hidden768_seq128` | +0.51% | -0.72% | +1.76% | +0.79% | 4 Linear | 4 Linear |
| `rms_norm/4096` | -2.98% | -4.09% | -1.87% | +0.92% | 4 Linear | 4 Linear |
| `rms_norm/896` | -0.27% | -2.17% | +1.66% | +1.68% | 4 Linear | 4 Linear |
| `silu_inplace/4096` | -1.35% | -1.98% | -0.72% | +0.55% | 4 Linear | 4 Linear |
| `silu_inplace/896` | -1.02% | -1.90% | -0.13% | +0.75% | 4 Linear | 4 Linear |
| `softmax_attention/128` | -1.53% | -1.94% | -1.13% | +0.29% | 4 Linear | 4 Linear |
| `softmax_attention/512` | +0.18% | -2.53% | +2.98% | +2.68% | 4 Linear | 4 Linear |

</details>

<details><summary>Run conditions</summary>

```
resolution: --quick
inference features: <none>
filters: inference='<all>' embed='<all>'
arm order: ABBA (base₁ → head₁ → head₂ → base₂)
locks:
  bench-window (/tmp/lion-bench-window.lock): acquired immediately (uncontended)
  Metal GPU (/tmp/lion-metal-gpu-test.lock): acquired immediately (uncontended)
ambient load:
  [quiet] before first arm:      idle 86.5% (floor 70.0%) ok
  [quiet] between order strata:  idle 86.4% (floor 70.0%) ok
  [quiet] after final arm:       idle 79.2% (floor 70.0%) ok
machine state: power ac, thermal nominal, gate passed at all three checkpoints
base_rustc = head_rustc = rustc 1.94.1 (e408947bf 2026-03-25)
base_cargo = head_cargo = cargo 1.94.1 (29ea6fb6a 2026-03-24)
base_criterion = head_criterion = 0.5.1
criterion_mode = quick
```

</details>

### What the run does not cover

No declared bench target executes the changed hashing code, so the clean table above is a statement about coverage, not about the change.

`ModelProvenance::new` is not called from any bench. The cache key is reachable in principle through `CachedEmbeddingService`, but the one bench that constructs it (`cached_embedding_delegate` in `crates/embed/benches/simd.rs`) builds it with capacity `0`, and `crates/embed/src/service/cached.rs` takes an explicit fast path that returns before any key is computed when the cache is disabled. So that group never reaches `compute_key` either.

This change therefore has no reachable bench group, and the A/B above measures only what the dependency-graph change did to unrelated code. The residual risk is stated rather than waived: an unreachable change is an unmeasured change, not a safe one. What bounds it here is that the two SHA-256 call sites are not on any hot path — one runs once per model load, the other once per uncached text — and that removing a dependency cannot slow code that never called it.

### Why the A/B was re-run

The first A/B was taken before the version bump. That bump edits `Cargo.toml` and `Cargo.lock`, which are build inputs, and the structural-waiver argument in `CLAUDE.md` explicitly does not extend to build inputs. Rather than argue that a version string cannot move numbers, the comparison was re-run at the bumped head. The run provenance above records `base_criterion = head_criterion = 0.5.1` and identical toolchain on both arms, so the locked bench invocation resolved the same harness on both sides.

### Head currency

The measured head is `fcbb662b3d`. The current head is `9b3913c0e9`. The delta between them is markdown-only:

```
$ git diff --name-only fcbb662b3d..9b3913c0e9
crates/embed/CHANGELOG.md
crates/embed/CONFIG.md
crates/embed/DESIGN.md
```

Filtering that list for non-markdown paths returns nothing, while the same filter over the whole branch returns `Cargo.lock Cargo.toml crates/embed/Cargo.toml crates/embed/src/cache.rs crates/embed/src/model.rs crates/inference/Cargo.toml crates/tune/Cargo.toml`, which is the control showing the filter reports non-markdown when non-markdown is present. The measurement carries forward to the current head.

ohdearquant:
Head-currency update for the bench-compare disposition above.

That disposition measured `fcbb662b3d`. The head has since moved to `5de9ccfce2` across three
prose commits. The delta remains markdown-only:

```
$ git diff --name-only fcbb662b3d..5de9ccfce2
crates/embed/CHANGELOG.md
crates/embed/CONFIG.md
crates/embed/DESIGN.md
```

Filtering that list for non-markdown paths returns nothing. The same filter over the whole branch
returns `Cargo.lock Cargo.toml crates/embed/Cargo.toml crates/embed/src/cache.rs
crates/embed/src/model.rs crates/inference/Cargo.toml crates/tune/Cargo.toml`, which is the
control showing the filter does report non-markdown when non-markdown is present.

No build input, no runtime source, and no bench-target definition differs between the measured
head and the current one, so the measurement and its coverage caveat both carry forward unchanged.

PR #1411, opened 2026-08-12
fix(inference): reject over-window embeddings input before truncation

### LINKED ISSUE #1408: serve: /v1/embeddings context-window check is unreachable for text when max_position_embeddings >= 4096
## Summary

The per-item context-window check in `crates/inference/src/serve/embeddings.rs` compares a **post-truncation** token count against the model's context window. Because the tokenizer has already truncated the input to `DEFAULT_BPE_MAX_SEQ_LEN` (4096) by the time the count is taken, the check cannot fire on any checkpoint whose context window exceeds 4096 tokens.

The user-visible effect: `POST /v1/embeddings` accepts text longer than the tokenizer's cap, silently embeds a **4096-token prefix**, and returns HTTP 200. The caller receives an embedding of a truncated document presented as an embedding of the whole document, with no error and no flag. `usage.prompt_tokens` reports the truncated count, so the response does not reveal the loss either.

All line references are at `18b2dc3e55309d8449914be4614e6f5514f3a1bb`.

## Mechanism

The count fed to the guard comes from `tokenize_len`:

```rust
// crates/inference/src/serve/embeddings.rs:123-124
pub fn tokenize_len(&self, text: &str) -> usize {
    self.tokenizer.tokenize(text).real_length
}
```

`real_length` is measured after truncation. `BpeTokenizer::tokenize` routes through `tokenize_to_ids`, which truncates to `max_seq_len` before the length is taken (`crates/inference/src/tokenizer/bpe.rs`, `tokenize_to_ids_into`, the `scratch.ids.truncate(self.inner.max_seq_len)` calls). `EmbeddingModel` builds its tokenizer with `BpeTokenizer::from_tokenizer_json` and never calls `with_max_seq_len`, so the cap is the default:

```rust
// crates/inference/src/tokenizer/bpe.rs:22
pub(crate) const DEFAULT_BPE_MAX_SEQ_LEN: usize = 4_096;
```

The limit it is compared against is derived from the checkpoint instead:

```rust
// crates/inference/src/serve/embeddings.rs:132-133
pub fn max_context(&self) -> usize {
    self.config.max_position_embeddings.min(8192)
}
```

And the check is a strict greater-than (`embeddings.rs:469,474`), reached from the text branch at `embeddings.rs:510`:

```rust
let token_count = embedder.tokenize_len(text);
check_item_fits_window(index, token_count, max_context)?;
```

So `token_count <= 4096` always, while the rejection needs `token_count > max_context`. **Whenever `max_position_embeddings >= 4096`, the two ranges do not overlap and the text branch of the guard is unreachable.**

This is a structural property of the two constants, not a property of any particular input: no text of any length can trigger it on such a checkpoint.

A Qwen3.5 checkpoint in local use reports `text_config.max_position_embeddings = 262144`, giving `max_context() = 8192` — exactly twice the tokenizer cap.

## Why the existing test does not catch it

`embed_items_rejects_text_item_over_context_window` (`embeddings.rs:784-795`) passes, but for a reason that does not generalize. Its fixture asserts `model.max_context() == 512` (`embeddings.rs:787`), which is **below** the 4096 truncation cap. A 600-token input therefore survives tokenization intact and does trip the guard.

The test only exercises the case where the context window is smaller than the tokenizer cap. The failing case is the opposite ordering, and no test covers it. A regression test for this fix has to pin a fixture whose context window exceeds 4096, otherwise it passes with or without the fix.

## Scope

The image branch (`embeddings.rs:518`) is not affected in the same way: its count comes from `image_scaffold_token_count`, whose padding term is derived from image geometry rather than from the tokenizer, so it is not bounded by `max_seq_len` and can still exceed the window.

`crates/inference/src/bin/lattice/serve.rs` reaches this module for its `/v1/embeddings` route, so that route is affected on any large-context checkpoint.

## Suggested fix

Measure the pre-truncation length. The tokenizer already computes it before truncating — the truncation sites log `original_len` — so the value exists and only needs to be carried out rather than recomputed. Comparing that against `max_context()` makes the guard reachable and turns silent truncation into the `context_length_exceeded` rejection the code already documents.

Worth deciding explicitly as part of the fix: whether the intended contract is to reject at `max_context()` or at the tokenizer cap, since text between 4096 and 8192 tokens is currently truncated by the tokenizer regardless of what the guard concludes. Rejecting at the window while the tokenizer still truncates at 4096 would move the silent-truncation boundary rather than remove it.

A related fix for the standalone `lattice_serve` binary's own embeddings route is in flight in #1355; this report is about the shared module, which that change does not touch.

ohdearquant:
Fixes #1408.

## The bug

`POST /v1/embeddings` silently truncates input between 4097 and 8192 tokens, embeds the
prefix, and returns 200. The rejection path that is supposed to catch over-long input is
unreachable.

Two limits disagree, and the guard reads the wrong side of the disagreement:

- the admission guard compares against `min(max_position_embeddings, 8192)`, which is 8192
  for the Qwen3.5 checkpoints (`max_position_embeddings = 262144`)
- the tokenizer the embeddings path builds was capped at `DEFAULT_BPE_MAX_SEQ_LEN` (4096),
  because it was constructed without `with_max_seq_len`

The guard then compared a **post**-truncation count against the window. A post-truncation
count can never exceed 4096, and the window is 8192, so the two ranges do not overlap and
the rejection is unreachable by construction. Input in that band is truncated by the
tokenizer, embedded from its prefix, and returned as a success.

The failure is silent in the worst direction: the caller receives a 200 and a
well-formed, correctly normalized vector that represents only the first part of its input.

## The fix

1. Cap the embeddings tokenizer to mirror the admission window
   (`.with_max_seq_len(config.max_position_embeddings.min(8192))`), so truncation and
   rejection coincide rather than straddling a gap.
2. Admit on the **pre**-truncation count, so an over-window item is rejected instead of
   being silently shortened.
3. Keep `usage.prompt_tokens` reporting the count actually embedded, not the count
   submitted. These are deliberately different numbers: admission is about what the caller
   sent, usage is about what the server did.

The second commit is a follow-up with no behavioural change: the text branch was calling
`tokenize()` twice per item, once for each count, when a single `TokenizedInput` already
carries both `real_length` and `pre_truncation_len`. It now tokenizes once and destructures
both, matching the shape already used elsewhere in the crate. Same two numbers reach the
same two call sites.

## Tests

Two regression tests, both against a fixture whose context window (1000) sits above its
tokenizer cap (995), which is the configuration that makes the gap expressible:

- `embed_items_rejects_text_item_whose_pre_truncation_length_exceeds_a_window_above_the_tokenizer_cap`
- `embed_items_accepted_prompt_tokens_matches_the_count_actually_embedded_when_truncated`

**Correction to an earlier version of this description.** It said both tests were
"checked for mutation sensitivity", which overstated the coverage. The mutation check
reverted the *admission-count* change and both tests reddened, so that property is
genuinely guarded. It did not cover the *tokenizer cap* at the loader: both tests build
their model through `EmbeddingModel::new` and set `.with_max_seq_len(5)` on their own
fixture, so reverting the production cap line leaves them green. This change has more than
one property and the mutation ranged over one of them.

The follow-up has now landed at this head. It adds one mutation arm per property and
closes the two further defects: the public `EmbeddingModel::new` now enforces the same
cap invariant the loader establishes (every constructor routes through it), and
`lattice-embed`'s `VisionEmbeddingModel` — which constructed the same tokenizer
uncapped — now enforces the identical pre-truncation admission on its own public
surface. `VisionEmbeddingModel::embed_text`, `embed_image`, and `embed_image_metal`
all reject input whose pre-truncation token count — plus, for the image entry points,
the scaffold tokens the image itself consumes (two delimiters + the merged-patch
pads) — exceeds `min(max_position_embeddings, 8192)`, returning
`EmbedError::InvalidInput` naming both counts. Previously `embed_text` silently
truncated exactly like the wire path did, and the two image entry points (whose
`prompt` parameter is unrestricted, unlike the wire route's image branch, which always
passes an empty prompt) had the same hole.

One pre-existing test changed its expectation as a direct consequence of the fix:
`embed_text_maps_context_overflow_to_inference_failed` asserted that an over-window
prompt fails deep inside the forward pass as `InferenceFailed` — the behaviour this
change removes. It is now `embed_text_rejects_prompt_exceeding_context_window_before_truncation`,
same fixture and input, asserting the admission rejection.

Mutation coverage at this head, per property, each arm observed red then restored green:

- removing the admission call in `embed_text` reddens its rejection test;
- removing the admission + scaffold-reserve calls in `embed_image` (and separately in
  `embed_image_metal`) reddens each rejection test;
- flipping the boundary comparison (`>` to `>=`) reddens both exactly-at-window
  acceptance tests (text and image).

The one arm not independently testable is `embed_image_metal` succeeding exactly at the
boundary: proving it requires a Metal ViT dispatch on a real device. Its admission call
is textually identical to `embed_image`'s and runs before the CPU/Metal dispatch fork,
and the metal rejection test proves the dispatch is unreached when admission fails.

Gates run at the final head: `cargo fmt --all -- --check`, `cargo clippy --workspace -- -D warnings`,
doc lint, `cargo test -p lattice-embed --lib` (341 passed, includes all `vision::tests`),
`cargo test -p lattice-inference --lib serve::embeddings::tests::`.

## bench-compare disposition

No A/B table. This is the structural-waiver case, and the argument is an absent call path,
not a compiled-out module — `serve` is in `lattice-inference`'s default feature set
(`default = ["std", "download", "serve"]`), so the changed code **is** compiled into the
bench binaries. Compilation is not the predicate; reachability is.

Population searched: all 32 declared bench targets in the workspace, enumerated two ways
that agree — the `[[bench]]` tables across every `crates/*/Cargo.toml`, and the 32
`crates/*/benches/*.rs` files on disk. Not the two targets a default A/B run would cover.

Result: zero of the 32 reach `serve::embeddings`. No bench source names
`serve::embeddings`, `embed_items`, `tokenize_lengths`, `tokenize_len`, or
`check_item_fits_window`. The search reaches those files — 27 of the 32 match `criterion`
in the same invocation shape, so an empty result is a real absence rather than a broken
query.

The same search covers the second changed file (`crates/embed/src/vision.rs`): zero bench
sources reference `VisionEmbeddingModel`, `embed_text_vlm_f16`, `embed_image_from_bytes`,
`preprocess_qwen35_image`, or `pooled_embed` (22 `criterion` matches in the same traversal
as the control). `crates/embed/benches/embeddings.rs` benchmarks only the plain vector-math
utilities in `lattice_embed::utils`, not the VLM path.

The nearest miss is worth naming, because it looks like a hit and is not:
`crates/inference/benches/tokenizer_bench.rs` imports
`lattice_inference::{BpeTokenizer, Tokenizer, WordPieceTokenizer}` — the tokenizer types
directly. This change alters a tokenizer *construction site* inside the embeddings module;
it does not touch `BpeTokenizer`'s own code, so that bench times identical executed code at
both base and head. Likewise `crates/embed/benches/embeddings.rs` imports
`lattice_embed::utils`, a different crate.

Build inputs are identical between base and head: the diff touches exactly two files
(`crates/inference/src/serve/embeddings.rs` and `crates/embed/src/vision.rs`), with no
manifest, lockfile, feature, profile, build-script, or bench-target-definition change.
That is what makes the waiver available at all; a build-input change has no absent call
path to name.

Residual risk, stated rather than implied: an unreachable change is not a safe change, it is
an unmeasured one. The added work on the request path is one extra comparison per item and,
after the second commit, one fewer `tokenize()` call than the first commit performed. On the
library path, `embed_text` gains one extra tokenize pass per call, and the image entry points
gain one extra image-preprocessing pass per call (the same duplicate-preprocessing shape the
wire route already uses for its usage accounting) so admission can run before the expensive
ViT forward. Nothing in the bench suite would show any of these.

### DIFF
--- a/crates/embed/src/vision.rs
+++ b/crates/embed/src/vision.rs
@@ -39,6 +39,30 @@ use std::path::Path;
 
 pub use lattice_inference::forward::cpu_f16::PoolingStrategy;
 
+/// Raises `tokenizer`'s own truncation cap to `config`'s
+/// `max_position_embeddings.min(8192)` decode window when it sits below it,
+/// otherwise returns `tokenizer` unchanged. Only ever raises the cap, never
+/// lowers it.
+///
+/// Mirrors `lattice_inference::serve::embeddings`'s identically-named
+/// guard (issue #1408): without it, a caller-supplied `BpeTokenizer` left at
+/// its constructor default (4096) truncates any prompt between 4097 tokens
+/// and this checkpoint's real window before [`embed_text_vlm_f16`] or
+/// [`embed_image_from_bytes_f16`] ever see it, silently discarding content
+/// the model could otherwise process. Unlike the `lattice_inference` sibling,
+/// this crate has no admission guard ahead of the forward pass to reject an
+/// over-window input outright -- the tokenizer's own cap is the only thing
+/// standing between "process it" and "silently shorten it", so it must never
+/// sit below the checkpoint's window.
+fn capped_tokenizer(tokenizer: BpeTokenizer, config: &Qwen35Config) -> BpeTokenizer {
+    let max_context = config.max_position_embeddings.min(8192);
+    if tokenizer.max_seq_len() < max_context {
+        tokenizer.with_max_seq_len(max_context)
+    } else {
+        tokenizer
+    }
+}
+
 #[cfg(test)]
 thread_local! {
     static AFTER_VISUAL_LOAD_HOOK: std::cell::RefCell<Option<Box<dyn FnOnce()>>> =
@@ -100,12 +124,21 @@ impl VisionEmbeddingModel {
     /// Use this when the checkpoint spans multiple safetensors shards (not
     /// supported by [`Self::from_directory`]) or when components are shared
     /// across other in-process model instances.
+    ///
+    /// Raises `tokenizer`'s own truncation cap to `config`'s context window
+    /// when it sits below it (see [`capped_tokenizer`]) -- see that
+    /// function's docs for why: without this, a caller-supplied tokenizer
+    /// left at its default cap silently truncates any prompt between 4097
+    /// tokens and the checkpoint's real window before it ever reaches the
+    /// forward pass. [`Self::from_directory`] goes through this constructor
+    /// too, so it gets the same guarantee.
     pub fn new(
         weights: F16ModelWeights,
         config: Qwen35Config,
         vision_weights: Qwen35VisionWeights,
         tokenizer: BpeTokenizer,
     ) -> Self {
+        let tokenizer = capped_tokenizer(tokenizer, &config);
         Self {
             weights,
             config,
@@ -656,8 +689,17 @@ mod tests {
     }
 
     fn write_tiny_vlm_checkpoint(dir: &Path, indexed: bool) {
-        let config = r#"{
-            "text_config": {
+        write_tiny_vlm_checkpoint_with_max_position_embeddings(dir, indexed, 512);
+    }
+
+    fn write_tiny_vlm_checkpoint_with_max_position_embeddings(
+        dir: &Path,
+        indexed: bool,
+        max_position_embeddings: usize,
+    ) {
+        let config = format!(
+            r#"{{
+            "text_config": {{
                 "hidden_size": 8,
                 "num_hidden_layers": 1,
                 "vocab_size": 16,
@@ -668,12 +710,12 @@ mod tests {
                 "head_dim": 8,
                 "rope_theta": 10000000.0,
                 "partial_rotary_factor": 1.0,
-                "rope_parameters": {
+                "rope_parameters": {{
                     "rope_theta": 10000000.0,
                     "partial_rotary_factor": 1.0,
                     "mrope_section": [2, 1, 1],
                     "mrope_interleaved": true
-                },
+                }},
                 "linear_num_key_heads": 2,
                 "linear_num_value_heads": 2,
                 "linear_key_head_dim": 32,
@@ -684,9 +726,9 @@ mod tests {
                 "layer_types": ["full_attention"],
                 "layer_mask": [true],
                 "eos_token_id": 15,
-                "max_position_embeddings": 512
-            },
-            "vision_config": {
+                "max_position_embeddings": {max_position_embeddings}
+            }},
+            "vision_config": {{
                 "depth": 1,
                 "hidden_size": 8,
                 "num_heads": 2,
@@ -697,13 +739,14 @@ mod tests {
                 "num_position_embeddings": 16,
                 "in_channels": 3,
                 "deepstack_visual_indexes": []
-            },
+            }},
             "image_token_id": 9,
             "vision_start_token_id": 10,
             "vision_end_token_id": 11,
             "tie_word_embeddings": true
-        }"#;
-        std::fs::write(dir.join("config.json"), config).expect("write config.json");
+        }}"#
+        );
+        std::fs::write(dir.join("config.json"), &config).expect("write config.json");
         write_tiny_tokenizer_json(dir);
 
         let shapes = tiny_vlm_checkpoint_shapes();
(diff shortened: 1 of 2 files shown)


oceanwaves630 (approved):
Approving on green CI across the full matrix at dd76c70c.

PR #1416, opened 2026-08-17
feat(embed): add sealed attestation contracts

ohdearquant:
> _PR description authored by Codex (OpenAI agent) on behalf of @ohdearquant._

## Summary

- Add the first dormant implementation slice for proposed ADR-088, stacked on draft spec PR #1415.
- Freeze only the bounded streaming `CheckpointAttestor`, the opaque 1..=4096-byte attestation report, and the four-axis native resource-budget value contract.
- Reject report-bound violations and retained-plus-transient accounting overflow with typed errors.

This PR does not add directory capabilities, snapshotting, model loading, `CpuPinned`, a preparer, a drain supervisor, a prepared embedding service, or any live provider-path behavior.

## TDD evidence

- RED: the two integration tests initially failed only because the new public contracts and typed errors did not exist.
- GREEN: the exact minimum/maximum report bounds and exact `u64::MAX` aggregate budget ceiling pass.
- Mutation sensitivity: allowing an empty report made the lower-bound test fail; replacing checked addition with saturating addition made the overflow test fail. Both mutations were reverted and the focused tests passed again.

## Test plan

- [x] `RUSTC_WRAPPER= cargo test -p lattice-embed --test prepared_attestation_api --test prepared_resource_budget_api -j4` — 4 passed
- [x] `RUSTC_WRAPPER= cargo test -p lattice-embed --lib -j4` — 338 passed
- [x] `RUSTC_WRAPPER= cargo clippy -p lattice-embed --all-targets --all-features -j4 -- -D warnings`
- [x] `cargo fmt --all -- --check`
- [x] repository pre-commit checks — workspace check, 191 tracked Markdown files, recursive discovery, capability matrix, and absolute-path checks passed

An exploratory all-feature test run passed 337/338 library tests; the only failure was the pre-existing Metal-only vision test because no Metal device could be initialized on this machine. The same library suite under default features passed 338/338.

## bench-compare disposition

Structural waiver: all 30 declared inference/embed targets were searched; performance-relevant build inputs are unchanged, and no target calls the added constructors, trait, errors, or re-exports. This slice is therefore unmeasured rather than proven performance-neutral; any later reachable prepare/load/encode path requires targeted paired A/B evidence.

## ADR

- Proposed ADR-088: `docs/adr/ADR-088-sealed-native-embedding-attestation.md`
- Benchmark governance: ADR-087
- Parallel child drafts: dormant CPU policy #1417 and private resource supervisor #1418

[...]

PR #1417, opened 2026-08-17
feat(inference): add dormant BERT CPU policy contract

ohdearquant:

## Summary

- Add a dormant, additive BERT CPU capability/profile contract stacked on #1416.
- Capture the process-stable scalar/AVX2/FMA/AVX-512F/NEON capability set in a private-field value object.
- Add an `Auto | Pinned(profile)` request value while keeping every existing `BertModel` constructor and kernel path on the legacy dispatcher.

Possessing this value is deliberately not proof that a BERT model was loaded or executed with pinned kernels. No model or kernel constructor accepts the policy in this slice.

## TDD evidence

- RED: the public integration test initially failed because the two root-exported policy types did not exist.
- GREEN: host capture/policy tests pass 2/2 and the host-independent synthetic capability matrix passes 1/1.
- Mutation sensitivity: changing `Pinned(profile)` to return `None` made the exact retention test fail; restoring it returned the focused suite to green.

## Test plan

- [x] `RUSTC_WRAPPER= cargo test -p lattice-inference --test bert_cpu_kernel_policy_api -j4` — 2 passed
- [x] `RUSTC_WRAPPER= cargo test -p lattice-inference --lib synthetic_capability_rows_are_copied_without_aliasing_fields -j4` — 1 passed
- [x] `RUSTC_WRAPPER= cargo clippy -p lattice-inference --all-targets -j4 -- -D warnings`
- [x] `git diff --check`
- [x] repository pre-commit checks

Exploratory `--all-features` clippy is not claimed green: Rust 1.94 reports eight pre-existing warnings in unchanged GPU files (`manual_is_multiple_of`, `manual_div_ceil`, redundant closure, and manual slice-size calculation). This PR does not edit those unrelated paths.

## bench-compare disposition

Structural waiver: all 30 declared inference/embed targets were searched. No target constructs these values, and no model, forward path, or numeric kernel accepts them. The slice is unmeasured rather than proven performance-neutral; the first reachable pinned prepare/load/encode slice must add dedicated paired A/B Criterion evidence.

## ADR

- Proposed ADR-088: `docs/adr/ADR-088-sealed-native-embedding-attestation.md`
- Parent contract draft: #1416
- Benchmark governance: ADR-087

This remains a stacked draft and must not merge before ADR-088 is accepted.

- [x] PR body starts with the required agent attribution
- [x] Commit contains the required Codex co-author trailer
- [x] Existing BERT constructors and numeric dispatch are unchanged
- [x] No execution or performance claim is made from capability metadata alone

## Out of scope

- Policy-aware primitive dispatchers
- Accelerate bypass and pinned numeric kernels
- FP-control guards and build identity
- Strict prepared BERT loading
- Snapshot/resource admission and lifecycle supervision
- Prepared embedding service or Khive integration


ohdearquant:
Keeping this one open on purpose, and recording why so it is not swept up with
its siblings.

The rest of the ADR-088 series was consolidated into #1467, which is based on
current `main` and carries the series tip as an ancestor. This change is not in
it: it branched off the same base as #1418 rather than continuing the chain, so `git merge-base --is-ancestor` puts
it outside the consolidated history, and neither
`crates/inference/src/forward/cpu/bert_policy.rs` nor
`crates/inference/tests/bert_cpu_kernel_policy_api.rs` exists at #1467's head.

Plan: once #1467 lands, rebase this onto `main` and take it on its own. It is
191 lines across four files, dormant by construction (`#[doc(hidden)]`,
`#[non_exhaustive]`, no constructor accepts `Pinned`), so it stands alone
without the rest of the series.

ohdearquant:
Closed in favour of #1477, which carries this commit unchanged on a branch cut
from current `main`.

Retargeting this branch onto `main` was tried first and produced the wrong diff:
#1467 landed as a squash, so the series commits below this one are not ancestors
of `main`, and the retargeted comparison came back at twenty files including a
revert of `main`'s ADR-088 text to the series' earlier draft. Cherry-picking the
single commit gives the intended four files and +191 lines with no conflict.

Verified on the new branch before opening it: format and clippy clean, both the
unit test and the integration test pass, and the aliasing assertion is
load-bearing — pointing `avx2_enabled` at the FMA field makes it fail, and
restoring the line makes it pass again.

PR #1418, opened 2026-08-17
feat(embed): add dormant native resource supervisor

ohdearquant:

## Summary

- Add a private, dormant four-axis native resource-admission substrate stacked on #1416.
- Atomically account preparation slots, encode slots, retained bytes, and transient-work bytes under one mutex.
- Add cancellation-safe admission futures, FIFO per-class wake entitlement with cross-class alternation, RAII release guards, irreversible two-phase drain closure, and bounded waiter notification.

This slice does not expose a preparer or prepared service and does not claim full ADR-088 D8 conformance.

## TDD evidence

- RED: the original 20 focused tests first failed at compile time because the private supervisor, admission/lifecycle errors, futures, guards, and drain did not exist. Five scheduler regressions then failed against the first implementation: unbounded eligible wakes, same-class newcomer overtaking, missing cross-class alternation, repeated post-drain admission wakes, and failure to hand off a second slot released while the first entitlement was outstanding.
- GREEN: all 25 admission/cancellation/FIFO-wake/drain/storage/invariant tests pass.
- Mutation sensitivity: reintroducing an all-eligible wake scan, selecting the newest class waiter, bypassing entitlement, disabling class alternation, repeating close notifications, and removing the successful-reservation handoff each made its exact regression test fail. Restoring each mutation returned the focused suite to green. The earlier final-release drain-wake mutation remains covered too.

## Test plan

- [x] `RUSTC_WRAPPER= cargo test -j4 -p lattice-embed --lib resource::tests` — 25 passed
- [x] `RUSTC_WRAPPER= cargo clippy -p lattice-embed --all-targets --all-features -j4 -- -D warnings`
- [x] `git diff --check`
- [x] repository pre-commit checks

## Resource and cancellation properties

- Single-request oversize fails on first poll without reserving any axis.
- Admission reserves all applicable axes atomically or waits with no partial lease.
- Dropping a queued future removes its registered waker and reserves no capacity.
- Guard drop uses checked release accounting, selects at most one FIFO entitlement atomically, and wakes it after unlocking.
- An exact entitlement prevents a repolled or newly arriving caller from stealing released capacity; alternating class preference guarantees progress across preparation and encode queues when both heads are eligible.
- Successful reservation atomically hands off any additional capacity that became free while its entitlement was outstanding, so independent releases cannot become latent.
- Retained-only pressure cannot head-of-line block an eligible work-only encode.
- `begin_drain` closes both classes before observation, wakes queued callers as `Closed`, and is idempotent.
- `wait_released` is typed before drain begins, removes cancelled waiters, and wakes only after the final guard releases.
- Cancelled waiter high-water storage is released rather than ratcheting domain memory.

## bench-compare disposition

Structural waiver: all 30 declared benchmark targets (25 inference targets and 5 embed targets) were searched, and none reaches this private dormant module. The perf build inputs are unchanged. This slice is unmeasured, not performance-neutral; the first reachable prepare/encode integration requires dedicated paired A/B evidence.

## ADR

- Proposed ADR-088: `docs/adr/ADR-088-sealed-native-embedding-attestation.md`
- Parent public-contract draft: #1416
- Parallel dormant CPU-policy draft: #1417

This remains a stacked draft and must not merge before ADR-088 is accepted.

## Explicitly deferred before D8 conformance

- Transfer of retained capacity into a published prepared-object lease
- Preparation-only terminal namespace faults while existing encodes continue
- Residue debt and aggregate cleanup errors
- Job-before-spawn tracking and native completion ownership
- Publication/fault/drain linearization with prepared object cleanup

- [x] PR body starts with the required attribution
- [x] Commit contains the required Codex co-author trailer
- [x] No public API or live service behavior is added

## Child draft

- #1420 adds the private checked preparation-ceilings, BERT/tensor formulas, and worst-case charge substrate on top of this supervisor.

### DIFF
--- a/crates/embed/src/service/prepared.rs
+++ b/crates/embed/src/service/prepared.rs
@@ -124,3 +124,6 @@ pub trait CheckpointAttestor: Send + 'static {
     /// Finalizes this fresh pass and transfers its immutable report bytes.
     fn finish(self) -> Result<OpaqueAttestationReport>;
 }
+
+#[cfg_attr(not(test), allow(dead_code))]
+mod resource;
--- a/crates/embed/src/service/prepared/resource.rs
+++ b/crates/embed/src/service/prepared/resource.rs
@@ -5,6 +5,7 @@
 //! cleanup errors remain outside this substrate and must land before prepared services use it.
 
 use super::NativeResourceBudget;
+use std::collections::BTreeMap;
 use std::future::Future;
 use std::pin::Pin;
 use std::sync::{Arc, Mutex, MutexGuard};
@@ -25,6 +26,15 @@ pub(super) enum ResourceLifecycleError {
 enum AdmissionClass {
     Preparation,
     Encode,
+}
+
+impl AdmissionClass {
+    fn other(self) -> Self {
+        match self {
+            Self::Preparation => Self::Encode,
+            Self::Encode => Self::Preparation,
+        }
+    }
 }
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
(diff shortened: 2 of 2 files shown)


ohdearquant:

PR #1420, opened 2026-08-17
feat(embed): add checked preparation limits

ohdearquant:
> _PR description authored by Codex (OpenAI agent) on behalf of @ohdearquant._

## Summary

- add a dormant, crate-private checked-arithmetic substrate for ADR-088 native preparation limits
- group inventory, tensor, BERT, parser, and resource ceilings without freezing a public constructor
- validate all cross-group relations and every allocation-indexable `u64` cap before later filesystem or parser work
- compute checked BERT tensor-count, tensor-element, fused-QKV, tensor-census, and worst-case retained/transient charges
- account for one retained attestation report plus the second transient report that coexists during pre/post comparison

This is intentionally a private leaf stacked on #1418. It has no public re-export, filesystem access, parser/tokenizer/inference dependency, model allocation, or production caller. ADR-088 remains Proposed, so this PR stays Draft and makes no claim that prepared loading or attestation is implemented.

## Dependency order

- stacked on #1418 (`codex/adr-160-native-resource-supervisor`)
- depends transitively on the dormant public contract in #1416 and proposed ADR-088 in #1415
- does not widen the public API or change legacy embedding behavior

## TDD and mutation evidence

The first focused compile was RED on 22 intentionally missing limits/types/formula symbols. The private implementation then made the focused suite GREEN.

Six production mutations were applied one at a time, observed RED, and restored:

1. `actual > limit` → `actual >= limit`
2. required-tensor coefficient `16` → `15`
3. removal of the second transient attestation-report charge
4. checked charge addition → wrapping addition
5. removal of the constructor's injected platform-cap validation
6. authoritative public report bound `4096` → `4095`

The final 20-test matrix covers all 32 limit axes, seven relations plus framed-header overflow, the constructor and all 18 platform-sized `u64` routes, seven BERT geometry axes, epsilon/head invariants, pooler and integer boundaries, three source dtypes, census failure atomicity, authoritative report-bound coupling, and independent charge-term deltas.

## Verification

```text
RUSTC_WRAPPER= cargo test -p lattice-embed --lib service::prepared::limits::tests -j4
20 passed; 0 failed

RUSTC_WRAPPER= cargo test -p lattice-embed --lib -j4
383 passed; 0 failed

RUSTC_WRAPPER= cargo clippy -p lattice-embed --all-targets --all-features -j4 -- -D warnings
PASS

RUSTC_WRAPPER= .githooks/pre-commit
[...]

PR #1421, opened 2026-08-17
feat(embed): add canonical tensor inventory digest

ohdearquant:

## Summary

- add a private, dormant SHA-256 tensor-inventory framing leaf for ADR-088
- reuse the bounded preparation census from #1420 for count, name, rank, dimension, element, metadata, and payload accounting
- sort caller-owned entries by raw UTF-8 name bytes, reject duplicates and unsupported source dtypes, and hash the exact versioned big-endian protocol
- pin the protocol with an independently computed golden plus mutation-sensitive ordering, field, Unicode, framing, limit, and overflow tests

## Stack and scope

- stacked on #1420 (`codex/adr-160-native-preparation-limits`)
- ADR-088 remains **Proposed**, so this PR remains **Draft**
- no public export, loader/parser/filesystem access, descriptor publication, model construction, inference call, or production caller
- no manifest or lockfile change; `sha2` was already a direct `lattice-embed` dependency

## Verification

- compile-first RED: focused test compilation failed on the deliberately absent digest and narrow limits seams
- focused GREEN: 13 passed
- `RUSTC_WRAPPER= cargo test -p lattice-embed --lib -j4`: 396 passed
- `RUSTC_WRAPPER= cargo clippy -p lattice-embed --all-targets --all-features -j4 -- -D warnings`: passed
- `.githooks/pre-commit`: passed, including workspace checks and 191 tracked Markdown files
- `cargo fmt --all -- --check`: passed
- `git diff --check`: passed
- independent final review: 0 Blocker / 0 High / 0 Medium

Six production mutations each made the focused guard fail and were restored:

1. disable canonical sorting
2. omit the domain separator
3. omit the tensor count
4. switch rank framing to little-endian
5. omit the dtype tag
6. accept duplicate tensor names

## bench-compare disposition

Structural reachability was checked across all 25 declared `lattice-inference` benches and all 5 declared `lattice-embed` benches. None imports or calls the private `prepared::tensor_inventory` module, and this PR changes no manifest, feature, benchmark target, build script, or production call path. An A/B benchmark would not execute the changed code.

This dormant code is **unmeasured, not performance-neutral**. Runtime cost, in-place sort cost, and hashing throughput remain residual risks for the later loader-wiring PR; that PR must add a reachable benchmark disposition before activation.

## Child draft

- #1422 — checked private prepared BERT geometry and logical payload facts

### DIFF
--- a/crates/embed/src/service/prepared.rs
+++ b/crates/embed/src/service/prepared.rs
@@ -130,3 +130,6 @@ mod resource;
 
 #[cfg_attr(not(test), allow(dead_code))]
 mod limits;
+
+#[cfg_attr(not(test), allow(dead_code))]
+mod tensor_inventory;
--- a/crates/embed/src/service/prepared/limits.rs
+++ b/crates/embed/src/service/prepared/limits.rs
@@ -9,7 +9,7 @@ use std::num::{NonZeroU64, NonZeroUsize};
 const ATTESTATION_CHUNK_BYTES: u64 = 1_048_576;
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
-enum LimitAxis {
+pub(super) enum LimitAxis {
     Files,
     PathBytes,
     TotalPathBytes,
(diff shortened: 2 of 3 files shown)


ohdearquant:

PR #1422, opened 2026-08-17
feat(inference): add checked prepared BERT facts

ohdearquant:

## Draft / governance gate

This stacked implementation PR remains **Draft** while ADR-088 is Proposed. It must not merge before that ADR is accepted and its earlier stack dependencies are ready.

Base: `codex/adr-160-tensor-inventory-digest` (`7a708b6db887b59d4902fe218b90a377b6786429`).

## Summary

- add a private, dormant checked-facts module for future sealed BERT preparation
- validate raw BERT geometry as bounded `u64`/`f64` facts before any platform-sized realization
- compute exact required-subset tensor counts/elements and fused-QKV elements with checked arithmetic
- account for mapped, copied/widened, and fused logical weight payload bytes without claiming RSS, peak, or complete retained-memory bounds
- provide one validated bridge to the existing `BertConfig`

## Safety boundary

The production helpers perform no parsing, filesystem access, heap allocation, model loading, serving, registry publication, or production routing. The slice exports no public or doc-hidden API and has no live production caller. Tests may exercise the existing fused-QKV builder. Extra tensors remain outside the required-subset census and are covered later by complete-inventory digest and mapped-file validation.

## TDD and validation

- compile-first RED: 120 expected missing-symbol errors
- final focused module tests: 16 passed
- full lib suite before the final test-only hardening: 2,637 passed, 17 ignored, 0 failed
- `cargo clippy -p lattice-inference --lib --tests -j4 -- -D warnings`: passed
- `cargo fmt --all -- --check`: passed
- `git diff --check`: passed
- repository pre-commit hook: passed

All-target/all-feature clippy was also attempted. The new slice was corrected for its one finding and then passed its default lib/test clippy surface; the broader command remains blocked by eight pre-existing, unrelated GPU/WGPU lints in `forward/gpu/inner/dims.rs`, `forward/gpu/inner/util.rs`, and `forward/gpu_gemm.rs`. This PR does not modify those paths.

Mutation checks were observed RED and restored for the required-count coefficient, the `4H²`, `2HI`, `9H`, pooler, and fused-QKV terms; F16 width; mapped-file lower-bound behavior; zero-copy disposition; platform validation; half-pooler rejection; and hidden-layer/attention-head projection.

## bench-compare disposition

This is private dormant validation/accounting code with no production call edge. None of the 25 declared `lattice-inference` benchmark targets reaches it.

Evaluated at PR head `ecf92f6b5331664cbc365cc1836d5ef341c1a14b` against the 25-target `lattice-inference` benchmark surface at parent `7a708b6db887b59d4902fe218b90a377b6786429`; this statement is void if either ref moves.

The reached `bert.rs` hunk adds only the private module item and modifies no existing item. Recursively, the new module defines only private inert data/functions plus `cfg(test)` tests: no global/static initializer, linker/codegen attribute, macro, trait impl on an existing type, or production callsite.

Feature-gated targets inspected: `decode_attn_bench`, `kv_cache_layout_bench`, `metal_decode_bench`, `cross_turn_prefix_cache_bench`, `mtp_decode`, `backward_stage0`, and `f16_convert_bench`.

`e2e_bench` reaches the legacy BERT load/encode path but not this dormant module; `inference_bench` uses its own synthetic path. Runtime cost is therefore **unmeasured, not performance-neutral**. A future loader/inspector callsite must add a matching benchmark disposition and measurements before activation.

## Out of scope

- SafeTensors header/range inspection
- required tensor name, shape, duplicate, and encoder-layer-index validation
- tokenizer/config parsing
- sealed snapshot construction
- CPU-pinned model loading or execution
- admission, drain, cleanup, registry, cache, or serving integration

## Child draft

- #1423 — checked dormant SafeTensors framing facts

### DIFF
--- a/crates/inference/src/model/bert.rs
+++ b/crates/inference/src/model/bert.rs
@@ -20,6 +20,9 @@ use std::fs;
 use std::path::Path;
 use tracing::warn;
 
+#[cfg_attr(not(test), allow(dead_code))]
+mod prepared_facts;
+
 /// Per-layer fused Q/K/V weight+bias, built once at model-load time from a
 /// `TransformerLayerWeights` layer's separate `query`/`key`/`value` tensors.
 ///
--- a/crates/inference/src/model/bert/prepared_facts.rs
+++ b/crates/inference/src/model/bert/prepared_facts.rs
@@ -1,0 +1,1231 @@
+//! Dormant checked BERT facts for sealed native preparation.
+//!
+//! This module deliberately performs no parsing, filesystem, model loading,
+//! allocation, or serving work. It validates raw numeric facts and computes a
+//! narrowly scoped logical weight-footprint estimate for a later prepared
+//! loader.
+
+use super::BertConfig;
+use std::num::{NonZeroU64, NonZeroUsize};
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(super) enum BertFactAxis {
+    VocabSize,
+    HiddenSize,
+    HiddenLayers,
+    AttentionHeads,
+    IntermediateSize,
+    PositionEmbeddings,
+    TypeVocabSize,
+    RequiredTensors,
+    RequiredElements,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(super) enum BertFactExpression {
+    RequiredTensorCount,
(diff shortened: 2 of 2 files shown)


ohdearquant:

PR #1423, opened 2026-08-17
feat(inference): add checked SafeTensors framing facts

ohdearquant:
> _PR description authored by Codex (OpenAI agent) on behalf of @ohdearquant._

## Draft / governance gate

This stacked implementation PR remains **Draft** while ADR-088 is Proposed. It must not merge before ADR-088 is accepted and the earlier stack dependencies are ready. It adds no public or doc-hidden API and does not activate prepared loading, attestation, or any downstream migration.

Stacked on #1422 at base `ecf92f6b5331664cbc365cc1836d5ef341c1a14b`.

Frozen framing-module SHA-256: `265424bba29808f1206b9d8dad32f7975deec729be8cb58fca4e0997a05376b8`.

## Summary

- add a dormant, crate-private scalar planner for SafeTensors prefix/header framing
- derive the prepared 4 MiB hard cap through a new inert crate-private accessor to the unchanged legacy loader constant
- validate caller file/header ceilings against the hard cap, `min(usize::MAX, isize::MAX)`, and checked `8 + max_header_bytes <= max_weight_file_bytes`
- decode the exact eight-byte little-endian header length and produce checked header/data ranges
- pin typed failure precedence for caller caps, platform span, short files, empty headers, arithmetic overflow, and headers extending past the declared file

## Safety and scope boundary

The production helper performs no I/O, allocation, mmap, JSON parsing, tensor/layout validation, model loading, or serving work. It has no live production caller.

A successful plan is not an inspected or trusted checkpoint. It does not bind the prefix and declared length to one opened handle, provide TOCTOU protection, acquire an admission lease, read header bytes, validate an inventory, or authorize mapping or loading.

## Legacy hazards deliberately unchanged

This PR neither changes nor blesses the existing `SafetensorsFile` path as prepared admission:

- the live legacy parser still converts the untrusted `u64` header length with `as usize` before ADR-088 caller/platform validation
- the live loader still maps or owns the complete backing before parsing the header
- legacy metadata materialization is not governed by ADR-088 caller ceilings for tensor count, decoded-name bytes, rank, dimensions, retained metadata, or parse-work bytes
- legacy loading, error behavior, and benchmark call paths remain unchanged

Those are activation gates for a later same-handle bounded reader/parser, not claims made by this framing-only slice.

## TDD and mutation evidence

The compile-first RED exited 101 with 25 planned missing-symbol errors and one unused-import warning. The final focused suite is GREEN at 8/8.

Nine production mutations were applied individually, observed RED, and restored:

[...]

PR #1424, opened 2026-08-17
feat(inference): add bounded SafeTensors inventory facts

ohdearquant:

## Draft / governance gate

ADR-088 remains **Proposed**. This stacked implementation PR therefore remains **Draft** and must not merge before that ADR is accepted and the earlier stack dependencies are ready. It adds no public API and no live prepared-mode caller.

Stacked on #1423 at base `8bd4a23c5ef24cd4016c5b46cc4b1cacca2ccace`.

Frozen inventory-module SHA-256: `ad7d23252461409296be01dc4ef137f4e0acb179105d605e0c2a9dda2118f9c4`.

## Summary

- add a dormant, crate-private two-pass parser for an already-framed SafeTensors header slice
- validate UTF-8 and a closed prepared-v1 JSON grammar without pass-one allocation
- enforce tensor-count, decoded-name, rank, dimension, aggregate-element, retained-metadata, and parse-work limits before any reserve
- fallibly reserve five exact logical arenas, then materialize decoded tensor names, dimensions, tensor records, and metadata keys
- accept only F32/F16/BF16, validate checked shape/source-byte arithmetic and exact entry-local ranges, and require globally contiguous full payload coverage
- reject decoded tensor-name and metadata-key aliases deterministically, including JSON escape aliases
- emit canonical raw-name-sorted inventory facts plus disjoint retained-inventory and transient parse-work accounting
- expose F32 absolute file-offset alignment only as a fact, never as zero-copy or mapping authority

## Safety and scope boundary

The production helper performs no filesystem I/O, prefix/header reads, payload reads, mapping, model loading, BERT-required-tensor matching, inventory hashing, descriptor construction, attestation, admission leasing, publication, or serving work. It has no live production caller.

A successful value is an inspected header inventory, not a trusted checkpoint. It does not bind bytes to one opened handle, close reread/TOCTOU races, authorize a mapping, or prove that an aligned F32 tensor is zero-copy eligible. Logical accounting excludes `Vec` headers, capacity slack, allocator overhead, and fixed stack state; a future live resource domain must reserve those conservatively.

This PR does not wire the parser into the existing loader or repair any legacy loader hazard. Same-handle bounded reads, exact mapped-prefix/header revalidation, payload integrity, snapshot sealing, and model publication remain activation gates.

## Closed grammar and accounting policy

- tensor objects contain exactly one each of `dtype`, `shape`, and `data_offsets`; unknown or duplicate members fail typed
- only one optional top-level `__metadata__` string-to-string object is accepted; duplicate decoded keys fail
- tensor names remain raw decoded UTF-8 bytes with no path interpretation or Unicode normalization
- scalar rank zero is one element; zero dimensions are allowed; all products and byte sums are checked
- zero-length ranges are legal only at the current sorted payload frontier
- retained inventory bytes are record bytes + decoded tensor-name bytes + dimension bytes
- parse-work bytes are the borrowed header length + metadata-key span bytes + decoded metadata-key bytes
- metadata values are streamed and never retained; their decoded byte count is proven not to exceed the bounded header

## TDD and mutation evidence

The compile-first RED exited 101 with 57 planned missing-symbol errors and one unused-import warning. The final focused suite is GREEN at 13/13.

Seventeen production mutations were applied individually, observed RED, and restored:

1. F32 source width `4` → `2`
2. limit admission `actual > limit` → `actual >= limit`
3. omit decoded tensor-name bytes from retained accounting
4. omit decoded metadata-key bytes from parse-work accounting
5. global frontier equality check `!=` → overlap-only `<`
6. disable decoded tensor-name duplicate rejection
7. disable decoded metadata-key duplicate rejection
8. omit relative tensor start from the absolute F32 offset
9. replace final canonical name sort with data-offset order
10. checked tensor multiplication → wrapping multiplication
11. swap the first two arena reservations
12. consume a mismatching byte before reporting its syntax position
13. platform byte-span selection `min(usize, isize)` → `max(usize, isize)`
14. omit dtype from semantic inventory equality
15. omit per-tensor data range from semantic inventory equality
16. omit parse-work bytes from semantic inventory equality
17. omit decoded metadata-value bytes from semantic inventory equality

## Verification

```text
Focused prepared_safetensors_inventory tests
13 passed; 0 failed

RUSTC_WRAPPER= cargo test -p lattice-inference --lib -j4 --quiet
2,659 passed; 17 ignored; 0 failed
32.93s

PASS

cargo fmt --all -- --check
PASS

git diff --check
PASS

repository pre-commit hook
PASS
```

All-target/all-feature clippy was also attempted. It remains blocked only by eight pre-existing, unrelated GPU/WGPU lints:

- `forward/gpu/inner/dims.rs`: two `manual_is_multiple_of`
- `forward/gpu/inner/util.rs`: one `manual_div_ceil`
- `forward/gpu_gemm.rs`: one `redundant_closure`, two `manual_div_ceil`, one `manual_slice_size_calculation`, and one `manual_is_multiple_of`

This PR does not modify those paths. Two independent final read-only reviews found 0 Blocker, 0 High, and 0 Medium findings on the frozen diff.

Evaluated at PR head `a6e36177d627994eca41abbb961158e106ab2700` against the 25-target `lattice-inference` benchmark surface at base `8bd4a23c5ef24cd4016c5b46cc4b1cacca2ccace`; this proof is void if either ref moves.

The diff changes no manifest, dependency or lockfile version, feature set, profile, build script, generated source, benchmark definition, or harness input. The reached `weights/mod.rs` hunk adds only one crate-private dormant module item and modifies no existing item. Recursive inspection of the new module found only new private/crate-private items and `cfg(test)` tests: no global/static initializer, linker/codegen attribute, macro expansion affecting existing code, unsafe code, trait implementation on an existing type, or production callsite.

`e2e_bench` reaches the unchanged legacy BERT loader and `f16_convert_bench` reaches the unchanged legacy `SafetensorsFile` path; neither imports or calls the new dormant parser. This slice is unmeasured, not performance-neutral.

The first same-handle reader, mapper, loader callsite, or prepared-mode activation must re-evaluate reachability and provide targeted measurement.

## Explicit exclusions / gates before activation

- source-directory capability, fixed inventory, snapshot, and domain-lock policy
- same-handle declared-length/prefix acquisition and exact bounded header read
- exact prefix/header rebind after mapping and before fact consumption
- payload-byte integrity and attestation
- required BERT tensor name, layer-index, duplicate, shape, pooler, and cross-artifact validation
- complete-inventory digest and downstream descriptor construction
- resource-domain leasing for allocator/container overhead and live parse/load work
- loader integration, F32 zero-copy disposition, model construction, publication, and serving

## Child draft

- #1425 — dormant required-BERT tensor inventory matcher

### DIFF
--- a/crates/inference/src/weights/mod.rs
+++ b/crates/inference/src/weights/mod.rs
@@ -7,6 +7,8 @@ pub(crate) mod ingress;
 pub(crate) mod mmap_trust;
 #[cfg_attr(not(test), allow(dead_code))]
 pub(crate) mod prepared_safetensors;
+#[cfg_attr(not(test), allow(dead_code))]
+pub(crate) mod prepared_safetensors_inventory;
 pub mod q3_weights;
 pub mod q4_weights;
 pub mod q8_weights;
--- a/crates/inference/src/weights/prepared_safetensors_inventory.rs
+++ b/crates/inference/src/weights/prepared_safetensors_inventory.rs
@@ -1,0 +1,2691 @@
+//! Dormant, bounded SafeTensors inventory facts for sealed native preparation.
+//!
+//! This module parses an already-framed header slice. It performs no filesystem
+//! access, mapping, payload reads, model loading, or serving work and has no live
+//! production caller. Pass one validates grammar and computes a zero-allocation
+//! census. Only then does pass two fallibly reserve bounded arenas, materialize
+//! decoded names and metadata keys, reject duplicates, and validate exact global
+//! payload coverage.
+//!
+//! A successful value is an inspected header inventory, not a trusted checkpoint.
+//! It does not bind bytes to an opened handle, close TOCTOU, authorize a mapping,
+//! or prove that an F32 tensor is actually eligible for zero-copy loading.
+//! Retained-inventory and parse-work byte facts are disjoint logical accounting
+//! terms. They exclude `Vec` headers, capacity slack, allocator overhead, and fixed
+//! stack state; a future live resource domain must reserve those conservatively.
+
+use std::num::NonZeroU64;
+use std::ops::Range;
+
+use super::prepared_safetensors::PreparedSafetensorsHeaderPlan;
+
+const CANONICAL_INVENTORY_U32_FIELD_MAX: u64 = u32::MAX as u64;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum InventoryAxis {
+    Tensors,
+    TensorNameBytes,
+    Rank,
+    Dimension,
+    AggregateElements,
+    MetadataBytes,
+    SafetensorsParseWorkBytes,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum InventoryExpression {
+    TensorCount,
+    DecodedNameBytes,
+    DimensionCount,
+    TensorElements,
+    TensorSourceBytes,
+    AggregateElements,
+    MetadataBytes,
+    SafetensorsParseWorkBytes,
+    AbsoluteDataOffset,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum TensorMember {
+    Dtype,
+    Shape,
+    DataOffsets,
+}
+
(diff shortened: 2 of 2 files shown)


ohdearquant:

PR #1425, opened 2026-08-17
feat(inference): match prepared BERT tensor inventory

ohdearquant:

## Draft / governance gate

ADR-088 remains **Proposed**. This stacked implementation PR therefore remains **Draft** and must not merge before that ADR is accepted and the earlier stack dependencies are ready. It adds no public API and no live prepared-mode caller.

Stacked on #1424 at base `a6e36177d627994eca41abbb961158e106ab2700`.

Frozen matcher-module SHA-256: `d0e0178c08917634f98fb9994df0dac5d7979866fa98a122d88456c584389989`.

## Summary

- add a dormant private matcher for the required prepared-BERT tensor closure
- derive optional pooler presence from the bounded inventory before checked geometry analysis
- recognize the five fixed embedding/LayerNorm names, all 16 exact tensor roles for every configured encoder layer, and the optional pooler pair
- require canonical ASCII layer indices in `0..L`, exact ranks, and exact dimensions from sealed config geometry
- prove completeness through upstream unique decoded names plus exact recognized-slot cardinality, without a per-layer bitset or heap allocation
- accept unrelated tensors as extras so they remain in the complete inventory for a later identity digest without entering required-only counts or conversion work
- return checked geometry plus a required-only dtype census; every required F32 tensor remains conservatively classified as copied

## Safety and scope boundary

The production helper performs three allocation-free linear inventory scans with O(1) auxiliary state: pooler discovery, required-slot cardinality/closed layer grammar, then exact shape and dtype accounting. It performs no SafeTensors, JSON, or config parsing, filesystem I/O, payload reads, mapping, model loading, tensor materialization, descriptor construction, attestation, publication, or serving work. It has no live production caller.

The input is the already bounded, duplicate-free `PreparedSafetensorsInventoryFacts` from #1424. This matcher does not claim to accept or repair duplicate names independently. A required F32 file offset marked four-byte aligned is still counted as copied: offset alignment does not prove a verified mmap base, host-endian disposition, or zero-copy eligibility.

Unrecognized names are permitted extras. A layer name becomes part of the closed required grammar only when it has exactly `encoder.layer.<INDEX>.<KNOWN_SUFFIX>`, where `<KNOWN_SUFFIX>` is one of the 16 loader roles. Unknown or nested suffixes remain extras; a recognized suffix with an empty, non-ASCII, leading-zero, overflowing, or out-of-range index fails typed without echoing attacker-controlled names.

## Exact required geometry

- word / position / token-type embeddings: `[V,H]` / `[P,H]` / `[T,H]`
- embedding LayerNorm weight and bias: `[H]`
- each layer query/key/value and attention-output weights: `[H,H]`; biases: `[H]`
- each layer attention-output and output LayerNorm pairs: `[H]`
- intermediate weight/bias: `[I,H]` / `[I]`
- output weight/bias: `[H,I]` / `[H]`
- optional pooler weight/bias: both present as `[H,H]` / `[H]`

The matcher derives the pooler pair first and passes it into the existing checked geometry analyzer, so callers cannot supply a geometry claim that disagrees with the inventory. Geometry errors precede layer/count/shape failures; recognized-layer grammar precedes subset cardinality; cardinality precedes shape and dtype accounting.

## TDD and mutation evidence

The compile-first RED exited 101 with 31 missing production-symbol errors and one unused-import warning. The final focused suite is GREEN at 13/13.

Ten production mutations were applied individually, observed RED, and restored to the frozen SHA:

1. weaken exact layer-suffix equality to `ends_with`, admitting suffix smuggling
2. map query-weight names to the key-weight typed role
3. validate a two-dimensional tensor's second dimension through axis zero
4. weaken the layer boundary from `index >= L` to `index > L`
5. disable canonical leading-zero rejection
6. classify every required F32 element as zero-copy instead of copied
7. merge BF16 elements into the F16 accounting bucket
8. treat the pooler bias name as pooler weight presence
9. replace checked accumulation with wrapping addition
10. transpose the required intermediate-weight shape from `[I,H]` to `[H,I]`

## Verification

```text
Focused prepared_tensor_inventory tests
13 passed; 0 failed

RUSTC_WRAPPER= cargo test -p lattice-inference --lib -j4
2,672 passed; 17 ignored; 0 failed
38.88s

PASS

cargo fmt --all -- --check
PASS

git diff --check / git diff --cached --check
PASS

repository pre-commit hook
PASS
```

All-target/all-feature clippy was not repeated because the parent #1424 already records the unchanged inherited GPU/WGPU lint failures, and this diff reaches none of those paths. Two independent final read-only reviews found 0 Blocker, 0 High, and 0 Medium findings on the frozen diff.

Evaluated at PR head `2469f7d0024abe012076062fbacf9d0f4d05199d` against the 25-target `lattice-inference` benchmark surface at base `a6e36177d627994eca41abbb961158e106ab2700`; this proof is void if either ref moves.

The diff changes no manifest, dependency or lockfile version, feature set, profile, build script, generated source, benchmark definition, or harness input. The reached `model/bert.rs` hunk adds only one private dormant module item and modifies no existing item. Recursive inspection of the new module found only new private items and `cfg(test)` tests: no global/static initializer, linker/codegen attribute, exported symbol, unsafe code, macro expansion affecting existing code, trait implementation on an existing type, allocator hook, or production callsite.

`e2e_bench` reaches the unchanged legacy BERT loader and `f16_convert_bench` reaches the unchanged legacy `SafetensorsFile` path; neither imports or calls the new dormant matcher. This slice is unmeasured, not performance-neutral.

The first same-handle reader, mapper, loader callsite, or prepared-mode activation must re-evaluate reachability and, if reachable, provide targeted measurement.

## Explicit exclusions / gates before activation

- same-handle bounded reads, snapshot capability, payload integrity, and TOCTOU closure
- exact required-tensor payload selection or a mapping authority
- actual zero-copy/materialization disposition and live resource leasing
- tokenizer/config parsing and vocabulary/token-id/sequence-cap cross-validation
- complete-inventory digest, effective descriptor, attestation, or publication
- loader integration, `BertModel` construction, caches, or serving
- shards/index, Qwen, Metal, remote providers, or public API

## Child draft

- #1426 — dormant bounded prepared-BERT config parser

### DIFF
--- a/crates/inference/src/model/bert.rs
+++ b/crates/inference/src/model/bert.rs
@@ -23,6 +23,9 @@ use tracing::warn;
 #[cfg_attr(not(test), allow(dead_code))]
 mod prepared_facts;
 
+#[cfg_attr(not(test), allow(dead_code))]
+mod prepared_tensor_inventory;
+
 /// Per-layer fused Q/K/V weight+bias, built once at model-load time from a
 /// `TransformerLayerWeights` layer's separate `query`/`key`/`value` tensors.
 ///
--- a/crates/inference/src/model/bert/prepared_tensor_inventory.rs
+++ b/crates/inference/src/model/bert/prepared_tensor_inventory.rs
@@ -1,0 +1,1040 @@
+//! Dormant required-tensor matching for sealed BERT preparation.
+//!
+//! This module consumes an already bounded, duplicate-free SafeTensors inventory.
+//! It derives pooler presence, validates checked BERT geometry, and matches the
+//! exact required tensor-name and shape closure without allocating. Unrelated
+//! tensors remain permitted and identity-bearing in the complete inventory.
+//!
+//! The returned census conservatively treats every required F32 tensor as copied.
+//! A file-offset alignment fact alone does not authorize a mapping or zero-copy
+//! disposition. This module performs no filesystem access, payload reads,
+//! mapping, model loading, publication, or serving work and has no live caller.
+
+use super::prepared_facts::{
+    BertGeometryLimits, BertPoolerMembers, PreparedBertFactError, PreparedBertGeometry,
+    RawBertConfigFacts, RequiredBertTensorCensus, analyze_prepared_geometry,
+};
+use crate::weights::prepared_safetensors_inventory::{
+    PreparedSafetensorsDtype, PreparedSafetensorsInventoryFacts, PreparedSafetensorsTensorView,
+};
+
+const LAYER_PREFIX: &[u8] = b"encoder.layer.";
+
+const LAYER_SUFFIXES: [(&[u8], PreparedBertLayerTensor); 16] = [
+    (
+        b".attention.self.query.weight",
+        PreparedBertLayerTensor::QueryWeight,
+    ),
+    (
+        b".attention.self.query.bias",
+        PreparedBertLayerTensor::QueryBias,
+    ),
+    (
+        b".attention.self.key.weight",
+        PreparedBertLayerTensor::KeyWeight,
+    ),
+    (
+        b".attention.self.key.bias",
+        PreparedBertLayerTensor::KeyBias,
+    ),
+    (
+        b".attention.self.value.weight",
+        PreparedBertLayerTensor::ValueWeight,
+    ),
+    (
+        b".attention.self.value.bias",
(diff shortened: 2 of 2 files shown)


ohdearquant:

PR #1431, opened 2026-08-17
feat(inference): classify prepared tokenizer JSON models

ohdearquant:
> _PR description authored by Codex (OpenAI agent) on behalf of @ohdearquant._

## Governance and stack

Draft implementation work for proposed Lattice ADR-088 and Khive ADR-160. Do not merge until the governing specification is accepted and the parent stack is ready.

Stacked on #1430. This slice is private, dormant, and has no production caller.

## What this adds

- a classifier for an already-decoded selected `tokenizer.json` `model.type`
- exact accepted spellings: `WordPiece`, `BPE`, `Unigram`, and `SentencePieceUnigram`
- three canonical facts: WordPiece, BPE, and Unigram
- typed missing, empty, and unsupported failures; wrong case is unsupported and never falls back

This slice deliberately starts after bounded JSON decoding. It does not inspect bytes, parse a document, validate vocabularies or token IDs, construct a tokenizer, inspect lower layouts, or claim content validation.

There is no filesystem/path access, allocation, attestation, loader integration, publication, serving callsite, or legacy behavior change.

## TDD evidence

Compile-first RED exited 101 with 6 missing production-symbol errors and one expected unused-import warning. GREEN locks all four supported spellings, their three canonical results, and missing/empty/unknown/wrong-case failures.

## Verification

```text
Focused prepared_tokenizer_json_model tests
2 passed; 0 failed

RUSTC_WRAPPER= cargo clippy -p lattice-inference --lib --tests -j4 -- -D warnings
PASS

cargo fmt --all -- --check
PASS

git diff --check
PASS

repository pre-commit hook
PASS

focused Blocker/High production review
0 Blocker; 0 High
```

## ADR-087 benchmark disposition

Evaluated at PR head `8ef720c124c5b97901c962c3d19f4a4163d1913a` against the 25-target `lattice-inference` benchmark surface at base `fa48e9e4974dfcd672b761069c1c40142a449046`; this proof is void if either ref moves.

The diff changes no manifest, dependency or lockfile version, feature set, profile, build script, generated source, benchmark definition, or harness input. The reached `model/bert.rs` hunk adds only one private dormant module item and modifies no existing item. Recursively, the new module contains only new private fixed-size enums, inherent/derived implementations, one borrowed-string classifier, and `cfg(test)` tests: no production caller, heap allocation, global/static initializer, linker/codegen attribute, exported symbol, unsafe/extern code, macro registration/export, allocator hook, or trait implementation on an existing type.

[...]

PR #1432, opened 2026-08-17
feat(inference): extract prepared config sequence caps

ohdearquant:

## Governance and stack

Draft implementation work for proposed Lattice ADR-088 and Khive ADR-160. Do not merge until the governing specification is accepted and the parent stack is ready.

Stacked on #1431. This slice is private, dormant, and has no production caller.

## What this adds

- prepared `config.json` sequence-candidate facts for `model_max_length` and the already-required `max_position_embeddings`
- exact in-file precedence: a present `model_max_length` wins; absence selects `max_position_embeddings`
- an owned deferred typed fault for a present string, zero, negative, fractional, exponent, or overflowing `model_max_length`, with no fallback to `max_position_embeddings`
- decoded-key duplicate detection, including literal/escaped aliases

Candidate-value faults are retained in the parsed facts instead of failing the outer geometry parse. This preserves whole-file precedence: a later valid `tokenizer_config.json` candidate may shadow the config candidate without observing its deferred fault. A literal/escaped duplicate remains a document-level config parser error.

This slice does not parse tokenizer configuration, cap at 2048, resolve cross-file precedence, compare against geometry/position rows, or construct a descriptor.

There is no filesystem/path access, allocation, attestation, loader integration, publication, serving callsite, or legacy behavior change.

## TDD evidence

Final-contract compile-first RED exited 101 with six missing production-symbol errors: four for `sequence_cap_candidate` and two for the deferred candidate error type. GREEN locks valid/absent precedence, deferred invalid-value faults without fallback, and literal/escaped duplicate rejection.

## Verification

```text
Focused prepared_config tests
16 passed; 0 failed

PASS

cargo fmt --all -- --check
PASS

git diff --check
PASS

repository pre-commit hook
PASS

focused Blocker/High production review
0 Blocker; 0 High
```

Evaluated at PR head `b94de84ec861b9e5de83b7482c17bceb9d54619a` against the 25-target `lattice-inference` benchmark surface at base `8ef720c124c5b97901c962c3d19f4a4163d1913a`; this proof is void if either ref moves.

The diff changes no manifest, dependency or lockfile version, feature set, profile, build script, generated source, benchmark definition, or harness input. It changes only the private dormant `prepared_config` module, which is outside ADR-087's corrected default A/B file surface, has no production caller, and is absent from every benchmark target's import and call surface. No reached benchmark file or symbol changes.

This is a pinned D3-A Step-1 unreachability disposition, not a Form-1 item-addition proof: existing dormant parser items change. The compiled crate still carries binary/codegen residual risk. This slice is unmeasured, not performance-neutral. Any live prepared-mode callsite must re-evaluate reachability and provide targeted measurement when required.

## Explicit exclusions

- tokenizer-config parsing or tokenizer-config-to-config cross-file precedence
- the 2048 cap, geometry/position-row validation, or realized descriptor fields
- filesystem inventory, handles, snapshots, copy, or TOCTOU closure
- attestation, model loading, publication, or serving
- public API, legacy behavior changes, Qwen, Metal, or remote providers

## Child draft

- #1433 — dormant parsed-facts sequence-cap adapter

### DIFF
--- a/crates/inference/src/model/bert/prepared_config.rs
+++ b/crates/inference/src/model/bert/prepared_config.rs
@@ -1,6 +1,9 @@
 //! Dormant checked parsing contract for prepared BERT `config.json` bytes.
 
 use super::prepared_facts::RawBertConfigFacts;
+use super::prepared_sequence_cap::{
+    PreparedBertSequenceCapKey, RawPreparedBertSequenceCapCandidate,
+};
 use std::num::NonZeroU64;
 use std::ops::Range;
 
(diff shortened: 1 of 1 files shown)


ohdearquant:

PR #1434, opened 2026-08-17
feat(inference): finalize prepared sequence cap facts

ohdearquant:
> _PR description authored by Codex (OpenAI agent) on behalf of @ohdearquant._

## Governance and stack

Draft implementation work for proposed Lattice ADR-088 and Khive ADR-160. Do not merge until the governing specification is accepted and the parent stack is ready.

Stacked on #1433. This slice is private, dormant, and has no production caller.

## What this adds

- finalization of an already-resolved sequence cap against `MatchedBertInventoryFacts`
- proof that the capped value does not exceed the position-embedding rows validated from the required tensor inventory
- realized facts preserving source, key, raw value, capped value, and validated position-row count
- a typed failure when the capped value exceeds validated rows

The API deliberately requires matched inventory facts rather than free geometry or a caller-supplied row count. This keeps actual tensor-shape validation in the authority chain.

This slice does not parse bytes, inspect files, build a tokenizer/model, construct a descriptor, or make the prepared path live.

There is no filesystem/path access, allocation, attestation, loader integration, publication, serving callsite, or legacy behavior change.

## TDD evidence

Compile-first RED exited 101 with four missing type/function errors and one expected unused-import warning. GREEN locks equality, cap-before-row comparison, exact overflow cases, and preservation of every resolved fact.

## Verification

```text
Focused prepared_sequence_cap_finalizer tests
2 passed; 0 failed

RUSTC_WRAPPER= cargo clippy -p lattice-inference --lib --tests -j4 -- -D warnings
PASS

cargo fmt --all -- --check
PASS

git diff --check
PASS

repository pre-commit hook
PASS

focused Blocker/High production review
0 Blocker; 0 High
```

## ADR-087 benchmark disposition

Evaluated at PR head `8f0afdd846ae26d58a2b3fb05ef0d17ac01a071a` against the 25-target `lattice-inference` benchmark surface at base `219c110acfef43e23de9801ee08114af31b9c445`; this proof is void if either ref moves.

The diff changes no manifest, dependency or lockfile version, feature set, profile, build script, generated source, benchmark definition, or harness input. The reached `model/bert.rs` hunk adds only one private dormant module item and modifies no existing item. Recursively, the new module contains only new private fixed-size facts/errors, ordinary accessors/finalization functions, and `cfg(test)` tests: no production caller, heap allocation, global/static initializer, linker/codegen attribute, exported symbol, unsafe/extern code, macro registration/export, allocator hook, or trait implementation on an existing type.

[...]

PR #1435, opened 2026-08-17
feat(inference): parse prepared tokenizer JSON model facts

ohdearquant:

## Governance and stack

Draft implementation work for proposed Lattice ADR-088 and Khive ADR-160. Do not merge until the governing specification is accepted and the parent stack is ready.

Stacked on #1434. This slice is private, dormant, and has no production caller.

## What this adds

- bounded, zero-allocation parsing of raw `tokenizer.json` bytes for the direct root `model.type` fact
- independent byte and logical two-pass work limits, checked arithmetic, platform-span admission, and fixed nesting depth 64
- complete strict JSON syntax validation before semantic extraction
- decoded direct-path matching, escaped-key/value handling, and duplicate `model`/`model.type` rejection
- reuse of #1431's exact four-spelling classifier and three canonical model facts

This is a model-type probe only. It does not validate vocabularies, merges, added/special token IDs, tokenizer construction, or any lower-priority representation.

There is no filesystem/path access, allocation, attestation, loader integration, publication, serving callsite, or legacy behavior change.

## TDD evidence

Compile-first RED exited 101 with 44 missing parser API/type errors. GREEN locks strict syntax-first parsing, direct decoded path semantics, duplicate rejection, the four supported spellings, fail-closed classification, and exact resource boundaries.

## Verification

```text
Focused prepared_tokenizer_json_model tests
7 passed; 0 failed

PASS

cargo fmt --all -- --check
PASS

git diff --check
PASS

repository pre-commit hook
PASS

focused Blocker/High production review
0 Blocker; 0 High
```

Evaluated at PR head `f7cc28ccf5042de34005cdab217ffab50932447e` against the 25-target `lattice-inference` benchmark surface at base `8f0afdd846ae26d58a2b3fb05ef0d17ac01a071a`; this proof is void if either ref moves.

The diff changes no manifest, dependency or lockfile version, feature set, profile, build script, generated source, benchmark definition, or harness input. It changes only the existing private dormant `prepared_tokenizer_json_model` file, which is outside ADR-087's corrected default A/B file surface, has no production caller, and is absent from every benchmark target's import and call surface. No reached benchmark file or symbol changes.

Production changes append only new private constants, fixed-size types, inherent/derived implementations, and ordinary parser functions. They add no heap allocation, global/static initializer, linker/codegen attribute, exported symbol, unsafe/extern code, macro registration/export, allocator hook, or trait implementation on an existing type.

This is a pinned D3-A Step-1 unreachability disposition. A compiled-artifact residual remains. This slice is unmeasured, not performance-neutral. Any live prepared-mode callsite must re-evaluate reachability and provide targeted measurement when required.

## Explicit exclusions

- tokenizer vocabulary/cardinality, merges, ordinary/added/special IDs, and tokenizer construction
- lower-priority layout content parsing or fallback
- filesystem inventory, handles, snapshots, copy, or TOCTOU closure
- attestation, model loading, publication, or serving
- public API, legacy behavior changes, Qwen, Metal, or remote providers

## Child draft

- #1436 — bounded shared `vocab.txt` dense-ID facts

### DIFF
--- a/crates/inference/src/model/bert/prepared_tokenizer_json_model.rs
+++ b/crates/inference/src/model/bert/prepared_tokenizer_json_model.rs
@@ -68,3 +68,1163 @@ mod tests {
         }
     }
 }
+
+#[cfg(test)]
+mod parser_tests {
+    use super::*;
+    use std::num::NonZeroU64;
+
+    fn nz(value: u64) -> NonZeroU64 {
+        NonZeroU64::new(value).unwrap()
+    }
+
+    fn exact_limits(bytes: &[u8]) -> PreparedBertTokenizerJsonLimits {
+        let len = u64::try_from(bytes.len()).unwrap();
+        PreparedBertTokenizerJsonLimits::try_new(nz(len), nz(len * 2)).unwrap()
+    }
+
+    fn parse(
+        bytes: &[u8],
+    ) -> Result<ParsedBertTokenizerJsonModelFacts, PreparedBertTokenizerJsonParseError> {
+        parse_prepared_bert_tokenizer_json_model(bytes, &exact_limits(bytes))
+    }
+
+    #[test]
+    fn four_supported_direct_model_type_spellings_parse_including_escapes() {
+        use PreparedBertTokenizerJsonModel::{Bpe, Unigram, WordPiece};
+
+        for (json, expected) in [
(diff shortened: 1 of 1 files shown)


ohdearquant:

PR #1496, opened 2026-09-09
feat(tune): time every training phase, and refuse a timing run that shares its machine

ohdearquant:
## Why

The exact-gradient trainer printed exactly one duration, its step loop, and that figure turns out to
cover about a third of a run. Everything before the loop — the two prefix-cache builds, the
consistency check, and the two baseline scoring passes — was real work that printed nothing. A cost
model built from the step loop alone understates by a term that grows with the pool and the held-out
size, which is precisely the term a scheduling decision depends on.

Measured at a representative size (32 train samples, 22 held out), the newly-reported phases:

| phase | cost |
|---|---|
| train prefix cache | 596.0s (1815 completion positions / 32 samples) |
| held-out prefix cache | 590.3s (1082 positions / 22 samples) |
| consistency check | 23.6s |
| baseline train pass | 377.6s |
| baseline valid pass | 398.2s |

## What changed

**Phase timing** (`full_driver.rs`). Each phase now states its own duration, which turns an external
wall clock into a *check* on this log rather than the only place the time can be found.

**`time_run.sh`** puts an external wall clock around a whole command and records the program's own
figure beside it rather than instead of it, so the gap between the two is visible.

**`--exclusive PATTERN`**, and this one is here because of a specific failure. A run of this trainer
took 3994s. The identical run, same binary and same inputs, on an idle machine took 615s. The driver
*had* checked for competing processes, and printed:

```
=== nothing of ours running: 1 ===
```

The label asserts the answer; the value beside it contradicts it. The run proceeded on a machine it
was sharing with another copy of itself. A printed count is not a guard. `--exclusive` refuses with a
distinct exit code and names the offending pids, and it snapshots the process table before any grep
runs so the search cannot match its own arguments. Competing counts are recorded at start *and* end,
because a machine that was quiet at t=0 can acquire a competitor at minute 13.

The stray process came from an earlier run that had been discarded as invalid. Discarding a
*measurement* says nothing about the *process*, which keeps running and keeps consuming the resource
the next measurement needs.

## Tests

`test_time_run_exclusive.sh`, eleven arms, including:

- a refused run must not execute its command (the fail-open arm)
- a non-matching pattern must still proceed, so the refusal is a difference and not a constant
- the command's own exit code survives the wrapper
- **the decoy must carry the marker being searched for.** The first version did not: `bash -c` execs
  a single simple command directly, so the marker never reached the process table, the guard
  correctly found nothing, and the test reported the guard as failing open. A decoy that does not
  carry the property under test tests the test, not the subject.

## Also

ADR-079's record is corrected. The GDN LoRA weight-gradient gap it named as open closed one day
after it was written, by #792 and #1322, and the record was never updated. The original text is kept
for provenance rather than rewritten, since it was accurate on the day it was written.

The remaining scripts size a run from a real one, score the paired metric, build a zero-B adapter for
identity checks, and verify split integrity.


ohdearquant:
Closing: this branch was cut from a stale base and replays work that already landed in #1495, so it would revert 81 lines of full_driver.rs plus changes in three other files. Only the exclusivity guard and its tests are genuinely new; reopening those on top of current main.

Issue #1517, opened 2026-09-10
bench(inference): six Metal bench targets self-deadlock under bench-compare — the harness holds the GPU lock the target acquires


ohdearquant:
## What happens

Any `crates/inference` bench target that calls `measurement::gpu_test_lock()` stalls for 30 minutes
and then panics when it is run through `scripts/bench-compare.sh` or `scripts/bench-command.sh`.

Both entry points run the measured command under `scripts/lib/bench_supervision.py`, which acquires
its locks through `scripts/lib/bench-locks.py`. That file takes two machine-wide locks
(`bench-locks.py:66-67`, applied at `:255`):

```
BENCH_WINDOW = "/tmp/lion-bench-window.lock"
GPU_LOCK     = "/tmp/lion-metal-gpu-test.lock"
```

The bench binary then opens the **same path** in `measurement.rs::gpu_test_lock()` and polls
`File::try_lock` every 500 ms until `GPU_MACHINE_LOCK_TIMEOUT` (`measurement.rs:22`, 30 minutes), at
which point it panics telling the reader that "a Metal test run elsewhere on this machine is
wedged". The other run is its own parent.

## Measured

Running `bench-compare.sh` with `BENCHES_INFERENCE=metal_decode_bench
CARGO_FEATURES_INFERENCE=metal-gpu,f16 BENCH_GROUPS_INFERENCE=lora_rank8`, on an idle machine that
passed the harness's own preflight (AC, thermal nominal, idle 98.8%):

`lsof` on the lock while the bench was "running" — three processes, one inode:

```
python3.1 25018  4u  REG  ...  18191356  .../lion-metal-gpu-test.lock   <- bench-locks.py
python3.1 25028  4u  REG  ...  18191356  .../lion-metal-gpu-test.lock   <- bench_supervision.py
metal_dec 26219  3w  REG  ...  18191356  .../lion-metal-gpu-test.lock   <- the bench, queued
```

`sample` of the bench process, 1567 of 1568 samples on one stack:

```
metal_decode_bench::main
  metal_decode_bench::load_q4_state
    lattice_inference::measurement::gpu_test_lock
      std::thread::functions::sleep -> nanosleep -> __semwait_signal
```

Duty cycle after four minutes of wall time: `etime 04:01, time 0:00.03, state SN`. Thirty
milliseconds of CPU. The harness log prints nothing after `--- Building + benching BASE ---`, so from
the outside it is indistinguishable from a slow build.

## Population

[...]

PR #1524, opened 2026-09-10
fix(inference): hold one GPU lock across each benchmark target list

### LINKED ISSUE #1523: bench: three Metal benchmarks hold gpu_test_lock only during setup, so the measured region runs unlocked
Three of the six GPU benchmarks bind their `gpu_test_lock()` guard inside a **setup function**, so the guard drops when that function returns and the measured region runs with no machine lock held. `gpu_test_lock()`'s documented contract is whole-operation serialization; these call sites serialize model loading and nothing else.

## Per-file disposition

Enumerated over every tracked bench source that calls `gpu_test_lock` (6 files, 9 call sites):

| File | Site | Guard scope | Covers the measured region? |
| --- | --- | --- | --- |
| `benches/metal_decode_bench.rs` | `:91`, `:103` | `load_q4_state()` / `load_q8_state()` | **No** — both return the state; `forward_prefill` / `forward_step` run after |
| `benches/lm_head_bench.rs` | `:176`, `:205` | `setup_lm_head_fixture()` | **No** — returns `Ok(Some(fixture))`; the fixture is benched later |
| `benches/cross_turn_prefix_cache_bench.rs` | `:64` | `load_state_and_tokenizer()` | **No** — returns the state |
| `benches/decode_attn_bench.rs` | `:25` | `bench_locked_metal_decode()` body | Yes |
| `benches/topk_readback.rs` | `:31` | `bench_locked_metal_groups()` body | Yes |
| `benches/mtp_decode.rs` | `:78`, `:174` | the `#[cfg(...)]` block that also contains the measurement | Yes |

`metal_decode_bench.rs` is the clearest instance: the guard is bound at `:91` inside `load_q4_state`, that function returns at `:93`, and the benched work is at `:203-209` — `state.forward_prefill(...)` followed by `state.forward_step(...)` inside `b.iter(...)`.

## Why this matters more than a missing lock usually would

The lock exists so a timing run is not contending with another GPU process. For these three benchmarks the isolation applied to checkpoint loading, which is not timed, and not to the measured decode, which is. Any number produced by them could have been taken while another Metal process was running, and nothing in the output distinguishes a contended run from a clean one.

That interacts badly with the project's own rule that a performance claim needs an A/B: the A/B ran, and the isolation its validity depends on was not held for the measured portion.

## Suggested shape

Hold the guard in the benchmark function for the whole span that touches the device, and have setup helpers return without owning it — either by acquiring in the caller before setup and keeping it alive across the measurement, or by returning the guard alongside the state so its lifetime is the caller's choice. The three already-correct files show both working shapes.

Worth noting for whoever picks this up: extending these guards is a prerequisite for any change that stops an outer supervisor from holding the GPU lock on the benchmark's behalf, because such a change relies on the benchmark protecting its own measured region. Related: #1517.

Found by a source audit of every `gpu_test_lock` call site, not by an observed contended measurement — no claim is made here that a specific published number was contaminated, only that the guarantee was absent.

ohdearquant:
> _Implementation authored by Codex (OpenAI agent) on behalf of @ohdearquant; this description and the verification below are the maintainer's._

Three Metal benchmarks bound their GPU lock inside a setup function, so the guard dropped when that function returned and the measured region ran unprotected. Two further shapes were found while fixing those: `mtp_decode` took a separate guard inside each of its two arms, releasing the lock between the two groups whose *difference* is the benchmark's entire output, and `topk_readback` ran five timed CPU groups in the same binary outside its Metal group's guard. Contamination that lands on one arm of a comparison biases the difference; it does not average out, which makes the `mtp_decode` shape worse than a plainly missing lock.

All six reviewed Metal benchmark targets now register a single wrapper whose guard spans the complete ordered target list. `topk_readback` keeps its four CPU groups, six Metal groups and final CPU group in their original order — eleven calls, unchanged sequence — and the five CPU groups are deliberately inside the span, because the harness's promise is timing isolation for the whole run and not only GPU exclusivity. `mtp_decode` gets one entry point over both arms rather than a guard per arm.

The source contract in `metal_measurement_lock_contract.rs` now checks Criterion registration and explicit caller chains, not just guard call sites. That matters because two of the three defect shapes above are not properties of a call site at all: every individual guard binding in `mtp_decode` is textbook-correct, and so is the guarded target in `topk_readback`. Both are properties of the `criterion_group!` registration — what else the binary runs, and in what order, relative to where the lock is held. Rejection cases cover setup-only and per-arm guards, partial spans, registration bypasses, aliases and shadowing, deferred and conditional calls, and non-bare arm arguments.

### Validation

Re-run by the maintainer on the host at this head, not relayed:

- `cargo test --locked -p lattice-inference --test metal_measurement_lock_contract` → **63 passed, 0 failed**.
- `cargo check --locked -p lattice-inference --bench metal_decode_bench --bench lm_head_bench --bench cross_turn_prefix_cache_bench --bench mtp_decode --bench topk_readback --bench decode_attn_bench --features metal-gpu,f16,bench-internals` → clean. No GPU work executed.
- Mutation control, expectation registered before it ran: restoring the parent commit's five bench sources under this PR's contract test must fail. It failed, 61 passed / 2 failed, naming all five files. Restoring the sources returned 63/63 with a clean tree at the same head. A contract that still passed against the defect it describes would be decoration.

### Bench-compare disposition: no structural waiver is available here, and the comparison it requires is blocked

An earlier version of this description claimed the ADR-058 structural exception on a cfg-gate argument. That was wrong, and it is corrected here rather than quietly dropped; the original wording is quoted in a comment below.

`CLAUDE.md` at this head excludes exactly this kind of change from the exception:

> a change to a bench target's own definition or to the bench harness has no absent call path to name and no `cfg` or `required-features` gate to name closed: the change alters what gets compiled or how the runner invokes it before any function body executes. No `cfg`-gate or absent-call-path proof is available for these

and the reachability rule it sits under says the exception applies

> Only when no declared target reaches the change

This diff *is* the declared targets — changed registrations and guard spans in five of them — so it is reached by construction. A cfg gate describes what happens outside the Metal configuration; it says nothing about the configuration in which these benchmarks are actually measured.

**Why the required comparison is not simply run.** Both sanctioned entry points execute the measured command under a supervisor that holds `/tmp/lion-metal-gpu-test.lock`, and all six of these targets acquire that same lock inside the binary. The run stalls for thirty minutes and then panics, blaming an unrelated process. That is #1517, and it is a measurement-path limitation, not an unreachability proof — the distinction matters, because treating a blocked instrument as an absent call path is how an invalid waiver gets written.

**Why this merges with the comparison owed rather than held.** The #1517 fix hands lock ownership from the supervisor to the target per invocation, and that is only sound once every one of these six protects its whole measured run — which is what this PR establishes. Holding this for a measurement that #1517 blocks, while #1517 waits on this, is a deadlock in the process rather than a safeguard. The comparison is tracked as owed and runs on the measurement host once #1517 lands.

**Residual risk, plainly.** This change is unmeasured, not shown safe. What it can move is benchmark timings, by removing contamination the previous registration permitted; the direction and size are unknown until the comparison runs. It touches no production code path — the diff is confined to `benches/`, the contract test, and one documentation paragraph. An earlier version of this description asserted that an A/B "would compare a known-contaminated arm against a clean one". That overstates the evidence: the missing span *permits* contamination on the base arm, and no run was measured to establish that it occurred.

### Scope, stated so it is not overread

The new coverage check closes admission for benchmarks: the Metal-marker inventory filtered to `benches/` must equal the set of files carrying a harness contract, so a new Metal benchmark cannot be added without one. This is the gap that let the original defect through — the pre-existing lifetime check admits a target only when it lexically contains `Device::system_default`, `new_command_buffer` or `new_command_buffer_with_unretained_references`, and `metal_decode_bench.rs` contains none of them because it drives the GPU through an already-constructed state object. A guard that skips a subject produces the same output as a guard that passes it.

That narrow filter still governs `examples/` and `src/bin/`. Counting lexically over each target's own source — the contract itself parses the module closure, so the exact split may move by a target or two — 6 of the 38 non-bench inventory entries carry one of those three markers and 32 do not. That is a reach count, not a defect count: deciding it needs a per-target classification, and the obvious-looking shortcut of excluding anything named like a production binary does not survive contact with the source. `src/bin/lattice/main.rs` is the counterexample — its `prune-score` subcommand computes dense and pruned perplexity, takes their delta and gates on it (`prune_score.rs:167-171,253-254`), so it is a mixed-purpose measurement target at mode scope. The reach is named here rather than left to be inferred from the fact that a check exists.

One constraint the diff imposes, fail-closed by design: the six reviewed Metal benches must use the short `criterion_group!(name, target)` form. The long `name = ...; config = ...; targets = ...` form is rejected with a parse error rather than being evaluated for ownership, which is why `topk_readback` moved to the short form with the default config it already used.

Fixes #1523.


ohdearquant:
## Correction: the bench-compare disposition I published was an invalid waiver, and I have replaced it

The original description said, under **Bench-compare disposition**:

> This takes the ADR-058 structural exception rather than an A/B, and here is what proves it rather than asserting it:
>
> - **cfg gate.** Every guard binding removed or added sits under `#[cfg(all(target_os = "macos", feature = "metal-gpu"))]` [...]
> - **identical effective source.** Outside the gate the only surviving change is the registration entry point [...]
>
> Inside the gate the change is *intended* to move timings, by removing contamination the previous registration permitted. A before/after A/B would therefore compare a known-contaminated arm against a clean one [...]

That does not meet this repository's rule, and the rule says so directly. `CLAUDE.md` at this head:

> a change to a bench target's own definition or to the bench harness has no absent call path to name and no `cfg` or `required-features` gate to name closed [...] No `cfg`-gate or absent-call-path proof is available for these

The exception applies **"Only when no declared target reaches the change"**. This diff *is* five declared targets. A cfg gate answers what happens outside the Metal configuration; the benchmarks are measured inside it. So the argument was sound about a neighbouring question and wrong about the one the rule asks.

Two smaller things in the same text, both mine:

- *"an A/B would compare a known-contaminated arm against a clean one"* overstates the evidence. The missing span **permits** contamination on the base arm; no run was measured to show it occurred. A permitted failure mode is not a measured one.
- The scope paragraph used "production binaries are not measurement harnesses" with `src/bin/lattice/main.rs` as an example. Wrong: its `prune-score` subcommand computes dense and pruned perplexity, takes the delta and gates on it (`prune_score.rs:167-171`, `:253-254`). It is a mixed-purpose measurement target. The corrected paragraph names it as the counterexample instead.

The replacement disposition states that no waiver is available, that the required comparison is blocked by #1517 through both sanctioned entry points, that this is a measurement-path limitation rather than an unreachability proof, and that the comparison is **owed** rather than waived, with the residual risk stated as unmeasured rather than safe.

Recorded here rather than by editing quietly, because the wrong text was the part a reader would have relied on, and because "the instructed route cannot execute" is exactly the situation in which an invalid structural waiver is most tempting to write.

One further correction, to a claim I made privately about the implementation evidence and should not leave standing: I characterised the twelve pre-existing rejection controls as covering synthetic fixtures rather than the real prior sources. That is wrong. Five of them restore the actual parent-commit bytes of the five changed benchmark files — I verified those snapshots byte-identical to the parent commit — and five more retain the new wrapper while restoring the old setup-only or per-arm guards. My own reversal on the host is replication of that, not a missing category of test.

PR #1529, opened 2026-09-11
docs: stop asserting a bench-target count that nothing re-derives

ohdearquant:
## What

`CLAUDE.md` asserted that `crates/inference` declares 23 bench targets. At this ref the manifest
declares 25.

## Why remove the number instead of correcting it

The sentence's argument is that the two targets `bench-compare` runs by default are not the
population a reachability proof must search. That argument does not need a count — the count is the
only part of it that can rot, and it rotted within the life of the paragraph.

Updating 23 to 25 would reset the same clock. People who add a `[[bench]]` table do not grep prose
for counts, and nothing in CI recomputes one written in a document. So the literal is replaced by
the command that produces the current figure at whatever ref the reader is working from:

```
git show <ref>:crates/inference/Cargo.toml | grep -c '^\[\[bench\]\]'
```

The replacement text also records that the old literal was wrong. That is deliberate rather than
tidy: a reader who is told only the correct method may still write the next number down, whereas one
who is shown that this exact sentence already carried a stale count has a reason not to.

## Verification

- 25 is `git show HEAD:crates/inference/Cargo.toml | grep -c '^\[\[bench\]\]'`, read at the ref
  rather than from the working tree.
- Documentation only. No crate source, manifest, feature, profile, or bench definition is touched,
  so there is nothing here a benchmark could execute.


ohdearquant (commented):

Verdict on head ab8d4662c92acf7d472549e8155bf80f7406bdd3: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

PR #1531, opened 2026-09-11
docs(adr): ADR-079 Amendment 1 — G2's re-entry trigger has fired; fix the reload contract

ohdearquant:
ADR-079's G2 gap ("in-process gate reload") was written with an explicit re-entry
condition: it fires "if in-process closure becomes a requirement". The adaptive-loop
work now queued is that requirement, so the trigger has fired while the gap itself is
still open. This records that, and pins the contract the implementation will build
against, following the ADR's own rule that a gap closure is a recorded status update
rather than a silent flip.

No code changes. The ADR status line gains an "amended" marker and Amendment 1 is
appended.

## What the amendment establishes, read from source rather than from the gap text

**The closure does not invert the dependency direction.** `update_router` is in
`lattice-tune`, `AdapterRouter` in `lattice-inference`, and inference does not depend on
tune — which reads like an obstacle and is not one. `RouterDelta.network_bytes` is
already a complete `Network::to_bytes()` blob (`crates/tune/src/lora/router_update.rs:141-147`)
and `AdapterRouter`'s field is a fann `Network`, so a reload path needs only
`Network::from_bytes`, already reachable from inference through `lattice-fann` under
`mixture` (`crates/inference/Cargo.toml:51`). The leaf-crate rule is unaffected.

**The obvious failure mode to specify does not exist, and the amendment says so.** The
natural thing to write is that `reload` must reject a mismatched gate or `route` would
mis-select. `route` already validates both dimensions at use time and fails closed:
`RouterError::InputSizeMismatch` on a context-vector width mismatch
(`crates/inference/src/mixture.rs:158-164`), and `RouterError::GateTooNarrow` when the
gate is narrower than `k` (`:171-178`, whose comment says the check exists so a narrow
gate "beats a panic inside `select_nth_unstable`"). A mismatched reload cannot produce a
wrong adapter selection. This is stated explicitly in the amendment because an
implementer would otherwise accept the plausible version from an ADR without re-reading
the routing body.

**What dimension checking at reload actually buys is blast radius.** A mismatched swap
turns every subsequent request into an error, and `reload` is the last moment a caller
can still choose to keep serving on the previous gate. So the contract is: reject a
differing width and leave `self` unmutated — the gate swaps wholly or not at all, and a
bad refit costs one failed reload rather than an outage.

One behaviour is deliberately left alone: a gate narrower than the adapter pool with `k`
small enough to stay inside `GateTooNarrow` silently restricts routing to the first
`num_outputs` adapters. That is pre-existing, reachable through `new` today, and the
equal-dimension rule means reload cannot introduce it.

## Test contract

G2 names the test as a sequence (route → feedback → refit → reload → route). The
amendment restates it as an assertion — the second `route` returns a *different*
selection for the same context and pool, attributable to the feedback — because the
sequence as written would pass against a `reload` that discards its argument. A second
arm pins the no-partial-mutation clause: reload a wrong-width gate, assert the error,
assert `route` still succeeds on the original gate.

### DIFF
--- a/docs/adr/ADR-079-adaptive-micro-lora-brain.md
+++ b/docs/adr/ADR-079-adaptive-micro-lora-brain.md
@@ -1,6 +1,6 @@
 # ADR-079: Adaptive Micro-LoRA Brain — the Composed Train → Govern → Compose → Route → Consume Loop
 
-**Status**: Proposed
+**Status**: Proposed (amended 2026-09-10 — see Amendment 1)
 **Date**: 2026-07-09
 **Crate**: lattice-tune / lattice-inference / lattice-fann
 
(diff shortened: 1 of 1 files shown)


ohdearquant:
Follow-up evidence found while scoping the serving-side work, after this was opened. It does not
change the amendment; it makes one of its clauses more load-bearing than the text claims.

`MetalQwen35State` is `!Send` and is confined to a dedicated worker thread in `lattice_serve`
(`crates/inference/src/bin/lattice_serve.rs:39`, constructed at `:1769-1795`). So in a serving
runtime, a refit gate produced anywhere else has to cross a thread boundary to take effect —
`AdapterRouter::reload` will be called from that worker, never from a request handler.

Two consequences:

1. The amendment's "reject a differing width **and** leave `self` unmutated" rule matters more
   across a channel than it does in-process. On the near side a caller could inspect the router
   after a failed swap; across a channel the returned error is the only thing it gets. A partial
   mutation there is unobservable by construction.

2. It reinforces the "between requests" framing rather than complicating it. The worker's queue is
   typed to a *generation job* (`crates/inference/src/serve/metal_worker.rs:239`) carrying an
   admission permit with an exactly-once release contract, so a reload is naturally a control
   message drained between jobs — which is the boundary G2 describes, made mechanical instead of
   maintained by convention.

Nothing here needs the ADR text to change. Recording it so the implementation does not rediscover
the boundary from scratch.

PR #1537, opened 2026-09-11
feat(data): micro-LoRA corpus curation and parser-verified DSL synthesis

ohdearquant:
Adds three data-curation pieces under `scripts/microlora/`, plus tests for each lane:

- `mine_commit_msg.py` — reproducible curation of a commit-message corpus, with trailer and
  PR-suffix stripping and length gates.
- `synth_khive_dsl.py` — synthesis of request-DSL pairs that are verified against a parser
  rather than assumed well-formed.
- `dsl_validator.rs` + `build_dsl_validator.py` — the validator behind that verification, brought
  under the formatting-coverage gate at its own edition.
- `curation_guard.py` — the sensitive-content screen and the output-path guard, one definition
  shared by both lanes.

All files are additions. The only modification to existing content is a single line in
`scripts/rust-fmt-coverage.allow`.

## Status

This branch was opened before its logic had been checked. That check is done and the head
carries its fixes:

- Recorded-call rows are screened for addresses and credential shapes before any text reaches a
  prompt or completion; a flagged row is skipped whole, nothing is redacted, because the
  completion is the recorded text. The screen now covers URL credentials, no-TLD addresses,
  Slack/Stripe/AWS/Google/JWT token shapes and unquoted assignments, and leaves decorators,
  scoped package names, environment reads and placeholder values alone.
- The commit miner refuses an output path a repository tracks or does not ignore, refuses an
  existing directory holding unrelated files, and writes splits without following a symlink at
  the output name. The DSL lane already had this guard; it is now shared.
- A registered verb with neither a reviewed template nor a listed exclusion stops generation
  before any row is written, and the refusal names every such verb. The verb families the live
  registry gained after the templates were written are listed with reasons.
- The `git.push` template matches the live schema (`expected_local`/`expected_remote`), and the
  type vocabulary accepts the spellings the live capture uses.

The credential probes in `test_mine_commit_msg.py` are assembled at runtime rather than written as
literals so that no line of a public test file scans as a token; a literal-shape grep over the file
is empty while the same grep over the assembled strings matches every probe.

Each fix has a test that fails with the fix reversed. Against a fresh 129-verb capture and a
validator built at the current parser: 86 verbs templated, 43 excluded, 0 template/schema
mismatches; a full generation run produced 3850 rows with exact duplicates as the only
candidate filter and no parser rejections. Local suite: 39 tests, 0 failures (7 skips are the
arms that need a capture and a validator on the command line).

## Follow-ups, not in this PR

- Several existing tests in `test_synth_khive_dsl.py` stay green under a named mutation of the
  code they cover (the split-guard empty-verb branch, the schema field-shape check, the registry
  total check, the unknown-child refusal, the AST round-trip marker, among others). They need
  fixtures under which the mutation is visible.
- `check_split_integrity.py` checks prompt/completion overlap and length balance between two
  files; it does not re-derive the whole-verb partition the generator enforces, so a same-verb
  row in two splits passes it. The README now says so; extending it is separate work.
- Schema checking is over flat parameter types; conditional `input_schema` constraints, enums
  and bounds are not validated.
- `dsl_validator.rs` renders arrays and objects twice (`render_arg` and `render_value`).


ohdearquant (commented):

Verdict on head 0d28711f1a253c3fc96cc613ae803830bdca6e36: REQUEST-CHANGES, 3 blocking findings. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

PR #1561, opened 2026-09-14
fix(inference): reject an mrope_section whose declared lane counts are not the ones realized

### LINKED ISSUE #1507: mrope_section passes a sum check, then the lane assignment silently uses a different T/H/W split
`build_mrope_tables` in `crates/inference/src/vision/qwen35_mrope.rs` validates `mrope_section`
with two checks — exactly three entries, and a sum equal to `rope_half` — and then assigns lanes
with a pattern those checks cannot constrain:

```rust
let axis_val = match (i % 3, i / 3) {
    (1, section_idx) if section_idx < mrope_section[1] => h,
    (2, section_idx) if section_idx < mrope_section[2] => w,
    _ => t,
};
```

`mrope_section[0]` is never read. T is whatever the H and W caps leave behind. So the declared
counts are honoured only when they happen to equal the cyclic schedule that `rope_half` induces;
outside that set the config is accepted and a different schedule is silently used.

Realized T/H/W for `rope_half = 32`, by running the match above:

| declared `mrope_section` | sum | realized T/H/W | |
|---|---:|---|---|
| `[11, 11, 10]` | 32 | 11 / 11 / 10 | matches (this is the shipped 0.8B config) |
| `[12, 10, 10]` | 32 | 12 / 10 / 10 | matches |
| `[32, 0, 0]` | 32 | 32 / 0 / 0 | matches |
| `[10, 11, 11]` | 32 | 11 / 11 / 10 | **diverges** |
| `[11, 10, 11]` | 32 | 12 / 10 / 10 | **diverges** |
| `[0, 16, 16]` | 32 | 11 / 11 / 10 | **diverges** — a config asking for zero time lanes gets eleven |

Every row passes both gates. Nothing downstream re-checks: `build_mrope_tables` returns the tables
immediately after the loop, and no other site compares a realized schedule against the declared
section.

The doc comment above the function is accurate about the mechanism — it says H and W overwrite
lanes "up to its declared section length" — while the same comment introduces the three entries as
"(T, H, W) section lengths". Both readings are in the same paragraph, and only one of them is what
the code does.

The shipped Qwen3.5-0.8B config is `[11, 11, 10]`, which is in the matching set, so nothing in
production is affected today. The exposure is a third-party or future checkpoint whose section is
sum-valid and outside it: it loads, runs, and produces rotary tables that do not match its own
config, with no error and no warning.

Either fix is small and they are not equivalent:

1. **Make the gate express the constraint.** Compute the realized schedule and reject when it
   differs from the declared section. This turns a silent substitution into a load-time error and
   is the fail-closed reading of what `mrope_section` means.
2. **Make the contract match the code.** Document that `mrope_section[1]` and `[2]` are caps
   applied to a fixed `i % 3` interleave, that `[0]` is unread, and that T is the remainder — then
   the accepted set is deliberate rather than accidental.

(1) is the safer default. Whichever is chosen, `mrope_section[0]` being unread should be visible
at the gate rather than inferable only from the match arms.

ohdearquant:
Closes #1507.

## The defect

`build_cos_sin` validates `mrope_section` with two checks — exactly three entries and a sum equal to `rope_half` — then assigns lanes with a pattern neither check can constrain:

```rust
let axis_val = match (i % 3, i / 3) {
    (1, section_idx) if section_idx < mrope_section[1] => h,
    (2, section_idx) if section_idx < mrope_section[2] => w,
    _ => t,
};
```

`mrope_section[0]` is never read, and T is whatever the H and W caps leave behind. A sum-valid section outside the set the cyclic schedule happens to induce is accepted, and a different schedule is silently used. Reproduced by extracting the match arms unchanged into a standalone program:

| declared | `rope_half` | realized |
|---|---:|---|
| `[11, 11, 10]` | 32 | `[11, 11, 10]` |
| `[10, 11, 11]` | 32 | `[11, 11, 10]` |
| `[11, 10, 11]` | 32 | `[12, 10, 10]` |
| `[0, 16, 16]` | 32 | `[11, 11, 10]` |
| `[16, 24, 24]` | 64 | `[22, 21, 21]` |
| `[22, 21, 21]` | 64 | `[22, 21, 21]` |
| `[20, 6, 6]` | 32 | `[20, 6, 6]` |

The shipped Qwen3.5-0.8B config is `[11, 11, 10]`, which realizes itself, so nothing in production is affected today. The exposure is a checkpoint whose section is sum-valid and outside that set: it loads, runs, and produces rotary tables that do not match its own config, with no error and no warning.

## The fix

Compute the realized `(T, H, W)` counts by running the same lane assignment the table loop uses, and reject when the realized triple differs from the declared one. The error names both triples and `rope_half`. All three declared entries are now read at the gate, so the unread-entry property is gone by construction rather than by comment. The doc comment carried two different contracts in one paragraph — "(T, H, W) section lengths" and a saturating strided assignment — and now states the one that is enforced.

The lane-assignment loop, the table arithmetic, and every accepted config's output are unchanged.

## Tests

Expected outcomes were recorded before running.

| arm | before fix | after fix |
| --- | --- | --- |
| shipped `[11, 11, 10]`, half 32 | passes (preservation control) | passes, 192/192 f32 bit patterns unchanged |
| `[10, 11, 11]`, half 32 | fails: returns `Ok` | rejected, realized `[11, 11, 10]` |
| `[11, 10, 11]`, half 32 | fails: returns `Ok` | rejected, realized `[12, 10, 10]` |
| `[0, 16, 16]`, half 32 | fails: returns `Ok` | rejected, realized `[11, 11, 10]` |
| `[16, 24, 24]`, half 64 | fails: returns `Ok` | rejected, realized `[22, 21, 21]` |
| `[20, 6, 6]` half 32, `[22, 21, 21]` half 64 | passes (acceptance controls) | passes |

The acceptance controls matter: without them the gate could be a single-value allowlist wearing a computation. Each rejection case runs with both empty positions and one position, so the validation cannot depend on entering the token loop — the pre-fix build returns empty tables successfully for the empty case.

`[16, 24, 24]` at half 64 was previously a positive sample in `lane_schedule_matches_hf_saturating_overwrite`; it is a rejection case now, and the reference helper's own model is unchanged. Every lane, not just the counts, was compared against that helper across all seven cases with no disagreement.

**Mutation.** The gate was reverse-applied as a patch, the source touched to force a rebuild, and the suite re-run: 21 passed, the four rejection tests failed with `Ok`. Restoring the gate and touching again: 25 passed. A test that still passes with the fix reverted is decoration, so this was run rather than asserted.

## bench-compare disposition — no structural waiver is claimed

All 30 declared `[[bench]]` targets (25 in `lattice-inference`, 5 in `lattice-embed`, counted at the base ref) were searched, together with `lattice-fann`'s and `lattice-transport`'s, including non-default `required-features` and `cfg` branches. None reaches `build_cos_sin`, `build_mrope_tables`, `build_decode_cos_sin`, or any multimodal prefill entry point.

That absence does not settle it. `crates/inference/Cargo.toml` also declares a `[[bin]]` measurement target, `bench_vision_prefill_ab` (`required-features = ["f16", "metal-gpu", "bench-internals"]`), whose timed region calls `bench_support::run_multimodal_prefill`, whose first statement is `request.build_mrope_tables(cfg)` — which calls `build_cos_sin`. The new validation loop therefore runs inside a real measured region, once per prefill. A `[[bench]]`-only search would have reported a clean waiver here, which is exactly the miss this target exists to catch.

A paired base/head comparison of that target was run on the bench host in ABBA order; the full method, pinned config and numbers are in a comment below. Result: no measurable effect — head minus base is +22.548 ms (+0.150%) on the `always_emit_head` arm and -0.030 ms (-0.000%) on the `emit_skip` arm, while the two passes of a single arm disagree by up to 26.907 ms, so neither difference is separable from drift. The mechanism bounds it independently: 32 integer comparisons once per a ~15-second prefill cannot cost milliseconds. No build input differs between the arms — `git diff --name-only` against the base lists only `crates/inference/src/vision/qwen35_mrope.rs`, and manifests, lockfile, features, profiles, build scripts, generated source, bench-target definitions and the harness scripts are byte-identical on both sides, with the harness identity checked rather than assumed.

## Gates

- `cargo fmt --all -- --check` — clean
- `cargo test -p lattice-inference --lib vision::qwen35_mrope` — 25 passed
- `cargo test -p lattice-inference --lib vision::` — 185 passed
- `scripts/lint-docs.sh`, `scripts/lint-absolute-paths.sh` — clean


ohdearquant:
## Paired base/head measurement of the reachable target

**Method first.** `bench_vision_prefill_ab` is a `[[bin]]`, not a Criterion target, so there is no baseline to save or diff and no route through `bench-compare`'s paired machinery. It also cannot run under `scripts/bench-command.sh`: that wrapper takes `/tmp/lion-metal-gpu-test.lock` and the target takes the same lock itself at `bench_vision_prefill_ab.rs:136`, so the two deadlock — filed as #1563 with the measured evidence. `--gpu-handoff` cannot rescue it either, because `scripts/lib/bench_admission.py` admits six Criterion targets by name.

So the run was driven directly: the machine bench window held exclusively for the whole comparison, the binary owning the GPU lock as its own code expects, and every arm gated by `scripts/lib/quiet-probe.py --floor 70` both before and after. Arms ran ABBA — base, head, head, base — so a monotonic drift across the comparison lands on both sides rather than on one.

**Pinned config.** Release build, `--locked`, `--features f16,metal-gpu,bench-internals`, per-arm target directories (no shared `CARGO_TARGET_DIR`; the two binaries differ in size, 1,822,112 vs 1,821,312 bytes, which is the control that the arms are different code). `BENCH_WARMUP=2`, `BENCH_REPEATS=7`, `visual_rows=1024` (actual 1024), `prompt_tokens=1027`, `cache_len=4096`, Qwen3.5-0.8B safetensors checkpoint whose `mrope_section` is `[11, 11, 10]` with `head_dim=256`, `partial_rotary_factor=0.25` → `rope_half=32`. The `scripts/` trees of the two arms were asserted byte-identical with `diff -r` before anything was built, so the harness cannot differ across the comparison. Every arm returned rc 0; idle ranged 96.5–98.7% against the 70% floor at all eight checkpoints.

To reproduce, with `$BASE` and `$HEAD` checked out side by side:

```
cargo build --release --locked -p lattice-inference \
  --features f16,metal-gpu,bench-internals --bin bench_vision_prefill_ab
BENCH_WARMUP=2 BENCH_REPEATS=7 ./target/release/bench_vision_prefill_ab
```

run once per tree per pass, in base/head/head/base order, with the bench window held across all four and an idle probe between them.

**Medians of 7 repeats per arm, then averaged across the two passes.**

| timed arm | base | head | head − base |
|---|---:|---:|---:|
| `always_emit_head` | 15037.425 ms | 15059.972 ms | +22.548 ms (+0.150%) |
| `emit_skip` | 10216.321 ms | 10216.291 ms | −0.030 ms (−0.000%) |

**Read against the drift, which is the point of running ABBA.** The two passes of a *single* arm disagree by up to 26.907 ms (`always_emit_head`, head p1 → p2) and 18.892 ms (`emit_skip`, base p1 → p2). The within-arm pass-to-pass movement is the same size as, or larger than, the head-minus-base difference in both timed arms, and the two arms disagree in sign. Nothing here is separable from drift.

**The mechanism bounds it independently, and that matters more than the table.** The added validation is a loop of `rope_half` = 32 integer comparisons, executed once per prefill inside a region that takes about 15 seconds. A real cost of +22 ms would be roughly six orders of magnitude above what 32 integer operations can cost. The measurement cannot resolve the change; the mechanism says there is nothing to resolve. Both statements are needed, because either alone is weaker than it looks.

**Disposition: no measurable effect, and no effect is mechanically possible at this magnitude.**

**Confounds, including the ones I could have controlled and didn't.** This route certifies less than `bench-compare.sh` does: it has no macOS cooldown, AC-power, thermal-state or HID-idle checkpoint, and no in-phase load sampling within an arm, so a load that starts and ends between two probes is invisible to it. Those checks exist only inside the wrapper this target cannot use (#1563), which is the reason rather than an oversight. The `emit_skip` base arm carries one low outlier, 10124.929 ms against siblings clustered near 10205–10232 ms; medians are reported precisely because they are robust to it, and the pooled standard deviations (base 27.05 ms, head 8.69 ms) show where it lands. Finally, the binary prints its own `delta_pct` near 32% — that is its internal `always_emit_head` versus `emit_skip` comparison, a different axis entirely from base versus head, and it is not the number above.

**One thing the run establishes beyond timing:** the head arm loaded the real checkpoint and returned rc 0 with full `RESULT` output, so the new gate accepts the shipped `[11, 11, 10]` configuration on a live model rather than only in a fixture test.

PR #1567, opened 2026-09-14
chore(release): check package size against the crates.io limit before publishing

ohdearquant:
## What

Adds a pre-publish check that brackets every publishable crate's packaged archive
against the crates.io 10 MiB upload limit, wires it into `make publish` and a
standalone `make package-size-check`, documents it as a release step, and ships a
self-test for it.

## Why

The v0.10.0 release nearly failed at upload. `lattice-inference` measured
**10,493,367 bytes across 420 files** against a **10,485,760** byte limit, about
7.6 KB over, after two test-only tokenizer fixture directories worth 43 MB raw
landed in the package. Excluding them in the crate manifest brought it to
**3,010,549 bytes across 406 files** with every other fixture still present.

That was caught by hand one step before an irreversible publish, and nothing in
the repository would have caught it otherwise. `cargo publish --dry-run`
validates only the leaf tier, because cargo cannot package a crate whose internal
path dependencies are not yet live on the registry — and the crate at risk is in
the second tier. A rejection there leaves the first tier already published and
immutable, with the version number spent.

The trend is what makes this worth a gate rather than a note: `lattice-inference`
published at 7.52 MiB at v0.7.1 and 7.70 MiB at v0.9.0, so the margin was
shrinking release over release before v0.10.0 closed it.

## How it measures

`cargo package --list` builds nothing, needs no registry, and names exactly the
files that would be archived. The check streams that list through tar into gzip
at level 6 — cargo's default, and the level to use here, since level 9 shaves
enough off the estimate to flatter a crate that is close to the limit. Cargo's
three synthesized entries (`Cargo.lock`, `Cargo.toml.orig`,
`.cargo_vcs_info.json`) are bracketed with real content rather than guessed at,
and symlinks are dereferenced because cargo packages their target's contents.

The set of crates checked is derived from `cargo metadata` rather than listed in
the script, and an empty set is refused. A list maintained separately from the
workspace is a gate that can end up checking the wrong set while still reporting
success.

Two outcomes are refusals rather than results, and both exit non-zero:

- **`cargo package --list` failed** — reported with cargo's own message. A dirty
  working tree is the usual cause, and `cargo publish` refuses the same tree.
- **A listed file is not on disk** — the signature of a measurement taken from the
  wrong directory. `cargo package --list` prints crate-relative paths, so summing
  sizes from the workspace root finds almost nothing and would otherwise report a
  comfortably small crate.

Neither refusal reads as an absence of a problem.

## Evidence

Four arms against the real workspace at this base:

| arm | expectation | result |
|---|---|---|
| dirty tree, no `--allow-dirty` | refuse, naming the file | refused, named `tests/paddleocr_vl_adhoc_ocr.rs` |
| all five crates as shipped | all under the limit | fann 132,056 · transport 79,118 · inference 3,010,549 · embed 454,855 · tune 338,809 |
| manifest `exclude` entries removed | catch the over-limit state | `10,493,367 B  100.1%  headroom -7,607 B  420 files  [OVER LIMIT]`, rc 1 |
| run from the workspace root | refuse, not report | `401 listed files are not on disk (first: BASELINE.toml); refusing to report a size` |
| a crate reached through a symlink | match what cargo archives | bracket `10,490,174 B [OVER LIMIT]` against cargo's own `10,490,174` byte archive |

The third arm is the one that matters: it reproduces the state v0.10.0 would have
published, and confirms the upload would have been rejected rather than merely
being close. The manifest was restored by inverse mutation and verified
byte-identical afterwards.

The fourth row guards a fail-open in the gate itself, found after the first cut
of this change. Cargo dereferences a symlink when it packages; Python's tarfile writes a
zero-byte link header by default. On that fixture the bracket first reported
5,245,520 bytes and passed a crate cargo archives at 10,490,174. With
dereferencing on it reports 10,490,174, equal to cargo's archive byte for byte.
No crate in this workspace contains a symlink today, so it was latent rather than
active, but a size gate that undercounts is worse than no gate: it answers the
question it was installed to answer, and answers it wrong.

`scripts/package-size-check-selftest.sh` asserts seven states and runs in CI
beside the other harness self-tests. The over-limit and symlink arms use
incompressible bytes so they cannot pass for the wrong reason — a fixture of
zeros would gzip to nothing and prove the opposite of what it claims. Two arms
drive the shipping wrapper end to end against a stub cargo, covering crate
discovery and refusal propagation without needing a Rust toolchain. Every
guarded behaviour was confirmed by reverting it and watching that arm alone
fail: raising the limit constant, removing dereferencing, emptying discovery,
and swallowing cargo's exit status.

Guards re-run locally at this head: `tests/test_bench_measurement_supervision.py`
(78 tests, OK — the measurement inventory still resolves every script under
`scripts/`, and the three added here are not measurement entry points, so none
needs an inventory entry), `make lint-docs`, `shellcheck -x` on all three shell
files.

## Also in this change

`docs/RELEASE.md`'s bump-and-yank verification was wrong in the reassuring
direction. It told a maintainer to run a bare `curl` against crates.io and read a
`latest_unyanked` field. crates.io refuses a request that does not identify its
caller and answers HTTP 403 with a JSON error object, so code looking for a
version in that object finds none — a refused read renders as "this version was
never published". The field does not exist on the crate object either. It now
points at the checked command, which runs a control first for exactly that
failure.


ohdearquant (commented):

Verdict on head d4931af198308cc1216629794e080b6f9fb27c93: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

PR #1590, opened 2026-09-15
docs(adr): record the Gemma row's added scope and name the Metal acceptance gate

ohdearquant:
Amends the rollout ledger on three points found while characterizing the existing generation
entry points at `f5a8bf2923a038aa055275c2423c2e23deb7514c`. Documentation only; no code changes.

**The Gemma row is not a mirror of the canonical CPU row.** The shared contract is token budget,
cancellation, sampling policy and output stream. The canonical CPU entry point already carries all
four: `generate_streaming_with_cancel(prompt, gen_cfg, on_token, should_cancel)`, where
`GenerateConfig` holds `max_new_tokens` and the sampling fields. The Gemma path exposes
`generate_greedy` and `generate_greedy_with_probe` only, with no streaming and no cancellation,
and no consumer outside its own module and its end-to-end test. That row therefore has to add the
contract before the shared driver can bind to it, which makes it the larger of the two live rows.

**The extraction rows named no acceptance command.** The Metal source file is two mutually
exclusive implementations of the same type: an inner module gated on macOS with the Metal feature,
holding roughly 94% of the file and thirteen generation entry points, and a fallback gated on the
negation, holding two. Both define `generate` and `generate_with_speculation`, which compiles only
because the gates cannot both hold. The default feature set is `std`, `download` and `serve`, so an
ordinary workspace lint or test run compiles the fallback arm, not the module those rows move code
out of. The rows now name the feature-gated command, and a new subsection states the structure the
gate follows from. The same fact bounds the driver: `generate_streaming_with_cancel` has no
fallback counterpart today.

**The target-inventory paragraph overstated its source.** It described a manifest-derived inventory
as the complete declared inventory and relied on it to find consumers. `autobins` is unset and the
workspace edition is 2024, so Cargo also auto-discovers every `src/bin/*.rs` file as a target with
no manifest entry: seven further inference binaries beyond the thirteen explicit tables. Having no
declaration, they carry no `required-features`, so they are precisely the binaries that build under
the default feature set. The paragraph now says a survey takes the union. The inventory artifact
itself scopes correctly to explicit manifest tables; the sentence citing it was the wider claim.

Scope of this pull request: 1 file, 1 commit (`gh pr diff --name-only | wc -l` = 1;
`gh pr view --json commits` = 1). The markdown reformat was verified to be alignment-only by
normalizing runs of spaces and dashes and requiring byte equality, with controls confirming the
normalizer still separates different text and still collapses pure alignment.

No measurement is claimed and none was run: an acceptance clause states a requirement, and the
feature-gated command's shape follows the existing Metal invocation form in this repository and the
cfg predicate read at the pinned ref. The machine was under memory pressure throughout, so nothing
was compiled for this change.

> _Comment authored by Claude Opus 5 (Anthropic) on behalf of @ohdearquant._


oceanwaves630 (approved):
R00-ROLLOUT.md sha256 c0fb5930…8fda7b at e579af33 matches the signed hashes.

PR #1641, opened 2026-09-17
docs(adr): accept ADR-091 and ADR-092 with sign-off edits

ohdearquant:

Accepts ADR-091 and ADR-092. The edits acceptance was conditioned on land in the same commit as the
status flips, so one diff shows both the corrected text and the flip. A flip landing without the
edits would record acceptance of text that was not accepted.

## ADR-091, learned mixture weights

**Decision 5 was circular and is restated as two acts.** As written, an evidence gate had to run
before any dependent code merged, while a failing run left the ADR at Proposed. The closed-loop run
that decides acceptance therefore needed code the ADR forbade merging, and the ADR could never leave
Proposed by its own rules. Now: acceptance fixes the contract; the mechanism code merges under that
accepted contract with the learned mode off by default and `1/k` named as the default weight policy;
the evidence gate governs the flip of that default rather than the status of the ADR; and a run that
fails the threshold leaves the default at `1/k`.

**The weight floor now reports what it dropped.** Decision 4 drops a caller-named adapter below
epsilon without renormalising. That behaviour is correct and is unchanged. What was missing is that
the caller was never told. Both paths now report the dropped adapters and the applied weight vector,
so a request naming an adapter that was absent from the blend is distinguishable from one naming an
adapter that was present at a small weight. Those are different facts about the caller's own request.

**Decision 3 names the falsifier for the entropy floor.** The floor cannot tell collapse from a
correctly sharp distribution, so it is a guard on refits rather than a guard on truth, and the
arbiter is Decision 5's held-out metric. A rejected refit whose held-out metric was improving is the
falsifier for the floor value. The ADR previously gave a floor value with nothing that would ever
change it.

Decision 4's resolution (learned weights normalise because their magnitudes are arbitrary, caller
weights do not because their magnitudes are the request, and both are floored) and the
reachability/`k = 1` paragraph are byte-identical to the base commit.

## ADR-092, model-owned decoder execution

**Completion evidence is now ordered, because the two halves are not equally forgeable.** The
real-consumer execution controls land first, as the gate for the migration PRs: token output plus a
private marker emitted inside the common controller, proved on Qwen CPU, Qwen Metal and Gemma E2B
CPU; a deliberate bypass of the controller must fail; the prefix-cache route repeated under the same
proof; the batch-verifier route excluded. The source-level `pipeline_boundary_contract` resolver
lands as its own step. Its absence does not block a migration PR that already carries the execution
controls, and once it lands its controls are required in that same PR. What ADR-087's and ADR-090's
existing guards already measure of this boundary is stated before the resolver is built rather than
after.

The source-pin recheck at the implementation base stays a stated precondition. ADR-009 and ADR-090
are not modified.

## Notes on the diff

Both status lines flip to `Accepted (2026-09-17)`, each preserving its own trailing-character
convention: ADR-092 keeps its Markdown hard break, ADR-091 has none. The diff shows a status change
and not a whitespace change.

Two prose fixes ride along beyond the conditioned edits. Neither document used em dashes before this
change and neither does now. And ADR-092's risk note no longer says implementation is subject to
acceptance of a Proposed decision, which the flip itself made false.

The resolver's required controls are now referred to by name rather than by count. The count was
ambiguous: six rejection cases, plus a must-pass decoy and the must-fail arms that prove the resolver
can still see anything. A count in prose is never re-derived by the people who add to the list, so it
can only decay, and the sentence does not need it.

Docs only, two files. `lint-docs` passes over 199 tracked Markdown files; no Rust gates apply.

Scope, as the branch diff against the target rather than as a reading of it: **2 files, 1 commit**
(`gh pr diff --name-only | wc -l`, `gh pr view --json commits`). Nothing under `crates/`, so the
bench-compare disposition does not apply to this change.


oceanwaves630 (approved):
Approved at b140c0a9.

All seven required checks pass at this head: actionlint, app-binaries-gate, cargo-audit-gate, cargo-deny, ci-gate, parity-gate, typos. The remaining checks are skipped by the documentation path filters, and a skipped check is counted as not run rather than as a pass.

Documentation only: ADR-091 and ADR-092, nothing under crates/.

PR #1669, opened 2026-09-20
feat(inference): tau floor and entropy-floor refit rejection for the LoRA mixture router (ADR-091)

ohdearquant:

Scope: 3 commits, 1 file changed (+906 / -22), all in `crates/inference/src/mixture.rs`.

ADR-091 gives the router a weight policy instead of a hardcoded `1/k`, and its Decision 3 names
three collapse guards that must all be present before the learned `Softmax` policy is preferred in
production. The first commit landed the weight mechanism and the `epsilon` floor. This adds two of
the three guards, and stops at the third for a stated reason.

**Guard (a), a floor on `tau`.** The existing check requires `tau` finite and strictly positive,
which `1e-9` satisfies while being an argmax with extra steps. `route` now refuses a `tau` below a
configurable floor (`set_tau_floor`, default `1e-2`) with a distinct `RouterError::TauBelowFloor`,
after the finite/positive check rather than in place of it. Refusing rather than clamping follows
the `epsilon` floor's existing drop-rather-than-damp precedent: a silently corrected temperature
would make a caller's configuration and the router's behaviour disagree without saying so.

**Guard (c), an entropy floor with refit rejection.** `submit_refit` takes a candidate gate and the
weight vectors a refit round produced, computes the round's mean weight-entropy, counts consecutive
rounds below the floor, and calls `reload` only on acceptance. A rejected refit leaves the live gate
serving its previous selection, which is what `reload`'s existing no-partial-mutation contract makes
cheap. The floor and the consecutive-round limit are both configurable; an empty round is an error
rather than a division by zero.

The counting state lives on the router and is reached through one new method, rather than in a
separate type, because a second object's lifecycle can drift out of step with the router it guards,
and a guard that something can forget to call is a guard that will be forgotten.

The entropy floor cannot distinguish collapse from a correctly sharp distribution, and the ADR says
so; it decides which gate is loaded, not whether the resulting policy is good. That sentence is
recorded next to the floor's definition so the number reads as policy rather than as truth.

**Guard (b), load-balance and z-loss over the weight distribution, is not implemented, and the
module documentation says so.** `load_balance_aux_loss` takes a probability distribution and
composes with a weight vector directly. `router_z_loss` is derived for unbounded pre-softmax
logits; applying it to a bounded, normalised weight vector is a redefinition rather than a
composition, and the surrounding code uses both as gradient terms inside the trainer's step over
the full logit vector. That leaves at least two architecturally different readings of the ADR
clause, and nothing in the repository adjudicates between them. Writing either one here would
settle a specification question inside an implementation, so no code was written for it. The
default weight policy remains `Uniform`; nothing is enabled by this PR.

**Tests.** 34 in `mixture::`, including both sides of the tau floor (rejected below, accepted at
and immediately above), entropy on flat, skewed and one-hot vectors, and the rejection sequence
with its controls: `M-1` consecutive collapsed rounds does not reject, a good round in the middle
resets the count, and a rejected round leaves the previous selection intact. Three mutation arms,
one per landed behaviour, each predicted before running and matched.

Note for reviewers: `mixture` is off by default, so `cargo test -p lattice-inference --lib
mixture::` without `--features mixture` runs zero tests and exits successfully. The same applies to
`cargo clippy` — the module is simply absent from a default build. Both were run with the feature.

**bench-compare disposition: structural waiver.** `crates/inference/Cargo.toml` declares 38
measurement targets at this ref (25 `[[bench]]`, 13 `[[bin]]`); none names `mixture` in its
`required-features`, and `pub mod mixture` is `#[cfg(feature = "mixture")]` with `mixture` absent
from the default feature set, so no declared target compiles this module, let alone executes it. No
source file outside `mixture.rs` references `mixture::`. The manifest, lockfile, feature set,
profile and bench definitions are unchanged by this PR. Residual risk stated plainly: an
unreachable change is unmeasured, not safe — if the `mixture` feature is later added to a bench
target's `required-features`, this code becomes measurable and has no baseline.


ohdearquant:
Superseded by #1670, same change on a correct base.

This branch was cut before the weight-mechanism work reached main, so it carried its own copy of that commit. The content is byte-identical to what main now has, but as a separate commit object, which left the branch conflicting and with no merge ref — so no required check ever ran on it, and the pending state here was the third one, "the gates do not exist at this head", rather than a slow queue.

#1670 is the two guard commits cherry-picked onto current main. The resulting file is byte-identical to the state gated here, verified by checksum before pushing.

PR #1672, opened 2026-09-20
docs(adr): scope ADR-091 Decision 3 weight-distribution clause to the load-balance term

ohdearquant:

Scope: 1 commit, 1 file changed, docs only.

ADR-091 Decision 3's second bullet asks for "the load-balance and z-loss terms ... applied to the
weight distribution, not only to the selection logits". Read as written it puts both terms on the
same vector, and that is wrong for one of them. This amendment scopes the phrase to the
load-balance term and names the misreading rather than quietly avoiding it, following the
convention the 2026-09-19 amendment set in the same document.

**The z-loss stays on the raw logits.** It is the *square* of log-sum-exp over the pre-softmax
scores, `(log Σ exp(s_i))²`, with gradient `2·lse·p_j`, and its purpose is to tax a gate that grows
its logits without bound — something only an unnormalised vector can do. Over a normalised weight
vector the whole chain collapses. For `k` adapters whose weights sum to one, `lse` runs from
`log(k·e^(1/k))` at uniform to `log(e + k − 1)` at one-hot, so the objective runs `2.678` to `3.040`
at `k = 4` and spans `0.091` on a value near `17.5` at `k = 64`. The gradient degenerates further,
and the gradient is what a guard works through: `2·lse·p_j` becomes a near-constant multiple of
`p_j`, varying by under `0.3%` at `k = 64` across the entire space from uniform to one-hot. A term
that pushes every weight down in proportion to itself by the same factor whatever the shape carries
no information about the shape, so moving it there would not be a stricter guard but a guard that
reports success by arithmetic.

**The load-balance term does move, and the ordering is the substance.** It runs on
`softmax(selected logits / tau)` taken **before** Decision 2's `epsilon` drop and renormalisation.
An adapter the floor has dropped has no gradient path through the drop, so a balance term computed
on the post-floor vector cannot move a weight that has already gone to zero — and that is exactly
the case the guard exists for, since a router collapsing onto one adapter is what drives the others
under `epsilon`. Computed before the floor, the collapsing adapters are still present and still
carry gradient.

`tau` enters as the `1/tau` scale, bounded below by guard (a). Without that floor the scale is
unbounded and the balance gradient can be made arbitrarily small by sharpening, which is the
failure guard (a) exists to prevent. Decision 3 says the three mechanisms are load-bearing on each
other; this is one of the places where that is mechanical rather than rhetorical.

**Guard (b) is a gradient term, not a monitor.** Decision 3 states the three mechanisms *bound*
collapse, and a quantity that is only computed and reported bounds nothing. It belongs in the
objective the trainer already applies.

The amendment builds on the 2026-09-19 one rather than restating it: the batch form and the
trainer-private route-frequency EMA are unchanged. "Only the second argument moves" is the tidy way
to state the rest and it is false, so the amendment specifies the three things it hides.
`load_balance_aux_gradient` refuses unequal lengths, its gradient indexes the gate's output layer,
and it differentiates through a plain softmax; a vector over the selected top-`k` satisfies none of
those. The weight vector is therefore projected into the gate's full output space (zero off the
selected set), the returned gradient is scaled by `1/tau` to reach the logits the trainer updates,
and because `step`/`rloo_step` carry neither the selected indices nor `tau` and cannot gain
parameters without a major break, the learned path routes through a new trainer entry point beside
them. The route-frequency EMA folds the same projected vector, so it tracks weighted traffic rather
than selection counts.

A consequence and a falsifier are both recorded in the text. The consequence: the two terms now
take arguments from different stages of the same forward pass, so a future change that reorders
selection and weighting silently changes what the guard measures, and a test should tie each
argument to its stage rather than to a vector of the right length — the projection makes every
vector in the chain the same length, so length is exactly the property that can no longer catch a
mistake. The falsifier: a closed-loop run
in which routing collapses onto one adapter while the load-balance gradient stays near zero, which
would mean the pre-floor placement is not reaching the collapsing adapters after all.

No code changes here, and no default changes. The learned weight policy remains off; ADR-091's own
gate keeps it unenabled until all three guards exist.


oceanwaves630 (approved):
Spec sign-off at 5982245a. Read the amendment against the 2026-09-19 one and the live trainer: the z-loss keeps its raw-logit arguments; the load-balance term takes softmax(selected logits / tau) before the epsilon drop, through the batch gradient form with route_freqs unchanged; both stay gradient terms in the trainer step. The bounds on log-sum-exp over a normalised k-vector check out: uniform gives log k + 1/k, one-hot gives log(e + k - 1), and convexity puts every other point between them. The stated consequence (arguments from two stages of one forward pass) and the named falsifier are the right shape. Accepted; implementation may proceed against this text.

ohdearquant (commented):

Verdict on head 5982245a347d40c97161c8edf44d1521043c1490: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

oceanwaves630 (approved):
Spec sign-off at 842f517a, replacing the one at 5982245a, which accepted a call shape the helper refuses and a range computed on the inner log-sum-exp rather than its square.

Checked against the trainer source: load_balance_aux_gradient refuses unequal lengths and returns K * p_j * (f_j - sum_i f_i p_i), so a full-space projection with zeros off the selected set gives exactly zero to unselected entries and the f-mean over the selected set, with the softmax Jacobian restricted to that set giving the same expression. The 1/tau factor is the chain rule for softmax(z / tau) and sits with the caller. observe_routing returns silently on a length mismatch, so the projection is what keeps the fold live. The squared objective's range at k = 4 and its near-constant gradient multiplier at k = 64 reproduce. The additive entry point beside step and rloo_step is the right call given the frozen signatures, and it is what binds the guard to the learned path.

Accepted; implementation may proceed against this text.

PR #1673, opened 2026-09-20
test(inference): widen the silent-return scan to benches and examples, and stop it following symlinks

ohdearquant:

Scope: 1 commit, 1 file changed.

The crate-source scan that guards against a silent `env::var(...)` / bare `return;` checkpoint site
read two roots, `src` and `tests`. `benches` and `examples` carry 56 `.rs` files between them and
were outside it. A checkpoint-reading site does not care which Cargo target it compiles into, so
the exclusion was arbitrary rather than principled.

**This is prophylactic, not a fix.** None of the 56 files carries the shape today, measured with the
scanner's own predicate reimplemented against the same fixture that the in-tree control uses (the
fixture resolves to line 2, as the committed control asserts). Saying so is the point: the
population this widens onto is the one nobody is currently watching, which is where the next
instance lands.

**The walk no longer follows symlinks.** It classified entries with `Path::is_dir`, which follows
the link and answers about the target. Measured: a link at `crates/inference/src/<name>` pointing at
`examples`, with a bad site planted in `examples`, made the pre-change scan report a hit at
`src/<name>/probe.rs` — a file under a root the scan had not named. A link pointing at an ancestor
is worse, since the pending queue carries no visited set and the walk does not terminate. Entries
are now classified from the directory entry's own file type, and skipped symlinks are collected and
asserted empty rather than dropped: skipping a symlink and reporting a clean tree are otherwise the
same observable, which is the fail-open direction for a guard whose output is an absence.

**The population assert is now per root.** It was a crate-wide `scanned.len() > 1`. One large root
keeps that plausible while another contributes nothing, and a root that silently reads empty is
exactly the widening failure this change is supposed to make visible.

Mutation arms, predicted before running and reconciled after:

| arm | predicted | observed |
| --- | --- | --- |
| bad site in `benches/`, widened roots | scan test red | red, names the `benches` path |
| same site, roots reverted to `src`/`tests` | scan test **green** — the site is invisible | green |
| bad site in `examples/`, widened roots | red | red, names the `examples` path |
| directory symlink under `src/`, no bad site | red on the symlink assert | red, names the link |
| pre-change `is_dir` + symlink into `examples` + site there | red naming a path **through** the link | red at `src/<name>/probe.rs` |

The second row is the one that makes the first meaningful: the identical site passes without the
widening, so the widening and not something else is what catches it.

Cost: the widened scan reads 331 files and the test reports `0.07s` across five consecutive runs
(1 test each run; the filter is the full exact path, since a truncated `--exact` name runs zero
tests and still prints `ok`). The narrow baseline is 275 files, so the added work cannot exceed that
0.07s total and the delta sits below the harness's own two-decimal reporting resolution — stated as
a bound rather than measured as a difference, because a differential below instrument resolution is
not a measurement.

Gates: `cargo fmt --all -- --check` and `cargo clippy --workspace --all-targets -- -D warnings` via
the repo pre-commit hook, both clean. `cargo test -p lattice-inference --lib test_support::` reports
**8 passed; 0 failed**.

### bench-compare disposition

Structural waiver, and the search is stated rather than asserted. `crates/inference/Cargo.toml` at
this ref declares 25 `[[bench]]` and 13 `[[bin]]` targets; both counts are read from the manifest at
the ref rather than from a directory listing, since the `[[bin]]` measurement targets whose timed
regions call into model code are invisible to a `benches/` glob.

No declared target reaches the change, and the proof is at compilation rather than call path:
`mod test_support` is declared `#[cfg(test)]` in `lib.rs`, and the edited code sits inside a
`#[cfg(test)] mod tests` within it. Benches and bins compile against the library without `cfg(test)`,
so the module is absent from every one of those 38 targets' effective source, whatever their
features. Base and head binaries are byte-identical inputs for all of them.

The build-input clause is also satisfied: the diff is one `.rs` file, with no manifest, dependency,
lockfile, feature, profile, build-script or bench-definition change.

Residual risk, stated because it is the point of the disposition: an unreachable change is an
unmeasured one, not a safe one. What is unmeasured here is the added wall time on a `cargo test`
run, and that is bounded above by the 0.07s figure quoted above rather than by this waiver.


ohdearquant (commented):

PR #1678, opened 2026-09-20
test(inference): stop the silent-return scan following a symlinked root

ohdearquant:
**Scope: 1 file, 1 commit.** Read from the API rather than described:

    gh pr diff 1678 -R ohdearquant/lattice --name-only | wc -l   ->  1
    gh pr view 1678 -R ohdearquant/lattice --json commits --jq '.commits|length'  ->  1

## The defect

`rust_sources_under` classifies every directory entry from the entry's own file type, and its
doc comment states the rule explicitly, naming `Path::is_dir` as the form that must not be
used because it follows the link and answers about the target. Its one caller then validated
each of the four scan roots with `root.is_dir()`.

The rule was stated in one place and broken in the other, three dozen lines apart. A symlinked
`benches` passes the root check, is walked, and contributes `.rs` files from outside the crate
— the same fail-open the entry-level classification exists to close, escaped one level up. The
guard then reports "no silent sites remain" over a population it never read, or over one it
was never meant to reach.

This is the sibling-invocation-path shape: a rule enforced at the site that was being edited
and not at the one call site outside it.

## The fix and why the test has teeth

`is_real_dir` asks `symlink_metadata`, which does not follow.

The test asserts the symlink is rejected **and** that `Path::is_dir` accepts it. The second
assertion is the load-bearing one: without it, swapping the helper back for the naive form
leaves the test green on the real-directory case alone, and a guard test that cannot express
the defect it guards is decoration. Asserting both pins the difference between the two forms
rather than the behaviour of whichever one we happen to call.

Verified by mutation, with the prediction written before the run: with `path.is_dir()` in the
helper body, the rejection assertion fails and nothing else does; restored by inverse edit with
the file's SHA-256 asserted back.

One process note, because it is the reason the arm is trustworthy rather than a detail: the
first attempt at that arm was filtered as `test_support::is_real_dir`, which matches no test —
it ran 0 tests and printed `ok` with exit 0, which is the identical output to a passing arm. It
was re-run with the full path `test_support::tests::is_real_dir_rejects_a_symlink_to_a_directory`
and a same-invocation control showing a non-zero count on the unmutated tree first.

## Gates

    cargo fmt --all -- --check                                      rc=0
    cargo clippy -p lattice-inference --all-targets -- -D warnings  rc=0
    cargo test -p lattice-inference --lib test_support::   9 passed, 0 failed

## bench-compare disposition

**Waiver, and the search that supports it.** The change is confined to a `#[cfg(test)]` module
inside `crates/inference/src/test_support.rs`: a new private helper and one call site, both
compiled out of any non-test build. There is no runtime call path to reach, because there is no
runtime artifact containing the code.

That is the one case where the structural argument is genuinely available rather than
convenient, and it is stronger than the usual unreachable-call-path form: the usual form argues
that no bench group executes a line that is present in the binary, while here the line is not in
the binary. The change's build inputs are unchanged too — no manifest, dependency, lockfile,
feature, profile, build-script or bench-target definition is touched, which is the condition the
structural waiver actually requires.

Residual risk, stated because an unmeasured change is not a safe one: none for runtime
performance, by the above. The real risk this change carries is the opposite kind — it makes the
guard STRICTER, so a repository that legitimately contained a symlinked scan root would now fail
this test rather than quietly scanning through it. That is the intended direction and there is no
such root today.


ohdearquant (commented):

PR #1679, opened 2026-09-20
test(inference): allocation arm for the shared decoder driver, with its controls in CI (ADR-090 R03 row D)

ohdearquant:

Adds the allocation instrument for ADR-090's shared decoder driver, and runs its
checkpoint-free controls in CI. No shipped code changes: `crates/*/src/` is untouched.

## Scope

    gh pr diff <n> --name-only | wc -l   ->  3
    gh pr view <n> --json commits        ->  2

Three files, two commits, +352 -0. The manifest diff is a comment block and one
additive `[[test]]` table; no dependency, feature, profile, build-script or bench-target
definition changed.

## What lands

`crates/inference/tests/decoder_allocation_arm.rs` — a thread-local counting
`#[global_allocator]` over `Qwen35Model::generate`, recording call count, total bytes,
count of allocations at or above 256 KiB, and the largest single allocation, for each of
the four prompt/config cases the CPU greedy golden already pins.

It is an A/B instrument. A global allocator is per binary, so the comparison is this
target run at two refs with the printed records diffed; nothing in it asserts a cross-ref
delta, because nothing in it can see one. That is why it is `test = false` — that, and a
measurement body which needs a 1.6 GiB bf16 checkpoint no runner has.

Thread-local rather than process-wide counters follow the precedent in
`crates/tune/tests/lora_apply_no_alloc.rs`: the test body runs on a harness thread while
other harness threads allocate during the measured window, and a process-wide counter
attributes that to the subject.

## Evidence: base vs head across the driver change

base `8952beea` (the driver's parent) vs head `7bbe0632` (the driver, #1677).
Arms proven distinct before measuring — `crates/inference/src/decoder/driver.rs` is
present at head and absent at base — because an A/B whose two arms could be the same
binary reports a perfect "no change". The head arm was run twice and the two records are
byte-identical; allocation counts are deterministic given the same inputs and code, which
is why one run per arm is evidence here and would not be for timing.

| case | calls base → head | Δ calls | Δ bytes |
| --- | --- | --- | --- |
| short_factual | 183 → 186 | +3 | +68 |
| long_prose | 383 → 386 | +3 | +480 |
| short_factual_reasoning_budget_4 | 197 → 201 | +4 | +176 |
| long_prose_reasoning_budget_4 | 383 → 387 | +4 | +588 |

| case | large_calls (≥256 KiB) base → head | max_bytes base → head |
| --- | --- | --- |
| short_factual | 21 → 21 | 1,048,576 → 1,048,576 |
| long_prose | 45 → 45 | 2,654,208 → 2,654,208 |
| short_factual_reasoning_budget_4 | 21 → 21 | 1,048,576 → 1,048,576 |
| long_prose_reasoning_budget_4 | 45 → 45 | 2,654,208 → 2,654,208 |

**The load-bearing table is the second one, and it is unchanged everywhere.** D1's rule is
not "allocate less", it is "do not materialize an owned vocabulary-sized vector to cross
this interface". Large-allocation count and largest single allocation are identical on
every case, so the driver introduced no buffer of that class. Measured, not argued.

The driver costs 3 to 4 additional small allocations per generation, 68 to 588 bytes, on a
base of 183 to 387 calls and roughly 25 to 45 MB — consistent with its own bookkeeping
(result struct, trace, logprob-offset vector, boxed session) and with nothing else. The
budgeted cases cost one allocation more than their unbudgeted twins on both prompts.

The pre-existing owned vocabulary-width `Vec<f32>` returned by
`prefill_tokens_batched_for_generate` is present on **both** arms, is inside the unchanged
`large_calls` figures, and is not reported as a driver cost.

### Reproducing the base arm

The base arm is an **overlay**, and an earlier version of this header hid that. This file does not
exist at any ref it is compared against, so "run it at the base ref" is not a checkout and a reader
following it literally finds no such target. The recipe is now in the file's header: detached
worktree at the base ref, copy the file in, append the same `[[test]]` block this branch adds, run
there.

The recipe was executed verbatim before being written down, and its first version was wrong -- it
failed at the copy step because the copy reads out of the current checkout and I ran it from one
that predates this target. The corrected text says to run it from the head side and says why.

Run through the recipe exactly as documented, at base `8952beea`:

    record  short_factual                     183  24952988  21  1048576
    record  long_prose                        383  44756356  45  2654208
    record  short_factual_reasoning_budget_4  197  25937255  21  1048576
    record  long_prose_reasoning_budget_4     383  44748255  45  2654208

Byte-identical to the base column in the table above, so the comparison is reproducible rather
than merely asserted to be.

## Controls, and the mutation arm on the CI step

Two checkpoint-free controls run in CI (`-- controls::`):

- `counter_sees_zeroed_allocations` — an ordinary `Vec::with_capacity` must be counted
  first, in the same body. Without that, a zero from the zeroed-path check would be
  indistinguishable from a dead counter, and a dead counter reports "no allocations",
  which reads like good news. The second half pins that `GlobalAlloc::alloc_zeroed`'s
  default still delegates to `alloc`; an override forwarding to `System.alloc_zeroed`
  would blind this arm to exactly the buffers it exists to see.
- `vocabulary_sized_allocation_is_detected` — materializes a 151,936-entry `f32` buffer,
  the width D1 forbids across the interface, and shows the histogram catching it. A
  detector that has never registered a large allocation proves nothing about a run in
  which none appears.

They live in a `controls` module rather than behind a `--skip` of the measurement, so a
checkpoint-needing test added later is not dragged into the CI step.

The CI step asserts the passing count. `cargo test` prints `0 passed` and exits 0 for a
renamed module, a misspelled filter, a deleted test, or a target behind a cfg the command
did not enable, and every one of those renders as a green step.

| arm | predicted | observed |
| --- | --- | --- |
| step body verbatim | rc=0, parses 2 | rc=0, `2 passed; 1 filtered out` |
| filter renamed `controls::` → `controlz::` | rc=1 | `0 passed`, cargo rc=0, **step rc=1** |

## The timing half of this row, reported here and deliberately kept out of the disposition above

This row's evidence is a pair: allocation and timing. The table above is the allocation half. The
timing half has now run, and it measures the same base/head pair -- the decoder driver -- not this
PR's own diff, which contains no shipped code at all. It is reported here because it completes the
row, and it is fenced off from the disposition below because conflating them would let a
measurement of someone else's diff look like a measurement of this one.

192 samples, four cases, 6 ABBA blocks each, across **two independent runs**. 192/192 generated
token ids matched the golden's expected ids for their case, so a timing number from a run that
generated different text could not pass silently. Arms proven distinct by binary hash first.

Pooled over both runs, n = 12 paired block deltas per case, t critical (df=11) = 2.20:

| case | Δ ms | t | Δ% | 95% CI (ms) |
| --- | --- | --- | --- | --- |
| short_factual | +2.21 | 1.44 | +0.27% | [−1.17, +5.59] |
| long_prose | −0.96 | −0.65 | −0.08% | [−4.21, +2.30] |
| short_factual_reasoning_budget_4 | −5.88 | −1.76 | −0.56% | [−13.22, +1.47] |
| long_prose_reasoning_budget_4 | −6.25 | −2.42 | −0.41% | [−11.92, −0.58] |

No case shows a reproducible difference. Three intervals contain zero; the fourth crosses in the
faster direction, and with four cases tested at .05 the chance of one spurious crossing is about
19%, so it is reported rather than claimed. The instrument's resolution is roughly ±3 to ±13 ms at
95% against generation times of 816 to 1508 ms.

**Why two runs and not twelve blocks.** Run 1 on its own reported +4.8 ms on `short_factual`
(+0.58%) with all six block deltas positive -- about four standard errors, which reads as a small
real regression. The replication gave −0.3 ms with mixed signs. The arithmetic was correct and the
denominator was not: a block-diff standard deviation measures variation inside one phase and is
blind to the drift between phases, which was the larger term here. More blocks inside run 1 would
have tightened an interval centred on an artifact.

## bench-compare disposition

**Waiver, on the structural ground, and the searched population is named.**

Runtime source: zero lines under `crates/*/src/` change, so there is no call path anywhere
that reaches an edited line — not an unreachable change, an absent one.

Build inputs, which a call-path argument cannot cover: the only manifest change is an
additive `[[test]]` table plus comments. `test` and `bench` are separate manifest flags, so
this was measured rather than assumed — `cargo bench -p lattice-inference --no-run`
selects the lib, the bins and the declared bench targets, and no integration-test target
appears in that selection, `decoder_allocation_arm` included. Every `cargo bench` call
site in the tree (`scripts/bench-ci.sh`, `scripts/bench-gate.sh`,
`scripts/lib/bench-compare-impl.sh`, `.github/workflows/bench-update.yml`) additionally
passes an explicit `--bench <name>`. No dependency, lockfile, feature, default-feature,
profile, build-script, generated-source or harness change is in the diff.

Residual risk, stated because that is the point of the disposition: an unmeasured change
is not a safe change. What is unmeasured here is the timing half of this row, which is
separate work and needs the quiet bench box. An allocation count is deterministic and
machine-independent, so that constraint does not apply to the arm above — that is a
statement about which measurement the rule governs, not a shortcut around it.

## Not in this PR

Prefill mode (serial vs batched) is absent from the matrix, and the reason is stronger
than "this target cannot reach it".

`force_serial_prefill()` has two definitions: under `#[cfg(test)]` it reads the
`FORCE_SERIAL_PREFILL` atomic, and under `#[cfg(not(test))]` it is `false`. So in every
build that is not the crate's own `--lib` test binary — every `[[bin]]`, every integration
test, and the shipped library — the serial-prefill branch is a constant-false dead path.
The atomic is additionally `pub(crate)`, but that barrier is not even needed.

This is worth stating precisely because the obvious next move is wrong: the serial
dimension does **not** become reachable by moving the measurement into a timing binary,
since a bin is also a separate crate linking the ordinary rlib. Measuring that dimension
requires either an in-crate `#[cfg(test)]` measurement — which would put a global allocator
over the whole crate unit-test binary — or promoting the hatch behind a non-default feature,
which is a public-surface change that should carry its own justification rather than ride
inside a measurement change.


ohdearquant (commented):

Verdict on head a6de8561f9f170556fe585834bccff589c0a6593: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

PR #1682, opened 2026-09-20
test(inference): contract test that no deprecation `since` names an unreleased version

ohdearquant:

Five `#[deprecated(since = ...)]` declarations named `0.11.1` while the workspace
sat at an unreleased `0.11.0` against a published `0.10.0`. They were corrected by
reading the registry by hand. Nothing checks the invariant, which is why it drifted
and why the correction had to be remembered rather than caught.

## What it does

Parses this crate's `src`, `tests`, `benches` and `examples` sources with `syn` and
fails on any `#[deprecated(since = ...)]` value ahead of `CARGO_PKG_VERSION`.

Two things are reported rather than skipped, because skipping either makes it
indistinguishable from a compliant declaration: a value that is not a dotted integer
version, and a file that does not parse. A `cfg_attr` carrying a `deprecated` payload
is reported as unresolved, since this reader does not expand it.

## Why it parses instead of matching text

The first version of this check was lexical, requiring a plain `"` after `since =`.
Three spellings that `rustc` accepts read as ABSENT to it:

```rust
#[deprecated(since = r"0.12.0")]
#[deprecated(since = r#"0.12.0"#)]
#[deprecated(
    since
        = "0.12.0"
)]
```

All three compile — checked with `rustc --edition 2024` on a fixture carrying each —
so the cheapest way to silence that guard was to write a raw string. That is the
fail-open direction for a guard whose entire job is to notice a new declaration, and
the three spellings are instances of a class rather than a list to patch.

Parsing also retires a property the lexical version depended on. It could not be
allowed to see its own fixtures, so every fixture had to spell its quote `\"` inside
a string literal. A string literal in a function body is not an attribute, so the
fixtures are now ordinary Rust, and one arm pins exactly that.

`syn` and `proc-macro2` with `span-locations` are already dev-dependencies here, and
`forward/metal_qwen35.rs` already uses `syn::parse_file` in its own test code, so this
is the crate's existing instrument rather than a new one.

## One walk, not two

The root list, the real-directory check, the per-root population assert, the
skipped-symlink assert and the file reads now live in one helper that both crate-source
scans use. Each previously carried its own copy — the sibling invocation path this
repo's guidance names, where a traversal fix has to land twice or the two drift. The
asserts live in the helper rather than in the callers, so a caller cannot report an
absence over a tree it never established it read.

That also reads the tree once instead of twice. Parsing is the dominant cost either
way: the filtered suite runs in 4.6 s against 0.1 s for the lexical version, over 334
files. That is a deliberate trade and it is stated rather than buried.

## Scope of the check, stated rather than implied

`crates/embed` carries one `since` declaration this scan does not reach. Reaching it
would mean walking out of the crate, which is exactly the fail-open the walker's root
check exists to prevent. Every crate here inherits `[workspace.package].version`, so
the invariant is the same there; the check is not.

## Controls

Four arms are must-MATCH, each planted at a real declaration site in
`crates/inference/src/model/mod.rs`, run against the committed code, then reverted by
inverse edit with the file confirmed byte-identical to `HEAD` afterwards:

| mutation | result |
| --- | --- |
| `since = "0.11.0"` → `"0.12.0"` | FAILS, naming `model/mod.rs:27` and both versions |
| `since = "0.11.0"` → `r"0.12.0"` | FAILS — the raw-string spelling the lexical form missed |
| `since` / `=` split across lines | FAILS, naming line 28, the literal's own line |
| `since = "0.11.0"` → `"next"` | FAILS as an unreadable value, not skipped |
| the reader's attribute predicate broken by one character | FAILS on the must-match control, not on a clean result |

The must-match control runs in the same pass that produces the absence: 7 real
declarations across 334 files, 0 of them fixtures. Without it a reader that has stopped
seeing attributes reports a clean crate.

The reader has its own spelling arms — plain, raw, hashed raw, split across lines, on a
type alias rather than a function — and five near-misses that must NOT match:
`licensed_since = "9.9.9"`, a doc comment, a string containing an attribute, bare
`#[deprecated]`, and `#[deprecated = "why"]`.

The comparison test pins the equal case as PASSING. A `since` naming the version that
first contains the deprecation is the correct declaration, so a guard written with `>=`
would condemn every correct site and be reverted rather than fixed.

## bench-compare disposition

Waived, structurally. The entire diff is inside `mod tests`, and `mod test_support` is
itself declared `#[cfg(test)]` at `crates/inference/src/lib.rs:143`. Bench targets are
separate crates linking the rlib built without `cfg(test)`, so this file does not exist
in any of them: base and head bench binaries are built from identical effective source.
No build input changed either — one file, no manifest, lockfile, feature, profile, build
script or bench-target definition is touched.

The residual risk is the honest one: an unreachable change is not a safe change, it is
an unmeasured one. Here the unreachability is total rather than a call-path argument.

## Verification run

```
cargo test -p lattice-inference --lib test_support::
test result: ok. 12 passed; 0 failed
cargo clippy -p lattice-inference --all-targets -- -D warnings   # clean
cargo fmt --all -- --check                                       # clean
```

`scripts/ci.sh:33` runs `cargo test --workspace`, which is what the `ci` job invokes, so
this test is reached by CI at default features without a new step.


ohdearquant (commented):

Verdict on head 40f6432152c3cc7492b542abb79f42dfb8134507: REQUEST-CHANGES, 1 blocking finding. Finding details are delivered to the review's recipients rather than posted here. Do not merge this head while blocking findings are outstanding; a pipeline comment on a newer head supersedes this one.

Issue #818, opened 2026-07-10
lint: require a same-line issue number on every TODO/FIXME/HACK marker


ohdearquant:
> _Issue authored by Claude (Anthropic agent) on behalf of @ohdearquant._

## What

Add `scripts/lint-source-markers.sh`, scanning Rust, shell, and Python
source under `crates/`, `apps/`, and `scripts/`. It matches comment-leading
`TODO`, `FIXME`, `HACK`, and `XXX` tokens and fails when the same line lacks
a same-line numeric issue reference in the form `#\d+` or `lattice#\d+`
(e.g. `TODO(#123): ...` or `TODO(lattice#123): ...`). Use the same explicit
ripgrep exit-code handling as the ignore-reason lint (`0` = violation found,
`1` = clean, any other code = lint failure, not a silent pass). Do not
introduce an allowlist/baseline file for existing exceptions — an actionable
marker either carries a live issue reference or it does not belong in the
tree as a `TODO`/`FIXME`/`HACK`/`XXX`.

This is a **syntactic provenance lint**, not a live-issue oracle. It cannot
tell whether a referenced issue is open, closed, or transferred — the
workspace already has four `TODO(#1958)` sites
(`crates/inference/src/forward/batch_prefill.rs:371`,
`crates/inference/src/forward/metal.rs:133`,
`crates/inference/src/forward/metal_qwen35.rs:1463,1494`) that are
syntactically valid under this lint despite pointing at a dead/closed issue
number. Re-pointing those four sites to a live issue is explicitly **not**
part of this lint's scope — it is owned by whichever code cluster touches
that Metal-roadmap work next. Do not add a GitHub API call to make this lint
issue-liveness-aware; that would make a deterministic, offline source lint
depend on network access and API rate limits inside required CI.

False positives to exclude by matching comment-leading marker tokens only,
not arbitrary substrings: literal `\uXXXX` escape sequences, and prose like
"see issue #123" that isn't a `TODO(...)`/`FIXME(...)`/`HACK(...)` marker.

Internal milestone names (`FP-165`, `FP-095`, `FANN-M5`, `i2`, and similar)
are not issue links and must fail this lint once landed — see the companion
marker-disposition issues in this cluster that file real issues for the
sites currently using those names, so that by the time this lint lands there
is a zero-baseline to gate on.

## Why

A marker with no numeric issue reference (an internal milestone name, or no
reference at all) is unresolvable provenance: there is no way to look up
[...]

PR #1686, opened 2026-09-20
chore(ci): run the source-marker provenance lint in the doc-lint pipeline

ohdearquant:

`scripts/lint-source-markers.sh` had no invoker. No Makefile target, no CI step, no pre-commit
call: the script existed, was self-tested, and ran nowhere. Its sibling `lint-absolute-paths.sh`
is invoked from `scripts/lint-docs.sh`, and `scripts/ci.sh` runs `lint-docs.sh`, so the sibling is
genuinely enforced. This wires the marker lint into the same place, selftest first, exactly like
the two blocks above it.

A guard-shaped script that nothing runs is worse than an absent one: a reader greps, finds the
file, and counts it as coverage.

## What changed

- `scripts/lint-docs.sh`: a `=== Source Marker Provenance (#818) ===` block running
  `lint-source-markers.sh --selftest` then `lint-source-markers.sh`, beside the existing
  absolute-path block.
- `.github/workflows/ci.yml`: installs ripgrep in the job that runs the pipeline. See below.
- `tests/test_lint_docs_modes.py`: the dispatch-contract fixture gains the new check's stub and row.
- `crates/fann/src/gpu/network.rs`: the single violation the lint reports, cleared so the lint
  lands green instead of red.
- `README.md` and the script header: the prerequisite, and the check list that had gone one short.

## The violation was a stale marker, not a deferred task

The line was `FIXME(FANN-M5): wire per-layer CPU softmax fallback`, on the `Activation::Softmax`
arm of the wgpu forward path. It asks for a fallback covering a network shape that cannot be
constructed:

- `Network::new` rejects a Softmax activation on any layer but the last.
- `NetworkBuilder::build` reaches that constructor, and there is a test pinning the rejection.
- `Network::from_bytes` also ends in `Network::new`, so the deserialization path is gated by the
  same check rather than by a parallel one.
- `GpuNetwork::new` takes an already-constructed `Network`, so the GPU path can only be handed a
  validated one.

So the arm is reachable only for the final layer, whose softmax is applied on CPU during readback.
The comment now states that invariant and names where it is enforced. Nothing in the executable
path changed.

## Two environment claims that were wrong, and what the first CI run said

The first push of this branch went red on three checks in under two minutes, and both causes were
assumptions about environments I had not read.

**ripgrep is on no runner in this matrix.** `rg: not found` came back from ubuntu-latest,
ubuntu-24.04-arm and macos-latest alike, so `scripts/ci.sh` failed on every leg at the doc lint.
The lint itself behaved correctly — exit 2, loud, no clean report for a tree it never scanned —
but a gate that cannot run is not a gate. The fix installs ripgrep once per leg and then runs
`rg --version`, because a silent install failure would otherwise resurface later as a lint error
that reads like a lint defect rather than a missing tool. Porting the script off ripgrep was the
alternative and was rejected here: its scan engine, its exit-code contract and roughly twenty
selftest arms are built around ripgrep's semantics and its tracked-file default, and swapping the
engine inside a wiring change is how a guard quietly gets weaker.

**A hermetic fixture pins the dispatch sequence.** `tests/test_lint_docs_modes.py` writes stubs
for each sub-check into a sandbox and asserts the exact order `lint-docs.sh` calls them, so a new
invocation fails the suite until the fixture knows about it. That is the suite working as designed.
It is also the thing to check before adding a call to a script other tests model: the change looked
complete from the lint's side and was incomplete from the fixture's.

## Evidence

Control first, because a lint that passes proves nothing about a lint that can fail. With a bare
`// TODO:` appended to the same file as a deliberate control, `./scripts/lint-docs.sh` exits 1 and
prints:

```
lint-source-markers: found TODO/FIXME/HACK/XXX markers without a same-line issue reference (#N or lattice#N):
crates/fann/src/gpu/network.rs:486:// TODO: deliberate control marker with no issue reference
```

The control was then removed and the file verified byte-identical to its pre-control state by
checksum. With the tree clean, the same script exits 0 and ends `=== Doc Lint Passed ===`, the new
block reporting `lint-source-markers: clean` after its selftest passes.

`actionlint .github/workflows/ci.yml` is clean. All five Python suites that the `actionlint` job
runs pass locally, including the two that parse the workflow file this PR edits, plus the orphaned
suite gate and its selftest.

## bench-compare disposition

Waived, and the argument does not depend on a reachability search: the `crates/fann/` diff is
comment-only on both sides. Every added line in that file matches a comment opener and every
removed line does too, checked mechanically on the diff rather than by reading it. No bench group,
declared or otherwise, can execute a changed byte, because no changed byte is executable. Every
other changed file is a workflow, a script, a test fixture or Markdown, all outside every Cargo
target.

This is weaker than it looks in one respect worth stating: a comment-only change cannot move the
numbers, but it also cannot prove anything about the code it describes. The invariant claim above
rests on the four call sites named, not on this section.

## Scope

6 files, 2 commits, re-derived from `gh pr diff --name-only | wc -l` and
`gh pr view --json commits` after the push rather than counted from the bullet list above,
which is how the first number written here came out one short.


ohdearquant (commented):

PR #1690, opened 2026-09-21
feat(inference): time the LoRA per-selection-change swap cost

ohdearquant:

`bench_lora_mixture` measured the CPU blend cost and end-to-end decode tok/s but nothing
between them, so the cost a serving path actually pays when the adapter selection changes was
unmeasured. This adds `SWAP_BENCH`, which times that change as three phases in the order the
residency registry performs them — blend, GPU unload, GPU upload — and refreshes the
documentation section, whose recorded output was two months stale.

Scope: 2 files, 2 commits, +149 −46.

## Why three phases and not one

The phases are avoidable independently. A path that changes only the mixture weights over a
fixed adapter set could skip the blend and would still pay the unload and the upload. A single
combined figure cannot express that difference, so it cannot be used to decide whether such a
path is worth building.

## Measured

Apple M4, 16 GB, AC power, release build with `--features metal-gpu,f16`, exclusive bench
window held for the whole run, `qwen3.5-0.8b-q4`. Decode runs at 135.5 tok/s, so one token is
7.38 ms. Summing the three phases gives the cost of one selection change:

| r | k | blend µs | unload µs | load µs | total µs | share of one token |
|---|---|---------:|----------:|--------:|---------:|-------------------:|
| 1 | 1 | 14.1 | 12.6 | 29.8 | 56.5 | 0.77% |
| 1 | 4 | 55.4 | 18.5 | 64.8 | 138.7 | 1.9% |
| 1 | 8 | 131.8 | 20.8 | 106.5 | 259.1 | 3.5% |
| 2 | 1 | 25.6 | 13.0 | 36.2 | 74.8 | 1.0% |
| 2 | 4 | 109.7 | 20.1 | 101.7 | 231.5 | 3.1% |
| 2 | 8 | 261.9 | 23.8 | 161.4 | 447.1 | 6.1% |

This half of the bench builds `q_proj` on the 6 full-attention layers, so these are 6-entry
adapters. A full adapter over every layer and every projection has many times the entries and
is not covered by these figures.

## bench-compare disposition

No `make bench-compare` table, and no structural waiver either. `bench_lora_mixture` is a
`[[bin]]` with a plain `fn main()`, not a Criterion harness, so it has no route through
`bench-compare`'s `--save-baseline`/`--baseline` pair, which drives `lattice-inference` and
`lattice-embed` Criterion targets only. The structural waiver is also unavailable: this change
edits a measurement target's own definition, which has no absent call path to name.

So the disposition is a hand-paired before/after of the target itself, ABBA, both binaries
built from their own revisions and run under the exclusive bench window. The target acquires
the Metal GPU lock itself, so it is run directly rather than through `bench-command.sh`, which
would deadlock against it.

`base1` ran at 73.9% ambient idle immediately after a build and fails the idle floor this
repository's bench tooling enforces. It is excluded on that condition, stated before the arms
were read, and its numbers are below so the exclusion is checkable. The other three arms ran
at 89.4–90.8% idle.

- `BLEND_BENCH`, base2 against the mean of the two head arms, six groups: −0.2% to +2.1%, five
  of six under 0.2%. This loop is untouched by the diff.
- `DECODE_BENCH`, same comparison, six groups: +0.3% to +0.6%, all the same direction.

The uniform +0.4% on decode is ordering drift and is not reported as an effect: the diff adds
work *before* the decode measurement and cannot mechanically make decode faster. The two head
arms differ from each other by up to 0.2% on the same binary, which is the spread this sits
inside. Read as decode unchanged.

<details>
<summary>ABBA arms, raw</summary>

```
ARM base1  idle 73.9%   (excluded: below the idle floor, post-build)
BLEND r=1 k=1 49.6 | r=1 k=4 188.8 | r=1 k=8 402.4 | r=2 k=1 71.2 | r=2 k=4 286.6 | r=2 k=8 596.4
DECODE 135.5 135.4 135.1 135.5 135.3 134.9

ARM head1  idle 89.4%
BLEND 31.6 120.8 289.8 54.0 239.3 569.8
DECODE 135.4 135.5 135.1 135.5 135.3 134.9

ARM head2  idle 90.8%
BLEND 31.5 120.8 290.1 54.1 239.2 570.1
DECODE 135.6 135.6 135.3 135.6 135.3 135.1

ARM base2  idle 89.4%
BLEND 30.9 121.0 290.1 54.0 238.9 569.8
DECODE 135.1 135.0 134.7 134.8 134.8 134.4
```

</details>

## Documentation

The `bench_lora_mixture` section recorded a run that crashed on a GDN layer and a `layers=56`
figure derived from a hardcoded 28-layer count. Both halves derive their layer indices from
`config.json`'s `layer_types` now, #637 closed with that fix, and the real count for this
config is 6 full-attention layers. The recorded output is replaced with the run above, the
machine conditions are stated beside the figures, and the failure-mode entry no longer tells
readers not to fix a bug that is already fixed.

One gotcha added while assembling the model directory: the checkpoint loader opens weight files
with `O_NOFOLLOW` and refuses any path whose final component is a symlink, so a model directory
built out of symlinks fails at the first weight. Hard links work; the guard covers the final
component only.

## Also

`BENCH_ITERS=0` parsed fine and then divided every per-iteration figure by zero, printing `NaN`
into fields a reader parses as microseconds. It is refused now.


ohdearquant (commented):

